PR #17, opened 2022-02-18
Feat/json

dexturr:
JSON highlighting for Markets, asset, proposals, genesis (i.e. the less important routes that are likely to just be JSON for now).

JSON highlighting is in Vega colors:
Screenshot 2022-02-18 at 17 54 33


mattrussell36 (approved):
Looks great

PR #20, opened 2022-02-23
Feature/12 add tailwindcss

bglownia:


mattrussell36 (approved):
Looks good to me. I'm surprised we need all the typescript config for the tailwind lib package!

PR #36, opened 2022-03-02
Feat/18 block explorer transaction UI first pass

sam-keen:


apps/explorer/src/app/components/txs/txs-per-block/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const {
+    state: { data: decodedBlockData },
+  } = useFetch<ChainExplorerTxResponse[]>(DATA_SOURCES.chainExplorerUrl, {
+    method: 'POST',
+    mode: 'cors',
+    headers: {
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify({
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
dexturr, line 19:
```suggestion
```

apps/explorer/src/app/hooks/use-fetch.tsx
```diff
@@ -72,15 +73,23 @@ function useFetch<T = unknown>(url?: string, options?: RequestInit): State<T> {
 
     void fetchData();
 
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
dexturr, line 76:
```suggestion
    // Do nothing if the url is not given
    // eslint-disable-next-line react-hooks/exhaustive-deps
```

apps/explorer/src/app/routes/router-config.tsx
```diff
@@ -50,6 +52,17 @@ const routerConfig = [
...
+    path: Routes.PENDING,
+    name: 'Pending txs',
+    element: <Txs />,
+    children: [
+      {
+        index: true,
+        element: <PendingTxs />,
+      },
+    ],
+  },
```
dexturr, line 65: Should be at `txs/pending` rather than it's own individual route

apps/explorer/src/app/components/blocks/home/blocksTable.tsx
```diff
@@ -0,0 +1,52 @@
+import { TendermintBlockchainResponse } from '../../../routes/blocks/tendermint-blockchain-response';
```
dexturr, line 1: Prefer `kebab-casing` for file names over `camelCase` or `PascalCase`, there's some other files kicking around with the same issue:

* txContent -> tx-content 
* txDetails -> tx-detials
* ect..

apps/explorer/src/app/components/blocks/home/blocksTable.tsx
```diff
@@ -0,0 +1,52 @@
...
+export const BlocksTable = ({ data, showTransactions }: BlocksProps) => {
+  if (!data?.result) {
+    return <>No block data</>;
+  }
+
+  return (
+    <Table>
+      {data.result?.block_metas?.map((block) => {
+        return (
+          <>
```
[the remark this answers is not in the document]
sam-keen: This was a tricky one - I can't do a key prop on a fragment, but I need to conditionally return a second table row, so it needs a parent. However, I'm inside a table so there's not an appropriate tag. Webstorm didn't shout at me as it usually would so I hoped I'd got away with it :D

apps/explorer/src/app/components/blocks/home/blocksTable.tsx
sam-keen: will do!

apps/explorer/src/app/components/blocks/home/blocksTable.tsx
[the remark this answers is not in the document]
dexturr: You can do a key prop in a fragment

https://reactjs.org/docs/fragments.html#keyed-fragments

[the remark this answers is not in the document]
sam-keen: Oh nice, thanks, just not with the shorthand. Noted!

dexturr (approved):

PR #53, opened 2022-03-07
Chore/build netlify

### LINKED ISSUE #52: Fix netlify redirects
As the block explorer is a SPA we need redirects back to index.html

There is an issue with netlify in that it is not picking up config files from subdirectories.

Is is not possible to set a config file path or redirects in the UI

A small script to move the app level config to the root should fix the issues, but is inelegant and annoying

dexturr:
Closes #52

### DIFF
--- a/apps/explorer/project.json
+++ b/apps/explorer/project.json
@@ -75,7 +75,7 @@
         ]
       }
     },
-    "build:netlify": {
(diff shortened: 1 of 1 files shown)


edd (approved):
Not ideal, but totally workable!

PR #55, opened 2022-03-07
Test/cypress bdd integration

### LINKED ISSUE #38: Map out E2E testing framework
Create pages objects and test ids for all the pages on the block explorer site incl. common methods used for testing

jtsang586:
Closes #38 

Tests add for navigating and verifying all the main pages of block explorer. 

Fully expecting a lot of tests to fail when more gets added to the explorer which will be picked up and fixed in future PRs.


apps/explorer-e2e/src/support/pages/base-page.js
```diff
@@ -0,0 +1,79 @@
...
+
+  navigateToNetworkParameters() {
+    cy.get(`a[href='${this.networkParametersUrl}']`).click();
+  }
+
+  navigateToValidators() {
+    cy.get(`a[href='${this.validatorsUrl}']`).click();
+  }
+
+  search(searchText) {
```
Ditmir-Vega, line 50: should update the nming to be more explicit 'i. searchForTransaction , searchForMarket, searchForBlock ...

apps/explorer-e2e/src/support/pages/transactions-page.js
```diff
@@ -0,0 +1,37 @@
...
+    cy.getByTestId(this.validatorLink).first().should('not.be.empty');
+    cy.getByTestId(this.blockTime).first().should('not.be.empty');
+  }
+
+  validateRefreshBtn() {
+    cy.getByTestId(this.blockHeight)
+      .first()
+      .invoke('text')
+      .then((blockHeightTxt) => {
+        cy.wait(1000); // eslint-disable-line cypress/no-unnecessary-waiting
```
Ditmir-Vega, line 24: anything we can do to make this more implicit ?

Ditmir-Vega:
1. Do you have a CI / Yml file for running this on PR's ?
2. Is there a cypress config file in here ? i couldn't see one

apps/explorer-e2e/src/support/pages/base-page.js
jtsang586: In future iterations the inherited pages will have more specific methods for searching. Currently AFAIK the search doesn't work properly so this isn't being used at all yet.

apps/explorer-e2e/src/support/pages/transactions-page.js
jtsang586: Yeah I'd like to avoid the wait if possible. I can't think of a much better way to do this. Possibly putting it in a loop to keep clicking the refresh button until the block has changed but I think that's overkill for this.

jtsang586:
> 1. Do you have a CI / Yml file for running this on PR's ?
> 2. Is there a cypress config file in here ? i couldn't see one

1. Not in the PR, got a separate ticket to add tests to CI 
2. Cypress file is here: frontend-monorepo/apps/explorer-e2e/cypress.json

PR #58, opened 2022-03-07
Feat/41 blocks styling

### LINKED ISSUE #41: blocks & blocks/:id styling

sam-keen:
Closes #41 

Styling for the first pass at the block home and block:id routes.

Screenshot 2022-03-07 at 19 59 23

Screenshot 2022-03-07 at 19 59 12


apps/explorer/src/app/components/jump-to-block/index.tsx
```diff
@@ -20,8 +20,20 @@ export const JumpToBlock = () => {
...
-      <input type={'submit'} value={'Go'} />
+      <label
+        htmlFor="block-input"
+        className="block uppercase text-h5 font-bold"
+      >
+        Jump to block
+      </label>
+      <input
+        id="block-input"
+        type={'tel'}
```
mattrussell36, line 31: No need for curly braces here. Same goes for other props in this PR

```suggestion
        type='tel'
```

apps/explorer/src/app/components/blocks/home/blocks-refetch.tsx
```diff
@@ -0,0 +1,11 @@
+interface BlocksRefetchProps {
+  refetch: () => void;
+}
+
+export const BlocksRefetch = ({ refetch }: BlocksRefetchProps) => {
+  return (
+    <button onClick={() => refetch()} className="underline mb-28">
```
mattrussell36, line 7: I can see this being a <ButtonLink /> component in UI toolkit. Basically something we semantically need to be a button but looks like a text link

apps/explorer/src/app/components/blocks/home/blocks-table.tsx
```diff
@@ -12,31 +12,34 @@ interface BlocksProps {
 
 export const BlocksTable = ({ data, showTransactions }: BlocksProps) => {
   if (!data?.result) {
-    return <>No block data</>;
+    return <div className="mb-28">Awaiting block data</div>;
```
mattrussell36, line 15: <p> tag here?

apps/explorer/src/app/components/jump-to-block/index.tsx
```diff
@@ -20,8 +20,20 @@ export const JumpToBlock = () => {
...
     <form onSubmit={handleSubmit}>
-      <input type={'tel'} name={'blockNumber'} placeholder={'Block number'} />
-      <input type={'submit'} value={'Go'} />
+      <label
+        htmlFor="block-input"
+        className="block uppercase text-h5 font-bold"
+      >
+        Jump to block
+      </label>
+      <input
```
mattrussell36, line 29: You could use the input component from ui toolkit now? 

You'll need to either:
- Update the input to allow rendering of a integrated button. 
- Or (and IMO the simpler/better solution for now) just use the input and button components from ui toolkit wrapped in a container with display flex.

apps/explorer/src/app/components/blocks/home/blocks-table.tsx
```diff
@@ -12,31 +12,34 @@ interface BlocksProps {
...
+              <td className="pl-4 py-2">
+                <Link
+                  to={`/blocks/${block.header?.height}`}
+                  className="text-vega-yellow"
+                >
                   {block.header?.height}
                 </Link>
               </td>
-              <td>
+              <td className="px-8 text-center">
```
mattrussell36, line 32: Lets avoid duplicating these classNames with a table cell component?

```
// in compoments/table/table.tsx
const TableCell = ({ children, className }) => {
  const cellClasses = classnames(
    'px-8', // default styles
    className
  )
  return <td className={cellClasses}>{children}</td>
}

// then use
<TableCell className="text-center">
```

apps/explorer/src/app/components/truncate/truncate.tsx
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Truncates a string of text from the center showing a specified number
+ * of characters from the start and end. Optionally takes a children as
+ * a render props so truncated text can be used inside other elements such
+ * as links
+ */
+export function TruncateInline({
+  text,
+  className,
+  style,
```
mattrussell36, line 23: Looks like we dont need a style prop

apps/explorer/src/app/routes/blocks/home/index.tsx
```diff
@@ -15,8 +15,8 @@ const Blocks = () => {
   return (
     <>
       <section>
-        <h1>Blocks</h1>
-        <button onClick={() => refetch()}>Refresh to see latest blocks</button>
+        <h1 className="route-header">Blocks</h1>
```
mattrussell36, line 18: Lets try and embrace tailwind a bit more and just make a <RouteHeader /> component. Added benefit to getting auto completion rather than having to remember/hunt down a className.

```suggestion
        <RouteHeader>Blocks</RouteHeader>
```

apps/explorer/src/styles.css
```diff
@@ -1,4 +1,26 @@
 /* You can add global styles to this file, and also import other style files */
 @tailwind base;
 @tailwind components;
+
+.route-header {
```
mattrussell36, line 5: As per other comment, lets embrace tailwind a bit more. Lets just add components for these styles.

mattrussell36 (changes requested):

dexturr:
https://github.com/vegaprotocol/frontend-monorepo/pull/58#issuecomment-1061074310 

Will be fixed by a rebase

apps/explorer/src/app/components/blocks/home/blocks-table.tsx
sam-keen: Yeah, I wanted block positioning properties and a bit of margin, felt that a div was better than having to use a longhand fragment with a classname and extra CSS to avoid being inline

apps/explorer/src/app/components/txs/txs-per-block/index.tsx
```diff
@@ -24,28 +27,44 @@ export const TxsPerBlock = ({ blockHeight }: TxsPerBlockProps) => {
...
+            <td>Transaction</td>
+            <td>From</td>
+            <td>Type</td>
+          </tr>
+        </thead>
+        <tbody>
+          {decodedBlockData &&
+            decodedBlockData.map(({ TxHash, PubKey, Type }, index) => {
+              return (
+                <tr key={index}>
```
dexturr, line 43:
```suggestion
                <tr key={TxHash}>
```

apps/explorer/src/app/routes/blocks/id/index.tsx
```diff
@@ -23,24 +23,29 @@ const Block = () => {
...
+        <tr className="table-bordered-tr">
+          <td className="table-bordered-td">Mined by</td>
+          <td className="table-bordered-td">
+            <Link
+              className="text-vega-yellow"
+              to={`/validators/${header.proposer_address}`}
+            >
               {header.proposer_address}
             </Link>
           </td>
```
edd, line 37: **Nitpick (non-blocking)**: These should be `<th scope=“col”>` elements, as they’re headers. You could also wrap the `<tr>` in a `<thead>` block, but it’s also non-essential. This definitely doesn’t block anything.

apps/explorer/src/app/components/txs/txs-per-block/index.tsx
```diff
@@ -24,28 +27,44 @@ export const TxsPerBlock = ({ blockHeight }: TxsPerBlockProps) => {
...
-                </td>
-                <td>{PubKey}</td>
-                <td>{Type}</td>
-              </tr>
-            );
-          })}
-      </tbody>
-    </table>
+    <div className="overflow-x-auto whitespace-nowrap mb-28">
+      <table className="w-full">
```
dexturr, line 31: chore (blocking): HTML here is invalid (divs used as children of this component and that is invalid. IMO this shouldn't be a table and just a header with a content body below it

apps/explorer/src/app/components/blocks/home/blocks-refetch.tsx
sam-keen: Good shout. I've made a ticket for this: https://github.com/vegaprotocol/frontend-monorepo/issues/75

apps/explorer/src/app/routes/blocks/home/index.tsx
```diff
@@ -15,8 +15,8 @@ const Blocks = () => {
...
+        <BlocksRefetch refetch={refetch} />
         <BlocksTable data={data} />
       </section>
```
dexturr, line 21: chore (blocking): jump to block should be in section, remove wrapping fragment

dexturr: same in other route

apps/explorer/src/app/components/blocks/home/blocks-table.tsx
dexturr, line 15: chore (blocking): We will render this in an error state as well, should handle errors with feedback

apps/explorer/src/app/components/txs/txs-per-block/index.tsx
```diff
@@ -24,28 +27,44 @@ export const TxsPerBlock = ({ blockHeight }: TxsPerBlockProps) => {
...
-                  <Link to={`/txs/${TxHash}`}>{TxHash}</Link>
-                </td>
-                <td>{PubKey}</td>
-                <td>{Type}</td>
-              </tr>
-            );
-          })}
-      </tbody>
-    </table>
+    <div className="overflow-x-auto whitespace-nowrap mb-28">
```
dexturr, line 30: chore (blocking): needs loading/error states if this is going to be loading data

dexturr: As there may be >300 transactions per block so loading is going to take a while

apps/explorer/src/app/components/blocks/home/blocks-table.tsx
sam-keen: No problem. There's too much variety in the classes needed for defaults though, it's only a couple that use `px-8`, so I'd end up having to add extra classes to the other cells to undo the default. I'll make a component anyway and we can discuss whether to do more after.

apps/explorer/src/app/components/jump-to-block/index.tsx
sam-keen: Will do. I'm going to get this branch sorted without it for now as I want to unblock Edd, but I will revisit this. https://github.com/vegaprotocol/frontend-monorepo/issues/76

apps/explorer/src/app/components/blocks/home/blocks-table.tsx
```diff
@@ -12,31 +12,34 @@ interface BlocksProps {
 
 export const BlocksTable = ({ data, showTransactions }: BlocksProps) => {
   if (!data?.result) {
-    return <>No block data</>;
+    return <div className="mb-28">Awaiting block data</div>;
   }
 
   return (
     <Table>
```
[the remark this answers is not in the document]
sam-keen: @dexturr This data has to be loaded as a second step as it relies on the block data from a previous fetch. I can hoist it if strongly recommended, but I think that having the data loaded at the component level gives us that nice locality of behaviour.

apps/explorer/src/app/components/truncate/truncate.tsx
```diff
@@ -0,0 +1,54 @@
+import * as React from 'react';
+
+const ELLIPSIS = '\u2026';
```
dexturr, line 3: praise (non-blocking): nicely done using proper ellipseses? Elipplisi? Elpipsisesses? Falndereseses?

apps/explorer/src/app/routes/blocks/id/index.tsx
```diff
@@ -23,24 +29,29 @@ const Block = () => {
...
-          <td>
-            <Link to={`/validators/${header.proposer_address}`}>
+      <RouteTitle>BLOCK {block}</RouteTitle>
+      <Table className="mb-28">
+        <TableRow modifier="bordered">
+          <TableHeader scope="row">Mined by</TableHeader>
+          <TableCell modifier="bordered">
+            <Link
+              className="text-vega-yellow"
+              to={`/validators/${header.proposer_address}`}
```
dexturr, line 39: chore (blocking): let's redirect to the all validators page until we build single valdiator pages

```suggestion
              to={"/validators"}
```

apps/explorer/src/app/components/render-fetched/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+  if (loading) {
+    return <StatusMessage className={className}>Loading...</StatusMessage>;
+  }
+
+  if (error) {
+    return <StatusMessage className={className}>Error: {error}</StatusMessage>;
+  }
+
+  // eslint-disable-next-line react/jsx-no-useless-fragment
+  return <>{children}</>;
```
dexturr, line 26: question (non-blocking): can we just:
```suggestion
  return children;
```
and then we don't need the fragment?

apps/explorer/src/app/components/table/index.tsx
```diff
@@ -1,13 +1,88 @@
...
+    <th className={cellClasses} {...props}>
+      {children}
+    </th>
+  );
+};
+
+export const TableRow = ({
+  children,
+  className,
+  dataTestId,
```
dexturr, line 58: chore (blocking): As you  are spreading props, you don't need this. Just pass `data-testid="foo"` to the consumer and it will end up where you spread it (i.e. on the `tr`).

Few other instances of this

apps/explorer/src/app/components/blocks/block-data.tsx
```diff
@@ -0,0 +1,59 @@
...
+      className={className}
+    >
+      <TableRow dataTestId="block-row" modifier="background">
+        <TableCell
+          dataTestId="block-height"
+          className="pl-4 py-2"
+          aria-label="Block height"
+        >
+          <Link
+            to={`/blocks/${block.header?.height}`}
```
dexturr, line 25: chore (blocking): Instead of using `/blocks` hard codded can we use the `ROUTES.Blocks` from the router config so if we need to change routes it's only in a single place

dexturr: Other isntances of this in other places in the PR

apps/explorer/src/app/components/blocks/block-data.tsx
```diff
@@ -0,0 +1,59 @@
...
+          {block.num_txs === '1'
+            ? '1 transaction'
+            : `${block.num_txs} transactions`}
+        </TableCell>
+        <TableCell
+          dataTestId="validator-link"
+          className="px-8 text-center"
+          aria-label="Validator"
+        >
+          <Link to={`/validators/${block.header?.proposer_address}`}>
```
dexturr, line 45: Same as above
```suggestion
          <Link to={"/validators"}>
```

apps/explorer/src/app/components/blocks/home/blocks-table.tsx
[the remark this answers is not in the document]
dexturr: As we are expecting more TXs soon I think rendering early is better so i like your solution more now

apps/explorer/src/app/components/truncate/truncate.tsx
sam-keen: Haha! Bouncing it over to @mattrussell36 as this is his work :)

PR #62, opened 2022-03-08
Test/e2 e ci

### LINKED ISSUE #57: Add E2E tests to GA
Add tests to the CI pipeline so tests are run on every push to PR

jtsang586:
Run Cypress tests in CI

Closes #57

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -30,9 +30,7 @@ jobs:
         uses: actions/setup-node@v2
         with:
           node-version: 16
-      - name: Install root dependencies and run tests
+      - name: Install root dependencies
         run: yarn install
-      - name: Build app
-        run: npx nx affected:e2e --target=build --parallel=3
-      - name: Run E2E tests
-        run: npx nx affected:e2e --target=test --parallel=2 --record --key ${{ secrets.CYPRESS_RECORD_KEY }}
+      - name: Run Cypress tests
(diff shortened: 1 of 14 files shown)


.github/workflows/cypress.yml
```diff
@@ -0,0 +1,36 @@
...
+jobs:
+  cypress-run:
+    name: Run end-to-end tests
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
```
dexturr, line 22: chore (blocking): this ref will not exist if this is not a pull request but the tirggers are also on push. You will need to add a separate job for on push

dexturr: For push I suggest using the same approach they use here: https://nx.dev/ci/monorepo-ci-github-actions defining 2 steps one for push (lines: 13-24) and one for pull request (line 26) you have the PR one but we will need the other as well to ensure that nx affected also works properly on push

dexturr (changes requested):
Great stuff in general dude :+1: nicely done, one small edge case which is hard to miss though push events will cause this to fail, have put some suggestion on how to fix this down 👍 

If you need to test the push flow you can always change the trigger to work on push of your branch only while testing the CI.

.github/workflows/cypress.yml
jtsang586: Done!

dexturr (approved):

PR #82, opened 2022-03-14
Feat/63 Deal ticket

### LINKED ISSUE #63: Deal ticket
- Market/Limit order types
- Quantity (cannot be 0, must be whole int)
- Long/short armament
- Pending confirmation in Vega wallet state
- Wallet rejection state
- Success state showing Txhash (linking to explorer)

#### Ticket type specific fields
- Market
  - TIF (IOC, FOK)
- Limit
  - Price
  - TIF (IOC, FOK, GTT, GTC, GFA, GFN)
  - If GTT, datetime selection shown

NOTE:
- The existing console deal ticket is here https://github.com/vegaprotocol/console/blob/develop/src/routes/trading/windows/deal-ticket/deal-ticket.tsx. Built using blueprint, this will need migration to ui-toolkit/tailwind.

mattrussell36:
- Adds deal ticket as own lib package
- Expands market page query to contain data needed for the deal ticket
- Updates `Dialog` and `FormGroup` components in the ui-toolkit to take more props to handle different styles 
- Adds a `ui-toolkit/utils/intent.ts` for intent based types and helpers.
- Adds a splash component to render centered text for loading/error/not found messages
- Deletes react-helpers/lib/trading and just keeps a map of trading components in the trade layout components.
- Adds `addDecimal` and `removeDecimal` functiosn to a new `decimals` directory in react-helpers

Note:
- For now if the wallet returns an error I'm just printing it as is on screen. There wallet doesn't seem to return consistent errors so we might want to bring this up and fix in a follow up ticket. 

Closes #63


libs/deal-ticket/src/type-selector.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+interface TypeSelectorProps {
+  order: Order;
+  onSelect: (type: OrderType) => void;
+}
+
+export const TypeSelector = ({ order, onSelect }: TypeSelectorProps) => {
+  return (
+    <FormGroup label="Order type">
+      <div className="flex gap-8">
```
sam-keen, line 12: The component inside the FormGroup is identical to the one inside https://github.com/vegaprotocol/frontend-monorepo/pull/82/files#diff-44dcb60c83770bb6d57d8e259e005c644fc29177f0be35c135086f208aa6a94aR12-R32

sam-keen (commented):
I don't yet have the knowledge to review the business logic side, but it's looking good, just a few tweaks from me.

libs/deal-ticket/src/order-dialog.tsx
```diff
@@ -0,0 +1,143 @@
...
+    }
+  }, [foundOrder, setStatus]);
+
+  // TODO: When wallets support confirming transactions return UI for 'awaiting confirmation' step
+
+  // Rejected by wallet
+  if (status === VegaTxStatus.Rejected) {
+    return (
+      <OrderDialogWrapper icon={<Icon name="warning-sign" size={20} />}>
+        <h1 className="text-h4">Order rejected by wallet</h1>
```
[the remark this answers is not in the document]
mattrussell36: I think it makes sense to just integrate it to OrderDialogWrapper

libs/ui-toolkit/src/components/dialog/index.tsx
```diff
@@ -1,19 +1,37 @@
...
-export function Dialog({ children, open, setOpen, title }: DialogProps) {
+export function Dialog({
+  children,
+  open,
+  setOpen,
+  title,
+  intent,
+}: DialogProps) {
+  const contentClasses = classNames(
+    'fixed w-[520px] top-[60px] p-28 left-[calc(50%-260px)]',
```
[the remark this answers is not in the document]
mattrussell36: I think I've overwritten some of Barts styles here too which I need to amend. What do you mean by documentation here?

libs/deal-ticket/src/order-dialog.tsx
```diff
@@ -0,0 +1,143 @@
...
+    variables: { partyId: keypair?.pub || '' },
+    skip: !keypair?.pub,
+    onSubscriptionData: ({ subscriptionData }) => {
+      if (!subscriptionData.data.busEvents.length) {
+        return;
+      }
+
+      // No types available for the subscription result
+      // eslint-disable-next-line
+      const matchingOrder = subscriptionData.data.busEvents.find((e: any) => {
```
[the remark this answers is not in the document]
mattrussell36: The event here is not a standard JS event but an [event type from the graphql schema](https://github.com/vegaprotocol/data-node/blob/9692153b546aea53fc0b5c5601fcbcf7bc6f5f49/gateway/graphql/schema.graphql#L2732).

Normally we'd pass a generic type to the useSubscription hook which would ensure `e` is correctly typed. But the graphql type generation exists in the trading app and not dealticket lib. So I kind of see this as temporary until we figure out our data layer solution between apps and libs/components

apps/trading/components/splash/splash.tsx
```diff
@@ -0,0 +1,14 @@
+import classNames from 'classnames';
```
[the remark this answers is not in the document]
mattrussell36: Moved to ui-toolkit with a story

libs/deal-ticket/src/deal-ticket.tsx
```diff
@@ -0,0 +1,260 @@
...
+  );
+
+  const handleSubmit = (e: FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+    submit(order);
+  };
+
+  const getDialogIntent = (status: VegaTxStatus) => {
+    if (status === VegaTxStatus.Rejected) {
+      return 'danger';
```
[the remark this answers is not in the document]
mattrussell36: I've update ui-toolkit/utils/intent to export an intent enum

libs/deal-ticket/src/side-selector.tsx
```diff
@@ -0,0 +1,35 @@
...
+  return (
+    <FormGroup label="Direction">
+      <div className="flex gap-8">
+        {Object.entries(OrderSide).map(([key, value]) => {
+          const isSelected = value === order.side;
+          return (
+            <Button
+              onClick={() => onSelect(value)}
+              className="flex-1"
+              type="button"
```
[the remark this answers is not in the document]
mattrussell36: updated button to have a default type of 'button'

libs/deal-ticket/src/expiry-selector.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { format } from 'date-fns';
+
+interface ExpirySelectorProps {
+  order: Order;
+  onSelect: (expiration: Date | null) => void;
+}
+
+export const ExpirySelector = ({ order, onSelect }: ExpirySelectorProps) => {
+  const date = order.expiration ? new Date(order.expiration) : new Date();
+  const dateFormatted = format(date, "yyyy-MM-dd'T'HH:mm");
```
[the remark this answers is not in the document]
mattrussell36: I think this already gets handled by input type date. React (and I think in browsers generally regardless of framework) require the format "yyyy-MM-dd'T'HH:mm". Then the browser seems to handle rendering the correct format.

So for me in the US I correctly see 'MM/dd/yyyy, HH:mmm PM'.

libs/deal-ticket/src/deal-ticket.tsx
```diff
@@ -0,0 +1,260 @@
...
+      status === VegaTxStatus.Pending ||
+      status === VegaTxStatus.Rejected
+    ) {
+      setOrderDialogOpen(true);
+    }
+  }, [status]);
+
+  let ticket = null;
+
+  if (order.type === 'TYPE_MARKET') {
```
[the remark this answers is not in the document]
mattrussell36: Done. I've also moved all the order types to lib/wallet/types.ts. The reason I think they should live there is that these types (eg the string 'TYPE_MARKET') are the values that are accepted by the wallet to issue an order submission transaction.

Annoyingly they are slightly different to what gets returned by the API, so I'm not sure if having them there is the best solution, but I think its fine for now.

libs/deal-ticket/src/order-dialog.tsx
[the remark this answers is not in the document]
mattrussell36: I went ahead and moved the subscription (and use-order-submit hook) into the trading app so we get type safety on the subscription

libs/ui-toolkit/src/components/dialog/index.tsx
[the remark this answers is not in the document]
sam-keen: It's not obvious why you'd need to do calc(50%-260px). I.e. why it's necessary, whether 260px is a special value etc. So perhaps you could pull these types of values into a var, or add a comment.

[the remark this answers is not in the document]
mattrussell36: Okay added a couple comments here 👍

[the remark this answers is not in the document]
sam-keen: Nice one :)

libs/deal-ticket/src/deal-ticket.tsx
```diff
@@ -0,0 +1,104 @@
...
+      };
+    };
+  };
+  tradingMode:
+    | 'BatchAuction'
+    | 'Continuous'
+    | 'MonitoringAuction'
+    | 'OpeningAuction';
+  state:
+    | 'Active'
```
dexturr, line 37: can we not use an enum or type for this?

mattrussell36: The problem is the enums are generated in apps/trading and shouldn't be imported into the the deal ticket package. We need them in a shared package

mattrussell36: I've gone ahead and added a types lib

libs/ui-toolkit/src/components/dialog/dialog.tsx
```diff
@@ -0,0 +1,52 @@
...
+}: DialogProps) {
+  const contentClasses = classNames(
+    // Positions the modal in the center of screen
+    'fixed w-[520px] px-28 py-24 top-[50%] left-[50%] translate-x-[-50%] translate-y-[-50%]',
+    // Need to apply background and text colors again as content is rendered in a portal
+    'dark:bg-black dark:text-white-95 bg-white text-black-95',
+    // For some reason if I use getIntentShadow from utils/intent the styles arent applied
+    'shadow-callout',
+    {
+      'shadow-intent-danger': intent === Intent.Danger,
```
dexturr, line 30: Can we use the intent util for this?

libs/deal-ticket/src/submit-button.tsx
```diff
@@ -0,0 +1,77 @@
...
+      return 'Market is no longer active';
+    }
+
+    if (market.tradingMode !== 'Continuous') {
+      if (order.type === OrderType.Market) {
+        return 'Only limit orders are permitted when market is in auction';
+      }
+
+      if (
+        order.timeInForce === OrderTimeInForce.FOK ||
```
bglownia, line 49: It will start work if we will add another OrderTimeInForce type
```
if (![OrderTimeInForce.GTT, OrderTimeInForce.GTC, OrderTimeInForce.GFA].includes(order.timeInForce))
```

libs/deal-ticket/src/submit-button.tsx
```diff
@@ -0,0 +1,77 @@
...
+    <>
+      <Button
+        className="w-full"
+        variant="primary"
+        type="submit"
+        disabled={disabled}
+      >
+        {transactionStatus === 'pending' ? 'Pending...' : 'Place order'}
+      </Button>
+      {invalidText && (
```
bglownia, line 72: We got <InputError /> component for this

libs/deal-ticket/src/submit-button.tsx
```diff
@@ -0,0 +1,77 @@
...
+        <p className="text-intent-danger text-center my-4">{invalidText}</p>
```
bglownia, line 73: We should use single-direction margin to make margin collapse work
IMHO it should look like this to 
```
<Button className="... mb-8"/>
{invalidText && <InputError className="mb-8"}
```

libs/ui-toolkit/src/components/dialog/dialog.tsx
bglownia: Or even more. We should use `<Callout>` component

package.json
```diff
@@ -26,9 +26,12 @@
     "ag-grid-react": "^27.0.1",
     "apollo": "^2.33.9",
     "autoprefixer": "^10.4.2",
+    "bignumber.js": "^9.0.2",
     "classnames": "^2.3.1",
     "core-js": "^3.6.5",
+    "date-fns": "^2.28.0",
```
bglownia, line 32: I think we can leave without date-fns and just use INTL format date

bglownia (approved):

libs/ui-toolkit/src/components/dialog/dialog.tsx
mattrussell36: Switched to using intent shadow helper (which is now working)

libs/deal-ticket/src/expiry-selector.tsx
```diff
@@ -0,0 +1,27 @@
...
+interface ExpirySelectorProps {
+  order: Order;
+  onSelect: (expiration: Date | null) => void;
+}
+
+export const ExpirySelector = ({ order, onSelect }: ExpirySelectorProps) => {
+  const date = order.expiration ? new Date(order.expiration) : new Date();
+  const dateFormatted = formatForInput(date);
+  const minDate = formatForInput(date);
+  console.log(dateFormatted, minDate);
```
dexturr, line 14:
```suggestion
```

dexturr (approved):

libs/deal-ticket/src/expiry-selector.tsx
mattrussell36: oopsie

PR #85, opened 2022-03-15
Feat/40 Styles for transaction and transaction details page

sam-keen:
Closes #40 

Styling for txs and txs:id routes

Screenshot 2022-03-16 at 15 41 31

Screenshot 2022-03-16 at 15 41 42

PR #89, opened 2022-03-16
Feat/84 Order list

mattrussell36:
Orderlist table first pass

- Adds order list container component
- Adds use-orders hook
- Updates useApplyGridTransaction hook to only update a row if its data changed
- Adds order-list package


apps/trading/components/order-list-container/order-list-container.spec.tsx
```diff
@@ -0,0 +1,42 @@
+import '@testing-library/jest-dom';
```
dexturr, line 1: chore (blocking): Don't like having to import this is every test file.

Can you add a setupsTests.ts file and call it from the jest config using the `setupAfterEnv` config property.

I think @sam-keen has a PR that does exactly this in #85.

I think there is another package that needs this as well

libs/graphql/apollo.config.js
```diff
@@ -2,7 +2,7 @@ module.exports = {
   client: {
     service: {
       name: 'vega',
-      url: process.env.NX_VEGA_URL,
+      url: process.env.NX_VEGA_URL || 'https://lb.testnet.vega.xyz/query',
```
dexturr, line 5: IMO if we want a default we should provide it in a comitted .env rather than just falling back

apps/trading/components/order-list-container/order-list-container.tsx
```diff
@@ -0,0 +1,17 @@
...
+  if (error) {
+    return <Splash>Something went wrong: {error.message}</Splash>;
+  }
+
+  if (loading) {
+    return <Splash>Loading...</Splash>;
+  }
+
+  return <OrderList orders={orders} />;
+};
```
dexturr, line 17: Can this be more generic? Async loader component that lives in UI toolkit that takes loading/error/children and just renders out the correct states rather than tying it directly to the order list container

apps/trading/hooks/use-orders.ts
```diff
@@ -0,0 +1,149 @@
...
+  }
+`;
+
+interface UseOrders {
+  orders: OrderFields[];
+  error: Error | null;
+  loading: boolean;
+}
+
+export const useOrdersImpl = () => {
```
dexturr, line 69: What does Impl mean? can this just be useOrders?

libs/order-list/src/order-list.spec.tsx
```diff
@@ -0,0 +1,137 @@
...
+  );
+  expect(cells[1]).toHaveTextContent('+10');
+  expect(cells[2]).toHaveTextContent(marketOrder.type);
+  expect(cells[3]).toHaveTextContent(marketOrder.status);
+  expect(cells[4]).toHaveTextContent('5');
+  expect(cells[5]).toHaveTextContent('-');
+  expect(cells[6]).toHaveTextContent(marketOrder.timeInForce);
+  expect(cells[7]).toHaveTextContent(
+    getDateTimeFormat().format(new Date(marketOrder.createdAt))
+  );
```
dexturr, line 96: optional (non-blocking): tidier

```
const expectedValues = ['+10', ...., getDateTimeFormat().format(new Date(marketOrder.createdAt))]
cells.forEach((cell, i) => expect(cell).toHaveTextContent(expected[i]))
```

dexturr: Again below

package.json
```diff
@@ -21,6 +21,7 @@
     "@radix-ui/react-tabs": "^0.1.5",
     "@sentry/react": "^6.18.1",
     "@sentry/tracing": "^6.18.1",
+    "@types/lodash.isequal": "^4.5.5",
```
dexturr, line 24: minor (non-blocking): Can we just add @types/loadash instead of needing for each function. As types are stripped at compile time we will have no worries about bundle bloat for this.

dexturr: Others below that could be removed

dexturr: Or we can move to using the whole of lodash, as long as we import correctly it gets tree shaken

dexturr (changes requested):
Few non-blocking things that would be nice and one thing I'd like to get sorted for the test setup that won't take long 👍 

Mostly LGTM

apps/trading/hooks/use-orders.ts
mattrussell36: Impl here just means 'implementation'. This hook then gets passed to the `singletonHook` hook function whose result is named `useOrders` (see bottom of the file). This is so that there is only one instance of the hook and therefore one subscription started no matter how many times `useOrders` is used.

Also, `useOrdersImpl` is exported for testing purposes

libs/order-list/src/order-list.spec.tsx
```diff
@@ -0,0 +1,136 @@
+import '@testing-library/jest-dom';
```
dexturr, line 1:
```suggestion
```

package.json
```diff
@@ -30,13 +31,12 @@
...
     "core-js": "^3.6.5",
+    "date-fns": "^2.28.0",
     "env-cmd": "^10.1.0",
     "ethers": "^5.6.0",
     "graphql": "^15.7.2",
     "graphql-ws": "^5.6.3",
-    "lodash.debounce": "^4.0.8",
-    "lodash.memoize": "^4.1.2",
-    "lodash.once": "^4.1.1",
+    "lodash": "^4.17.21",
```
dexturr, line 39: follow up (non-blocking): now we are importing the whole of lodash we should add that linting rule that prevents developers from importing the whole thing

dexturr (approved):

package.json
dexturr: #115

apps/trading/specs/index.spec.tsx
```diff
@@ -1,26 +1,18 @@
...
-          selectPublicKey: jest.fn(),
-          connector: null,
-          sendTx: jest.fn(),
-        }}
-      >
-        <Index />
-      </VegaWalletContext.Provider>
-    );
+    const { baseElement } = render(<Index />);
     expect(baseElement).toBeTruthy();
```
sam-keen, line 16: As per https://vegaprotocol.slack.com/archives/C8WS7M50E/p1647868686724769 we should move to use `screen` where possible. As long as `<Index />` can accept props then I think this should work nicely:

```suggestion
    render(<Index data-testid='render-test' />);
    expect(screen.getByTestId('render-test')).toBeInTheDocument();
```

libs/order-list/src/order-list.tsx
```diff
@@ -0,0 +1,101 @@
...
+        headerName="Amount"
+        field="size"
+        valueFormatter={({ value, data }: ValueFormatterParams) => {
+          let prefix = '';
+          if (data.side === 'Buy') {
+            prefix = '+';
+          } else if (data.side === 'Sell') {
+            prefix = '-';
+          }
+          return prefix + value;
```
sam-keen, line 52: Presumably `data.side` can only be 'Buy' or 'Sell'?

```suggestion
          return `data.side === 'Buy' ? "+" : "-"` + value;
```

sam-keen (approved):
LGTM, one question and couple of minor suggestions

apps/trading/hooks/use-orders.ts
```diff
@@ -0,0 +1,149 @@
...
+
+      setLoading(true);
+
+      try {
+        const res = await client.query<Orders, OrdersVariables>({
+          query: ORDERS_QUERY,
+          variables: { partyId: keypair.pub },
+        });
+
+        if (!res.data.party?.orders.length) return;
```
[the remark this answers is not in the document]
mattrussell36: No, its likely that this condition will never. However, this guards against an update being emitted with an empty array in which case we can just skip the upate

PR #91, opened 2022-03-17
Test/block page

### LINKED ISSUE #92: Fix and add to existing block explorer tests with the added styling to blocks

jtsang586:
- Added a blocks page separate from transactions with test
- Added transaction details test
- Corrected network param test

Closes #92


dexturr (approved):
Looks reasonable, but tests are failing. Let me know if you need me to take a look

apps/explorer/src/app/components/blocks/block-data.tsx
```diff
@@ -16,7 +16,7 @@ export const BlockData = ({ block, className }: BlockProps) => {
       aria-label={`Data for block ${block.header?.height}`}
       className={className}
     >
-      <TableRow data-testid="block-row" modifier="background">
+      <TableRow data-testid="block-list" modifier="background">
```
mattrussell36, line 19: I feel like the testid naming is a bit off here. This element should surely should be labelled as 'block-row'. Likewise in blocks-data.tsx block-table is being applied to a list item element? Seems to be the wrong way around to me?

mattrussell36 (changes requested):

apps/explorer/src/app/components/blocks/block-data.tsx
jtsang586: Yeah I agree with this. Done

PR #93, opened 2022-03-17
Feature/29 market list table

bglownia:


libs/react-helpers/src/lib/grid-cells/price-cell.tsx
```diff
@@ -17,25 +17,24 @@ export const PriceCell = ({
...
+        <span data-testid="postfix">{' ' + postfix.trim()}</span>
       )}
     </span>
-  )
-}
+  );
+};
 
-PriceCell.displayName = 'PriceCell'
+PriceCell.displayName = 'PriceCell';
```
dexturr, line 19: question (non-blocking): should we be enforcing this for all components?

libs/react-helpers/src/lib/grid-cells/flash-cell.tsx
```diff
@@ -8,84 +8,84 @@
...
 }
 
 /**
-* Get value from previous render
-*/
+ * Get value from previous render
+ */
 function usePrevious<T>(value: T): T | undefined {
-  const ref = React.useRef<T | undefined>()
+  const ref = React.useRef<T | undefined>();
```
dexturr: chore (non-clocking): as a follow up I think we should enable full project linting on CI, noticing a lot of files getting in with little linting errors and not sure how

libs/react-helpers/src/lib/grid-cells/price-cell.tsx
```diff
@@ -17,25 +17,24 @@ export const PriceCell = ({
   decimalPlaces = 0,
   value,
   className = '',
-  invalid = false
+  invalid = false,
 }: IPriceCellProps) => {
-  if (!value || isNaN(Number(value)) || invalid) return <span>-</span>
-  const formattedNumber = getNumberFormat(decimalPlaces).format(value)
+  if (!value || isNaN(Number(value)) || invalid) return <span>-</span>;
```
dexturr: minor (non-blocking): edge case here of `value = 0` will return "-" when I think we want it to return a formatted price.

dexturr (approved):
Minor things, nothing blocking LGTM

dexturr:
Diff looks funky, not sure what's happeneing here but appears to have all the deal ticket changes in it as well? Maybe just needs a rebase

libs/react-helpers/src/lib/grid-cells/flash-cell.tsx
bglownia: Agree. There is precommit hook but it's possible that some code changes will sneak to master unnoticed

libs/react-helpers/src/lib/grid-cells/price-cell.tsx
bglownia: Good point, I will add it to post merge fix list

PR #96, opened 2022-03-17
Fix/Set initially selected pubkey after connect

mattrussell36:
Fixes a but where if you didn't have a selected pubkey in local storage you would be left in an unconnected state even after connecting.

### DIFF
--- a/libs/wallet/src/provider.tsx
+++ b/libs/wallet/src/provider.tsx
@@ -56,7 +56,6 @@ export const VegaWalletProvider = ({ children }: VegaWalletProviderProps) => {
 
         return publicKeysWithName;
       } catch (err) {
-        console.error(err);
         return null;
       }
     },
--- a/libs/wallet/src/provider.tsx
+++ b/libs/wallet/src/provider.tsx
@@ -30,29 +30,38 @@ export const VegaWalletProvider = ({ children }: VegaWalletProviderProps) => {
   // Reference to the current connector instance
   const connector = useRef<VegaConnector | null>(null);
 
-  const connect = useCallback(async (c: VegaConnector) => {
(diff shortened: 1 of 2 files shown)


libs/wallet/src/provider.tsx
```diff
@@ -30,29 +30,38 @@ export const VegaWalletProvider = ({ children }: VegaWalletProviderProps) => {
...
+
+        if (!res) {
+          return null;
+        }
+
+        const publicKeysWithName = res.map((pk) => {
+          const nameMeta = pk.meta?.find((m) => m.key === 'name');
+          return {
+            ...pk,
+            name: nameMeta?.value ? nameMeta.value : 'None',
```
dexturr, line 47: chore (non-blocking): should show truncated key name if no name is set

libs/wallet/src/provider.tsx
```diff
@@ -30,29 +30,38 @@ export const VegaWalletProvider = ({ children }: VegaWalletProviderProps) => {
...
+
+        setKeypairs(publicKeysWithName);
+
+        if (publicKey === null) {
+          setPublicKey(publicKeysWithName[0].pub);
+        }
+
+        return publicKeysWithName;
+      } catch (err) {
+        console.error(err);
```
dexturr, line 59:
```suggestion
```

dexturr (approved):

libs/wallet/src/provider.tsx
mattrussell36: I'm not sure on this one. I feel like name should be left null otherwise it could lead to some inconsistent looking UIs if you were to say loop over all keys and render out the name?

PR #107, opened 2022-03-21
106 - migrate mainnet stats to monorepo

### LINKED ISSUE #106: Migrate mainnet-stats into monorepo

sam-keen:
Closes #106


apps/stats-mainnet/src/styles/styles.css
```diff
@@ -0,0 +1,31 @@
+@tailwind base;
+@tailwind components;
+
+@font-face {
+  font-family: 'AlphaLyrae-Medium';
+  src: url('./AlphaLyrae-Medium.woff');
```
dexturr, line 6: follow up (non-blocking): we need to add alpha lyrae support to the UI toolkit

apps/stats-mainnet/src/index.html
```diff
@@ -0,0 +1,52 @@
...
+      // Set dark mode preference indicator here to avoid FOUC
+      if (
+        localStorage.theme === 'dark' ||
+        (!('theme' in localStorage) &&
+          window.matchMedia('(prefers-color-scheme: dark)').matches)
+      ) {
+        document.documentElement.classList.add('dark-mode-preferred');
+      } else {
+        document.documentElement.classList.remove('dark-mode-preferred');
+      }
```
dexturr, line 35: chore (blocking): A hook now exists for this

apps/stats-mainnet/src/components/images/light-mode-toggle.tsx
```diff
@@ -0,0 +1,10 @@
+export const LightModeToggle = () => {
```
dexturr, line 1: chore (blocking): Use UI toolkit switcher that now exists

apps/stats-mainnet-e2e/cypress.json
```diff
@@ -0,0 +1,12 @@
+{
```
dexturr, line 1: chore (blocking): make cypress tests run on a port that is not going to be used by other tests. Currently in use:
* 3000
* 4200

dexturr: You can compare this to either BE or tradiung, one of them is using non-default ports for the config

apps/stats-mainnet-e2e/cypress.json
```diff
@@ -0,0 +1,12 @@
...
+  "fileServerFolder": ".",
```
dexturr, line 2: follow up (non-blocking): let's create a ticket to add some cypress tests for this

apps/stats-mainnet/src/styles/styles.css
sam-keen: Yes indeedy. Bart and I were talking about that this morning, we're going to find a way to have it accessible for the whole monorepo

apps/stats-mainnet/.browserslistrc
```diff
@@ -0,0 +1,16 @@
+# This file is used by:
+# 1. autoprefixer to adjust CSS to support the below specified browsers
+# 2. babel preset-env to adjust included polyfills
+#
+# For additional information regarding the format and rule options, please see:
+# https://github.com/browserslist/browserslist#queries
+#
+# If you need to support different browsers in production, you may tweak the list below.
+
+last 1 Chrome version
```
dexturr, line 10: follow up (non-blocking): need a meeting to decide our browser support. THis does not include brave, which I would consider probably in our list of supported browsers. Probably will still work as it's backed onto Chrome V8 but something we should nail down as a team

apps/stats-mainnet/src/index.html
sam-keen: Does the hook avoid a FOUC? I did this intentionally at this early stage

apps/stats-mainnet/src/config/types.ts
```diff
@@ -0,0 +1,41 @@
...
+  blockDuration: string;
+  status: string;
+  vegaTime: string;
+  appVersion: string;
+  chainVersion: string;
+  uptime: string;
+  chainId: string;
+}
+
+export type value = any;
```
dexturr, line 21: question (non-blocking): can we do:
```suggestion
export type value = string | number | null | undefined;
```

apps/stats-mainnet-e2e/cypress.json
dexturr: #109

apps/stats-mainnet/src/index.html
dexturr: It does not avoid a FOUC, but it should be up to the hook to fix it. 

Happy to ticket this as a follow up in that case if fixing the hook would be a lot of work and we don't want a FOUC right now. I think @campbellssource noticed this when we were looking at the Trading UI on Friday?

apps/stats-mainnet/src/styles/styles.css
dexturr: #110

apps/stats-mainnet/src/index.html
sam-keen: Sounds good, I'd like to get this on block explorer asap

apps/stats-mainnet/src/components/images/light-mode-toggle.tsx
sam-keen: This isn't currently compatible with the way I've set up the dark mode, will deal with this as part of #113

apps/stats-mainnet/.browserslistrc
sam-keen: Good shout

libs/ui-toolkit/src/components/vega-logo/vega-logo.spec.tsx
```diff
@@ -0,0 +1,10 @@
+import { render } from '@testing-library/react';
```
dexturr, line 1: Use screen ;) MWAHAHAHA

dexturr (approved):

libs/mainnet-stats-manager/src/components/good-threshold-indicator/good-threshold-indicator.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export const GoodThresholdIndicator = ({
+  goodThreshold,
+  value,
+}: GoodThresholdIndicatorProps) => {
+  return (
+    <div
+      className={`inline-block w-8 h-8 mb-[0.15rem] mr-8 rounded ${
+        (goodThreshold &&
+          (goodThreshold(value) ? 'bg-intent-success' : 'bg-intent-danger')) ||
```
mattrussell36, line 16: I don't see these classes used or imported anywhere else?

mattrussell36 (approved):
LGTM. I'd prefer to not tie the stats app and lib to mainnet by the name stats-mainnet and mainnet-stats-manager. Can this be changed so that we can specify the environment using .env. Happy for that to be a follow up though.

libs/mainnet-stats-manager/src/components/good-threshold-indicator/good-threshold-indicator.tsx
sam-keen: They're in the theme, I think this is the intended use for them

PR #114, opened 2022-03-21
Feature/29 market list table improvments

bglownia:


apps/trading/.eslintrc.json
```diff
@@ -19,7 +19,15 @@
     },
     {
       "files": ["*.js", "*.jsx"],
-      "rules": {}
+      "rules": {
+        "unicorn/filename-case": [
+          "error",
+          {
+            "case": "kebabCase",
+            "ignore": ["react-singleton-hook/**/*.js"]
```
dexturr, line 27: chore (blocking): This doesn't support globs 😢  https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/rules/filename-case.js#L155-L159

package.json
```diff
@@ -54,6 +53,7 @@
     "uuid": "^8.3.2"
   },
   "devDependencies": {
+    "@apollo/react-testing": "^4.0.0",
```
dexturr, line 56: Well spotted

apps/trading/pages/markets/index.page.tsx
```diff
@@ -26,4 +26,6 @@ const Markets = () => {
   );
 };
 
-export default Markets;
+const TwoMarkets = () => (<><div style={{height: '50%'}}><Markets /></div><div style={{height: '50%'}}><Markets /></div></>)
```
mattrussell36, line 29: chore (blocking): Whats going on here? Looks like some temp test code?

libs/react-helpers/src/lib/grid-cells/price-cell.tsx
```diff
@@ -6,12 +6,11 @@ export interface IPriceCellProps {
 }
 
 export const PriceCell = ({ value, valueFormatted }: IPriceCellProps) => {
-  if (!value || isNaN(Number(value))) return <span>-</span>;
+  if ((!value && value !== 0) || isNaN(Number(value)))
```
mattrussell36, line 9: Another code style/consistency question here 😉, but I'd prefer to always have curly braces and not rely in JS indentation

apps/trading/specs/index.spec.tsx
```diff
@@ -1,26 +1,29 @@
...
-          selectPublicKey: jest.fn(),
-          connector: null,
-          sendTx: jest.fn(),
-        }}
-      >
-        <Index />
-      </VegaWalletContext.Provider>
-    );
-    expect(baseElement).toBeTruthy();
+  it('should render successfully', async () => {
```
mattrussell36, line 9: chore (blocking): We can remove the mocked VegaWalletContext.Provider wrapping component, its not required for the test anymore

mattrussell36 (changes requested):

apps/trading/pages/markets/index.page.tsx
```diff
@@ -1,17 +1,78 @@
...
+
+            if (rowNode) {
+              const updatedData = produce(
+                rowNode.data.data,
+                (draft: Markets_markets_data) => assign(draft, delta)
+              );
+              if (updatedData !== rowNode.data.data) {
+                update.push({ ...rowNode.data, data: delta });
+              }
+            } /* else {
```
mattrussell36, line 54: Can we remove the commented out code

apps/trading/pages/index.page.tsx
```diff
@@ -4,14 +4,68 @@ import {
...
+const Grid = () => {
   const rowData = [
     { make: 'Toyota', model: 'Celica', price: 35000 },
     { make: 'Ford', model: 'Mondeo', price: 32000 },
     { make: 'Porsche', model: 'Boxter', price: 72000 },
   ];
+  const ref = useRef(rowData);
+  const getRowNodeId = (data: { make: string }) => data.make;
+  const gridApi = useRef<GridApi | null>(null);
+  useEffect(() => {
```
mattrussell36, line 21: Can we clean this up?

mattrussell36 (changes requested):

libs/graphql/src/index.ts
```diff
@@ -14,3 +14,5 @@ export * from './__generated__/Orders';
 export * from './__generated__/OrderSub';
 export * from './__generated__/PartyAssetsQuery';
 export * from './__generated__/ProposalsQuery';
+
+export * from './data-providers';
```
edd, line 18: **thought (non-blocking)**:  I've finally got to answering in a slack thread about data-providers as an abstraction and I'm not a fan - I think the queries are better as close to the components that require the data as possible (https://www.apollographql.com/docs/react/data/operation-best-practices/#query-only-the-data-you-need-where-you-need-it). I'm going to mark this as non-blocking because the contents of this file can still be moved elsewhere before we go too far down the path of generic providers without everyone agreeing/agreeing that the correct balances are

apps/trading/pages/markets/index.page.tsx
```diff
@@ -1,17 +1,76 @@
...
+            if (rowNode) {
+              const updatedData = produce(
+                rowNode.data.data,
+                (draft: Markets_markets_data) => assign(draft, delta)
+              );
+              if (updatedData !== rowNode.data.data) {
+                update.push({ ...rowNode.data, data: delta });
+              }
+            }
+            // @TODO - else add new market
```
edd, line 55: **nitpick (non-blocking)**: Leftover todo

libs/react-helpers/src/lib/grid-cells/price-cell.test.tsx
```diff
@@ -0,0 +1,21 @@
+import { render, screen } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+import * as React from 'react';
+
+import { PriceCell } from './price-cell';
+
+describe('<PriceCell />', () => {
+  it('Displayes formatted value', () => {
```
edd, line 8: **nitpick (non-blocking)**

```suggestion
  it('Displays formatted value', () => {
```

edd (approved):
I've left some feedback, none of it blocking. As per the commento nt he data-providers definition, I think it could lead us to move queries to the wrong place, but code-wise the provider looks solid. My opinion counts for the least here, so I think it's mergeable as is - and then if we move where the queries/data munging happens the code can just migrate to the best place. 

Otherwise, 👍 . Happy to see some simple unit tests too.

apps/trading/pages/markets/index.page.tsx
mattrussell36: We discussed this one as a potential follow up due to how the markets data subscription is a nested part of the markets query which complicates things..

mattrussell36 (approved):

PR #132, opened 2022-03-24
frontend-monorepo-127 Generalize stats page to work for all environments

### LINKED ISSUE #127: Generalize stats page to work for all environments

sam-keen:
Closes #127


libs/network-stats/src/components/stats-manager/stats-manager.tsx
```diff
@@ -14,16 +14,81 @@ interface statsManagerProps {
...
+    title: 'Stagnet 1',
+    stats: 'https://n03.s.vega.xyz/datanode/rest/statistics',
+    nodes: 'https://n03.s.vega.xyz/datanode/rest/nodes-data',
+  },
+  stagnet2: {
+    title: 'Stagnet 2',
+    stats: 'https://n01.stagnet2.vega.xyz/datanode/rest/statistics',
+    nodes: 'https://n01.stagnet2.vega.xyz/datanode/rest/nodes-data',
+  },
+};
```
dexturr, line 49: chore (blocking): can't hard code this. We can add environments or remove them at any point, we don't want to have to deploy the frontend if an evironment is changed. You should use environment variables here, one fo the REST url (bear in mind for mainnet this could be anything) and a network (which also could be anything).

The API paths are consistent so you need:
```
const title = process.env['NX_VEGA_ENV']
const endpoint = process.env['NX_VEGA_REST']
return {
  title,
  stats: new URL(endpoint, '/statistics').href
  nodes: new URL(endpoint, '/nodes-data').href
}
```

Then adjust the different environment files

libs/network-stats/src/components/stats-manager/stats-manager.tsx
```diff
@@ -14,16 +14,81 @@ interface statsManagerProps {
...
+      return dataSources.stagnet1;
+    }
+
+    if (dataNodeUrl.includes('stagnet2')) {
+      return dataSources.stagnet2;
+    }
+  }
+
+  return dataSources.mainnet;
+};
```
dexturr, line 77: chore (blocking): we cannot detect network like this. Use the environement variable as described above with a map and a default of "unknown"

libs/network-stats/src/components/stats-manager/stats-manager.tsx
```diff
@@ -14,16 +14,81 @@ interface statsManagerProps {
...
+
 export const StatsManager = ({ className }: statsManagerProps) => {
+  const dataSource = useRef<dataSourceProps>(getDataSource());
```
dexturr, line 80: chore (blocking):
```suggestion
  const dataSource = useState<dataSourceProps>(getDataSource());
```

libs/network-stats/src/components/stats-manager/stats-manager.tsx
```diff
@@ -14,16 +14,81 @@ interface statsManagerProps {
...
   const [data, setData] = useState<IStructuredStats | null>(null);
   const [error, setError] = useState<Error | null>(null);
 
   useEffect(() => {
+    getDataSource();
```
dexturr, line 85: queston (blocking): why is this needed?

libs/network-stats/src/components/stats-manager/stats-manager.tsx
dexturr: Also this component shouldn't care about vega environments ect. hoist this logic to the consumer, take these variables as props to the components I think makes the most sense.

libs/network-stats/src/components/stats-manager/stats-manager.tsx
```diff
@@ -14,16 +14,81 @@ interface statsManagerProps {
   className?: string;
 }
 
+interface dataSourceProps {
```
dexturr, line 17: chore (blocking): interfaces should e PascalCase

apps/explorer/src/app/routes/home/index.tsx
```diff
@@ -1,4 +1,4 @@
-import { StatsManager } from '@vegaprotocol/mainnet-stats-manager';
+import { StatsManager } from '@vegaprotocol/network-stats';
```
dexturr, line 1: question (blocking): why is the package called network-stats when directory is called stats?

dexturr (changes requested):
The environment detection logic is not suitable

libs/network-stats/src/components/stats-manager/stats-manager.tsx
sam-keen: Ah that's an error - that was my first attempt and should have been moved when I switched to getting the data in the `useRef`

apps/explorer/src/app/routes/home/index.tsx
sam-keen: urgh - naming is hard! I considered calling the package the same thing as the app, because the app is essentially just a wrapper around the component. But that didn't feel like a great idea, so opted for something that was different but the same, if that makes sense :)

libs/network-stats/src/components/stats-manager/stats-manager.tsx
sam-keen: Great points, thanks

dexturr (approved):

bglownia (approved):

edd (approved):
All looks fine to me, but I am new to NX. Luckily @bglownia has approved already!

PR #143, opened 2022-03-25
Feat/103 deposits

mattrussell36:
- Adds deposits page and deposits lib
- Updates web3-container to fetch the ethereum config from network parameters.
- Adds hooks to react-helpers for handling contracts and ethereum transactions
- Adds a transaction-dialog to ui-toolkit
- Removes anything web3 related from ui-toolkit and refactors the etherscan link component to use a number for chainId


apps/trading/.env
```diff
@@ -19,6 +19,7 @@ NX_DEPLOY_URL=$DEPLOY_URL
 NX_DEPLOY_PRIME_URL=$DEPLOY_PRIME_URL
 
 # App configuration variables
+NX_VEGA_ENV = "TESTNET"
 NX_VEGA_URL = "https://lb.testnet.vega.xyz/query"
 NX_ETHEREUM_CHAIN_ID = 3
-NX_INFURA_ID = "4f846e79e13f44d1b51bbd7ed9edefb8"
\ No newline at end of file
+NX_INFURA_ID = "4f846e79e13f44d1b51bbd7ed9edefb8"
```
edd, line 25: **praise (non-blocking)**:  This made me go and check that it doesn't need to be a secret. It doesn't! Just to save anyone else going through the same check. @mattrussell36 knows best.

dexturr, line 25: chore (blocking): let's not tie ourselves explicitly to infura. Can we make this NX_ETH_PROVIDER_URL

dexturr: Reason being is that once we move to capsule this will be a local running node and so infura won't know anything about it

apps/trading/pages/portfolio/deposit/deposit-container.tsx
```diff
@@ -0,0 +1,60 @@
...
+            </Splash>
+          );
+        }
+
+        return (
+          <DepositManager
+            bridgeAddress={ethereumConfig.collateral_bridge_contract.address}
+            requiredConfirmations={ethereumConfig.confirmations}
+            // @ts-ignore TS not inferring on union type for contract address
+            assets={data.assets.filter((a) => a.source.__typename === 'ERC20')}
```
dexturr, line 53: nit[ick (non-blocking): id there a better way to do this for example an `as` or similar so we can avoid the TS ignore

libs/deposits/src/lib/use-allowance.ts
```diff
@@ -0,0 +1,28 @@
...
+      }
+
+      const res = await contract.allowance(account, bridgeAddress);
+
+      setAllowance(res);
+    };
+
+    run();
+  }, [contract, account, bridgeAddress]);
+
```
dexturr, line 26: chore (non-blocking): Will cause unmounted errors in race condition cases (same for all of below) should we have a `use-async` primitive or similar to prevent unmounted errors (see use fetch hook in block explorer for example. Also use fetch should move into react helpers)

libs/react-helpers/src/hooks/use-ethereum-transaction.test.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+  expect(result.current.status).toEqual(TxState.Requested);
+  expect(result.current.confirmations).toBe(0);
+
+  await waitFor(() => {
+    expect(result.current.status).toEqual(TxState.Pending);
+    expect(result.current.txHash).toEqual(MockContract.txHash);
+  });
+
+  await act(async () => {
```
dexturr, line 86: praise (non-blocking): nice tests

libs/react-helpers/src/hooks/use-ethereum-transaction.ts
```diff
@@ -0,0 +1,88 @@
...
+  return { perform, status, error, confirmations, txHash };
+};
+
+export class EthereumError extends Error {
+  code: number;
+
+  constructor(message: string, code: number) {
+    super(message);
+    this.code = code;
+    this.name = 'EthereumError';
```
dexturr, line 79: minor (non-blocking): feild set that is not declared in type

libs/react-helpers/src/hooks/use-ethereum-transaction.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+}
+
+export function isEthereumError(err: unknown): err is EthereumError {
```
dexturr, line 83: follow up (non-blocking): let's add a lint rule to be opionated about function statements/expressions. I don't care what it is as long as --fix works but it would be nice to be consistent.

libs/ui-toolkit/src/components/dialog/dialog.tsx
```diff
@@ -25,14 +24,25 @@ export function Dialog({
     'fixed w-[520px] px-28 py-24 top-[50%] left-[50%] translate-x-[-50%] translate-y-[-50%]',
     // Need to apply background and text colors again as content is rendered in a portal
     'dark:bg-black dark:text-white-95 bg-white text-black-95',
-    getIntentShadow(intent)
+    'shadow-callout',
```
dexturr, line 27: minor (non-blocking): I managed to get utility functions to work in BE. We can still use this if you prefer, you can check out #187 for the eventual implementation. I think the key was getting rid of the classnames function IIRC

libs/ui-toolkit/src/components/etherscan-link/etherscan-link.tsx
```diff
@@ -1,16 +1,15 @@
...
-const etherscanUrls: Record<EthereumChainId, string> = {
-  '0x1': 'https://etherscan.io',
-  '0x3': 'https://ropsten.etherscan.io',
-  '0x4': 'https://rinkeby.etherscan.io',
-  '0x5': 'https://goerli.etherscan.io',
-  '0x2a': 'https://kovan.etherscan.io',
+const etherscanUrls: Record<number, string> = {
+  1: 'https://etherscan.io',
+  3: 'https://ropsten.etherscan.io',
 };
```
dexturr, line 8: chore (non-blocking): should these be environment variables. General idea being the app can run against any chain without code changes, for capsule ect. However this does mean if we wanted an env switcher we would need multiple. Possibly a point for discussion, but I think our app is too ethereum environement aware atm

libs/ui-toolkit/src/components/input-error/input-error.tsx
```diff
@@ -39,6 +39,7 @@ export const InputError = ({
       data-testid="input-error-text"
       className={effectiveClassName}
       {...props}
+      role="alert"
```
dexturr, line 42: follow up (non-blocking): we should we add eslint react a11y plugin.

libs/ui-toolkit/src/components/transaction-dialog/transaction-dialog.tsx
```diff
@@ -0,0 +1,129 @@
...
+export interface TransactionDialogProps<TEvent = undefined> {
+  name: string;
+  status: TxState;
+  error: Error | null;
+  confirmations: number;
+  txHash: string | null;
+  requiredConfirmations?: number;
+  // Undefined means this dialog isn't expecting an additional event for a complete state, null means
+  // it is but hasn't been received yet
+  confirmationEvent?: TEvent | null;
```
dexturr, line 20: chore (non-blocking): seems convoluted can we enum this?

libs/deposits/src/lib/deposit-manager.tsx
```diff
@@ -0,0 +1,102 @@
...
+  const bridgeContract = useBridgeContract();
+
+  // Get users balance of the erc20 token selected
+  const { balanceOf, refetch } = useBalanceOfERC20Token(tokenContract);
+
+  // Get temporary deposit limits
+  const limits = useDepositLimits(bridgeContract, asset);
+
+  // Get allowance (approved spending limit of brdige contract) for the selected asset
+  const allowance = useAllowance(tokenContract, bridgeAddress);
```
dexturr, line 61: chore (non-blocking): these two hooks are reading from the chain and the bottom 2 are writing to the chain. Could we have names that reflect that? Otherwise it looks like these are all the same and they aren't

libs/deposits/src/lib/deposit-manager.tsx
```diff
@@ -0,0 +1,102 @@
...
+  bridgeAddress,
+  assets,
+  initialAssetId,
+}: DepositManagerProps) => {
+  const [assetId, setAssetId] = useState<string | undefined>(initialAssetId);
+
+  // Find the asset object from the select box
+  const asset = useMemo(() => {
+    const asset = assets?.find((a) => a.id === assetId);
+    return asset;
```
dexturr, line 45: nitpick (non-blocking): throw error

dexturr (changes requested):
LGTm, only the infura thing is blocking IMO everything else is nice to have or improvements or discussions. Good work :D

libs/ui-toolkit/src/components/transaction-dialog/transaction-dialog.tsx
```diff
@@ -0,0 +1,129 @@
...
+      intent,
+    };
+  };
+
+  useEffect(() => {
+    // Close dialog if error is due to user rejecting the tx
+    if (
+      status === TxState.Error &&
+      isEthereumError(error) &&
+      error.code === 4001
```
dexturr, line 103: minor (non-blocking): would be nice to util this or move it to a local function

libs/ui-toolkit/src/components/transaction-dialog/transaction-dialog.tsx
```diff
@@ -0,0 +1,129 @@
...
+      icon = <Icon name="tick" />;
+      intent = Intent.Success;
+    }
+
+    return {
+      title,
+      icon,
+      intent,
+    };
+  };
```
dexturr, line 96: chore (non-blocking): if we do the above can this can just become a map

libs/react-helpers/src/hooks/use-ethereum-transaction.ts
mattrussell36: Its already declared on the Error constructor. If we want to declare it anyway we need an override modifier

libs/react-helpers/src/hooks/use-ethereum-transaction.ts
mattrussell36: I'd like the consistency too but I've found in some cases that you can't use generics using an arrow function. I'll try and find the specific example.. I may have just been doing something silly

libs/ui-toolkit/src/components/etherscan-link/etherscan-link.tsx
mattrussell36: Agree. I've added env vars for this for now but I also don't think its right for the ui toolkit to consume those env vars

libs/ui-toolkit/src/components/input-error/input-error.tsx
mattrussell36: https://github.com/vegaprotocol/frontend-monorepo/issues/210

libs/ui-toolkit/src/components/transaction-dialog/transaction-dialog.tsx
mattrussell36: I'm going to just update this prop to be a `confirmed` boolean. My initial thinking was that we might want to display data from the confirmation event in the dialog (but for all types of events, hence the generic) right now though its only used as essentially a boolean on whether or not to show the final 'confirming...' row

libs/ui-toolkit/src/components/transaction-dialog/transaction-dialog.tsx
mattrussell36: Not following you here. Can you elaborate?

libs/deposits/src/lib/use-allowance.ts
mattrussell36: I've added a use-ethereum-read-contract hook, which looks very similar to use-fetch except it takes a function as an argument rather than a url and of course doesn't do any response handling.

edd (approved):
Bit of a quick-shallow-skim review but it looks well compartmentalised

libs/ui-toolkit/src/components/transaction-dialog/transaction-dialog.tsx
mattrussell36: Sorted, some additional logic needed to handle vega confirmation state but no duplicated success states now

PR #145, opened 2022-03-25
Task/Updated tests to pass when there are no transactions

### LINKED ISSUE #144: Update Block Explorer test - Slack integration
Currently there are Cypress tests failing on block explorer due to 0 transactions currently being generated on testnet.

In order to reduce the amount of noise from the failing tests the tests should be changed to not fail but still show a message for visibility if there are no transactions on the pages, possibly using a slack hook.

jtsang586:
Updated tests to pass when there are no transactions.

Added logs and screenshots when this happens but unfortunately doesn't look like there is a way to get visibility on this from Slack integration

Closes #144


apps/explorer-e2e/src/support/pages/transactions-page.js
```diff
@@ -48,35 +43,51 @@ export default class TransactionsPage extends BasePage {
...
+          .should('not.be.empty')
+          .invoke('text')
+          .then((txTypeTxt) => {
+            if (txTypeTxt == 'Order Submission') {
+              cy.get('.hljs-attr')
+                .should('have.length.at.least', 8)
+                .each(($propertyName) => {
+                  cy.wrap($propertyName).should('not.be.empty');
+                });
+              cy.get('span[style*="color"]')
```
sam-keen, line 68: I'm a bit dubious of targeting an element with a style property, I expect this could break easily on a change to the UI theme designs

apps/explorer-e2e/src/support/pages/transactions-page.js
```diff
@@ -48,35 +43,51 @@ export default class TransactionsPage extends BasePage {
...
+      }
+    });
   }
 
   clickOnTopTransaction() {
-    cy.getByTestId(this.transactionRow).first().find('a').first().click();
+    cy.wait(1000);
+    cy.get('body').then(($body) => {
+      if ($body.find(`[data-testid=${this.transactionRow}]`).length) {
+        cy.getByTestId(this.transactionRow).first().find('a').first().click();
```
sam-keen, line 86: Similarly, designs might change the order of the items in the row, could this be targeted with a test-id instead?

apps/explorer-e2e/src/support/pages/transactions-page.js
jtsang586: I totally agree finding element by style is not best practice. A while ago I did see if I could add test ids to the json data pulled in for the tx details but I don't think it was as easy as the other elements. I'll create a separate issue for that.

apps/explorer-e2e/src/support/pages/transactions-page.js
jtsang586: The order doesn't matter too in this test. Purpose of the test is just to click on a random transaction and assert the details. In the future when we have capsule running the test will probably switch to clicking on a specific transaction when we have more control over the environment.

mattrussell36 (changes requested):
Looks like the package has been setup with an additional directory which is a bit different to the other apps/libs 

Right now its: `"cypress-cypress-commands": "libs/cypress/cypress/commands"` Can we regenerate the package so its just `"cypress": "libs/cypress"`

With a directory structure something more like:
```
- libs/
  - cypress/
    - commands/
```

mattrussell36 (changes requested):
Still dont think the directory structure is 100% right. IMO it should be:
```
- apps/
- libs/
  - order-list/
  ...
  - cypress/
    - .babelrc
    - project.json
    ...
    - src/
      - index.ts
      - lib/
        - commands/
          - get-by-test-id.ts
          - slack.ts
```

workspace.json
```diff
@@ -1,6 +1,7 @@
 {
   "version": 2,
   "projects": {
+    "cypress-commands": "libs/cypress/commands",
```
mattrussell36, line 4: lib should be called just `cypress`

tsconfig.base.json
```diff
@@ -15,6 +15,7 @@
     "skipDefaultLibCheck": true,
     "baseUrl": ".",
     "paths": {
+      "@vegaprotocol/cypress/commands": ["libs/cypress/commands/src/index.ts"],
```
mattrussell36, line 18: should just be `@vegaprotocol/cypress`

mattrussell36 (approved):

libs/cypress/src/lib/commands/slack.ts
```diff
@@ -0,0 +1,26 @@
+// eslint-disable-next-line @typescript-eslint/no-namespace
+declare namespace Cypress {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  interface Chainable<Subject> {
+    slack(message: string): void;
+  }
+}
+
+export function addSlackCommand() {
+  // @ts-ignore - ignoring Cypress type error which gets resolved when Cypress uses the command
```
dexturr, line 10: Same as above

apps/explorer-e2e/src/support/pages/blocks-page.js
```diff
@@ -25,10 +26,16 @@ export default class BlocksPage extends BasePage {
   validateBlockValidatorPage() {
     cy.getByTestId(this.minedByValidator).should('not.be.empty');
     cy.getByTestId(this.blockTime).should('not.be.empty');
-    cy.get(`[data-testid=${this.transactionsRow}] > td`)
-      .should('have.length.above', 0)
-      .each(($cell) => {
-        cy.wrap($cell).should('not.be.empty');
-      });
+    cy.wait(1000);
```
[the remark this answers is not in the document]
jtsang586: The implicit wait is unfortunately needed now because I've changed the test to no longer wait for transactions to appear on the page - to stop the amount of failures happening when there's no transactions. 
So now the wait is there to stop it moving before _actual_ transactions had a chance to load. 
When we have capsule running I'm planning to change this back in the future

apps/explorer-e2e/src/support/pages/blocks-page.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable cypress/no-unnecessary-waiting */
```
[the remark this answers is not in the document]
jtsang586: Done

apps/explorer-e2e/src/support/pages/transactions-page.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable cypress/no-unnecessary-waiting */
```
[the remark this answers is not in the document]
jtsang586: Done

libs/cypress/src/lib/commands/get-by-test-id.ts
```diff
@@ -0,0 +1,14 @@
...
+  interface Chainable<Subject> {
+    getByTestId(selector: string): Chainable<JQuery<HTMLElement>>;
+  }
+}
+
+export function addGetTestIdcommand() {
+  // @ts-ignore - ignoring Cypress type error which gets resolved when Cypress uses the command
+  Cypress.Commands.add('getByTestId', (selector, ...args) => {
+    return cy.get(`[data-testid=${selector}]`, ...args);
+  });
```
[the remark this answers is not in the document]
jtsang586: Created a separate issue to solve this in the future: https://github.com/vegaprotocol/frontend-monorepo/issues/191

apps/explorer-e2e/src/support/pages/blocks-page.js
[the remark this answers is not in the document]
dexturr: Cool can we add TODOs everywhere this is with the reasoning why, just so we can be super clear :+1:

dexturr (approved):

dexturr:
`Run end-to-end tests - PR Expected — Waiting for status to be reported` will never have status reported. Merging.

PR #147, opened 2022-03-25
Feat/76 Use UI toolkit components for 'jump to block'

### LINKED ISSUE #76: Use UI toolkit components for 'jump to block'
[apps/explorer/src/app/components/jump-to-block/index.tsx](https://github.com/vegaprotocol/frontend-monorepo/pull/58/files/9739e4b467a7ffdc478bce9ea19f6cdfd80ad5d8#diff-b542aa8e472ce143dad62b04bdc4892fe7a929319dc7bca4a9c644d684e66255)

simple solution for now: just use the input and button components from ui toolkit wrapped in a container with display flex.

At the same time, remove reliance on custom class for styling.

sam-keen:
Closes #76


mattrussell36 (approved):

apps/explorer/src/app/routes/parties/home/index.tsx
```diff
@@ -28,21 +29,21 @@ export const JumpToParty = () => {
...
+          id="block-input"
+          name="partyId"
+          placeholder="Party id"
+          className="max-w-[200px]"
+        />
+        <Button variant="secondary" type="submit">
+          Go
+        </Button>
+      </div>
     </form>
```
dexturr, line 47: minor (non-blocking): can we make this generic that takes a label, place holder and submit handler and then consume it from jump to block. Possibly could also use for search 🤔

dexturr (approved):

apps/explorer/src/app/routes/parties/home/index.tsx
sam-keen: Nice idea!

PR #156, opened 2022-03-28
Feat/83 switching vega key

### LINKED ISSUE #83: Switching Vega Key
- Show currently connected key in the app chrome somewhere
- Provide a way to switch keys via dialog/drawer

mattrussell36:
- Move truncateByChars function to helpers
- Add wallet/manage-dialog component and state to manage it in _app.page.tsx
- Move tooltip component from stats to ui-toolkit and add alignment option
- Organise format functions by date, truncate, number and utils
- Add CopyWithTooltip component to ui-toolkit

Closes #83 

Screen Shot 2022-03-27 at 10 13 09 PM


campbellssource:
Just incase I don't get around to giving you a more detailed answer: here is the modal I had imagined that you get when you go to switch keys. [In Figma](https://www.figma.com/file/F995ESZ0QMpTBuQWGAgLRz/Vega-team-library?node-id=657%3A30616).
Screenshot 2022-03-28 at 13 58 44
Things it includes (perhaps for a different iteration):
- assets for given key (help the user find keys with assets without pogo jumping)
- search (asset or alais)
- Reminder of where and what type of wallet you are connected to.

libs/react-helpers/src/lib/format/number.ts
```diff
@@ -0,0 +1,27 @@
+import { BigNumber } from 'bignumber.js';
```
dexturr, line 1: follow up: do we want to make a utils/helpers package that is just TS rather than making this specific to react?

libs/ui-toolkit/src/components/copy-with-tooltip/copy-with-tooltip.spec.tsx
```diff
@@ -0,0 +1,24 @@
+import {
+  fireEvent,
+  render,
+  screen,
+  waitFor,
+  waitForElementToBeRemoved,
+} from '@testing-library/react';
+import { CopyWithTooltip, TOOLTIP_TIMEOUT } from './copy-with-tooltip';
+
+jest.useFakeTimers();
```
dexturr, line 10: question (blocking): why do we need to fake timers here?

dexturr: Oh I see, can we move this to a test or a before each jest.useRealTimers() at the end or in an after each

dexturr (approved):

PeteWilliams:
@mattrussell36 Here's a restyled version for the current implementation:

[Figma link, including light theme](https://www.figma.com/file/K49Nx7XfocShhHcwwQzQkm/Console-V2?node-id=699%3A6381)

image

For now, I've just included the functionality you've built, we can look at adding the additional stuff @campbellssource has proposed in the future, unless you think anything else is required now Michael?

libs/ui-toolkit/src/components/copy-with-tooltip/copy-with-tooltip.spec.tsx
mattrussell36: Hm just checking the docs on this and it says its a 'global operation and will affect other tests within the same file' that said its probably best to isolate it

libs/react-helpers/src/lib/format/number.ts
```diff
@@ -0,0 +1,27 @@
...
+import { getUserLocale } from './utils';
+
+export function addDecimal(value: string, decimals: number): string {
+  if (!decimals) return value;
+  return new BigNumber(value || 0)
+    .dividedBy(Math.pow(10, decimals))
+    .toFixed(decimals);
+}
+
+export function removeDecimal(value: string, decimals: number): string {
```
sam-keen, line 12: (non-blocking): I suggest the name 'moveDecimal' as that's closer to what the function's doing, and describes another way it could be used

sam-keen (approved):

libs/react-helpers/src/lib/format/number.ts
mattrussell36: Yeah IMO I think we can just rename react-helpers to utils?

libs/ui-toolkit/src/components/copy-with-tooltip/copy-with-tooltip.spec.tsx
mattrussell36: Moved to beforeAll and restored in afterAll

libs/react-helpers/src/lib/format/number.ts
mattrussell36: Lets do this as follow up 👍

libs/react-helpers/src/lib/format/number.ts
mattrussell36: Lets discuss tomorrow and make a follow up if necessary

PR #157, opened 2022-03-28
Feature/128 positions table

bglownia:


apps/trading/pages/markets/positions.tsx
```diff
@@ -0,0 +1,71 @@
+import { useRef, useCallback, useMemo } from 'react';
+import { produce } from 'immer';
+import assign from 'assign-deep';
```
mattrussell36, line 3: I've switched this in master to use lodash/merge, can we do the same here

libs/positions/src/lib/positions-table.tsx
```diff
@@ -0,0 +1,125 @@
+import { forwardRef, useMemo } from 'react';
+import type { ValueFormatterParams } from 'ag-grid-community';
+import {
+  PriceCell /*, MarginCell*/,
```
mattrussell36, line 4: Lets remove the comment here

libs/positions/src/lib/positions-table.spec.tsx
```diff
@@ -0,0 +1,22 @@
+import { render } from '@testing-library/react';
+import { act } from 'react-dom/test-utils';
+import { MockedProvider } from '@apollo/react-testing';
+import PositionsTable from './positions-table';
+
+describe('PositionsTable', () => {
```
mattrussell36, line 6: Would be good to add a test to render a single row and assert that the right columns are shown and they are correctly formatted, I've done something similar already in order-list.spec.tsx

mattrussell36 (changes requested):
A few things to change here. Would be nice to get a test in for the positions table, we also need to go back and add one for the market list

apps/trading/components/web3-container/web3-container.spec.tsx
```diff
@@ -17,13 +17,15 @@ jest.mock('@web3-react/core', () => {
...
 
-test('Prompt to connect opens dialog', () => {
+test('Prompt to connect opens dialog', async () => {
   mockHookValue = defaultHookValue;
-  render(
-    <Web3Container>
-      <div>Child</div>
-    </Web3Container>
-  );
+  await act(async () => {
```
dexturr, line 22: Why are most tests in this PR wrapped in await act and an async function that does not await anything or return a promise?

Can some or all of these just be:

```
    render(
      <Web3Container>
        <div>Child</div>
      </Web3Container>
    );
```

if not when do we need this any

apps/trading/pages/markets/index.page.tsx
```diff
@@ -1,77 +1,63 @@
...
+          rowNode.data.data,
+          (draft: Markets_markets_data) => merge(draft, delta)
+        );
+        if (updatedData !== rowNode.data.data) {
+          update.push({ ...rowNode.data, data: updatedData });
         }
       }
-    );
-  }, [client, initialized]);
+      // @TODO - else add new market
```
dexturr, line 38: todo?

dexturr (approved):

apps/trading/components/web3-container/web3-container.spec.tsx
bglownia: Because of lazy loaded ag-grid

apps/trading/pages/markets/index.page.tsx
bglownia: We need to handle corner case where new market is added. This is not critical because list of markets will be not changing very often right now, but it need to be fixed.

mattrussell36 (approved):

PR #165, opened 2022-03-30
Task/164 colocate queries

### LINKED ISSUE #164: Co-locate data providers with consuming package
- Colocate markets and positions data providers with the markets and positions libs respectively.
- Move generic data provider to react helpers
- Add generate scripts for each lib that has queries
- Only generate global types to the graphql lib
- Rename graphql package to 'types'
- Eslint rule to import types only
- Ensure we handle naming collisions between queries in different libs
- Possibly add eslint rule for query naming convention

mattrussell36:
This is a bit of a big one but the majority of changes are changing import paths to point to colocated generated types rather than referencing them from vegaprotocol/graphql. The main changes to review are:
- deal-ticket-container.tsx, a new wrapping component to call initial query
- order-list-container.tsx, prevents rendering children (and calling queries) unless public key provided
- order-list-manager, now calls useOrders
- positions-container.tsx, same as order-list-container.tsx
- positions-manager.tsx, now provides variables and update callback

*Summary*
- Moves data providers and queries to their respective libs
- Removes unnecessary fields from deal ticket query
- Prevents rendering of positions and orderlist if no keypair (partyid) provided, previously they got stuck in a loading state
- Deal ticket components updated to use global types
- AsyncRenderer moved to ui-toolkit
- useVegaTransaction moved to wallet
- Removed onRowClick from positions table.

Closes  #164


bglownia (approved):

libs/positions/src/lib/positions-container.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { useVegaWallet } from '@vegaprotocol/wallet';
+import { PositionsManager } from './positions-manager';
+
+export const PositionsContainer = () => {
+  const { keypair } = useVegaWallet();
+
+  if (!keypair) {
+    return (
+      <Splash>
+        <p>Please connect Vega wallet</p>
```
sam-keen, line 11: Duplicated in libs/order-list/src/order-list-container.tsx

sam-keen (approved):

PR #167, opened 2022-03-30
Feat/160 explorer theme switcher

### LINKED ISSUE #160: Use theme switch hook and UI component for explorer
Requires moving styling from scss into tailwind as an initial step

sam-keen:
Closes #160


bglownia:
For AlphaLyrae font files we should have separate lib that will produce fonts.css with
`@font-face { font-family: "AlphaLyrae-Medium"; src: url("./AlphaLyrae.woof") }`
Where font files are imported from package sourced from https://github.com/vegaprotocol/alpha-lyrae

Then deployment pipeline to static.vega.xyz/fonts.css and 
`<link rel="stylesheet" href="//static.vega.xyz/fonts.css">` in app index template

bglownia (approved):
I'm not sure if we need facade for <table> <tbody> <thead> <td> and <th>. But it was there before so I approve.

apps/explorer/src/app/App.scss
```diff
@@ -1,76 +0,0 @@
-@import './styles/colors';
```
dexturr, line 1: Noice

apps/explorer/src/app/components/truncate/truncated-link.tsx
```diff
@@ -0,0 +1,27 @@
+import { TruncateInline } from './truncate';
```
dexturr, line 1: praise (non-blocking): good shout

apps/explorer/src/app/components/table/index.tsx
```diff
@@ -25,6 +25,21 @@ interface TableCellProps extends ThHTMLAttributes<HTMLTableCellElement> {
 }
 
 export const Table = ({ children, className, ...props }: TableProps) => {
+  const classes = classnames(className, 'overflow-x-auto whitespace-nowrap');
```
dexturr, line 28: chore (non-blocking): I think we need to add our key/value table to UI toolkit from TFE as this covers most basic tables and would probably reduce what we needed to keep here in this project

dexturr (approved):

apps/explorer/src/app/components/table/index.tsx
sam-keen: Ah ok cool, I'll do that when I move on to governance shortly

sam-keen: #171

apps/stats/tailwind.config.js
```diff
@@ -1,6 +1,7 @@
 const { join } = require('path');
 const { createGlobPatternsForDependencies } = require('@nrwl/next/tailwind');
 const theme = require('../../libs/tailwindcss-config/src/theme');
+const vegaCustomClasses = require('../../libs/tailwindcss-config/src/vega-custom-classes');
```
dexturr, line 4: question (non-blocking): is there a reason we can't import fom the package and need to use relative imports? Shouldn't everything needed to use the package be exported from it 🤔 

Happy to do this as a follow up.

sam-keen: Good point!

Issue #168, opened 2022-03-30
AlphaLyrae font hosting to reuse across all apps 


bglownia:
AlphaLyrae font files we should have separate lib that will produce fonts.css with
`@font-face { font-family: "AlphaLyrae-Medium"; src: url("./AlphaLyrae.woof") }`
Where font files are imported from package sourced from https://github.com/vegaprotocol/alpha-lyrae

Then deployment pipeline to static.vega.xyz/fonts.css and 
`<link rel="stylesheet" href="//static.vega.xyz/fonts.css">` in app index template

PR #170, opened 2022-03-30
[#168] Add static app to host AlphaLyrae font

bglownia:


apps/static/src/environments/environment.prod.ts
```diff
@@ -0,0 +1,3 @@
+export const environment = {
```
dexturr, line 1: can we remove this file?

apps/static/src/environments/environment.ts
```diff
@@ -0,0 +1,6 @@
+// This file can be replaced during build by using the `fileReplacements` array.
```
dexturr, line 1: can we remove this file?

apps/static/src/app/app.element.ts
```diff
@@ -0,0 +1,383 @@
...
+
+  connectedCallback() {
+    const title = 'static';
+    this.innerHTML = `
+    <div class="wrapper">
+      <div class="container">
+        <!--  WELCOME  -->
+        <div id="welcome">
+          <h1>
+            <span> Hello there, </span>
```
mattrussell36, line 14: Can we clear out all this NX boilerplate code?

mattrussell36 (changes requested):

apps/static/src/app/app.element.ts
bglownia: I removed all references to it and just forgot to remove files itself. It's cleaned up

dexturr (approved):

PR #173, opened 2022-03-30
Task/Orders data provider for libs/order-list

mattrussell36:
- Moves all code aside from index.ts to order-lists/src/lib to match other libs
- Removes use-orders hooks and replaces it with orders-data-provider.ts
- Fixes display of loader in order dialog by adding a size prop to loader component


libs/order-list/src/lib/order-list-manager.tsx
```diff
@@ -0,0 +1,73 @@
...
+  // Apply updates to the table
+  const update = useCallback((delta: OrderSub_orders[]) => {
+    if (!gridRef.current) {
+      return false;
+    }
+
+    const update: OrderFields[] = [];
+    const add: OrderFields[] = [];
+
+    delta.forEach((d) => {
```
bglownia, line 28: We need to do same trick here as in provider update function
```
const sortedOrders = sortOrders(delta);
const incoming = uniqOrders(sortedOrders);
incoming .forEach((d) => {
```
or just `uniqOrders(sortOrders(delta)).forEach((d) => {`

bglownia (changes requested):

libs/order-list/src/lib/order-list-manager.tsx
mattrussell36: Added a prepareIncomingOrders function to be even a bit more DRY, its important the sort comes first so uniq grabs the correct order

bglownia (approved):

dexturr (approved):

dexturr:
LGTM think the linting errors will be fixed with a rebase

Issue #128, opened 2022-03-23
Positions table


mattrussell36:
- Show positions data in AgGrid
- Columns
  - Market
  - Amount (openVolume)
  - Average entry price
  - Mark price
  - Realised PNL
  
  
Note:
- Lets not show leverage yet
- No close button yet
  
Example queries to get the above data
```
export const POSITIONS_FRAGMENT = gql`
  fragment PositionDetails on Position {
    realisedPNL
    openVolume
    unrealisedPNL
    averageEntryPrice
    market {
      id
      name
      data {
        markPrice
      }
      decimalPlaces
      tradableInstrument {
        instrument {
          id
          name
          metadata {
            tags
          }
          code
          product {
            ... on Future {
              settlementAsset {
                id
                symbol
                name
                decimals
              }
              quoteName
            }
          }
        }
      }
    }
  }
`

export const POSITION_QUERY = gql`
  ${POSITIONS_FRAGMENT}
  query positions($partyId: ID!) {
    party(id: $partyId) {
      id
      positions {
        ...PositionDetails
      }
    }
  }
`

export const POSITIONS_SUB = gql`
  ${POSITIONS_FRAGMENT}
  subscription positionSubscribe($partyId: ID!) {
    positions(partyId: $partyId) {
      ...PositionDetails
    }
  }
`
```

gordsport:
Closing as in status `Done`

PR #178, opened 2022-03-31
[#128] Use native ag-grid sort instead on useMemo

bglownia:


libs/positions/src/lib/positions-table.tsx
```diff
@@ -21,22 +20,22 @@ interface PositionsTableProps {
...
+  nodeB: { data: Positions_party_positions },
+  isInverted: boolean
 ) => {
-  if (
-    a.market.tradableInstrument.instrument.name <
-    b.market.tradableInstrument.instrument.name
-  ) {
-    return -1;
+  const a = nodeA.data.market.tradableInstrument.instrument.name;
+  const b = nodeB.data.market.tradableInstrument.instrument.name;
```
dexturr, line 31: chore (blocking): Many markets may share the same instrument, we likely need a secondary sort. I suggest market by ID as we are guaranteed uniqueness between market/position.

bglownia: Good point. I've added secondary sort on market id

dexturr (approved):

mattrussell36 (approved):

PR #187, opened 2022-03-31
Task/componentize stats

dexturr:
Take reusable stuff from stats and componetize it


libs/ui-toolkit/src/utils/intent.ts
```diff
@@ -19,3 +26,13 @@ export const getIntentShadow = (intent?: Intent) => {
     'shadow-intent-help': intent === Intent.Help,
   });
 };
+
+export const getVariantBackground = (variant?: Variant) => {
+  return classNames('shadow-callout', {
```
mattrussell36, line 31: is shadow-callout meant to be added here?

libs/ui-toolkit/src/utils/intent.ts
```diff
@@ -9,6 +9,13 @@ export enum Intent {
   Help = 'help',
 }
 
+export enum Variant {
```
mattrussell36, line 12: I think we should probably just use Intent and not add a new Variant enum (if possible) as they both attempt to solve the same problem? We might need to consolidate how we use the intent enum first though so this could be fixed as an interation

dexturr: Hmmm yes I agree, we have variants that are different for button and lozenge, would be good to unify these

libs/network-stats/src/components/stats-manager/stats-manager.tsx
```diff
@@ -9,6 +9,7 @@ import { Table } from '../table';
 import { TableRow } from '../table-row';
 import { PromotedStats } from '../promoted-stats';
 import { PromotedStatsItem } from '../promoted-stats-item';
+import { Card } from '@vegaprotocol/ui-toolkit';
```
bglownia, line 12: 'Card' is defined but never used

libs/ui-toolkit/src/components/indicator/indicator.tsx
```diff
@@ -0,0 +1,11 @@
+import classNames from 'classnames';
+import type { Variant } from '../../utils/intent';
+import { getVariantBackground } from '../../utils/intent';
+
+export const Indicator = ({ variant }: { variant?: Variant }) => {
+  const names = classNames(
+    'inline-block w-8 h-8 mb-[0.15rem] mr-8 rounded',
```
bglownia, line 7: I don't like this custom spacing values. It's better to use to use mb-2. It's 0.125 rem which gives 2px. 0.15rem is 2.4px which will be rounded to 2px

mattrussell36 (approved):
LGTM. As I said I think we should do a follow up to sort out intents/variants

edd (approved):

sam-keen (commented):
LGTM

PR #193, opened 2022-04-01
[#185] Add accounts table

### LINKED ISSUE #185: Accounts table MVP
### MVP accounts table

- Fields
  - Asset (symbol)
  - Type
  - Balance
  - Market, if present, otherwise just a '-'
- Row id needs to be a combination of asset, type and market id fields
- Price needs to be formatted using asset.decimals
- `market` field is null for general accounts, so you will need to consider that when creating row ids
  
 Example query
 ```
fragment AccountFields on Account {
  type
  balance
  market {
    id
  }
  asset {
    symbol
    decimals
  }
}

query Accounts($partyId: ID!) {
  party(id: $partyId) {
    accounts {
      ...AccountFields
    }
  }
}

subscription AccountsSub($partyId: ID!) {
  accounts(partyId: $partyId) {
      ...AccountFields
  }
}
```

### PR DESCRIPTION
**Pull request by bglownia**
closes #185

### CODE REVIEW

**Review by mattrussell36 — APPROVED** [review]
Good stuff!

**Review by dexturr — APPROVED** [review]

File: libs/accounts/tsconfig.lib.json [hunk 1]
```diff
@@ -0,0 +1,28 @@
...
+    "**/*.test.js",
+    "**/*.spec.jsx",
+    "**/*.test.jsx"
+  ],
+  "include": [
+    "**/*.js",
+    "**/*.jsx",
+    "**/*.ts",
+    "**/*.tsx",
+    "../react-helpers/src/lib/grid-cells/summary-row.spec.ts"
```
mattrussell36, line 26: Can remove

libs/accounts/src/lib/accounts-table.tsx
```diff
@@ -0,0 +1,146 @@
...
+  data: Accounts_party_accounts & { __summaryRow?: boolean },
+  columnApi: ColumnApi
+) => {
+  if (data.__summaryRow) {
+    return null;
+  }
+  const sortColumnId = columnApi.getColumnState().find((c) => c.sort)?.colId;
+  switch (sortColumnId) {
+    case 'asset.symbol':
+      return data.asset.symbol;
```
mattrussell36, line 33: We should probably use asset.id for this. Its extremely unlikely but in theory I think we could end up with duplicate asset symbols

mattrussell36 (changes requested):

mattrussell36 (approved):

PR #199, opened 2022-04-04
Test added for stats page

### LINKED ISSUE #109: Add tests for mainnet stats page
follow up (non-blocking): let's create a ticket to add some cypress tests for this

_Originally posted by @dexturr in https://github.com/vegaprotocol/frontend-monorepo/pull/107#discussion_r831105173_

jtsang586:
Adds test for the state/home page on block explorer

Closes #109

### DIFF
--- a/apps/explorer-e2e/cypress.json
+++ b/apps/explorer-e2e/cypress.json
@@ -14,6 +14,7 @@
   "screenshotsFolder": "../../dist/cypress/apps/explorer-e2e/screenshots",
   "chromeWebSecurity": false,
   "env": {
+    "environment": "testnet",
     "tsConfig": "tsconfig.json",
     "TAGS": "not @todo and not @ignore and not @manual"
   }
--- a/apps/explorer-e2e/src/integration/home-page.feature
+++ b/apps/explorer-e2e/src/integration/home-page.feature
@@ -2,7 +2,7 @@ Feature: Home page
 
   Scenario: Stats page displayed correctly
     Given I am on the homepage
-    Then the stats for "testnet" is correctly displayed
+    Then the stats for deployed environment are correctly displayed
 
   Scenario Outline: Succesfful search for specific id by <IdType>
     Given I am on the homepage
--- a/apps/explorer-e2e/src/integration/home-page.feature
+++ b/apps/explorer-e2e/src/integration/home-page.feature
@@ -1,6 +1,10 @@
 Feature: Home page
 
-  Scenario Outline: Successful search for specific id by <IdType>
+  Scenario: Stats page displayed correctly
+    Given I am on the homepage
+    Then the stats for "testnet" is correctly displayed
+
+  Scenario Outline: Succesfful search for specific id by <IdType>
     Given I am on the homepage
     When I search for '<Id>'
     Then I am redirected to page containing id '<Id>'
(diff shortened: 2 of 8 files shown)


apps/explorer-e2e/src/support/pages/home-page.js
```diff
@@ -1,3 +1,67 @@
...
+      });
+  }
+
+  verifyStatsBlockHeightUpdating() {
+    cy.getByTestId(this.statsValue)
+      .eq(1)
+      .invoke('text')
+      .then((blockHeightTxt) => {
+        cy.wait(1000); // eslint-disable-line cypress/no-unnecessary-waiting
+        //Wait needed to allow blocks to change
```
dexturr, line 58: chore (blocking): can we not poll for this in some way rather than waiting a specific amount of time and let the test timeouts catch it 🤔

dexturr (changes requested):
LGTM, just would like to avoid the specific time wait and poll for a dom element or poll until the block height has increased in some way.

apps/explorer-e2e/src/support/pages/home-page.js
dexturr: Possibly intercepts are the way to go here, if you intercept the `statistics` query you can alias that and wait for that 👍

apps/explorer-e2e/cypress.json
```diff
@@ -14,6 +14,7 @@
   "screenshotsFolder": "../../dist/cypress/apps/explorer-e2e/screenshots",
   "chromeWebSecurity": false,
   "env": {
-    "tsConfig": "tsconfig.json"
+    "tsConfig": "tsconfig.json",
+    "TAGS": "not @todo and not @ignore and not @manual"
```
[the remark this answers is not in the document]
jtsang586: The way tags are used here is specific to cucumber preprocessor. The NX generators for Cypress don't include cucumber so afaik we'll need to keep adding these every new project unfortunately.

apps/explorer-e2e/src/support/pages/home-page.js
jtsang586: I messed around with intercept and found out Cypress was already waiting properly without the implicit wait 🤦‍♂️. It wasn't all wasted effort though. I've replaced a wait elsewhere with an intercept 👍🏼

dexturr: Nice!

apps/explorer-e2e/cypress.json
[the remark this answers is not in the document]
dexturr: Yeah so we can create a pre setup lib as a template and use an nx generator to mean we don't have to do this self same setup every single time:

https://nx.dev/generators/creating-files#creating-files-with-a-generator

but certainly a follow up and not for now

apps/explorer-e2e/src/integration/home-page.feature
```diff
@@ -1,5 +1,9 @@
 Feature: Home page
 
+  Scenario: Stats page displayed correctly
+    Given I am on the homepage
+    Then the stats for "testnet" is correctly displayed
```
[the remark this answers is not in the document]
jtsang586: I've added "testnet" as an environment variable. If we're going to be running these tests on multiple different environments then I'm happy to create a separate ticket to properly split them out into separate env files.

mattrussell36 (changes requested):
A few linting issues to resolve

dexturr (approved):

mattrussell36 (approved):

bglownia (approved):

dexturr:
Blocked by: #242

PR #217, opened 2022-04-07
Feat/152 Trade list MVP

### LINKED ISSUE #152: Trade list
- Order id
- Price (green if up from previous trade, red if down)
- Size (green if up from previous trade, red if down)
- Timestamp

mattrussell36:
- Adds lib/trades, with trades-container, trades-table and trades-data-provider
- Renders out trades-container on the trading page
- Also adds the accounts table to the trading page

Note this just displays the latest 50 trades. I have created #218 to capture the next iteration

### DIFF
--- a/libs/trades/src/lib/trades-table.tsx
+++ b/libs/trades/src/lib/trades-table.tsx
@@ -1,0 +1,13 @@
+import type { AgGridReact } from 'ag-grid-react';
+import { forwardRef } from 'react';
+import type { TradeFields } from './__generated__/TradeFields';
+
+interface TradesTableProps {
+  data: TradeFields[] | null;
+}
+
+export const TradesTable = forwardRef<AgGridReact, TradesTableProps>(
+  ({ data }, ref) => {
+    return <pre>{JSON.stringify(data, null, 2)}</pre>;
+  }
+);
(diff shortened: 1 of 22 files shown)


bglownia (approved):

libs/trades/src/lib/trades-table.tsx
```diff
@@ -0,0 +1,88 @@
...
+    let colorClass = '';
+
+    if (typeof rowIndex === 'number') {
+      const prevRowNode = api.getModel().getRow(rowIndex + 1);
+      const prevValue = prevRowNode?.data[dataKey];
+      const valueNum = new BigNumber(value);
+
+      if (valueNum.isGreaterThan(prevValue)) {
+        colorClass = UP_CLASS;
+      } else if (valueNum.isLessThan(prevValue)) {
```
sam-keen, line 31: Is it possible that `valueNum` could ever be equal to `prevValue`?

sam-keen (approved):

libs/trades/src/lib/trades-table.tsx
mattrussell36: It is! and in that case we don't apply a class so it just inherits the default text color (up, green, down red, no change white)

PR #228, opened 2022-04-07
Fix/Loader size, token refetch, order dialog and deposits subscription

mattrussell36:
- Fixes loader size for depoist transaction
- Refetch erc20 token balance after faucet
- Properly interpolate market name in order dialog
- Remove prepending 0x from pubkey for deposit bus event subscription

### DIFF
--- a/libs/deposits/src/lib/use-submit-deposit.ts
+++ b/libs/deposits/src/lib/use-submit-deposit.ts
@@ -6,7 +6,7 @@ import type {
 } from './__generated__/DepositEvent';
 import { DepositStatus } from '@vegaprotocol/types';
 import { useState } from 'react';
-import { useEthereumTransaction } from '@vegaprotocol/react-helpers';
+import { remove0x, useEthereumTransaction } from '@vegaprotocol/react-helpers';
 import type { VegaErc20Bridge } from '@vegaprotocol/smart-contracts-sdk';
 
 const DEPOSIT_EVENT_SUB = gql`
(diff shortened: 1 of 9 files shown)


libs/ui-toolkit/src/components/ag-grid/ag-grid-dynamic-themed.tsx
```diff
@@ -35,10 +35,12 @@ export const AgGridThemed = ({
       style={style}
     >
       {theme === 'dark' ? (
+        // @ts-ignore TODO: fix me
```
dexturr, line 38: ?

dexturr (commented):
Looks fine but a lot of TS ignores with fix me? If intended as follow ups then can we have tickets else can we fix in this PR?

libs/deposits/src/lib/use-submit-deposit.ts
```diff
@@ -57,7 +54,7 @@ export const useSubmitDeposit = (
   }, confirmations);
 
   useSubscription<DepositEvent, DepositEventVariables>(DEPOSIT_EVENT_SUB, {
-    variables: { partyId: partyId || '' },
+    variables: { partyId: partyId ? partyId.substring(2, partyId.length) : '' },
```
sam-keen, line 57: (chore, non-blocking): I think this would benefit from being reusable, and should include a check that the string begins with '0x'.

sam-keen (approved):

libs/deposits/src/lib/use-submit-deposit.ts
bglownia: ahoy to that!
`const remove0x = s => s.replace(/^0x/, '')`

bglownia (approved):

PR #211, opened 2022-04-05
Task/229 Stub api in trading e2e

mattrussell36:
- Split out vega wallet into own support class
- Add markets-page.feature (for market list page) and portfolio-page.feature
- Update market-order.feature to be deal-ticket.feature
- Add cy.intercepts for queries, with fixtures

Closes #239

PR #243, opened 2022-04-11
Fix for failing market tests timing out

jtsang586:
Increasing timeout for finding market to place order in to stop tests from failing

### DIFF
--- a/apps/trading-e2e/src/support/pages/markets-page.js
+++ b/apps/trading-e2e/src/support/pages/markets-page.js
@@ -21,7 +21,7 @@ export default class MarketPage extends BasePage {
   }
 
   clickOnMarket(marketText) {
-    cy.contains(marketText).click();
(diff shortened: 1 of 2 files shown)


dexturr (changes requested):
question (blocking): I am unsure how this would help? Surely on live now the market is either there or it isn't in which case the current timeout would be fine. Also will #211 remove the need for this or am I missing something?

jtsang586:
> question (blocking): I am unsure how this would help? Surely on live now the market is either there or it isn't in which case the current timeout would be fine. Also will #211 remove the need for this or am I missing something?

Currently the tests are failing even when the markets are showing which is really weird and annoying. But you're right that #211 will remove the need for this. 
I thought it was worth pushing a really small change to stop the tests failing right now because the failing tests are bothering me :triumph: . Either way I'm also happy to close the PR and wait for the stubs to go in.

PR #244, opened 2022-04-11
Add conditional IPFS deployment for projects

edd:
Adds a new tool that will deploy to IPFS if affected projects have a Fleek
configuration file, but only on commits to master.

- Remove unused publish step
- Add fleek file for stats, and script to run it if affected
- Genericise tool name, add workflow


tools/ipfs-deploy.js
```diff
@@ -0,0 +1,74 @@
...
+ * that have a .fleek.json file
+ */
+
+// Fleek CLI requires this variable to be set
+if (!process.env['FLEEK_API_KEY']) {
+  console.error('Error: FLEEK_API_KEY must be set');
+  process.exit(1);
+}
+
+const { existsSync } = require('fs');
```
dexturr, line 14: nitpick (non-blocking): I like requires at the top of a file so they can be easily seen as the dependecies.

dexturr (approved):

.github/workflows/ipfs.yml
```diff
@@ -0,0 +1,32 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
+          fetch-depth: 0
+      - name: Derive appropriate SHAs for base and head for `nx affected` commands
+        uses: nrwl/nx-set-shas@v2
+        with:
+          main-branch-name: master
```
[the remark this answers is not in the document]
edd: I originally had this as `master` and `master~1` And will put it back to that. I thought this did something smarterer. I’m not sure that I considered what smarterer would be.

tools/ipfs-deploy.js
edd: Good catch, not sure how this ended up here.

.github/workflows/ipfs.yml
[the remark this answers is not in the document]
edd: Done

.github/workflows/ipfs.yml
```diff
@@ -0,0 +1,28 @@
...
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
+          fetch-depth: 0
+      - name: Use Node.js 16
+        id: Node
+        uses: actions/setup-node@v2
+        with:
+          node-version: 16.14.0
+      - name: Run deploy for affected sites
+        run: yarn nx print-affected --base=origin/master~1 --head=origin/master --select=projects | ./tools/ipfs-deploy.js
```
bglownia, line 28: different between origin/master~1 and origin/master is always latest commit. If we will push 5 new commit to master and only changes from top commit will be taken to comparsion.
We should do --base=REF_TO_COMMIT_PREVIOUSLY_DEPLOYED --head=REF_TO_COMMIT_WHICH_WE_WANT_DEPLOY

We need to take information some somewhere which version is currently deployed (I will suggest git tag).
We can have tag for currently deployed version to IPSF and update it after each deployment
```
git checkout master
git pull
yarn nx print-affected --base=IPFS --head=master --select=projects | ./tools/ipfs-deploy.js
git tag -a IPFS -m "Some extra comment, maybe even date"
git push origin IPFS
```

tools/ipfs-deploy.js
```diff
@@ -0,0 +1,73 @@
...
+// This hangs for input, and process.exit ensures it only triggers one
+process.stdin.on('data', function (affectedProjectsString) {
+  // If there is no input, nothing is affected. Bail early.
+  if (affectedProjectsString.trim().length === 0) {
+    console.log('Success: No projects affected');
+    process.exit(0);
+  }
+
+  // Handle multiple projects or a single project being affected
+  let affectedProjects =
```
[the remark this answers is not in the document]
edd: Fair, but now redundant

.github/workflows/ipfs.yml
edd: It was a fair point. I've rewritten the script so that instead of this workflow, or the one it was before, it:

1. Checks the ./apps/** folder to see which one has a fleek config
2. Gets the fleek siteId for each project
3. Uses their API to fetch the last commit that triggered a build
4. Runs nx affected and checks to see if this project was affected
5. If it was, trigger a build using their API

It feels circular and weird, but when I started looking at writing an nx command that would return it's siteID, or better yet, the last commit ID, it didn't seem like there would be a good way to filter nx affected... so I bailed out and used this shell script style instead. I might be fighting the tools though.

tools/ipfs-deploy.js
```diff
@@ -0,0 +1,145 @@
...
+const { existsSync, readdirSync, lstatSync, readFileSync } = require('fs');
+const { execSync } = require('child_process');
+
+/**
+ * Parses the last commit hash out of the Fleek API response
+ * @param {String} siteId
+ * @returns string Last commit that triggered a build
+ */
+function getFleekLastBuildCommit(siteId) {
+  const curl = `curl 'https://api.fleek.co/graphql' --silent -X POST -H 'Accept: */*' -H 'Accept-Encoding: gzip, deflate, br' -H 'Content-Type: application/json' -H 'Authorization: ${process.env['FLEEK_API_KEY']}' --data-raw '{"query":"{getSiteById(siteId: \\"${siteId}\\"){publishedDeploy{repository{commit}}}}","variables":null}'`;
```
bglownia, line 27: Thanks got there will be fetch API available soon
https://nodejs.org/en/blog/release/v17.5.0/

tools/ipfs-deploy.js
```diff
@@ -0,0 +1,145 @@
...
+// Fleek CLI requires this variable to be set
+if (!process.env['FLEEK_API_KEY']) {
+  console.error('Error: FLEEK_API_KEY must be set');
+  process.exit(1);
+}
+
+readdirSync(projectPath).forEach((proj) => {
+  try {
+    const project = `${projectPath}${proj}`;
+    if (!lstatSync(project).isDirectory()) {
```
bglownia, line 68: We don't need this condition `!existsSync(config)` from line 74 is sufficient

tools/ipfs-deploy.js
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    // Now run nx affected
+    let isAffected;
+
+    try {
+      const affectedSinceCommit = execSync(
+        `yarn nx print-affected --base=${baseCommit} --head=master --select=projects`
+      );
+      isAffected = affectedSinceCommit.toString().indexOf(proj) !== -1;
```
bglownia, line 115: it can be that this will return false positive eg: `'trading-mobile, console'.indexof('trading')  !== -1`
It will be safer to do `affectedSinceCommit.toString().split(',').map(v => v.trim()).includes(proj)`

bglownia (approved):

tools/ipfs-deploy.js
edd: Hehe yeah. `require('https')` was so heavyweight. But more sensible. But... so inelegant. I expected you to reject the review based on this 😆

tools/ipfs-deploy.js
edd: Removed

tools/ipfs-deploy.js
edd: Good catch. Fixing...

edd:
Thanks for the great review - I've fixed all the feedback, so I'm merging. Let's... see what happens!

PR #250, opened 2022-04-12
Fix IPFS deploy step

edd:
After #244, and setting up the appropriate ENV var, I forgot to explicitly make it available in the build step that needs it.

- Update IPFS action


dexturr (approved):

dexturr:
question (non-blocking): is this going to be a required status check?

edd:
@dexturr I think... ~no~? ~Maybe~? ~Not yet~? Not now at least.

PR #253, opened 2022-04-12
Task/Deposit e2e testing

### LINKED ISSUE #252: Deposit e2e testing

Ditmir-Vega:
closes #252

### DIFF
--- a/apps/trading-e2e/src/integration/auction-orders.feature
+++ b/apps/trading-e2e/src/integration/auction-orders.feature
@@ -1,0 +1,6 @@
+Feature: Auction Orders
+
+  Scenario: Place a limit order
+    Given market is in continuous trading market
+
+  Scenario:
--- a/apps/trading-e2e/src/integration/auction-orders.feature
+++ b/apps/trading-e2e/src/integration/auction-orders.feature
@@ -1,6 +1,1 @@
 Feature: Auction Orders
-
-  Scenario: Place a limit order
-    Given market is in continuous trading market
-
-  Scenario:
--- a/apps/trading-e2e/src/integration/deposits.feature
+++ b/apps/trading-e2e/src/integration/deposits.feature
@@ -9,7 +9,7 @@ Feature: Deposits to vega wallet
     And the connect button is displayed
 
   @todo
-  Scenario Outline: user cannot deposit if approval amount is 0 (approval amount is 0)
+  Scenario: Cannot deposit if approved amount is 0 (approval amount is 0)
     And I connect my ethereum wallet
     And I connect my vega wallet
     When I set "0" tokens to be approved
@@ -25,7 +25,7 @@ Feature: Deposits to vega wallet
     Then I can see the deposit is unsuccessful
 
   @todo
-  Scenario: user cannot deposit if approval amount is lower than deposit ()
+  Scenario: Cannot deposit if approved amount is lower than deposit amount
     When I set "2" tokens to be approved
     And I approve the asset tokens
     And I can see the deposit form is displayed
@@ -41,7 +41,7 @@ Feature: Deposits to vega wallet
     And i try to deposit "5" vega
 
   @todo
-  Scenario: User can succesfully deposit (approval amount is greater than deposit)
+  Scenario: Can succesfully deposit (approved amount is greater than deposit)
     When I set "200000000" tokens to be approved
     And I approve the asset tokens
     And I can see the deposit form is displayed
@@ -56,7 +56,7 @@ Feature: Deposits to vega wallet
     And Balance is updated to reflect deposit amount
 
   @todo
-  Scenario: validation errors where no fields are filled in
+  Scenario: Validation errors where no fields are filled in
     When I submit a deposit form with empty fields
     Then I can see validation errors present
 
@@ -72,7 +72,7 @@ Feature: Deposits to vega wallet
     Then I can see the field is updated with the maximum amount of the asset from my wallet
 
   @todo
-  Scenario: transaction fails if attempting to deposit more than available in wallet
+  Scenario: Transaction fails if attempting to deposit more than available in wallet
     And I can see the deposit form is displayed
     And I select "" asset from the dropdown list
     When I enter the following details
@@ -83,7 +83,7 @@ Feature: Deposits to vega wallet
     Then an error message is shown stating not enough tokens in wallet to deposit
 
   @todo
-  Scenario: deposit to a vega wallet key which is not your own
+  Scenario: Deposit to a vega wallet key which is not your own
     And I can see the deposit form is displayed
     And I select "" asset from the dropdown list
     When I enter the following details
@@ -96,7 +96,7 @@ Feature: Deposits to vega wallet
     And Balance is updated to reflect deposit amount
 
   @todo
-  Scenario: deposit when wallet is not connected
+  Scenario: Deposit when vega wallet is not connected
     And I disconnect my vega wallet
     And I can see the deposit form is displayed
     And I select "" asset from the dropdown list
--- a/apps/trading-e2e/src/integration/deposits.feature
+++ b/apps/trading-e2e/src/integration/deposits.feature
@@ -1,0 +1,110 @@
+Feature: Deposits to vega wallet
+
+  Background:
+    Given I navigate to deposits page
+
+  @todo
+  Scenario: Eth wallet connect message displayed if no ethereum wallet connected
+    Then I can see a message stating no wallet is connected
+    And the connect button is displayed
+
+  @todo
+  Scenario Outline: user cannot deposit if approval amount is 0 (approval amount is 0)
+    And I connect my ethereum wallet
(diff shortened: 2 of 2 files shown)


apps/trading-e2e/src/integration/deposits.feature
```diff
@@ -0,0 +1,110 @@
...
+    When I enter the following details
+      | Field         | Value    |
+      | To (Vega key) | xxxxxxxx |
+      | Amount        | 50       |
+    And I click to the deposit the funds
+    And I approve the ethereum transaction
+    Then I can see the deposit is unsuccessful
+
+  @todo
+  Scenario: Cannot deposit if approved amount is lower than deposit amount
```
campbellssource, line 28: suggestion non-blocking: "User is prompted to approve more if the deposit amount is greater than the approved amount (and approve !=0)"

campbellssource:
fyi: non-blocking: these tests all assume that the asset you are depositing is ERC20 (which require approvals) in the future there may be some assets that don't require an approval

apps/trading-e2e/src/integration/deposits.feature
```diff
@@ -0,0 +1,110 @@
...
+    Then I can see the deposit is Successfull
+    And Balance is updated to reflect deposit amount
+
+  @todo
+  Scenario: Validation errors where no fields are filled in
+    When I submit a deposit form with empty fields
+    Then I can see validation errors present
+
+  @todo
+  Scenario: Deposit maximum amount you have in the wallet for a given asset
```
campbellssource, line 64: suggestion non-blocking: "use the 'Use Maximum' button to populate amount input with the balance in the connected wallet

apps/trading-e2e/src/integration/deposits.feature
```diff
@@ -0,0 +1,110 @@
...
+    And I select "" asset from the dropdown list
+    When I enter the following details
+      | Field         | Value    |
+      | To (Vega key) | xxxxxxxx |
+      | Amount        | 0        |
+    When I click the use maximum button
+    Then I can see the field is updated with the maximum amount of the asset from my wallet
+
+  @todo
+  Scenario: Transaction fails if attempting to deposit more than available in wallet
```
campbellssource, line 75: suggestion non-blocking: "user is warned if the the amount to deposit is greater than what is available in the connected wallet"

apps/trading-e2e/src/integration/deposits.feature
```diff
@@ -0,0 +1,110 @@
...
+    And I can see the deposit form is displayed
+    And I select "" asset from the dropdown list
+    When I enter the following details
+      | Field         | Value                      |
+      | To (Vega key) | VEGA KEY of another wallet |
+      | Amount        | 50                         |
+    And I click to the deposit the funds
+    And I approve the ethereum transaction
+    Then I can see the deposit is Successfull
+    And Balance is updated to reflect deposit amount
```
campbellssource, line 96: Question non-blocking: I think we talked about just testing that we see confirmation that a deposit is in progress rather than waiting for the deposit to be complete

apps/trading-e2e/src/integration/deposits.feature
```diff
@@ -0,0 +1,110 @@
...
+    And I click to the deposit the funds
+    Then an error message is shown stating not enough tokens in wallet to deposit
+
+  @todo
+  Scenario: Deposit to a vega wallet key which is not your own
+    And I can see the deposit form is displayed
+    And I select "" asset from the dropdown list
+    When I enter the following details
+      | Field         | Value                      |
+      | To (Vega key) | VEGA KEY of another wallet |
```
campbellssource, line 91: FYI: another use case is depositing to the treasury e.g. address `0x0`

campbellssource (changes requested):

apps/trading-e2e/src/integration/auction-orders.feature
```diff
@@ -0,0 +1 @@
+Feature: Auction Orders
```
jtsang586, line 1: For scenarios with auctions I think it makes more sense to group them with the other deal ticket tests. 
So I think this can be deleted.

apps/trading-e2e/src/integration/deposits.feature
Ditmir-Vega: @campbellssource is there another journey here where the user will be prompted to approve more if the deposit is higher than approval amount ?

i thought the approval step is done prior to any deposit attempted

apps/trading-e2e/src/integration/auction-orders.feature
Ditmir-Vega: depending on how many tests in total there are in deal ticket it might be better to split them out in seperate feature files .
When running tests in parralel  you don't want the features to be too long to run

apps/trading-e2e/src/integration/deposits.feature
```diff
@@ -0,0 +1,110 @@
...
+  Background:
+    Given I navigate to deposits page
+
+  @todo
+  Scenario: Eth wallet connect message displayed if no ethereum wallet connected
+    Then I can see a message stating no wallet is connected
+    And the connect button is displayed
+
+  @todo
+  Scenario: Cannot deposit if approved amount is 0 (approval amount is 0)
```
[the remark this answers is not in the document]
Ditmir-Vega: does this mean that if the user has set their approval limit to 0 , then they will be prompted to increase it ?

apps/trading-e2e/src/integration/deposits.feature
Ditmir-Vega: what is the difference between depositing to the treasury and depositing to another wallet .
Is there a difference in deposit journey ?

apps/trading-e2e/src/integration/deposits.feature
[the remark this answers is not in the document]
campbellssource: The approved amount is 0 by default. so this is more like a "prompt the user to approve for the first time"

apps/trading-e2e/src/integration/deposits.feature
campbellssource: it is this is if the first time approved I set it only to a bit more than my first deposit. but this scenario is a second attempt to deposit where the amount will over use my initial approved amount

apps/trading-e2e/src/integration/deposits.feature
campbellssource: no difference really was just letting you know some background

apps/trading-e2e/src/integration/deposits.feature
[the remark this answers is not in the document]
Ditmir-Vega: I think this is a different test than the one you just mentioned. This test just makes sure that the user cannot deposit if they are on 0 limit .

We can write a test to approve on the first time but we need capsule to always load a checkpoint as most likely in this case the user would have already approved the first time

apps/trading-e2e/src/integration/deposits.feature
```diff
@@ -0,0 +1,110 @@
+Feature: Deposits to vega wallet
+
+  Background:
+    Given I navigate to deposits page
+
+  @todo
+  Scenario: Eth wallet connect message displayed if no ethereum wallet connected
+    Then I can see a message stating no wallet is connected
+    And the connect button is displayed
+
```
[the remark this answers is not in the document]
mattrussell36: It might be better to have a separate `web3.feature` to test that case (and other web3 connection related things)

apps/trading-e2e/src/integration/deposits.feature
```diff
@@ -0,0 +1,113 @@
...
+  @todo
+  Scenario: Cannot deposit if approved amount is 0 (approval amount is 0)
+    And I connect my ethereum wallet
```
mattrussell36, line 13: Do we need a line for `Given I am on the deposit page` at the start of each of these scenarios?

apps/trading-e2e/src/integration/deposits.feature
```diff
@@ -0,0 +1,113 @@
...
+    And I approve the ethereum transaction
+  # The following step is valid and are commented out as currently it cannot be automated
+  # Then I can see the deposit is unsuccessful
+
+  @todo
+  Scenario: Cannot deposit if approved amount is lower than deposit amount
+    When I set "2" tokens to be approved
+    And I approve the asset tokens
+    And I can see the deposit form is displayed
+    And I select "" asset from the dropdown list
```
mattrussell36, line 33: Should you not specify what asset you are selecting here?

mattrussell36 (changes requested):

apps/trading-e2e/src/integration/deposits.feature
Ditmir-Vega: not yet .. we can fill in this blank when we start automating this scenario. we'll need to ensure the asset here is always available to select and we don't know if this would require extra work to set up etc .

PR #254, opened 2022-04-12
Test/234 switch public key

### LINKED ISSUE #234: UI Test - Switching Vega Key
Add E2E test for switching public key on connected Vegawallet

jtsang586:
Additional test for switching public key for connected Vega wallet as well as assertions that the pub key is displayed.

Closes #234

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -95,6 +95,8 @@ jobs:
         run: ./vegawallet import -w UI_Trading_Test --recovery-phrase-file ./recovery -p ./passphrase
       - name: Import config
         run: ./vegawallet network import --from-url https://raw.githubusercontent.com/vegaprotocol/networks/master/fairground/fairground.toml
+      - name: Create new public key
+        run: ./vegawallet generate key -w UI_Trading_Test -p ./passphrase
       - name: Start service
         run: ./vegawallet service run --network fairground  &
 
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -50,7 +50,7 @@ jobs:
         run: ./vegawallet import -w UI_Trading_Test --recovery-phrase-file ./recovery -p ./passphrase
       - name: Import config
         run: ./vegawallet network import --from-url https://raw.githubusercontent.com/vegaprotocol/networks/master/fairground/fairground.toml
-      - name: Create new public key
+      - name: Create public key 2
         run: ./vegawallet key generate -w UI_Trading_Test -p ./passphrase
       - name: Start service
         run: ./vegawallet service run --network fairground  &
@@ -97,7 +97,7 @@ jobs:
         run: ./vegawallet import -w UI_Trading_Test --recovery-phrase-file ./recovery -p ./passphrase
       - name: Import config
         run: ./vegawallet network import --from-url https://raw.githubusercontent.com/vegaprotocol/networks/master/fairground/fairground.toml
-      - name: Create new public key
+      - name: Create public key 2
         run: ./vegawallet key generate -w UI_Trading_Test -p ./passphrase
       - name: Start service
         run: ./vegawallet service run --network fairground  &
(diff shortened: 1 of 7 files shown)


.github/workflows/cypress.yml
```diff
@@ -50,6 +50,8 @@ jobs:
         run: ./vegawallet import -w UI_Trading_Test --recovery-phrase-file ./recovery -p ./passphrase
       - name: Import config
         run: ./vegawallet network import --from-url https://raw.githubusercontent.com/vegaprotocol/networks/master/fairground/fairground.toml
+      - name: Create new public key
+        run: ./vegawallet key generate -w UI_Trading_Test -p ./passphrase
```
dexturr, line 54: minot (non-blocking): I'd just call this key 2 as it i always just the second key

apps/trading-e2e/src/support/step_definitions/vega-wallet.step.js
```diff
@@ -28,3 +28,16 @@ When('I connect to Vega Wallet', () => {
   );
   vegaWallet.clickConnectVegaWallet();
 });
+
+When('I open wallet dialog', () => {
+  vegaWallet.validatePublicKeyDisplayed('47836c…c7d278'); // Default Test wallet pub key
```
dexturr, line 33: chore (blocking): can we move both these keys to the env file so developers can configure their own local wallet when running tests

.github/workflows/cypress.yml
jtsang586: That's renamed now

apps/trading-e2e/src/support/step_definitions/vega-wallet.step.js
jtsang586: I've added the second pub key into a Cypress env and now truncating it in the method. 👍🏼

apps/trading-e2e/tsconfig.json
```diff
@@ -12,7 +12,9 @@
     "noImplicitOverride": true,
     "noPropertyAccessFromIndexSignature": true,
     "noImplicitReturns": true,
-    "noFallthroughCasesInSwitch": true
+    "noFallthroughCasesInSwitch": true,
+    "jsx": "react-jsx",
```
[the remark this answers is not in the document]
jtsang586: I got JSX related errors when I tried to import  _truncateByChars_ from '@vegaprotocol/react-helpers'

Adding JSX to tsconfig fixed it. If there's a better way of doing this I'd be happy to know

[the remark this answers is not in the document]
mattrussell36: Suggested follow up, move react things (grid cells and theme context) to ui-toolkit, and rename react-helpers to utils?

apps/trading-e2e/src/support/step_definitions/trading-page.step.ts
```diff
@@ -55,14 +55,14 @@ Given('I am on the trading page for an active market', () => {
   mockMarket(MarketState.Active);
 
   cy.visit('/markets/market-id');
-  cy.wait('@Market');
+  cy.wait('@Market', { timeout: 3000 });
```
mattrussell36, line 58: question (non-blocking): Why are these needed now? Seemed to be working nicely without it before?

mattrussell36 (approved):

apps/trading-e2e/tsconfig.json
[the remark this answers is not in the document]
dexturr: IMO let's remove this, add another env property called truncatedVegaPubKey and use that :+1: to verify.

Feels odd to me to test that the key is displayed correctly we import the logic used to shorten in. If there was a bug with the logic these tests wouldn't find it. So I think I prefer it

apps/trading-e2e/src/support/step_definitions/trading-page.step.ts
jtsang586: I think I spotted it timing out a few times so I increased it by a bit. But looking now at the test run history I can't even find it anymore so probably not needed. Removed.

apps/trading-e2e/tsconfig.json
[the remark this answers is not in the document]
jtsang586: I just created an issue before I saw this. I'll go with Dex's suggestion and remove the import. Feel free to close the issue if not needed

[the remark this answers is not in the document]
jtsang586: I was tempted to just copy the truncate logic and use it for the test to reduce the amount of overhead if someone wanted to run the tests locally using their own wallets. I guess for now, ppl can just import the test wallet if they don't want to change the env variables. 
I've added the env variable now and removed the import and jsx from the config.

dexturr (approved):

PR #257, opened 2022-04-13
add scenarios for withdrawals in console v2

### LINKED ISSUE #259: Add scenarios for withdrawals - console v2

Ditmir-Vega:
closes #259

### DIFF
--- a/apps/trading-e2e/src/integration/withdrawals.feature
+++ b/apps/trading-e2e/src/integration/withdrawals.feature
@@ -1,0 +1,67 @@
+Feature: Withdrawals
+
+  Scenario: Can prepare a withdrawal
+    Given I navigate to withdrawals page
+    And I connect eth wallet
+    And I connect vega wallet
+    And I can see the withdrawals warning info
+    And I can see the from ethereum address field is automatically populated with current connected key
+    And I select asset 'tBtc'
+    And I input '100' tokens to withdraw
+    And the withdraw button is enabled
+    When I click the withdraw button
+    Then I am redirected to the finalise withdrawals page
+
+  Scenario: Can enter a different withdrawal address
+    Given I navigate to withdrawals page
+    And I can see an eth address is already filled in the withdraw to field
+    When I click the enter manually button
+    Then I can enter a new eth address
+
+  Scenario: Form field validation if fields are incomplete
+    Given I navigate to withdrawals page
+    When I click the withdraw button
+    Then I can see validation errors on incomplete fields
+
+  Scenario: Can prepare a withdrawal to send to another eth wallet
+    Given I navigate to withdrawals page
+    And I select asset 'tBtc'
+    And I can see an eth address is already filled in the withdraw to field
+    When I click the enter manually button
+    Then I can enter a new eth address
+    And I input '100' tokens to withdraw
+    And the withdraw button is enabled
+    When I click the withdraw button
+    Then I am redirected to the finalise withdrawals page
+
+  Scenario: Eth key validation on form
+    Given I navigate to withdrawals page
+    And I select asset 'tBtc'
+    When I click the enter manually button
+    And I enter eth address 'MMMMNNNN'
+    Then the invalid eth address error is shown
+
+  Scenario: Validation error if trying to withdraw more than available
+    Given I navigate to withdrawals page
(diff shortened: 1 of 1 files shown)


apps/trading-e2e/src/integration/withdrawals.feature
```diff
@@ -0,0 +1,67 @@
...
+    And I can see an eth address is already filled in the withdraw to field
+    When I click the enter manually button
+    Then I can enter a new eth address
+
+  Scenario: Form field validation if fields are incomplete
+    Given I navigate to withdrawals page
+    When I click the withdraw button
+    Then I can see validation errors on incomplete fields
+
+  Scenario: Can prepare a withdrawal to send to another eth wallet
```
campbellssource, line 26: I think this is a dupe of "Can enter a different withdrawal address"

apps/trading-e2e/src/integration/withdrawals.feature
```diff
@@ -0,0 +1,67 @@
+Feature: Withdrawals
+
+  Scenario: Can prepare a withdrawal
+    Given I navigate to withdrawals page
```
mattrussell36, line 4: I think its a bit superfluous to navigate to the page each time and just lengthens tests needlessly. I'd prefer to have a Given which just starts on the right page. We can have a different navigation feature which can navigate to each page and make sure that all works

```suggestion
    Given I am on the withdrawals page
```

apps/trading-e2e/src/integration/withdrawals.feature
```diff
@@ -0,0 +1,67 @@
...
+    And I connect eth wallet
+    And I connect vega wallet
+    And I can see the withdrawals warning info
```
mattrussell36, line 7: I think we should remove this step and instead have an additional scenario where the user has incomplete withdrawals.

It would look something like:
```
Given I am on the create withdrawal page
And I have incomplete withdrawals
Then I can navigate to the withdrawals page
```

apps/trading-e2e/src/integration/withdrawals.feature
```diff
@@ -0,0 +1,67 @@
...
+    And I can see the from ethereum address field is automatically populated with current connected key
+    And I select asset 'tBtc'
+    And I input '100' tokens to withdraw
+    And the withdraw button is enabled
+    When I click the withdraw button
+    Then I am redirected to the finalise withdrawals page
```
mattrussell36, line 13: This has changed a bit and the now the full withdrawal happens all on the form page. As a first pass I think we should just assert here that the transaction dialog shows with the correct txHash

apps/trading-e2e/src/integration/withdrawals.feature
```diff
@@ -0,0 +1,67 @@
...
+      | to        |
+      | created   |
+      | signature |
+    And I can see the finish withdrawals button
+
+  Scenario: Finish withdrawal to eth wallet
+    Given I navigate to finish withdrawals page
+    And I can see there are unfinished withdrawals
+    And I click on an unfinished withdrawal button
+    Then I approve transaction on ethereum
```
mattrussell36, line 66: We probably won't be able to test this until we have capsule

apps/trading-e2e/src/integration/withdrawals.feature
Ditmir-Vega: update as part of     feat/104-withdraws

apps/trading-e2e/src/integration/withdrawals.feature
Ditmir-Vega: update as part of     feat/104-withdraws

apps/trading-e2e/src/integration/withdrawals.feature
Ditmir-Vega: we already have such tests towards the bottom . in this scenario we are just checking we can prepare a withdrawal rather than complete it (phase 1)

apps/trading-e2e/src/integration/withdrawals.feature
Ditmir-Vega: I'll add a @manual or @ignore tag to this scenario

Ditmir-Vega:
moved to  feat/104-withdraws

PR #262, opened 2022-04-14
Test passing for trades list

### LINKED ISSUE #141: UI Testing - Trade list
Add Cypress test to cover order list related to: #84

jtsang586:
Add additional assertion to existing step for checking that the trades list is displayed with no missing values.

Once we have capsule running, we could potentially have some more detailed tests to verify specific trades getting updated. One for follow up ticket if possible to do. 

Closes #141

### DIFF
--- a/apps/trading-e2e/src/support/trading-windows/trades-list.ts
+++ b/apps/trading-e2e/src/support/trading-windows/trades-list.ts
@@ -16,7 +16,6 @@ export default class TradesList {
     cy.get(`[col-id=${this.colIdCreatedAt}]`).each(($tradeDateTime, index) => {
       if (index != 0) {
         //ignore header
-        cy.log(index);
         cy.wrap($tradeDateTime).invoke('text').should('match', dateTimeRegex);
       }
     });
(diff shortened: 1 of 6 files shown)


apps/trading-e2e/src/support/trading-windows/trades-list.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  verifyTradesListDisplayed() {
+    cy.get(`[col-id=${this.colIdPrice}]`).each(($tradePrice) => {
+      cy.wrap($tradePrice).invoke('text').should('not.be.empty');
+    });
+    cy.get(`[col-id=${this.colIdSize}]`).each(($tradeSize) => {
+      cy.wrap($tradeSize).invoke('text').should('not.be.empty');
+    });
+
+    const dateTimeRegex =
```
mattrussell36, line 14: nitpick (non blocking): Could just import in our date formatter function here

mattrussell36 (approved):
LGTM

apps/trading-e2e/src/support/trading-windows/trades-list.ts
jtsang586: I could be wrong but I think using regex to assert here is probably a bit more efficient.

dexturr (approved):

PR #266, opened 2022-04-19
Feature/151 orderbook

### LINKED ISSUE #151: Orderbook
### Acceptance Criteria

- Show full book (using new api with delta)
- Aggregation
  - Lowest. market tick size
  - Movement amount calculated using logarithmic scale (base 2 suggested constant)
  - Default to most precision (1 market tick)
- Layout handles auction mode (price crossing)
  - Auction
    - Mid point is indicative uncrossing price
    - Show uncrossing area on cumulative volume
  - Continuous
    - Mid point is midprice
- Histograms
  - Buy/Ask scales show the largest volume at level
  - Cumulative scales are showing sum of all orders of larger side (cumulative of course!)
  - Show cumulative volume number

### Notes

Generally the order book should show the **sum of orders at a given level**. 
for example in the following example there could be two parties with sell orders at 0.021 with volumes of 6 and 4 (meaning 10 in total.

```
price |  orders at price | sum of orders at this level and to the mid price* | side  
0.031 | 30 | 60 | sell 
0.020 | 20 | 30 | sell
0.021 | 10 | 10 | sell 

mid price: 0.020

0.019 |  5 |  5 | buy
0.018 |  5 | 10 | buy
0.009 |  5 | 15 | buy
```

## Aggregation 

Typically orders can be grouped or **aggregated** into a common level. ☝️ is shown at the tick level e.g. `0.001`.
this is shown aggregated at `0.01`
```
0.03 | 30 | 60 | sell 
0.02 | 30 | 40 | sell 

mid price: 0.020

0.01 | 10 | 15 | buy
0.09 |  5 | 15 | buy
```
Order aggregation may not be done in the first pass.

## Auction mode
The order book will need to behave significantly differently when the market is in limited order book mode Vs Auction
In limited order book mode prices below the best bid will always be bids and above the best ask will always be asks, because if orders cross they will be matched.

In auction the matching does not happen until the end of the auction.

The following image shows how a single order book could be constructed that works in both modes.

Screenshot 2022-04-06 at 16 15 15

This format is more common and space efficient if continuous trading is always the case.
Screenshot 2022-04-05 at 14 26 58

## Rows for each potential price or rows only where there are orders in the book.

In the above example there are no orders in the book between 0.03076 and 0.03073, and as a result there is no row. Should this be the case?

## Centring and scrolling
By default the order book should be centred on the mid price (or indicative uncrossing price if in auction) and when the mid changes the users should still be anchored on or near the mid. 

This would leave many prices out of view. a user may wish to see prices that are further away from the current mid. to see more prices there are a few options
- Aggregation (see above)
- scroll
- jump to sell side, buy side or mid

bglownia:
closes #151


bglownia:
closes #151

campbellssource:
Request (non-blocking): Find a way to stop the scroll bar from flashing each time the top/bottom of the book gets new rows

https://user-images.githubusercontent.com/13255539/164212991-5c378312-51d9-4203-adb9-9b351492e157.mov
in old console the scrollbars were all hidden with CSS.

campbellssource:
My top feature request (non-blocking): some sort of stickyness around the mid or indicative uncrossing.

campbellssource:
Question (non-blocking): What is is happening in the cumulative column at the moment? 
I don't understand why there is a Gap at A
What are the white bars indicating at B
Screenshot 2022-04-20 at 11 43 42

apps/trading/components/page-query-container/page-query-container.tsx
```diff
@@ -7,19 +7,22 @@ import { AsyncRenderer } from '@vegaprotocol/ui-toolkit';
 interface PageQueryContainerProps<TData, TVariables> {
   query: DocumentNode;
   options?: QueryHookOptions<TData, TVariables>;
-  children: (data: TData) => ReactNode;
+  render: (data: TData) => ReactNode;
```
bglownia, line 10: IMHO render function as a child is an anti pattern. It's much more clear to use `render()` than `children()` and <Component render={() => <Child/>}/> than <Component>{() => <Child/>}</Component>

libs/ui-toolkit/src/components/async-renderer/async-renderer.tsx
```diff
@@ -25,6 +27,6 @@ export function AsyncRenderer<T = object>({
   if (!data) {
     return <Splash>No data</Splash>;
   }
-
-  return <>{children(data)}</>;
+  // eslint-disable-next-line react/jsx-no-useless-fragment
+  return <>{render ? render(data) : children}</>;
```
bglownia, line 31: instead of 
`<AsyncRenderer data={data}>{(data) => <Child data={data}/>}<AsyncRenderer/>`
We can now use
`<AsyncRenderer data={data}render={(data) => <Child data={data} />}/>`
or 
`<AsyncRenderer  data={data}><Child data={data}/></AsyncRenderer >`

libs/market-list/src/lib/market-list-table.tsx
```diff
@@ -1,6 +1,6 @@
 import { forwardRef } from 'react';
 import type { ValueFormatterParams } from 'ag-grid-community';
-import { PriceCell, formatNumber, t } from '@vegaprotocol/react-helpers';
```
bglownia, line 3: PriceCell was renamed to PriceFlashCell, and PriceCell was added which formats price but don't flash on change.

libs/orderbook/src/lib/orderbook-manager.tsx
```diff
@@ -0,0 +1,81 @@
...
+        return false;
+      }
+      dataRef.current = updateCompactedData(
+        dataRef.current,
+        delta.sell,
+        delta.buy,
+        resolutionRef.current
+      );
+      const now = new Date().getTime();
+      if (now - lastUpdateRef.current > 1000) {
```
bglownia, line 40: Render not often than once per second.
Subscription pass single buy or sell update  in high frequency
We can also batch delta sell and buy to make this more efficient, and then run  updateCompactedData or request changes in API to batch updates and lower frequency

libs/orderbook/src/lib/orderbook-manager.tsx
```diff
@@ -0,0 +1,81 @@
...
+}
+
+export const OrderbookManager = ({
+  marketId,
+  resolution,
+}: OrderbookManagerProps) => {
+  const variables = useMemo(() => ({ marketId }), [marketId]);
+  const resolutionRef = useRef(resolution);
+  const lastUpdateRef = useRef(new Date().getTime());
+  const [orderbookData, setOrderbookData] = useState<OrderbookData[] | null>(
```
bglownia, line 22: In data provider only raw data is stored. It's compacted in the component and stored as orderbookData.ref.
This allows to:
 - use same data in more than one component (ie grapth which uses market depth also)
 - change resolution of price level grouping without need of refetching initial data
 
 It could be done in data provider getData and update methods, but I find this way much more  effective

libs/orderbook/src/lib/orderbook-manager.tsx
```diff
@@ -0,0 +1,81 @@
...
+    dataRef.current = compact(
+      data.depth.sell,
+      data.depth.buy,
+      resolution
+    );
+    setOrderbookData(dataRef.current);
+  }, [data, resolution]);
+
+  useEffect(() => {
+    resolutionRef.current = resolution;
```
bglownia, line 70: If resolution is changed we flush data from provider and they will go through above useEffect hook which will compact them.

libs/orderbook/src/lib/orderbook.spec.tsx
```diff
@@ -0,0 +1,10 @@
+import { render } from '@testing-library/react';
+
+import Orderbook from './orderbook';
+
+describe('Orderbook', () => {
+  it('should render successfully', () => {
+    const { baseElement } = render(<Orderbook data={null} decimalPlaces={4} />);
```
bglownia, line 7: I did not added any more module tests, because I expect changes in data presentation layer. Everything else is covered by unit tests.

libs/react-helpers/src/hooks/use-data-provider.ts
```diff
@@ -27,7 +33,9 @@ export function useDataProvider<Data, Delta>(
     [update]
   );
   useEffect(() => {
-    return dataProvider(callback, client, variables);
+    const { unsubscribe, flush } = dataProvider(callback, client, variables);
+    flushRef.current = flush;
+    return unsubscribe;
   }, [client, initialized, dataProvider, callback, variables]);
-  return { data, loading, error };
+  return { data, loading, error, flush };
```
bglownia, line 40: flush can be used to get updated data from provider even if component passes update function

libs/orderbook/src/lib/orderbook.spec.tsx
mattrussell36: Fine with me

libs/orderbook/src/lib/orderbook-manager.tsx
```diff
@@ -0,0 +1,77 @@
...
+  );
+  const dataRef = useRef<OrderbookData[] | null>(null);
+
+  // Apply updates to the table
+  const update = useCallback(
+    (delta: MarketDepthSubscription_marketDepthUpdate) => {
+      if (!dataRef.current) {
+        return false;
+      }
+      dataRef.current = updateCompactedData(
```
mattrussell36, line 33: chore (blocking): Its not immediately clear why we need dataRef and can't just use some react state. Could you add some comments to clarify

libs/ui-toolkit/src/components/async-renderer/async-renderer.tsx
mattrussell36: I mentioned I another comment but this thread is a better spot for it... 

I quite like enforcing to use at least a render prop (I don't mind if its `render` or `children`). This way in the body of the render prop function TS can infer that data is indeed truthy. Otherwise we end up writing pointless null checks
```
// TS cant infer data is truthy
<AsyncRenderer ... >
  {data && <Child />
</AsyncRenderer>

// TS can infer this though
<AsyncRenderer ... 
  render={(data) => (
    <Child data={data} />
  )
/>
```

libs/react-helpers/src/lib/grid/vol-cell.tsx
```diff
@@ -0,0 +1,42 @@
...
+  }
+  return (
+    <div className="relative" data-testid="vol">
+      <div
+        className="h-full absolute top-0 left-0"
+        style={{
+          width: relativeValue ? `${relativeValue}%` : '0%',
+          backgroundColor: type === 'bid' ? BID_COLOR : ASK_COLOR,
+        }}
+      ></div>
```
mattrussell36, line 30: nitpick (non-blocking): Can make this self clsoing
```suggestion
      />
```

libs/react-helpers/src/lib/grid/price-flash-cell.test.tsx
```diff
@@ -0,0 +1,21 @@
+import { render, screen } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+import * as React from 'react';
```
mattrussell36, line 3: chore (non-blocking): React 17 we shouln't need to import react unless its used, same goes for other files in the PR. (PS I realize we are unnecessarily importing it in other files 🕊️ )
```suggestion

```

libs/react-helpers/src/lib/grid/cumulative-vol-cell.tsx
```diff
@@ -0,0 +1,67 @@
...
+      <div
+        className="h-full absolute top-0 right-0"
+        style={{
+          width: `${relativeBid}%`,
+          backgroundColor:
+            relativeAsk && relativeAsk > relativeBid
+              ? INTERSECT_COLOR
+              : BID_COLOR,
+        }}
+      ></div>
```
mattrussell36, line 29: chore (non-blocking): Could self close some of these divs in this file
```suggestion
      />
```

libs/react-helpers/src/lib/grid/cumulative-vol-cell.tsx
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
+import type { ICellRendererParams } from 'ag-grid-community';
+
+import { BID_COLOR, ASK_COLOR } from './vol-cell';
+
+const INTERSECT_COLOR = 'darkgray';
```
mattrussell36, line 6: suggestion (blocking): Can you import one of the grays from our tailwind theme?

libs/react-helpers/src/lib/grid/vol-cell.tsx
```diff
@@ -0,0 +1,42 @@
...
+  value: number | bigint | null | undefined;
+  relativeValue?: number;
+  type: 'bid' | 'ask';
+}
+export interface IVolCellProps extends ICellRendererParams {
+  value: number | bigint | null | undefined;
+  valueFormatted: Omit<VolProps, 'value'>;
+}
+
+export const BID_COLOR = 'darkgreen';
```
mattrussell36, line 15: chore (non-blocking): Can we use tailwind theme colors here?

libs/react-helpers/src/lib/grid/cumulative-vol-cell.tsx
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
+import type { ICellRendererParams } from 'ag-grid-community';
+
+import { BID_COLOR, ASK_COLOR } from './vol-cell';
```
mattrussell36, line 4: suggstion (blocking): Use colors from tailwind theme

libs/orderbook/src/lib/orderbook-data.ts
```diff
@@ -0,0 +1,310 @@
...
+    data.cumulativeVol.relativeAsk = toPercentValue(
+      data.cumulativeVol.ask / cumulativeVol
+    );
+    data.cumulativeVol.relativeBid = toPercentValue(
+      data.cumulativeVol.bid / cumulativeVol
+    );
+  });
+};
+
+export const compact = (
```
mattrussell36, line 68: chore (blocking): Could really use some comments at the top of these functions outlining whats happening

libs/orderbook/src/lib/orderbook-data.ts
```diff
@@ -0,0 +1,310 @@
...
+        | MarketDepthSubscription_marketDepthUpdate_buy
+      )[]
+    | null,
+  resolution: number
+) => {
+  let cumulativeVol = 0;
+  const askOrderbookData = [...(sell ?? [])]
+    .sort((a, b) => Number(a.price) - Number(b.price))
+    .map<OrderbookData>((sell) => ({
+      price: Number(sell.price),
```
mattrussell36, line 87: suggestion (blocking, but could also potentially be some follow up work): Lots of math going on here with numbers that potentially have lots of decimal places. We should probably use BigNumber, this will have the knock on effect of having to use BigNumbers methods like .plus rather than regular JS operations.

```suggestion
      price: new BigNumber(sell.price),
```

mattrussell36 (changes requested):

libs/orderbook/src/lib/orderbook-row.tsx
```diff
@@ -0,0 +1,58 @@
...
+    relativeBidVol,
+    price,
+    askVol,
+    relativeAskVol,
+    decimalPlaces,
+    cumulativeRelativeAskVol,
+    cumulativeRelativeBidVol,
+  }: OrderbookRowProps) => {
+    return (
+      <>
```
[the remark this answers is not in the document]
bglownia: I'm using display: grid, I will need to switch to 
```
<div className="flex" role="row">
  <div className="grow-0 shrink-0 w-1/4"  role="gridcell"></div>
  <div className="grow-0 shrink-0 w-1/4" role="gridcell"></div>
  <div className="grow-0 shrink-0 w-1/4" role="gridcell"></div>
  <div className="grow-0 shrink-0 w-1/4" role="gridcell"></div>
</div>
```
instead of 
```
<>
  <div></div>
  <div></div>
  <div></div>
  <div></div>
</>
```

libs/ui-toolkit/src/components/async-renderer/async-renderer.tsx
bglownia: I know what you mean, but TS will not infer if data is null also when it will be used in render function
image

Only way to make it look nice is to allow <Child /> to accept nullish data, and then we can use both render function or child notation

libs/react-helpers/src/lib/grid/vol-cell.tsx
bglownia: @campbellssource
Bar background colors are not present on styleguide colors palette. We need to extend palette or amend colors.

libs/orderbook/src/lib/orderbook-data.ts
bglownia: I changed price type to string and I use BigNumber(price) for to compare which one is greater than the other

mattrussell36 (approved):

libs/react-helpers/src/lib/grid/cumulative-vol-cell.tsx
mattrussell36: https://github.com/vegaprotocol/frontend-monorepo/issues/297

libs/react-helpers/src/lib/grid/cumulative-vol-cell.tsx
mattrussell36: https://github.com/vegaprotocol/frontend-monorepo/issues/297

libs/react-helpers/src/lib/grid/vol-cell.tsx
mattrussell36: https://github.com/vegaprotocol/frontend-monorepo/issues/297

libs/ui-toolkit/src/components/async-renderer/async-renderer.tsx
mattrussell36: It will but you have to pass AsyncRenderer a generic type argument. Regardless I dont think this is important enough to block the PR 😄

libs/market-depth/src/lib/market-depth-mock.ts
```diff
@@ -0,0 +1,57 @@
...
+  return new Array(numberOfRecords)
+    .fill(null)
+    .map(() => depthRow(price - Math.round(Math.random() * distance)));
+};
+
+export const getMockedData = (id?: string): MarketDepth_market => ({
+  __typename: 'Market',
+  id: id || '',
+  decimalPlaces: 2,
+  // "positionDecimalPlaces": 0,
```
dexturr, line 42:
```suggestion
```

libs/react-helpers/src/lib/grid/vol-cell.tsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import type { ICellRendererParams } from 'ag-grid-community';
+import { PriceCell } from './price-cell';
+
+export interface VolProps {
+  value: number | bigint | null | undefined;
+  relativeValue?: number;
+  type: 'bid' | 'ask';
```
dexturr, line 8: chore (non-blocking): might be worth enuming this?

libs/react-helpers/src/lib/grid/price-flash-cell.tsx
```diff
@@ -0,0 +1,22 @@
...
+    return <span data-testid="price">-</span>;
+  }
+  return (
+    <span className="font-mono" data-testid="price">
+      <FlashCell value={Number(value)}>{valueFormatted}</FlashCell>
+    </span>
+  );
+};
+
+PriceFlashCell.displayName = 'PriceFlashCell';
```
dexturr, line 22: thought (non-blocking): I wonder if we can enforce this in some way? 🤔

libs/react-helpers/src/lib/generic-data-provider.ts
```diff
@@ -68,84 +83,122 @@ function makeDataProviderInternal<QueryData, Data, SubscriptionData, Delta>(
...
-        subscription = undefined;
-      }
-      data = null;
-      error = undefined;
-      loading = false;
+      reset();
     }
   };
 
+  //
```
dexturr, line 201:
```suggestion
```

dexturr (approved):
Great stuff 👏

PR #282, opened 2022-04-20
frontend-monorepo-267 Reduce total staked to 2 decimal precision

### LINKED ISSUE #267: Reduce total staked to 2 decimal precision
As discussed in the slack thread linked below. On the stats page the number for total staked is currently 18 d.p. which is far more than what people would realistically be interested in. 

https://vegaprotocol.slack.com/archives/C8WS7M50E/p1650373450856089

sam-keen:
Closes #267


dexturr (changes requested):
Screenshot 2022-04-20 at 15 31 36

We need to add a format decimal places to the number format function, this should set both minDecimals and maxDecimals in number.intl to ensure that it always renders the passed in number of d.p. probably want this to default to number of decimals to add to the number

libs/react-helpers/src/lib/format/number.ts
```diff
@@ -2,11 +2,15 @@ import { BigNumber } from 'bignumber.js';
 import memoize from 'lodash/memoize';
 import { getUserLocale } from './utils';
 
-export function addDecimal(value: string, decimals: number): string {
+export function addDecimal(
+  value: string,
+  decimals: number,
+  decimalPrecision?: number
```
dexturr, line 8: I forget the synax for default with TS so double check this but: suggest defaulting here

```suggestion
  decimalPrecision = decimals: number
```

dexturr (approved):

edd (approved):

PR #290, opened 2022-04-22
UI test for positions table

### LINKED ISSUE #191: Fix Ts ignores in Cypress library
chore (blocking): Far too many TS ingores for my comfort here. Let's jump on a call and see if we can sort this together, should be easy enough to sort out together :D

_Originally posted by @dexturr in https://github.com/vegaprotocol/frontend-monorepo/pull/145#discussion_r839811703_

Update: After spending a small of time to fix this it became clear it wasn't going to be an easy fix so converting into separate issue 

This might be useful to use as reference:
https://github.com/dmtrKovalenko/cypress-real-events

jtsang586:
UI test for validating the positions table is displayed using mocks. 

Closes #191

### DIFF
--- a/apps/trading-e2e/src/support/mocks/generate-positions.ts
+++ b/apps/trading-e2e/src/support/mocks/generate-positions.ts
@@ -3,7 +3,7 @@ import type { PartialDeep } from 'type-fest';
 import type {
   Positions,
   Positions_party_positions,
-} from '../lib/__generated__/Positions';
+} from '@vegaprotocol/positions';
 import { MarketTradingMode } from '@vegaprotocol/types';
 
 export const generatePositions = (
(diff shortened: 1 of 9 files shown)


apps/trading-e2e/src/support/mocks/generate-positions.ts
```diff
@@ -0,0 +1,128 @@
...
+        },
+        __typename: 'Market',
+      },
+      __typename: 'Position',
+    },
+  ];
+
+  const defaultResult = {
+    party: {
+      id: 'f9982447fb4128f9968f9981612c5ea85d19b62058ec2636efc812dcbbc745ca',
```
dexturr, line 121: chore (non-blocking): Can we pull this from the cypress environment? If the user is already configuring the connected key in env seems like a more sensible default

jtsang586: Done

mattrussell36 (approved):

bglownia (approved):

PR #295, opened 2022-04-25
Run nx format

dexturr:
Red build on master (mostly) due to prettier not being run on the token site before merging.

Run `nx format --all`


edd (approved):
Approved without properly reviewing because... who has time for reviewing automatic formatting.

jtsang586 (approved):

PR #308, opened 2022-04-26
Check for 2dp on total staked

jtsang586:
Small change to an existing test to check that total staked is now displayed as a number with 2 decimal precision. 

Closes: #267


dexturr (approved):

mattrussell36 (approved):
I think theres probably a better way of doing this, perhaps updating the data testid to contain the field id so we can target it directly, but this is fine for now

PR #311, opened 2022-04-27
Mock accounts for trading page

mattrussell36:
- Move accounts and positions into own feature
- Move accounts and positions steps to trading page
- Add mock query result for accounts

### DIFF
--- a/apps/trading-e2e/src/integration/accounts.feature
+++ b/apps/trading-e2e/src/integration/accounts.feature
@@ -1,7 +1,0 @@
-Feature: Accounts table 
-  Scenario: Accounts displayed when connected to wallet
-    Given I am on the trading page for an active market
-    And I connect to Vega Wallet
-    When I click on accounts tab
-    Then accounts are displayed
-    And I can see account for tEURO
--- a/apps/trading-e2e/src/integration/accounts.feature
+++ b/apps/trading-e2e/src/integration/accounts.feature
@@ -1,0 +1,7 @@
+Feature: Accounts table 
+  Scenario: Accounts displayed when connected to wallet
+    Given I am on the trading page for an active market
+    And I connect to Vega Wallet
+    When I click on accounts tab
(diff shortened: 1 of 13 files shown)


apps/trading-e2e/src/support/mocks/generate-accounts.ts
```diff
@@ -0,0 +1,30 @@
...
+import { AccountType } from '@vegaprotocol/types';
+import type { PartialDeep } from 'type-fest';
+
+export const generateAccounts = (
+  override?: PartialDeep<Accounts>
+): Accounts => {
+  const defaultAccounts: Accounts = {
+    party: {
+      __typename: 'Party',
+      id: 'party-id',
```
dexturr, line 12: chore (blocking): pull party id from config as per other generate files?

apps/trading-e2e/src/integration/accounts.feature
```diff
@@ -0,0 +1,7 @@
+Feature: Accounts table 
+  Scenario: Accounts displayed when connected to wallet
+    Given I am on the trading page for an active market
+    And I connect to Vega Wallet
+    When I click on accounts tab
+    Then accounts are displayed
+    And I can see account for tEURO
```
jtsang586, line 7: Question (non-blocking): What do you think about moving the tests for the trading tabs (accounts/orders/positions etc.) into a single trading page feature file? 
It's quite expensive time wise to have lots of feature files. Cypress tears down and reloads an instance of Chrome for each feature file in a test run so I'd prefer to have as less as possible - as long as it makes sense.

dexturr: This will become doubly true when capsule is a thing and we have to remove parallelisation from the tests (hopefully we can have it back at a later date)

dexturr (approved):

jtsang586 (approved):

PR #263, opened 2022-04-19
Feat/104 withdraws

mattrussell36:

PR #322, opened 2022-04-27
Allow any unknown asset query to be used with ERC20Asset type guard

mattrussell36:
Addresses: https://github.com/vegaprotocol/frontend-monorepo/pull/263#discussion_r858719566


dexturr (approved):

edd (approved):
👍 I was going to say it needs a comment, but it's readable. Good feedback, neat fix.

PR #334, opened 2022-05-03
Disable SRI for stats

edd:
Subresource Integrity breaks CSS for stats

Disabling the NX build option for SRI means the CSS loads correctly. We want SRI, but
having the stats page working is more important. This PR disables SRI. A followup ticket
will be created to work out what was wrong.

- Disable SRI in stats build
- Update fleek build command to match other projects


dexturr (approved):

sam-keen (approved):
Thanks for figuring this out Edd!

mattrussell36 (approved):

PR #336, opened 2022-05-03
Ci/generate queries

dexturr:


queries.graphql
```diff
@@ -0,0 +1,659 @@
+# File: apps/token/src/components/vega-wallet/hooks.ts
```
dexturr, line 1: Delete

.github/workflows/generate-queries.yml
```diff
@@ -0,0 +1,30 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Use Node.js 16
+        id: Node
+        uses: actions/setup-node@v2
+        with:
+          node-version: 16.14.0
```
sam-keen, line 22: Question (non-blocking): Is it possible to grab the node version from the package.json rather than hard code it?

scripts/get-queries.js
```diff
@@ -0,0 +1,38 @@
...
+  });
+};
+
+const run = async () => {
+  const files1 = await globPromise('apps/**/*.ts');
+  const files2 = await globPromise('libs/**/*.ts');
+  processFiles([...files1, ...files2]);
+};
+
+run();
```
sam-keen, line 38: Suggestion (non-blocking): bit cleaner as an IIFE

sam-keen (approved):

.github/workflows/generate-queries.yml
dexturr: Package specifies a range not a specific version so not really. We should move this to matrix at some point but for now let's see if this provides any value before we do that :+1:

mattrussell36 (approved):

PR #374, opened 2022-05-04
Fix for failing test in CI

jtsang586:
Small fix to see if it fixes a market test failing in CI

### DIFF
--- a/apps/trading-e2e/src/support/pages/markets-page.ts
+++ b/apps/trading-e2e/src/support/pages/markets-page.ts
@@ -57,7 +57,6 @@ export default class MarketPage extends BasePage {
   clickOnMarket(text: string) {
     cy.get(`[col-id=${this.marketStateColId}]`)
       .contains(text)
-      .should('be.visible')
       .click({ force: true });
   }
 }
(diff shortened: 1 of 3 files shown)


apps/trading-e2e/src/support/pages/markets-page.ts
```diff
@@ -54,6 +54,6 @@ export default class MarketPage extends BasePage {
   }
 
   clickOnMarket(text: string) {
-    cy.contains(text).click();
+    cy.contains(text).click({ force: true });
```
Ditmir-Vega, line 57: This function simply just finds any text within the page. we should add a data-testid to the markets table i.e 'data-testid="markets-table"' and then only click within the table for the market 

something a little like this 

```
cy.getByTestId('markets-table')
  .within((marketState: string) => {
    return cy.get[(role="row"]).contains(marketState).click()
  })
```

Please raise this as an improvement

Ditmir-Vega (approved):

apps/trading-e2e/src/support/pages/markets-page.ts
jtsang586: I don't think we can use test ids here. Markets change all the time so there's no way to know which one is active or suspended etc. The test just wants to click on any market that is active for example which I think is fine the way it's done here.

dexturr (approved):

apps/trading-e2e/src/support/pages/markets-page.ts
jtsang586: Done

PR #376, opened 2022-05-05
Feature/vega core

elmar-vega:
This PR is to add the bare functionality and setup for the simple trading app (name change pending).
We are using many of the components from the deal-ticket library and have kept refactors to a minimal. There should be no breaking changes for the other apps and where refactoring has taken place, the defaults have been kept.


apps/simple-trading-app/project.json
```diff
@@ -0,0 +1,73 @@
+{
```
dexturr, line 1: chore (blocking): somewhere in here you will need to set a different port from default for testing purposes, as the test run in parallel if everything tries to start on the same port we get errors

libs/deal-ticket/src/index.ts
```diff
@@ -1,5 +1,13 @@
+export * from './expiry-selector';
```
dexturr, line 1: question (non-blocking): why are we now exporting everything?

dexturr (changes requested):

apps/simple-trading-app/src/app/components/deal-ticket/deal-ticket-container.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+export const DealTicketContainer = ({ marketId }: DealTicketContainerProps) => {
+  const { data, loading, error } = useQuery(DEAL_TICKET_QUERY, {
+    variables: { marketId },
+  });
+
+  return (
+    <AsyncRenderer<DealTicketQuery> data={data} loading={loading} error={error}>
+      {data && data.market ? (
+        <DealTicketManager market={data.market}>
```
mattrussell36, line 45: suggestion (non-blocking) This is almost identical to deal-ticket-container in the lib/deal-ticket. Perhaps we add a render prop so we can provide different form components to be rendered but allow the 'container' to make the fetch request and the 'manager' to handle order state..

Something like:
```
// simple-trading-app
<DealTicketContainer
  render={({ market, submit }) => <DealTicketSteps market={market} submit={submit} />}
/>
// trading
<DealTicketContainer
  render={({ market, submit }) => <DealTicket market={market} submit={submit} />}
/>

```

apps/simple-trading-app/project.json
```diff
@@ -0,0 +1,73 @@
...
+  "root": "apps/simple-trading-app",
+  "sourceRoot": "apps/simple-trading-app/src",
+  "projectType": "application",
+  "targets": {
+    "build": {
+      "executor": "@nrwl/web:webpack",
```
macqbat, line 7: So finally are we sure, we don't want to do it as next.js?

package.json
```diff
@@ -13,6 +13,9 @@
   },
   "private": true,
   "dependencies": {
+    "@emotion/react": "^11.9.0",
+    "@emotion/styled": "^11.8.1",
+    "@mui/material": "^5.6.2",
```
macqbat, line 18: non-blocking: Do we really need this package (another one set of components) similar to blueprintjs and internal ui-toolkit? Wouldn't it be better to keep one "component facilitator" than many of them in one repo?

macqbat: moreover, tailwind.css alone offers many possibilities

macqbat (approved):

mattrussell36 (approved):

dexturr (approved):

PR #387, opened 2022-05-09
begin porting over old chain explorer api

### LINKED ISSUE #198: Import explorer/api to monorepo
https://github.com/vegaprotocol/explorer/tree/master/api is currently deployed to a different netlify site from the actual explorer. It's a bit messy. Let's add it here so we can redirect all of explorer.vega.trading to .xyz

Next: https://github.com/vegaprotocol/docs/pull/323

dexturr:
Closes #198


dexturr:
This PR would mean that the function is deployed for every single site. We could skirt around this by using bash scripts to remove functions for all apps that are not explorer how IMO this is worse than a single repo with the chain explorer api in it. Closing this PR.

PR #388, opened 2022-05-09
Feat/dockerize frontends

### LINKED ISSUE #296: Dockerizing frontends for capsule usage

dexturr:
Closes #296


apps/explorer/src/app/config/env.ts
```diff
@@ -0,0 +1,38 @@
...
+  envName: windowOrDefault('NX_VEGA_ENV'),
+  dataSources: {
+    chainExplorerUrl: windowOrDefault('NX_CHAIN_EXPLORER_URL'),
+    tendermintUrl: windowOrDefault('NX_TENDERMINT_URL'),
+    tendermintWebsocketUrl: windowOrDefault('NX_TENDERMINT_WEBSOCKET_URL'),
+    dataNodeUrl: windowOrDefault('NX_VEGA_URL'),
+    restEndpoint: windowOrDefault('NX_VEGA_REST'),
+  },
+  flags: {
+    assets: truthy.includes(windowOrDefault('NX_EXPLORER_ASSETS') || ''),
```
sam-keen, line 24: Question (non-blocking): Why do you need to pass in an empty string as an or? Presumably `windowOrDefault` will either return a string or an `undefined`, either of which will be falsey when checked against the `truthy` array?

sam-keen (approved):

sam-keen (commented):
Doesn't look like the funnest task so far! Looking good though, just one question at this stage

apps/explorer/src/app/config/env.ts
```diff
@@ -0,0 +1,34 @@
...
+  }
+  return (process.env[key] as string) || '';
+};
+
+const truthy = ['1', 'true'];
+
+export const ENV = {
+  // Data sources
+  // Environment
+  nodeEnv: windowOrDefault('NODE_ENV'),
```
notbot00, line 13: i think some env variables are build time only by nature, like the NODE_ENV here, for example. this is usually supposed to describe the node environment the bundle was built in, which we can't really change from within the runtime.

MadalinaRaicu (approved):

Issue #274, opened 2022-04-20
Allow Callout component to accept ReactNode or IconName


mattrussell36:
Currently Callouts only allow an IconName to be passed as the prop:
`<Callout iconName='hand-up' />`

However sometimes we need to be able to render something completely custom inside, for example a `<Loader />`:
`<Callout icon={<Loader />} />`

It would be good to allow Callouts's props to take either an `IconName` or a ReactNode for something custom. Props could end up something like:
```
interface CalloutProps {
  ...
  icon?: IconName | ReactNode;
}
```

Some logic will be needed to render an `<Icon />` if an IconName is passed in otherwise whatever is passed in.

- [ ] Update Callout to accept generic element for icons

PR #394, opened 2022-05-09
Feat/callout icon

notbot00:
# Related issues 🔗

Closes [#274](https://github.com/vegaprotocol/frontend-monorepo/issues/274)

# Description ℹ️

Adds an icon prop to the callout component.

# Technical

New prop called `icon` as a react node. The callout now takes either an optional `icon` or an optional `iconName`, and renders either the passed in element, or an Icon component with the passed in name.

Screen Shot 2022-05-09 at 7 23 51 PM

### DIFF
--- a/libs/ui-toolkit/src/components/callout/callout.tsx
+++ b/libs/ui-toolkit/src/components/callout/callout.tsx
@@ -1,23 +1,62 @@
+import type { ReactNode } from 'react';
 import classNames from 'classnames';
 import { getIntentShadow, Intent } from '../../utils/intent';
 import type { IconName } from '../icon';
 import { Icon } from '../icon';
 
-export interface CalloutProps {
+interface CalloutRootProps {
   children?: React.ReactNode;
   title?: React.ReactElement | string;
   intent?: Intent;
-  iconName?: IconName;
   headingLevel?: 1 | 2 | 3 | 4 | 5 | 6;
 }
 
-export function Callout({
-  children,
-  title,
-  intent = Intent.Help,
-  iconName,
-  headingLevel,
-}: CalloutProps) {
+interface CalloutWithoutIcon extends CalloutRootProps {
+  iconName?: never;
+  icon?: never;
+}
+
+interface CalloutPropsWithIconName extends CalloutRootProps {
+  children?: React.ReactNode;
+  title?: React.ReactElement | string;
+  intent?: Intent;
+  headingLevel?: 1 | 2 | 3 | 4 | 5 | 6;
+  iconName: IconName;
+  icon?: never;
+}
+
+interface CalloutPropsWithIcon extends CalloutRootProps {
+  children?: React.ReactNode;
+  title?: React.ReactElement | string;
+  intent?: Intent;
+  headingLevel?: 1 | 2 | 3 | 4 | 5 | 6;
+  iconName?: never;
+  icon: ReactNode;
+}
+
+type CalloutProps =
+  | CalloutWithoutIcon
+  | CalloutPropsWithIconName
+  | CalloutPropsWithIcon;
+
+const getIconElement = (props: CalloutProps) => {
+  if (props.iconName) {
+    return (
+      <Icon
+        name={props.iconName}
+        className="fill-current ml-8 mr-16 mt-8"
+        size={20}
+      />
+    );
+  }
+  return props.icon;
+};
+
+export function Callout(props: CalloutProps) {
+  const { children, title, intent = Intent.Help, headingLevel } = props;
+
+  const icon = getIconElement(props);
+
   const className = classNames(
     'border',
     'border-black',
@@ -27,15 +66,12 @@ export function Callout({
     'p-16',
     getIntentShadow(intent),
     {
-      flex: !!iconName,
+      flex: !!icon,
     }
   );
   const TitleTag: keyof JSX.IntrinsicElements = headingLevel
     ? `h${headingLevel}`
     : 'div';
-  const icon = iconName && (
-    <Icon name={iconName} className="fill-current ml-8 mr-16 mt-8" size={20} />
-  );
   const body = (
     <>
       {title && <TitleTag className="text-h5 mt-0 mb-8">{title}</TitleTag>}
--- a/libs/ui-toolkit/src/components/callout/callout.stories.tsx
+++ b/libs/ui-toolkit/src/components/callout/callout.stories.tsx
@@ -74,3 +74,18 @@ IconAndContent.args = {
     </div>
   ),
 };
+
+export const CustomIconAndContent = Template.bind({});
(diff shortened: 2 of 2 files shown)


libs/ui-toolkit/src/components/callout/callout.tsx
```diff
@@ -1,23 +1,62 @@
...
-  intent = Intent.Help,
-  iconName,
-  headingLevel,
-}: CalloutProps) {
+interface CalloutWithoutIcon extends CalloutRootProps {
+  iconName?: never;
+  icon?: never;
+}
+
+interface CalloutPropsWithIconName extends CalloutRootProps {
```
sam-keen, line 19: Note (non-blocking, see below for further (blocking) refactor comments): You're extending `CalloutRootProps`, so the only props you need in here are `iconName` and `icon`. Same for `CallourPropsWithIcon`.

libs/ui-toolkit/src/components/callout/callout.tsx
```diff
@@ -1,23 +1,62 @@
...
+interface CalloutPropsWithIcon extends CalloutRootProps {
+  children?: React.ReactNode;
+  title?: React.ReactElement | string;
+  intent?: Intent;
+  headingLevel?: 1 | 2 | 3 | 4 | 5 | 6;
+  iconName?: never;
+  icon: ReactNode;
+}
+
+type CalloutProps =
```
sam-keen, line 37: Chore (blocking): I like the self-documenting code approach, but as far as I can tell, the benefit of having these different interfaces is that you've defined a situation where you never have an `iconName` and `icon` prop simultaneously. This isn't necessary, as your code in `getIconElement` handles that scenario. Therefore, we only need a single interface:

`interface CalloutProps {
  children?: React.ReactNode;
  title?: React.ReactElement | string;
  intent?: Intent;
  headingLevel?: 1 | 2 | 3 | 4 | 5 | 6;
  iconName?: IconName;
  icon?: ReactNode;
}`

libs/ui-toolkit/src/components/callout/callout.tsx
```diff
@@ -1,23 +1,62 @@
...
+  | CalloutWithoutIcon
+  | CalloutPropsWithIconName
+  | CalloutPropsWithIcon;
+
+const getIconElement = (props: CalloutProps) => {
```
sam-keen, line 42: Chore (blocking): Let's reduce this to only have `iconName` and `icon` props passed in here

libs/ui-toolkit/src/components/callout/callout.tsx
```diff
@@ -1,23 +1,62 @@
...
+  if (props.iconName) {
+    return (
+      <Icon
+        name={props.iconName}
+        className="fill-current ml-8 mr-16 mt-8"
+        size={20}
```
sam-keen, line 48: Chore (blocking): This icon will normally convey semantic value, which makes this a good opportunity to improve accessibility. Let's include an optional `iconDescription` prop (optional because it's possible that the icon could be decorational, and can be ignored by assistive tech if so). If `iconDescription`, use it with the `aria-label` attribute. If not, add `aria-hidden="true"`. 

There's a good article in icon accessibility at https://fontawesome.com/docs/web/dig-deeper/accessibility

libs/ui-toolkit/src/components/callout/callout.tsx
```diff
@@ -27,15 +66,12 @@ export function Callout({
     'p-16',
     getIntentShadow(intent),
     {
-      flex: !!iconName,
+      flex: !!icon,
```
sam-keen, line 69: Chore (blocking): This will then need to become `flex: !!iconElement`

sam-keen (changes requested):

libs/ui-toolkit/src/components/callout/callout.tsx
notbot00: I think it is important to realize that my implementation is not equivalent with your suggestion. 
As a consumer of the Callout component, I'd like to avoid any potential confusion. If I do what you suggest here, that would allow to pass in both icon and iconName as props, but we never render both at the same time.
If we allow passing in both icon and iconName to the component, we would be left relying on the internal implementation to handle which would be overriding the other. In the runtime I am still making an arbitrary choice, of course, but it would not let tsc pass.
just had a look and realized you don't run typecheck in the pipelines, so not sure how much value this provides, other then assuming the devs have tsc hooked up in their IDE correctly and they see error messages in there.

libs/ui-toolkit/src/components/callout/callout.tsx
notbot00: okay, changed.

libs/ui-toolkit/src/components/callout/callout.tsx
```diff
@@ -1,23 +1,62 @@
...
+      />
+    );
+  }
+  return props.icon;
+};
+
+export function Callout(props: CalloutProps) {
+  const { children, title, intent = Intent.Help, headingLevel } = props;
+
+  const icon = getIconElement(props);
```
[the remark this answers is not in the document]
notbot00: changed

libs/ui-toolkit/src/components/callout/callout.tsx
notbot00: and changed

libs/ui-toolkit/src/components/callout/callout.tsx
notbot00: i've replied to this below, as it's about the exact same topic as your other comment i believe.

libs/ui-toolkit/src/components/callout/callout.tsx
notbot00: nice, added to the icon props

libs/ui-toolkit/src/components/callout/callout.tsx
sam-keen: No, this is an implementation detail - you're extending `CalloutRootProps`, but redefining the same props here (`headingLevel`, `title` etc). You do the same with `CalloutPropsWithIcon`. You've done it properly on  `CalloutWithoutIcon` however.

libs/ui-toolkit/src/components/callout/callout.tsx
sam-keen: Ok cool, if you make the change to take advantage of extending CalloutRootProps in each interface then this gets a tick from me

libs/ui-toolkit/src/components/callout/callout.tsx
notbot00: i see, sorry, missed that part. totally right, removed the redundancies.

mattrussell36 (approved):

sam-keen (approved):

PR #395, opened 2022-05-10
(DRAFT)POC different tests with different wallets - generate random wallet a…

### LINKED ISSUE #380: Multiple wallets needed for different states in trading app
Currently for the tests we have only 1 wallet set up to run the tests which is set in git.secrets as `CYPRESS_TRADING_TEST_VEGA_WALLET_PASSPHRASE`

Some of the tests require the wallet to not have an approved tokens state as we check for certain validations and unhappy paths with the user not approving tokens before attempting to deposit . Some deposit tests are failing because the wallet has been approved and the tests are looking for a not approved validation .

We need to set up 2 wallet mnemonics and store them as cypress envs : 
CYPRESS_TRADING_TEST_VEGA_WALLET_PASSPHRASE_APPROVED
CYPRESS_TRADING_TEST_VEGA_WALLET_PASSPHRASE_NOT_APPROVED

Ditmir-Vega:
…nd private key

# Related issues 🔗

Closes #380 

# Description ℹ️

- add function for generating random keys
- add step for testing different wallets/keys

### DIFF
--- a/libs/cypress/src/lib/eip1193-bridge.ts
+++ b/libs/cypress/src/lib/eip1193-bridge.ts
@@ -113,6 +113,12 @@ const getProvider = () =>
     Cypress.env('ethereumChainId')
   );
 
+const getRandomKey = () => {
+  const createRandomWallet = Wallet.createRandom();
+  const randomPrivateKey = createRandomWallet.mnemonic.phrase;
+  return Wallet.fromMnemonic(randomPrivateKey, getAccount(0)).privateKey;
+};
+
 export const createBridge = () => {
   const provider = getProvider();
   const privateKey = Wallet.fromMnemonic(
@@ -120,5 +126,11 @@ export const createBridge = () => {
     getAccount(0)
   ).privateKey;
   const signer = new Wallet(privateKey, provider);
+  return new CustomizedBridge(signer, provider);
+};
+
+export const createBridgeWithNewWallet = () => {
+  const provider = getProvider();
+  const signer = new Wallet(getRandomKey(), provider);
   return new CustomizedBridge(signer, provider);
 };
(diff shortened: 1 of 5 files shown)


libs/cypress/src/lib/eip1193-bridge.ts
```diff
@@ -113,6 +113,12 @@ const getProvider = () =>
     Cypress.env('ethereumChainId')
   );
 
+const getRandomKey = () => {
+  const createRandomWallet = Wallet.createRandom();
+  const randomPrivateKey = createRandomWallet.mnemonic.phrase;
+  return Wallet.fromMnemonic(randomPrivateKey, getAccount(0)).privateKey;
```
dexturr, line 119: chore (blocking): don't use a random key, rather use the same mnemonic and generate the second key from that mnemonic. 

In the `createBridge` function below I would allow it to take an integer as a parameter, which should be passed to the `getAccount` function instead of 0.

Then you can take the account number as a parameter in the `mockWeb3Provider` command in  `mock-web3-provider.ts` and for the tests where you need to be unapproved you set it so that for example.

key 0, used for trading and has everything enabled. This is the default unless specified
key 1, used to test forms for approval ect.

and then you can continue to add keys infinitely all from the same mnemonic.

It would be good to document what key is used for what in a README.md file somewhere :+1:

dexturr, line 119: chore (blocking): don't use a random key, rather use the same mnemonic and generate the second key from that mnemonic. 

In the `createBridge` function below I would allow it to take an integer as a parameter, which should be passed to the `getAccount` function instead of 0.

Then you can take the account number as a parameter in the `mockWeb3Provider` command in  `mock-web3-provider.ts` and for the tests where you need to be unapproved you set it so that for example.

key 0, used for trading and has everything enabled. This is the default unless specified
key 1, used to test forms for approval ect.

and then you can continue to add keys infinitely all from the same mnemonic.

It would be good to document what key is used for what in a README.md file somewhere :+1:

PR #396, opened 2022-05-11
Feat/304 token tailwind 2

sam-keen:
# Description ℹ️

This PR completes the move away from scss and BEM-style selectors to using Tailwind. Visual change should be fairly minimal, but we are also using components from ui-toolkit where appropriate, so there will be some minor discrepancies.


apps/token/src/components/locked-progress/locked-progress.tsx
```diff
@@ -1,11 +1,57 @@
...
 
+const ProgressContents = ({
+  light,
+  children,
+}: {
+  light: boolean;
+  children: React.ReactNode;
+}) => (
+  <div
+    className={`flex justify-between py-2 font-mono ${
```
mattrussell36, line 14: suggestion (non-blocking): This would be cleaner using the classnames function which we already use in other projects. Same goes for any other files where we are nesting ternaries in strings

mattrussell36 (approved):

dexturr (approved):

PR #399, opened 2022-05-12
Test/add test for successful deposit

### LINKED ISSUE #325: Console V2- Tests - Successful deposit
Add a test that successfully completes a deposit in console V2 

This ticket will aim to identify any GAP's / Blockers with being able to successfuly test a deposit 
This includes 
- eipBridge function to approve/confirm a transaction
- How long confirmation takes 
- If we need to build any custom functions to approve or set approval

Ditmir-Vega:
# Related issues 🔗

Closes #325 

# Description ℹ️

- Added tests for submitting a deposit 
- Added custom command for use in feature files that fills in the deposits form - dynamic table
- Updated some tests to use dyanmic form,

### DIFF
--- a/apps/trading-e2e/src/integration/deposits.feature
+++ b/apps/trading-e2e/src/integration/deposits.feature
@@ -58,18 +58,50 @@ Feature: Deposits to vega wallet
   # Then I can see the deposit is Successfull
   # And Balance is updated to reflect deposit amount
 
-  Scenario: Validation errors
-    # wallet is connected on before hook so this step may no longer be required
-    # Given I connect my Ethereum wallet
-    When I submit a deposit with empty fields
-    Then I can see validation errors present
-    And I enter an invalid public key
+  Scenario: Empty form Validation errors
+    When I enter the following deposit details in deposit form
+      | asset  |  |
+      | to     |  |
+      | amount |  |
+    And I submit the form
+    Then I can see empty form validation errors present
+
+  Scenario: Invalid deposit Public Key Validation error
+    When I enter the following deposit details in deposit form
+      | asset  | tBTC TEST                                                        |
+      | to     | zzz85edfa7ffdb6ed996ca912e9258998e47bf3515c885cf3c63fb56b15de36f |
+      | amount | 1                                                                |
+    And I submit the form
     Then Invalid Vega key is shown
-    And I enter an amount less than the minimum viable amount
+
+  Scenario: Deposit Amount too small validation
+    When I enter the following deposit details in deposit form
+      | asset  | tBTC TEST                                                        |
+      | to     | zzz85edfa7ffdb6ed996ca912e9258998e47bf3515c885cf3c63fb56b15de36f |
+      | amount | 0.00000000000000000000000000000000001                            |
+    And I submit the form
     Then Amount too small message shown
-    And I enter a valid amount
+
+  Scenario: Deposit Amount greater than approved amount validation
+    When I enter the following deposit details in deposit form
+      | asset  | tBTC TEST                                                        |
+      | to     | zzz85edfa7ffdb6ed996ca912e9258998e47bf3515c885cf3c63fb56b15de36f |
+      | amount | 788888888888888                                                  |
+    And I submit the form
+    And Insufficient amount message shown
+  # Then Amount too small message shown
+  # And I enter a valid amount
+  # And I submit the form
   # This next step is being skipped due to account having approved status
   # Then Not approved message shown
+
+  Scenario: Successful deposit
+    When I enter the following deposit details in deposit form
+      | asset  | tBTC TEST                                                        |
+      | to     | 99d423f2839cada5f75fbc60f50c8a63343de7a132bf81b282858fbba5d82714 |
+      | amount | 1                                                                |
+    And I submit the form
+    And I can see the 'deposit pending' modal is shown
 
   @todo
   Scenario: Use the 'Use Maximum' button to populate amount input with the balance in the connected wallet
--- a/apps/trading-e2e/src/integration/deposits.feature
+++ b/apps/trading-e2e/src/integration/deposits.feature
@@ -93,9 +93,9 @@ Feature: Deposits to vega wallet
 
   Scenario: Successful deposit
     When I enter the following deposit details in deposit form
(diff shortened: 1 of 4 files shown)


sam-keen (approved):

apps/trading-e2e/src/integration/deposits.feature
```diff
@@ -58,19 +58,51 @@ Feature: Deposits to vega wallet
...
-    And I enter a valid amount
+
+  Scenario: Deposit Amount greater than approved amount validation
+    When I enter the following deposit details in deposit form
+      | asset  | tBTC TEST                                                        |
+      | to     | zzz85edfa7ffdb6ed996ca912e9258998e47bf3515c885cf3c63fb56b15de36f |
+      | amount | 788888888888888                                                  |
+    And I submit the form
+    And Insufficient amount message shown
+  # Then Amount too small message shown
```
dexturr, line 92: covered by test above

```suggestion
```

apps/trading-e2e/src/integration/deposits.feature
```diff
@@ -58,19 +58,51 @@ Feature: Deposits to vega wallet
...
+  # And I enter a valid amount
+  # And I submit the form
   # This next step is being skipped due to account having approved status
   # Then Not approved message shown
 
+  Scenario: Successful deposit
+    When I enter the following deposit details in deposit form
+      | asset  | tBTC TEST                                                        |
+      | to     | 99d423f2839cada5f75fbc60f50c8a63343de7a132bf81b282858fbba5d82714 |
```
dexturr, line 101: chore (blocking): These asset ids need to be in the cypress environment file. Suggest 2 properties, assetName and assetId, with some magic strings as per other comment. Otherwise this is going to break in capsule and be very hard to run across different environments

Ditmir-Vega: I suggest this is not blocking .  we should visit this when capsule is integrated to this project and create this as an issue - maybe add it to an existing label for front end capsule testing related stuff. The reason i'm hesitant to add these things now is that they are fairly easy to resolve should we need and having capsule fail on them will highlight what needs to be done. Also the feature files here are supposed to be descriptive/readable so they can be reproduced manually having 
for example :
|asset | assetiD        |
|to      | validAddress|
is less readable than
|asset | tBTC        |
|to      | 99d423f2839cada5f75fbc60f50c8a63343de7a132bf81b282858fbba5d82714 |

If a test fails we can debug easier here

Also for capsule , wouldn't we always build something with deterministic data. like loading a checkpoint with those assets already in . Capsule should fit around our testing needs rather than have our tests optimised for capsule

dexturr: It is blocking for multiple reasons:

1. This is environment config, and cypress has an environment config system which we should be leveraging. This is not just about capsule it's also about best practice. 
2. This isn't just about capsule, as we are containerising our apps and there are already custom private networks people may want to run the tests against this has to be configurable
3. You don't have to do it this way I would suggest having a map of asset name to contract addresses or similar so you can do something along the lines of
```
|asset | tBTC |
|to | currentConnectedKey |
````
5. We have had blocking comments for changes regarding capsule on multiple PRs both dev and QA side and I see no reason why it should change for this PR alone. 
6. Capsule should fit our needs yes however the architecture of the system is such that the smart contracts and core are the data owners for things like asset ids. We cannot specify an id from the frontend and make capsule use it because the front end does not and should not control these things. Following the dependency inversion principle the frontend inverts these dependencies and lets let be specified by environment config or from the core. The tests should allow for the same, using cypresses environment system.

**Review by dexturr — CHANGES_REQUESTED** [review]
See comment

File: apps/trading-e2e/src/support/pages/deposits-page.ts [hunk 3]
```diff
@@ -49,6 +54,12 @@ export default class DepositsPage extends BasePage {
       .should('be.visible');
   }
 
+  veryInsufficientAmountAmountMessage() {
```
jtsang586, line 57: Typo?

apps/trading-e2e/src/integration/deposits.feature
```diff
@@ -58,19 +58,50 @@ Feature: Deposits to vega wallet
...
-    And I enter an invalid public key
+  Scenario: Empty form Validation errors
+    When I enter the following deposit details in deposit form
+      | asset  |  |
+      | to     |  |
+      | amount |  |
+    And I submit the form
+    Then I can see empty form validation errors present
+
+  Scenario: Invalid deposit Public Key Validation error
```
jtsang586, line 69: There's inconsistent casing on the scenario titles throughout the feature file. It should be capitals on first word of the title and lower case throughout except for Names etc.

apps/trading-e2e/src/support/pages/deposits-page.ts
```diff
@@ -49,6 +54,12 @@ export default class DepositsPage extends BasePage {
...
+    cy.getByTestId('input-error-text')
+      .contains('Insufficient amount in Ethereum wallet')
+      .should('be.visible');
```
jtsang586, line 60: I don't think `.contains` is needed here. The text should be asserted by the assertion:
`cy.getByTestId('input-error-text').should('have.text', 'Insufficient amount in Ethereum wallet')

apps/trading-e2e/src/support/pages/deposits-page.ts
Ditmir-Vega: yes will update thanks

apps/trading-e2e/src/integration/deposits.feature
Ditmir-Vega: This particular test just fills in the form and who to deposit the tokens to , it does not connect a specific wallet or key . Afaik if it's a valid key it will not matter what environment you are on /it should not matter what environment you are on. This is just the recipient of the deposit .

apps/trading-e2e/src/support/pages/deposits-page.ts
Ditmir-Vega: _input-error-text_ is present multiple times and have.text will fail , include.text should work in this case but then does the same thing as above

dexturr (approved):

Issue #273, opened 2022-04-19
Allow Input to accept prepend/append elements


mattrussell36:
Currently inputs only allow an IconName to be passed as the prop:
`<Input value='foo' appendIconName='hand-up' />`

However sometimes we need to be able to render something completely custom inside, a good example of this is for rendering currency symbols 'inside' the input:
`<Input value={amount} appendElement={<Lozenge>USD</Lozenge>} />`

It would be good to allow Input's props to take either a string (IconName) or a ReactNode for something custom. Props could end up something like:
```
interface InputProps extends InputHTMLAttributes<HTMLInputElement> {
  ...
  prependElement?: IconName | ReactNode;
  appendElement?: IconName | ReactNode;
}
```

If an IconName is passed we render an `<Icon />` component otherwise whatever is passed in.

- [ ] Update Input to accept generic element for prepending/appending
- [ ] Update [token-input](https://github.com/vegaprotocol/frontend-monorepo/blob/760753ea3d2f6cad6f3e2acb39f122b78097dd37/apps/token/src/components/token-input/token-input.tsx) to render currency

gordsport:
Closing as in `Done`

PR #402, opened 2022-05-16
Feat/Input prepend and append elements

notbot00:
# Related issues 🔗

Closes #[273](https://github.com/vegaprotocol/frontend-monorepo/issues/273)

# Description ℹ️

Allows custom react nodes to be rendered as part of the input. 

# Demo 📺

Screen Shot 2022-05-16 at 12 33 17 PM
Screen Shot 2022-05-16 at 12 33 09 PM

# Technical 👨‍🔧

Changes the Input component's interface to accept the following prop combinations: 
- `prependIconName` with an optional `prependIconDescription`, 
- `appendIconName` with an optional `appendIconDescription`, 
- `prependElement`
- `appendElement`

### DIFF
--- a/libs/ui-toolkit/src/components/input/input.tsx
+++ b/libs/ui-toolkit/src/components/input/input.tsx
@@ -1,4 +1,4 @@
-import type { InputHTMLAttributes } from 'react';
+import type { InputHTMLAttributes, ReactNode } from 'react';
 import { forwardRef } from 'react';
 import classNames from 'classnames';
 import type { IconName } from '../icon';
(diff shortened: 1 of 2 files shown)


libs/ui-toolkit/src/components/input/input.tsx
```diff
@@ -61,13 +106,70 @@ export const inputStyle = ({
...
+      appendIconDescription,
+      prependElement,
+      appendElement,
+      className,
+      hasError,
+      ...props
+    },
+    ref
+  ) => {
     className = `h-28 ${className}`;
```
mattrussell36, line 165: suggestion (non blocking): use classnames func here for the hasPrepended, hasAppended logic?

mattrussell36 (approved):

libs/ui-toolkit/src/components/input/input.tsx
notbot00: totally makes sense, changed it. also changed the className to be a variable, i think it's generally a bad practice to overwrite react props.

dexturr (approved):

PR #409, opened 2022-05-17
Fix/flakey cypress test runs

mattrussell36:
# Description ℹ️

Adds a wait for markets page tests which allows time for ag grids to render before making assertions

# Technical 👨‍🔧

This isn't the best solution, but I think good enough for now to get green builds

### DIFF
--- a/apps/trading-e2e/src/support/pages/markets-page.ts
+++ b/apps/trading-e2e/src/support/pages/markets-page.ts
@@ -24,6 +24,10 @@ export default class MarketPage extends BasePage {
       'Description',
     ];
 
+    // We need this to ensure that ag-grid is fully rendered before asserting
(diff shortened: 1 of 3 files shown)


sam-keen (approved):

apps/trading-e2e/src/support/pages/markets-page.ts
```diff
@@ -10,6 +10,9 @@ export default class MarketPage extends BasePage {
   marketStateColId = 'data';
 
   validateMarketsAreDisplayed() {
+    // We need this to ensure that ag-grid is fully rendered before asserting
+    // eslint-disable-next-line cypress/no-unnecessary-waiting
+    cy.wait(1000);
     cy.get('.ag-root-wrapper').should('be.visible');
```
ditmirhasani, line 16: something more robust like a testid would be better here
\

Issue #315, opened 2022-04-27
Token Frontend migration rough edges and issues


dexturr:
# Token frontend migration

## Questions:
* Are claim codes being used?
    * Twitter DM problem?

## Test cases from notion:
	* notion.so/vegaprotocol/Test-cases-fc43837bcf964581b7ee79a3ea61b49e

## Tasks

- [x] A bunch of generic regressions
- [x] Removing blueprint
- [x] Table of what is needed for each step
- [x] Sentry, ensure that we are pointing to the correct sentry project (Env vars need updating after deploy)
- [x] Callouts need to accept any react element
    * No loading elements in some of the callouts
- [x] Prepend and append elements in UI toolkit (lozenges with currency)
- [x] ~~Confirmation in transaction callouts~~
    * ~~Should we add Vega scan transactions in~~ Lets do this in this ticket https://github.com/vegaprotocol/frontend-monorepo/issues/534
- [x] Staking is broken investigate why?

## Adding deployments for all envs

- [ ] Stagnet1
- [ ] Stagnet2
- [ ] Devnet
- [ ] Testnet
- [ ] Mainnet

## Regression testing:

Key flows:
* Association
    * Wallet
    * Associate more
    * Vesting contract
    * Amounts credited correctly
    * Desktop wallets/Local wallets
    * Token input
        * Balances
        * Disabled stated
        * Use maximum
        * Approval
* Disassociation
    * Token input
        * Balances
        * Disabled stated
        * Use maximum
* Nominating
    * Remove
        * Remove now
        * Remove at end of epoch
    * Add
    * Click through into a validator
    * Token input
        * Balances
        * Disabled stated
        * Use maximum
    * Remove by dissassociation 
* Claiming
    * Targeted
    * Untargeted
* Governance
    * Empty state
    * Voting
    * Viewing proposals
    * Changing votes
* Rewards
    * Check that epochs are ending properly
    * Will need to do a transfer for this (May require developer to do this, possible core help)
* Withdrawals
    * Token input
        * Balances
        * Disabled stated
        * Use maximum
* Redeeming
* Locked/unlocked
* Tranches pages
* Balance formatting consistency
* Unconnected states
    * Vega only
    * Eth only
    * Both connected
    * Neither connected
    * Add to feature files in TFE that exist
* Pending transaction modals
* Meta mask state changing
    * Chain
    * Account
* Home page
* Mobile across all states
    * Drawer

candida-d:
[...]

PR #423, opened 2022-05-17
Chore/315 remove blueprint from token dapp

mattrussell36:
# Related issues 🔗

Partially addresses #315 

# Description ℹ️

Removes @blueprintjs/core and @blueprintjs/select dependencies from the token dapp, adding components where necessary to replace functionality

# Demo 📺

New radio buttons:
Screen Shot 2022-05-18 at 8 36 06 AM

# Technical 👨‍🔧

- Mobile menu drawer has been replaced with a custom dialog component
- Country select used to use the blueprints suggest component but I've opted to just replace this with a standard Select component from the ui-toolkit for now. This means we lose the type to search for country functionality.
- All imported blueprint styles now don't need to be bundled
- Fixes spacing on disassociate page


apps/token/src/components/connected-vega-key/connected-vega-key.tsx.tsx
```diff
@@ -9,7 +9,9 @@ export const ConnectedVegaKey = ({ pubKey }: { pubKey: string | null }) => {
       <strong data-testid="connected-vega-key-label">
```
sam-keen, line 9: Question (non-blocking): what's the deal with this file name - typo?

apps/token/src/components/country-selector/country-selector.tsx
```diff
@@ -36,30 +13,18 @@ export const CountrySelector = ({
...
-        popoverProps={{ minimal: true }}
-        noResults={<MenuItem disabled={true} text="No results." />}
-        itemPredicate={filterCountry}
-        fill={true}
-      />
+      <Select
+        value={code ? code : undefined}
+        onChange={(e) => onSelectCountry(e.target.value)}
+      >
+        {countryData.map((country) => {
```
sam-keen, line 20: Chore (non-blocking): implicit return and destructuring cleaner here IMO

apps/token/src/components/transactions-modal/transactions-modal.tsx
```diff
@@ -65,53 +48,52 @@ export const TransactionModal = () => {
...
+        <table className="w-full">
+          <thead>
+            <tr>
+              <TransactionModalTh>{t('transaction')}</TransactionModalTh>
+              <TransactionModalTh>{t('status')}</TransactionModalTh>
+            </tr>
+          </thead>
+          <tbody>
+            {transactions.map((t) => {
+              return (
```
sam-keen, line 80: Suggestion (non-blocking): could do implicit return here too (clean, reduces lines of code, easy to read, win win win IMO)

apps/token/src/routes/staking/staking-form.tsx
```diff
@@ -210,28 +209,26 @@ export const StakingForm = ({
       <h2 className="text-h4 mb-8">{t('Manage your stake')}</h2>
       <FormGroup>
         <RadioGroup
-          onChange={(e) => {
+          onChange={(value) => {
```
sam-keen, line 212: Question (non-blocking): Is the ts-ignore still necessary on 213 now?

sam-keen (approved):
A few non-blocking bits from me, looks good.

libs/tailwindcss-config/src/theme.js
```diff
@@ -126,7 +126,7 @@ module.exports = {
     DEFAULT: '0.225rem',
     md: '0.3rem',
     lg: '0.5rem',
-    full: '9999px',
+    full: '100%',
```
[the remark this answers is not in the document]
mattrussell36: Pretty sure that's not the case for border radius and it goes by the actual element. @bglownia 9999px seemed like weird arbitrary value for `rounded-full` any reason you use it over 100%?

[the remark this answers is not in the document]
sam-keen: Ah scratch my comment, I thought you were adjusting breakpoints (just saw the md, lg etc). Value of 100% seems like a good move to me.

PR #433, opened 2022-05-19
Feat/427: Serve explicit environments

### LINKED ISSUE #427: Env flag for the nx commands
Instead of manually having to copy over the correct env file's content when running an app, we could have a flag, which could work like:

```
nx run <app-name>:serve --env=stagnet1
```

This could make sure you have the right content in your local env file (defaulting to testnet?). 

There are probably several ways to achieve this, adding a [custom hook to webpack](https://webpack.js.org/api/compiler-hooks/#environment), using [nx run commands](https://nx.dev/packages/nx/executors/run-commands#nxrun-commands), etc.

EDIT: going to wait with this for the outcome of the spike [#14](https://github.com/vegaprotocol/frontend-monorepo/issues/14), might be redundant

notbot00:
# Related issues 🔗

Closes #427 

# Description ℹ️

Added a new executor which proxies the `@nrwl/web:dev-server`. Now when we run the serve, we can pass in a custom env, which will make the executor load the relevant `.env.{name}` from the project root. 
e.g.
```
yarn nx run token:serve --env=stagnet1
```

This will now pick up the env variables put in the `.env.stagnet1` file and applies it to the app.

There's an open issue related to this here: https://github.com/nrwl/nx/issues/7729

# Technical

The current solution works only for apps using the `@nrwl/web:dev-server`. This excludes the trading app, as that one uses a nextjs executor.


apps/explorer/project.json
```diff
@@ -51,6 +51,42 @@
...
+      "options": {
+        "command": "env.cmd -f ./apps/explorer/.env.stanget2 nx run explorer:serve"
+      }
+    },
+    "serve-mainnet": {
+      "executor": "@nrwl/workspace:run-commands",
+      "options": {
+        "command": "env.cmd -f ./apps/explorer/.env.mainnet nx run explorer:serve"
+      }
+    },
```
dexturr, line 89: question (blocking): can we not use the args forwarding in order to have a single command for this, with a fallback?

https://nx.dev/packages/workspace/executors/run-commands#arguments-forwarding

notbot00: i did try that out, and indeed it reduces the verbosity of the code. that way it would look something like:

```
    "serve-env": {
      "executor": "@nrwl/workspace:run-commands",
      "options": {
        "command": "env.cmd -f ./apps/explorer/.env.{args.env} nx run explorer:serve"
      }
    },
```

what i didn't like is the lack of the ability to provide a default the env arg (without any super ugly bash hacks), and the lack of ability to tap into the error messages. for example without passing in the --env flag to the above version, the error message would say `./apps/explorer/.env.undefined` doesn't exist, which is not very intuitive imho. also, wasn't really sure about the naming of the command either, serve-env seems kinda awkward.

notbot00: thinking of creating a script for this maybe?
```
"command": "node ./scripts/get-env.js {args.env} nx run ..."
```

mattrussell36: I'm in favor of a script if it means we can have a single 'serve' command for each project but still get nice error messages

notbot00: alright, going to do a bit of rework on this

notbot00: @mattrussell36 @dexturr , i think i managed to clean this up, added a custom nx executor to run the dev-servers, let me know what you think.

tools/executors/serve/impl.js
```diff
@@ -0,0 +1,366 @@
+'use strict';
```
notbot00, line 1: this is a generated file by tsc, from the `./impl.ts` file

notbot00: removed it from the repo and added a step to the postinstall step to have it compiled.

apps/explorer/project.json
dexturr: Nice :+1:

dexturr (approved):

MadalinaRaicu (approved):

dexturr (approved):

PR #435, opened 2022-05-19
Feature/simple trading drawer

elmar-vega:
# Related issues 🔗

Closes #[378]

# Description ℹ️

- Added a new drawer component to ui-toolkit
- Used the drawer component inside simple trading and added routes with some dummy pages

# Demo 📺

None

# Technical 👨‍🔧

None


libs/ui-toolkit/src/components/drawer/drawer.tsx
```diff
@@ -0,0 +1,66 @@
+import * as React from 'react';
+import { theme } from '@vegaprotocol/tailwindcss-config';
+import type { ReactElement } from 'react';
+import { useEffect } from 'react';
+import Drawer from '@mui/material/Drawer';
```
mattrussell36, line 5: fix (blocking): I'd prefer to not add @mui components to the toolkit.

I recently created a drawer for the token dapp using radix. I didn't do it in the ui-toolkit as I didn't think it was needed across projects, until now! 😄 Its not merged yet but you can [find it here](https://github.com/vegaprotocol/frontend-monorepo/pull/423/files#diff-9e246d25dc7e309c6afe507168fa7155fecfe9d316f8356fc500b04e772afdefR177)

It doesn't cover as much functionality as your implementation does but perhaps its a useful starting point for migrating this from using @mui to @radix-ui.

Alternatively, if you want to just roll with @mui for now, can you make this a local component for the simple-trading-app.

mattrussell36 (changes requested):

libs/ui-toolkit/src/components/drawer/drawer-toggle.tsx
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export const DrawerToggle = ({
+  onToggle,
+  variant = DRAWER_TOGGLE_VARIANTS.CLOSE,
+  className,
+}: Props) => {
+  const [iconName, setIconName] = useState(IconNames.MENU);
+  const classes = classNames('md:hidden', {
+    [`${className}`]: className,
```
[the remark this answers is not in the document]
elmar-vega: @mattrussell36 I tried that first gave me so many errors!!! 🤯 I will try again, might be my webstorm behaving silly

libs/ui-toolkit/src/components/drawer/drawer.tsx
elmar-vega: Moved into simple trading project for now after discussion. Future ticket will be created to refactor into tailwind ui and then add back.

apps/simple-trading-app/src/app/components/drawer/drawer.tsx
```diff
@@ -0,0 +1,66 @@
...
+  onToggle(): void;
+}
+
+export const NavigationDrawer = ({
+  isMenuOpen = false,
+  onToggle,
+  children,
+}: Props) => {
+  const [windowSize, setWindowSize] = React.useState({
+    width: window.innerWidth,
```
macqbat, line 19: non-blocking: suggestion for the future - something like this (resolutions of the viewport getter etc.) would be useful to have in external hook or provider

apps/simple-trading-app/src/app/components/main/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+  const splashLoading = (
+    <Splash>
+      <Loader />
+    </Splash>
+  );
+
+  return (
+    <main className="p-20 overflow-hidden">
+      <React.Suspense fallback={splashLoading}>{routes}</React.Suspense>
```
macqbat, line 28: a question (non-blocking): Has Suspense a sense only with lazy import? I can't see it here..

apps/simple-trading-app/src/app/components/drawer/drawer.tsx
```diff
@@ -0,0 +1,66 @@
...
+  };
+
+  return (
+    <Drawer
+      classes={drawerRootClasses}
+      variant={isMobile ? 'temporary' : 'permanent'}
+      open={isMobile ? isMenuOpen : true}
+      onClose={onToggle}
+      ModalProps={{
+        keepMounted: true, // Better open performance on mobile.
```
dexturr, line 60: question (non-blocking): There is no way to connect a vega wallet to a mobile phone currently and won't be for quite a while. So any mobile stuff will be read only, just in case you weren't aware 👍

apps/simple-trading-app/src/app/routes/router-config.tsx
```diff
@@ -0,0 +1,49 @@
...
+    element: <div>Markets</div>,
+  },
+  {
+    path: ROUTES.TRADING,
+    name: 'Trading',
+    text: t('Trading'),
+    element: (
+      <DealTicketContainer
+        marketId={
+          '41013c28d53a72225c07cf2660cdd415d9dd0e9317ec4574e77592332db35596'
```
dexturr, line 32: comment (non-blocking): as you probably know this will change next time testnet is restarted but I assume market select is coming in a new PR

apps/simple-trading-app/src/app/components/main/index.tsx
dexturr, line 28: nitpick (non-blocking): don't think you need suspense as the routes aren't being lazy loaded

dexturr (approved):

macqbat (approved):

apps/simple-trading-app/src/app/components/drawer/drawer.tsx
elmar-vega: Yes, absolutely 👍

apps/simple-trading-app/src/app/components/drawer/drawer.tsx
elmar-vega: 👍

apps/simple-trading-app/src/app/components/main/index.tsx
elmar-vega: I will remove it for now

apps/simple-trading-app/src/app/routes/router-config.tsx
elmar-vega: 👍

mattrussell36 (approved):

PR #438, opened 2022-05-20
Fix/429: app zindex

### LINKED ISSUE #429: Z-index issues
We have some z-index issues with the hovered candle data and border on chart and the Toggle component having higher values the overlay background.

Screen Shot 2022-05-18 at 6 01 04 PM

notbot00:
# Related issues 🔗

Closes #429 

# Description ℹ️

Fixes z-index.

# Demo 📺

before:
Screenshot 2022-05-20 at 15 53 54

after:
Screenshot 2022-05-20 at 15 53 40

# Technical 👨‍🔧

Scopes the app screen, so the z-indexes defined are not able to affect the dialogs, and other root level elements, portals, etc.

### DIFF
--- a/apps/trading/pages/_app.page.tsx
+++ b/apps/trading/pages/_app.page.tsx
@@ -45,7 +45,7 @@ function VegaTradingApp({ Component, pageProps }: AppProps) {
               />
               <link rel="stylesheet" href="https://static.vega.xyz/fonts.css" />
             </Head>
-            <div className="h-full dark:bg-black dark:text-white-60 bg-white text-black-60 grid grid-rows-[min-content,1fr]">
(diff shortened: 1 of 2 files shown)


dexturr (approved):

MadalinaRaicu:
not sure why there is a z-index on the buttons  in the first place

notbot00:
> not sure why there is a z-index on the buttons in the first place

not sure either, tbh. i was assuming we might need to use z-indexes in some cases, but they should be contained so that the modal is always on top of the app's root area. z-indexes are okay as long as they are tamed with a relative position somewhere up in the tree, imo.
but i'll have another look, and try to figure out what happens if i remove it from the buttons. might make sense not to have it on them if it's useless.

MadalinaRaicu (approved):

PR #442, opened 2022-05-20
Feat/175: add explorer link to ui toolbox

### LINKED ISSUE #175: Add BlockExplorer link component
EtherscanLink component currently consumes NX_ETHERSCAN_URL which is problematic if we ever want to publish the toolkit as it is environment specific at build time, so we need a better solution for easily linking to other apps.

Proposed solution:
- [ ] Add a generic Link component
- [ ] Add some method (hook/context/hoc) of easily consuming these env vars to construct links to both Etherscan and the Vega block explorer
- [ ] Replace usage of the EtherscanLink component with Link components so that it no longer uses env vars

notbot00:
# Related issues 🔗

Closes #175 

# Description ℹ️

Adds a link based on explorer entities, similarly to the etherscan-link in the ui-toolbox.

# Demo 📺

Screenshot 2022-05-20 at 19 00 24
Screenshot 2022-05-20 at 19 00 31

# Technical 👨‍🔧

Implemented this with using children, instead of "text" as a prop, felt a lot more semantic regarding how anchor tags behave in general, even if we plan to use it only with string values to render.

### DIFF
--- a/libs/ui-toolkit/src/components/explorer-link/explorer-link.tsx
+++ b/libs/ui-toolkit/src/components/explorer-link/explorer-link.tsx
@@ -3,7 +3,7 @@ import type { ReactNode, HTMLProps } from 'react';
 
 const EXPLORER_URL = process.env['NX_EXPLORER_URL'] as string;
 
-type ExplorerEntity = 'block' | 'party';
+type ExplorerEntity = 'block' | 'party' | 'transaction';
 
 type ExplorerLinkProps = HTMLProps<HTMLAnchorElement> & {
   entity: ExplorerEntity;
@@ -14,6 +14,7 @@ type ExplorerLinkProps = HTMLProps<HTMLAnchorElement> & {
 const entityUrlMap: Record<ExplorerEntity, string> = {
   block: 'blocks',
   party: 'parties',
+  transaction: 'txs',
 };
 
 /**
--- a/libs/ui-toolkit/src/components/explorer-link/explorer-link.tsx
+++ b/libs/ui-toolkit/src/components/explorer-link/explorer-link.tsx
@@ -1,0 +1,53 @@
+import classNames from 'classnames';
+import type { ReactNode, HTMLProps } from 'react';
+
+const EXPLORER_URL = process.env['NX_EXPLORER_URL'] as string;
+
+type ExplorerEntity = 'block' | 'party';
+
+type ExplorerLinkProps = HTMLProps<HTMLAnchorElement> & {
+  entity: ExplorerEntity;
+  id?: string;
+  children?: ReactNode;
+};
+
+const entityUrlMap: Record<ExplorerEntity, string> = {
+  block: 'blocks',
+  party: 'parties',
+};
+
+/**
+ * Form an HTML link tag pointing to an appropriate Explorer page
+ */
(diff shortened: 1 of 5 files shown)


libs/ui-toolkit/src/components/explorer-link/explorer-link.tsx
```diff
@@ -0,0 +1,53 @@
...
+type ExplorerEntity = 'block' | 'party';
+
+type ExplorerLinkProps = HTMLProps<HTMLAnchorElement> & {
+  entity: ExplorerEntity;
+  id?: string;
+  children?: ReactNode;
+};
+
+const entityUrlMap: Record<ExplorerEntity, string> = {
+  block: 'blocks',
```
dexturr, line 15: chore (blocking): transaction hash will also be needed here, `/txs`

libs/ui-toolkit/src/components/explorer-link/explorer-link.tsx
```diff
@@ -0,0 +1,53 @@
+import classNames from 'classnames';
+import type { ReactNode, HTMLProps } from 'react';
+
+const EXPLORER_URL = process.env['NX_EXPLORER_URL'] as string;
```
dexturr, line 4: chore (blocking): the eventual idea of this is to be able to publish this library so it should be independent of env config when building. IMO let's invert this dependency with a `base` prop and have the consumer pass in the base, which can be pulled from config.

dexturr (changes requested):
Looks good, couple of minor things

libs/ui-toolkit/src/components/explorer-link/explorer-link.tsx
notbot00: I see. I am not sure if that makes much sense, tbh. In a UI toolkit the general idea should be to stay as agnostic of the data as possible. from the UI's perspective a link is just a string. Considering this, the idea of having an environment variable was also a bit weird, considering the above, but i imagined it's kinda practical in some cases. that's how the `etherscan-link` is working currently ([here](https://github.com/vegaprotocol/frontend-monorepo/blob/master/libs/ui-toolkit/src/components/etherscan-link/etherscan-link.tsx#L4)). It would be nice to aim for some sort of consistency here I think. 

If we're going to create a different toolkit link component for all the possible hosts we want to allow links to point to, and couple it with certain data types, I think it is very counterproductive and also confusing. Assembling the actual href can be easily handled by the host application, it's just a few pieces of strings. But I think it would be nice to have a better understanding of what we're aiming for, maybe we can have a short discussion about this.

libs/ui-toolkit/src/components/explorer-link/explorer-link.tsx
notbot00: meanwhile, added this part, thank you! :)

libs/ui-toolkit/src/components/explorer-link/explorer-link.tsx
dexturr: We will have a follow up to discuss this with the wider team as discussed on slack

dexturr (approved):

mattrussell36 (approved):

PR #453, opened 2022-05-24
Feat/175: ui-toolkit links

### LINKED ISSUE #175: Add BlockExplorer link component

notbot00:
# Related issues 🔗

Closes #175 

# Description ℹ️

Find a way to deal with the usage of environment variables in the libraries, and create a generic Link component, then remove the EtherscanLink.

# Technical 👨‍🔧

- removes `EtherscanLink` from the ui-toolkit
- adds a `Link` component to the ui-toolkit
- adds `EnvironmentProvider` context provider & `useEnvironment` hook to the react-helpers lib
- replaces all the `EtherscanLink` occurrences with the `Link` component
- wraps the apps with an `EnvironmentProvider`
- remove all process.env references from libs and get the variables calling the `useEnvironment` hook

NOTE: there's one more env variable left hardcoded in the `types/apollo.config.js` which is not replaceable directly with a hook call. That will need some attention eventually.


libs/web3/src/lib/use-bridge-contract.ts
```diff
@@ -2,20 +2,22 @@ import type { Networks } from '@vegaprotocol/smart-contracts-sdk';
...
+  const { VEGA_ENV } = useEnvironment();
   const { provider } = useWeb3React();
 
   const contract = useMemo(() => {
     if (!provider) {
       return null;
     }
     return new VegaErc20Bridge(
-      process.env['NX_VEGA_ENV'] as Networks,
+      VEGA_ENV as Networks,
```
dexturr, line 16: chore (non-blocking): can we do this case in the useEnviornment hook? As VEGA_ENV should always be a network

libs/react-helpers/src/hooks/use-environment.tsx
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  return (
+    <EnvironmentContext.Provider value={environment as Environment}>
+      {children}
+    </EnvironmentContext.Provider>
+  );
+};
+
+export const useEnvironment = () => useContext(EnvironmentContext);
```
mattrussell36, line 84: question (non blocking): What error do you get if you haven't wrapped the app with the provider? I've been using this pattern which gives a nice error, though it may be redundant if react reports it nicely

```
export function useEnvironment() {
  const context = useContext(EnvironmentContext);
  if (context === undefined) {
    throw new Error('useEnvironment must be used within EnvironmentProvider');
  }
  return context;
}
```

apps/trading/pages/portfolio/deposit/deposit-container.tsx
```diff
@@ -46,6 +48,7 @@ export const DepositContainer = ({
             requiredConfirmations={ethereumConfig.confirmations}
             assets={data.assets}
             initialAssetId={assetId}
+            isFaucetable={VEGA_ENV === 'MAINNET'}
```
mattrussell36, line 51: fix: (blocking): Can only be faucetable if the environment is not mainnet
```suggestion
            isFaucetable={VEGA_ENV !== 'MAINNET'}
```

mattrussell36 (changes requested):

libs/react-helpers/src/hooks/use-environment.tsx
notbot00: tried to see what the default error message says, not very useful. great suggestion, added the code inside the hook, thanks!

Screenshot 2022-05-24 at 15 09 08

libs/react-helpers/src/hooks/use-environment.tsx
```diff
@@ -0,0 +1,125 @@
...
+      return process.env['NX_ETHERSCAN_URL'];
+  }
+};
+
+const transformValue = (key: EnvKey, value?: string) => {
+  switch (key) {
+    case 'VEGA_ENV':
+      return value as Networks;
+    case 'ETHEREUM_CHAIN_ID':
+      return Number(value) ?? 3;
```
dexturr, line 61: chore (blocking): I don't think we want this to default, let's throw an error if it's not defined. I'd rather be explicit than accidentally have someone create a mainnet environment and be pulling testnet data

notbot00: sure, absolutely. updated so we throw an error if any of the variables are missing. not sure which are the values we can  handle gracefully and still have a running app while they are missing. i'm guessing ultimately having a network switcher will make sure we have all the stuff we need defined.

mattrussell36 (approved):

MadalinaRaicu (approved):

Issue #318, opened 2022-02-03
Pull contract addresses from core


Ditmir-Vega:
go to https://n03.stagnet2.vega.xyz/playground

query{networkParameters{key value}}

 {
        "key": "blockchains.ethereumConfig",
        "value": "{\"network_id\": \"3\", \"chain_id\": \"3\", \"collateral_bridge_contract\": { \"address\": \"0xffC1eb64e22fd5E29816c633eE84088EEEe879E5\" }, \"staking_bridge_contract\": { \"address\": \"0x7896C9491962D5839783CB6e0492ECebd34Bb35F\", \"deployment_block_height\": 11177313 }, \"token_vesting_contract\": { \"address\": \"0x9F10cBeEf03A564Fb914c2010c0Cd55E9BB11406\", \"deployment_block_height\": 11177353 }, \"confirmations\": 3}"
      },

replaces the hardcoded values with the ones from gprahQL query

gordsport:
decommission the contracts we don't need anymore (claim codes / others?)

web3 wrapper does this

PR #458, opened 2022-05-25
feat: port over smart contracts repository into the monorepo

dexturr:
# Related issues 🔗

#318

# Description ℹ️

Port over the smart contracts SDK this is prerequisite to the above issue as well as some capsule work that is coming.

Next steps:
* Get the apps to consume this rather than the published SDK
* Remove the dependency on @vegaprotocol/smart-contracts-sdk

# Technical 👨‍🔧

A second look at these classes and I am annoyed by them, I think a better refactor is called for once we have got all the apps using this.


mattrussell36 (approved):

libs/smart-contracts/src/utils/decimals.ts
```diff
@@ -0,0 +1,10 @@
+import BigNumber from 'bignumber.js';
+
+BigNumber.config({ EXPONENTIAL_AT: 20000 });
+
+export function addDecimal(value: BigNumber, decimals: number): BigNumber {
```
notbot00, line 5: question (non-blocking): seen these two also in `lib/react-helpers/src/lib/format/numbers.ts` in a slightly different variation, is there a natural habitat where these number related helpers should live in the monorepo?

notbot00 (approved):

libs/smart-contracts/src/utils/decimals.ts
dexturr: At the moment they live in react helpers, however this is a ts only library and that currently exports components as well so I didn't want this to be dependent on that. i think we need a ts only utils package for date/number/other ts only helpers, but didn't want the scope creep in this PR. I will ticket now.

PR #476, opened 2022-05-25
Feat/248: portfolio layout

### LINKED ISSUE #248: Implement layout for portfolio page
Link to deposits and withdraws
use existing components
deal with connected/not connected state

notbot00:
# Related issues 🔗

Closes #248 

# Description ℹ️

Add a layout for the portfolio page with the correct sections

# Demo 📺

Screenshot 2022-05-26 at 08 42 33

# Technical 👨‍🔧

Added components (without any further tweaks for now):
- positions container to positions tab
- order list container to orders tab
- accounts manager into bottom collaterals tab
- withdrawals manager into bottom withdrawals tab
- added a wallet handler so when landing on the portfolio page the wallet modal would open if no vega wallet is connected.


apps/trading/pages/portfolio/index.page.tsx
```diff
@@ -1,26 +1,76 @@
...
+              </GridTabs>
+            </section>
+          </main>
+          <section className="fixed bottom-0 left-0 w-full h-[200px]">
+            <GridTabs group="collaterals">
+              <GridTab id="collateral" name={t('Collateral')}>
+                <AccountsContainer />
+              </GridTab>
+              <GridTab id="deposits" name={t('Deposits')}>
+                <h2>{t('Deposits...')}</h2>
```
mattrussell36, line 64: chore (blocking): Can we add a link to /portfolio/deposit here. Just so there is an easy way to navigate there for now

mattrussell36 (changes requested):

apps/trading/pages/portfolio/index.page.tsx
notbot00: sure, no problem, added one.

apps/trading/pages/portfolio/index.page.tsx
```diff
@@ -1,39 +1,79 @@
...
-        </AnchorButton>
-        <AnchorButton
-          data-testid="withdraw-tEuro"
-          href="/portfolio/withdraw?assetId=8b52d4a3a4b0ffe733cddbc2b67be273816cfeb6ca4c8b339bac03ffba08e4e4"
-        >
-          {t('Withdraw tEURO')}
-        </AnchorButton>
-      </div>
-    </div>
+    <Web3Container
```
mattrussell36, line 15: question (blocking): Any reason for wrapping with the Web3Container component here. Should only be needed for the deposit and withdraw/withdrawals pages respectivly?

notbot00: if i remember correctly, the provider is needed for the withdrawals table which is the bottom sections's last tab of the bottom section. the table looks quite different from the design, because it's just whatever we had for the withdrawals page for now, unmodified.

Screenshot 2022-06-01 at 08 49 01

MadalinaRaicu (approved):

Issue #445, opened 2022-05-23
console V2 Select Markets modal routing updates 


MadalinaRaicu:
When a user first visits the platform, the intention is to get them straight into the trading environment and allow them to see the app as soon as they land - but we also want them to select their preferred market before getting stuck in.

- [ ] we want the route to reflect that. 

<img width="1133" alt="Screenshot 2022-05-23 at 13 12 15" src="https://user-images.githubusercontent.com/16125548/169797555-5c884de6-c294-4291-bd5e-332ef8c03632.png">

MadalinaRaicu:
<img width="813" alt="image" src="https://user-images.githubusercontent.com/16125548/169837641-47c41788-5415-4cb3-8164-390dee591c19.png">
 fixed markPrice

Issue #456, opened 2022-05-24
Improve market navigation


PeteWilliams:
### Background

Currently, to change markets you need to click 'Trading' and then select from the list:

![Image](https://user-images.githubusercontent.com/1325606/169993258-1cccb0c6-e400-420c-9e15-493f0b98e92a.png)

![Image](https://user-images.githubusercontent.com/1325606/169993326-74f860d8-3c1c-4639-8ef9-925cb2a5cd7b.png)

I think we'll need something a bit more clear/intuitive before we go live.

The design we want to eventually have is to have a line of market information at the top which includes a link that opens a mega-dropdown of markets:

![Image](https://user-images.githubusercontent.com/1325606/169992743-62ba0239-e61a-4dd5-8d50-44cd891c8b27.png)

![Image](https://user-images.githubusercontent.com/1325606/169992859-c9165b18-c361-4bc6-9aa9-5bfac491c705.png)

### Proposal
I'd suggest we change the market title to use the proposed style and use it to open the [existing Markets table](https://alpha.console.vega.xyz/markets) in a modal.

![Image](https://user-images.githubusercontent.com/1325606/170057599-6b0df7a3-899d-4358-868e-ab94c38f8ec5.png)

![image](https://user-images.githubusercontent.com/1325606/170292368-36e9bfab-c9b4-494c-a100-73804a15984d.png)

The 'Trading' nav item no longer needs to go to the market list, so it would not be clickable when you are in the trading view. However, when you are on the portfolio page, clicking 'Trading' would then take you back to the trading view for the last market you had selected.

MadalinaRaicu:
This ticket https://github.com/vegaprotocol/frontend-monorepo/issues/445 gets completely changed by this piece of work. 
I will use some unmerged logic from it, and create another branch for this ticket.

PR #484, opened 2022-05-26
fix: [#445] shallow routing from index

MadalinaRaicu:
# Related issues 🔗

Closes [#445]

# Description ℹ️

When a user first visits the platform, the intention is to get them straight into the trading environment and allow them to see the app as soon as they land - but we also want them to select their preferred market before getting stuck in.
The default market in the overlay is the oldest market traded in continuous mode.
We want the route to reflect that.

# Demo 📺

image

# Technical 👨‍🔧

- [x] direct to the market URL - make it sharable

### DIFF
--- a/apps/trading/pages/index.page.tsx
+++ b/apps/trading/pages/index.page.tsx
@@ -32,6 +32,14 @@ export function Index() {
   // The default market selected in the platform behind the overlay
   // should be the oldest market that is currently trading in continuous mode(i.e. not in auction).
   const { data, error, loading } = useQuery<MarketsLanding>(MARKETS_QUERY);
+  if (data && !error && !loading) {
+    const marketId = marketList(data)[0]?.id;
+    window.history.replaceState(
+      data,
+      '',
+      marketId ? `/markets/${marketId}` : '/markets'
+    );
+  }
   return (
     <>
       <LandingDialog />
--- a/libs/market-list/src/lib/components/landing/landing-dialog.tsx
+++ b/libs/market-list/src/lib/components/landing/landing-dialog.tsx
@@ -1,4 +1,5 @@
 import { useQuery } from '@apollo/client';
+import { t } from '@vegaprotocol/react-helpers';
 import { Interval } from '@vegaprotocol/types';
 import { AsyncRenderer, Dialog, Intent } from '@vegaprotocol/ui-toolkit';
 import { useState } from 'react';
@@ -22,7 +23,7 @@ export const LandingDialog = () => {
     <AsyncRenderer loading={loading} error={error} data={data}>
       {
         <Dialog
-          title="Select a market to get started"
+          title={t('Select a market to get started')}
           intent={Intent.Prompt}
           open={open}
           onChange={setClose}
--- a/libs/market-list/src/lib/components/landing/select-market-list.tsx
+++ b/libs/market-list/src/lib/components/landing/select-market-list.tsx
@@ -1,8 +1,10 @@
 import {
   addDecimalsFormatNumber,
   PriceCell,
+  t,
 } from '@vegaprotocol/react-helpers';
 import { PriceCellChange, Sparkline } from '@vegaprotocol/ui-toolkit';
+import Link from 'next/link';
 import { mapDataToMarketList } from '../../utils';
 import type { MarketList } from '../markets-container/__generated__/MarketList';
 
@@ -46,9 +48,9 @@ export const SelectMarketList = ({ data }: SelectMarketListProps) => {
                     className={`hover:bg-black/20 dark:hover:bg-white/20 cursor-pointer relative`}
                   >
                     <td className={`${boldUnderlineClassNames} relative`}>
-                      <a href={`/markets/${id}`} className={stretchedLink}>
+                      <Link href={`/markets/${id}`} passHref={true}>
                         {marketName}
-                      </a>
+                      </Link>
                     </td>
                     <td className={tdClassNames}>
                       {lastPrice && (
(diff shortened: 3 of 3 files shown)


apps/trading/pages/index.page.tsx
```diff
@@ -32,6 +32,14 @@ export function Index() {
   // The default market selected in the platform behind the overlay
   // should be the oldest market that is currently trading in continuous mode(i.e. not in auction).
   const { data, error, loading } = useQuery<MarketsLanding>(MARKETS_QUERY);
+  if (data && !error && !loading) {
+    const marketId = marketList(data)[0]?.id;
+    window.history.replaceState(
```
MadalinaRaicu, line 37: the purpose was to make the link sharable and consistent with the oldest traded market found. 

after lots of things that I've tried, this is the most efficient way of maintaining index functionality, but shallow route without reloading, and without messing up with the windows history - so that going back won't redirect to '/'.

mattrussell36: question (blocking): Did you look at using the useRouter component from next? See here: https://nextjs.org/docs/api-reference/next/router.

Altering the window history directly in the render function is probably a bad idea. I would suggest putting this logic in a useEffect hook and using the replace function from useRouter.

```
const { replace } = useRouter()

useEffect(() => {
  if (marketId) {
    replace(`/markets/${marketId}`)
  }
}, [marketId])
```

mattrussell36 (changes requested):

apps/trading/pages/index.page.tsx
mattrussell36: With the above, I think you'll probably want to render the <LandingDialog /> on the market page. 

We end up with no index page at all and we just use it to redirect where we need to go, for example:
```
function Index() {
 const { data, error, loading} = useQuery(...)

  useEffect(() => {
    ... calculate desired market and redirect
  }, [data])
  
  return null
}
```

MadalinaRaicu: so that was the first approach I've tried.
this requires to render the modal on the market page only if it's the first launch of the app (not the first render of the market page). 

 and for this I needed to maintain some global state to show the select market modal only on landing. 
 
 I've also tried using hooks with localstorage, but that does not give me the desired behaviour.

MadalinaRaicu: I tried initially to use only `next/router` methods such as push and replace:
```
      replace(
        `....`,
        `/markets/${marketId}?portfolio=orders&trade=orderbook&chart=candles`,
        { shallow: true }
      );
```
And even with shallow set to true it would still do a proper redirect to the market page, there is a small glitch, and the modal closes as it does not know it is the first pop up

MadalinaRaicu: also - what I have here is subject to change - this is because in my next ticket (started to work on it already) I will launch another modal to select the markets (and it will open the grid that we have on the `/markets` page currently).  

https://github.com/vegaprotocol/frontend-monorepo/issues/456

MadalinaRaicu: > question (blocking): Did you look at using the useRouter hook from next? See here: https://nextjs.org/docs/api-reference/next/router.
> 
> Altering the window history directly in the render function is probably a bad idea. I would suggest putting this logic in a useEffect hook and using the replace function from useRouter.
> 
> ```
> const { replace } = useRouter()
> 
> useEffect(() => {
>   if (marketId) {
>     replace(`/markets/${marketId}`)
>   }
> }, [marketId])
> ```
I tried initially to use only next/router methods such as push and replace:

      replace(
        `....`,
        `/markets/${marketId}?portfolio=orders&trade=orderbook&chart=candles`,
        { shallow: true }
      );
And even with shallow set to true it would still do a proper redirect to the market page, there is a small glitch, and the modal closes as it does not know it is the first pop up

MadalinaRaicu: > so that was the first approach I've tried. this requires to render the modal on the market page only if it's the first launch of the app (not the first render of the market page).
> 
> and for this I needed to maintain some global state to show the select market modal only on landing (I don't think keeping global state is necessary right now, but I might find out that it is while working on the next ticket #456).
> 
> I've also tried using hooks with localstorage, but that does not give me the desired behaviour.

so that was the first approach I've tried.
this requires to render the modal on the market page only if it's the first launch of the app (not the first render of the market page).

and for this I needed to maintain some global state to show the select market modal only on landing (I don't think keeping global state is necessary right now, but I might find out that it is while working on the next ticket https://github.com/vegaprotocol/frontend-monorepo/issues/456).

I've also tried using hooks with localstorage, but that does not give me the desired behaviour.

MadalinaRaicu: also - what I have here is subject to change - this is because in my next ticket (started to work on it already) I will launch another modal to select the markets (and it will open the grid that we have on the /markets page currently).

https://github.com/vegaprotocol/frontend-monorepo/issues/456

MadalinaRaicu: > With the above, I think you'll probably want to render the on the market page.
> 
> We end up with no index page at all and we just use it to redirect where we need to go, for example:
> 
> ```
> function Index() {
>  const { data, error, loading} = useQuery(...)
> 
>   useEffect(() => {
>     ... calculate desired market and redirect
>   }, [data])
>   
>   return null
> }
> ```

so that was the first approach I've tried.
this requires to render the modal on the market page only if it's the first launch of the app (not the first render of the market page).

and for this I needed to maintain some global state to show the select market modal only on landing (I don't think keeping global state is necessary right now, but I might find out that it is while working on the next ticket https://github.com/vegaprotocol/frontend-monorepo/issues/456).

I've also tried using hooks with localstorage, but that does not give me the desired behaviour.

MadalinaRaicu: also - what I have here is subject to change - this is because in my next ticket (started to work on it already) I will launch another modal to select the markets (and it will open the grid that we have on the /markets page currently).

https://github.com/vegaprotocol/frontend-monorepo/issues/456

mattrussell36 (approved):
Lets merge and revisit with global state implementation

notbot00 (approved):

PR #486, opened 2022-05-26
Feat/372 fractional orders

### LINKED ISSUE #372: Fractional order sizes
### Fractional order sizes

Scope, as a minimum its required to support fractional orders:

- [ ]  is clear to the user when a number is not valid
- [ ] Market framework data should specify (position DP?) the asset decimal places

Notes:
positionDecimalPlaces: Int!
positionDecimalPlaces indicated the number of decimal places that an integer must be shifted in order to get a correct size (uint64). i.e. 0 means there are no fractional orders for the market, and order sizes are always whole sizes. 2 means sizes given as 10^2 * desired size, e.g. a desired size of 1.23 is represented as 123 in this market.

Core specs: https://github.com/vegaprotocol/specs-internal/blob/master/protocol/0052-FPOS-fractional_orders_positions.md

notbot00:
# Related issues 🔗

Closes #372 

# Description ℹ️

Allows the user to order fractional amounts using the `positionDecimalPlaces` of the market data. 

# Demo 📺

Screenshot 2022-05-26 at 19 23 41


libs/deal-ticket/src/use-order-submit.ts
```diff
@@ -35,12 +36,14 @@ const ORDER_EVENT_SUB = gql`
   }
 `;
 
-interface UseOrderSubmitMarket {
-  id: string;
-  decimalPlaces: number;
-}
+const calculateOrderSize = (
+  market: DealTicketQuery_market,
```
dexturr, line 40: chore (blocking): take a number rather than the whole markt object, as GQL we request different subsets of the market data this will likely lead to type collisions later

libs/react-helpers/src/lib/format/number.ts
```diff
@@ -2,6 +2,10 @@ import { BigNumber } from 'bignumber.js';
 import memoize from 'lodash/memoize';
 import { getUserLocale } from './utils';
 
+export function toDecimal(numberOfDecimals: number) {
+  return numberOfDecimals > 0 ? 1 / Math.pow(10, numberOfDecimals) : 1;
```
dexturr, line 6: nitpick (non-blocking): simplier
```suggestion
  return numberOfDecimals > 0 ? Math.pow(10, -numberOfDecimals) : 1;
```

notbot00: nice, thanks! actually realised the null check can also be omitted in both variations as `Math.pow(10, -0)` returns 1

libs/deal-ticket/src/use-order-submit.ts
notbot00: looking at this again it seems quite redundant as it's just a proxy function for the `removeDecimals`. removed this for now.
question tho: not entirely sure how the type generators work, but I'd expect them to generate the correct types based on the queries we write. the market query's generated type is `DealTicketQuery_market`, shouldn't this already cover all the variations we're requesting? i was assuming it does, hence the removal of the `UseOrderSubmitMarket` type.

apps/simple-trading-app/src/app/components/deal-ticket/deal-ticket-steps.tsx
```diff
@@ -2,7 +2,7 @@ import * as React from 'react';
 import type { FormEvent } from 'react';
```
dexturr, line 2: comment: cc @elmar-vega @macqbat, this may make adding fraction orders easier to your application 👍

libs/deal-ticket/src/deal-ticket.spec.tsx
```diff
@@ -83,7 +115,7 @@ it('Deal ticket defaults', () => {
 });
 
 it('Can edit deal ticket', () => {
```
dexturr, line 117: chore (non-blocking): would be good to add a test here for fraction order sizes and check everything is coming through correctly.

libs/deal-ticket/src/deal-ticket.tsx
```diff
@@ -26,9 +19,9 @@ export const DealTicket = ({
   market,
   submit,
   transactionStatus,
-  defaultOrder = DEFAULT_ORDER,
+  defaultOrder,
 }: DealTicketProps) => {
-  const [order, updateOrder] = useOrderState(defaultOrder);
+  const [order, updateOrder] = useOrderState(market, defaultOrder);
```
dexturr, line 24: chore (blocking): maybe I am being thick in some way but I am not seeing any validation on position size in here? Even with a step of 0.1 someone could enter 0.001, which we will need to catch and show an error for

dexturr: Probably as form validation 👍

libs/deal-ticket/src/use-order-submit.ts
dexturr: So the types will generate based on the query that is specified so for example if query A chooses feild X, Y a type will be generate to reflect that. The core tenant of GQL is to only request the data you need not everything unlike rest so query B may require fields Y and Z. So we end up with 2 types depending on what was requested, and it is very unlikely that the these types will overlap.

So when we generate functions that will act on say a market the best thing to do is create an interface that requires the specific fields that the function needs e.g.

```
getMarketLeverageName (market: ILeverageMarket) { return `${market.name} x${1/market.riskFactor}` }
```
with interface
```
ILeverageMarket {
riskFactor: string
name: string
}
```

or generate just take number parameters if feesable (as in this example).

Now if we generate a different query for the same entity that creates a different type we can still use this function even if the types do not overlap perfectly, as long as the 2 fields that the function requires takes. I anticipate we will request many different versions of markets, so we want this to be reusable.

More over this makes the boundary very defined between libraries and defines a specfic interface rather than a leeky abrstraction based on what is requested which is also good IMO.

dexturr: Hopefully that all makes sense :joy:

notbot00: it does, more or less. i think what makes it a little confusing is that currently i see only one source query for getting markets in this lib, so they seemingly always use the same type among the components. anyhow, this needs another look i think, did a pretty big refactor 😄

libs/deal-ticket/src/hooks/use-order-validation.spec.tsx
```diff
@@ -0,0 +1,196 @@
...
+);
+
+it.each`
+  tradingMode                            | errorMessage
+  ${MarketTradingMode.BatchAuction}      | ${ERROR.MARKET_CONTINUOUS_LIMIT}
+  ${MarketTradingMode.MonitoringAuction} | ${ERROR.MARKET_CONTINUOUS_LIMIT}
+  ${MarketTradingMode.OpeningAuction}    | ${ERROR.MARKET_CONTINUOUS_LIMIT}
+`(
+  'Returns an error message when trying to submit a non-limit order for a "$tradingMode" market',
+  async ({ tradingMode, errorMessage }) => {
```
mattrussell36, line 135: chore (non blocking): no need for async function here

libs/deal-ticket/src/utils/validate-size.ts
```diff
@@ -0,0 +1,12 @@
+export const ERROR_SIZE_DECIMAL = 'step';
+
+export const validateSize = (step: number) => {
+  const [_, stepDecimals = ''] = String(step).split('.');
+  return (value: string) => {
+    const [_, valueDecimals = ''] = value.split('.');
```
dexturr: follow up (non-blocking): This will not work depending on the localisation of the platform (ie. if commas are used instead of peroids. Will need to add logic for this.

PR #495, opened 2022-05-31
Feat/smart contracts addresses

### LINKED ISSUE #318: Pull contract addresses from core

mattrussell36:
# Related issues 🔗

#315
Closes #318 

# Description ℹ️

This PR makes the required changes for apps that use smart contracts to run with capsule enabling easier testing and development.

- Any addresses which can be fetched from network parameters now do so. To help enable this a useEthereumConfig hook has been added to the web3 lib.
- Contract classes have been:
  - simplified to more closely mirror the ABI
  - changed to not handle decimals
  - changed to no convert values to BigNumber objects (as we require decimals for this)
  - Additional any config from the smart-contracts lib has been removed.
- Contract addresses that remain hard coded have been moved to the useEnvironment hook

**Note**: As a result of the above changes, the pending transactions dialog shown when associating will need to be temporarily disabled.


MadalinaRaicu (approved):

apps/token/src/contexts/contracts/contracts-provider.tsx
```diff
@@ -1,29 +1,29 @@
...
 /**
  * Provides Vega Ethereum contract instances to its children.
  */
 export const ContractsProvider = ({ children }: { children: JSX.Element }) => {
-  const { ADDRESSES, VEGA_ENV } = useEnvironment();
   const { provider: activeProvider, account } = useWeb3React();
-  const [txs, setTxs] = React.useState<TxData[]>([]);
+  const { config } = useEthereumConfig();
+  const { VEGA_ENV, ADDRESSES } = useEnvironment();
+  // const [txs, setTxs] = React.useState<TxData[]>([]);
```
notbot00, line 26: this line looks like it can be removed

libs/react-helpers/src/lib/format/number.ts
```diff
@@ -50,3 +51,10 @@ export const formatNumberPercentage = (value: BigNumber, decimals?: number) => {
     typeof decimals === 'undefined' ? Math.max(value.dp(), 2) : decimals;
   return `${value.dp(decimalPlaces).toFormat(decimalPlaces)}%`;
 };
+
+export const convertEthersBigNum = (
```
dexturr, line 55: nitpick (non-blocking): I think I would prefer to see a type overload on addDecimal rather than a separate method

apps/token/src/contexts/contracts/contracts-provider.tsx
```diff
@@ -98,7 +94,8 @@ export const ContractsProvider = ({ children }: { children: JSX.Element }) => {
   }
 
   return (
-    <ContractsContext.Provider value={{ ...contracts, transactions: txs }}>
+    // TODO: re add transactions array
```
dexturr, line 97: chore (blocking): needed before merge IMO

apps/token/src/hooks/use-transaction.ts
```diff
@@ -60,7 +60,13 @@ export const useTransaction = (
     });
 
     try {
-      const tx = await performTransaction();
+      const result = performTransaction();
+
+      if (result === null) {
```
dexturr, line 65: question (blocking): I don't understand this case? Moroer if this happens won't we end up in an infinite pending state? Or is that desired?

libs/smart-contracts/src/contracts/token.ts
```diff
@@ -0,0 +1,45 @@
...
+  decimals(): Promise<number> {
+    return this.contract.decimals();
+  }
+  faucet() {
+    if (!this.faucetable) {
+      throw new Error('Cannot use faucet');
+    }
+
+    return this.contract.faucet();
+  }
```
dexturr, line 44: nitpick (non-blocking): Not sure about having fauctable as a prop, could we use one of:
1. Detect if the method is present?
2. Let the eth transaction fail and catch the error?
3. Have a separate class as these are two different contracts and so should have differing thin wrappers?

mattrussell36: I think we have to go with option 3 here. The ethers lib will only add the method to the instance of ethers.Contract if it exists in the ABI, which means we have to pass in the ABI to the constructor which I don't think is desirable. Having an additional wrapper makes most sense given with our future plans here.

apps/token/src/hooks/use-transaction.ts
mattrussell36: This is needed because of the approve function for collateral bridge. See here https://github.com/vegaprotocol/frontend-monorepo/pull/495/files#diff-f9779344e8902f6d965ac153e1f431ffbd102bc4749924cd8ecdd20d6c2f60caR45

We could change the rendering logic if needed to avoid having to the check the result of `performTransaction`

mattrussell36: You are right about not resetting tx state after the early exit. I've fixed that but perhaps it'll be nicer to do the above

apps/token/src/contexts/contracts/contracts-provider.tsx
mattrussell36: See comment here https://github.com/vegaprotocol/frontend-monorepo/pull/495#discussion_r889942639

apps/token/src/hooks/use-transaction.ts
mattrussell36: Gone ahead and passed in config via prop so we don't need to account for it not being available at tx execution time

dexturr (approved):

PR #502, opened 2022-06-03
Feat/6 Network Switcher

### LINKED ISSUE #6: Network switcher
- [ ] Add env vars with links to each other network for the given app
- [ ] Add a way to change network based on the list of links
- [ ] Find somewhere to open the dialog

### Notes:
During a scope discussion we covered:
- To simplify: network switch for Alpha Mainnet release takes you to a different URL (easiest approach for now and enhance this later on)
- IPFS only really needs to be mainnet if that helps simplify things until APIs become more stable (may not need to support on all internal/development networks if this simplifies first implementation) 

Further work captured in:

- https://github.com/vegaprotocol/frontend-monorepo/issues/464

notbot00:
# Related issues 🔗

Closes #6 

# Description ℹ️

Network switcher dialog which can be rendered on-demand.

# Demo 📺

Screenshot 2022-06-02 at 17 36 43
Screenshot 2022-06-02 at 17 36 36

# Technical 👨‍🔧

- moved the environment hook to a new network-swicther library
- added env variables for each app's deployed urls
- exposed a network switcher which can be rendered on-demand, added it to the trading app 

NOTE: finalising the UI is still in progress on this, but since the UI is a very small part of this PR, it's not blocking the merge imo


apps/explorer/.env
```diff
@@ -24,6 +24,7 @@ NX_TENDERMINT_WEBSOCKET_URL = "wss://lb.testnet.vega.xyz/tm/websocket"
 NX_VEGA_URL = "https://lb.testnet.vega.xyz/query"
 NX_VEGA_ENV = 'TESTNET'
 NX_VEGA_REST = 'https://lb.testnet.vega.xyz/datanode/rest'
+NX_VEGA_NETWORKS = "TESTNET=https://explorer.fairground.wtf;MAINNET=https://explorer.vega.xyz"
```
dexturr, line 27: chore (blocking): if we need a string to be parsed into code let's use JSON rather than this.

dexturr (changes requested):
questio

sam-keen (approved):

apps/explorer/.env
notbot00: changed this to JSON, thanks!

mattrussell36 (approved):
Looks good, only thing I would is that I would perhaps name then new lib something else. Perhaps just `libs/network` or `libs/environment`

notbot00:
> Looks good, only thing I would is that I would perhaps name then new lib something else. Perhaps just `libs/network` or `libs/environment`

yeah, it makes sense. i'm using this PR as a base for this one: https://github.com/vegaprotocol/frontend-monorepo/pull/507
I'd do a rename there, if that works @mattrussell36

dexturr (approved):

PR #507, opened 2022-06-06
Feat/464: Node Discovery

### LINKED ISSUE #464: Node discovery
- [ ] Add a env var for config file location
- [ ] If NX_VEGA_URL not defined, fetch a configuration file which contains list of nodes for each network
- [ ] Randomly select a node (promise race?)
- [ ] Check graphql compatibility (introspection query?)
- [ ] Load app if successful, fallback to splash screen error if not
- [ ] Allow custom node entry on splash screen
- [ ] Store selected node in localStorage

As an extension / enhancement to:

- https://github.com/vegaprotocol/frontend-monorepo/issues/6

NOTES: Requirements may change when incentivised datanode is specced / implemented. 

As a user I want to connect to a randomly selected (pre-configured?) node but also have the ability to add a custom preferred node

notbot00:
# Related issues 🔗

Closes #464 

# Description ℹ️

- depends on #502 

Enables connecting to a random host from a given pool, instead of only having a hardcoded environment variable for it. The code stays backward compatible, so when there is a NX_VEGA_URL in the environment, it bundles the code with that, and ignores any network configuration.

# Demo 📺

Different potential stages of the node discovery flow:

Screenshot 2022-06-10 at 11 54 08
Screenshot 2022-06-10 at 11 36 35
Screenshot 2022-06-10 at 11 36 50
Screenshot 2022-06-10 at 11 55 04
Screenshot 2022-06-10 at 11 37 01
Screenshot 2022-06-10 at 11 52 48

# Technical 👨‍🔧

- the discovery stages above in a best-case scenario happen really fast, the user won't see the loading stages, but the Network Loader component is made so it can take an optional app skeleton to show while sorting out the node connection, in case we'll need it
- enable using network configurations from external json files (currently supports a list of hosts to be loaded)
- when there is not NX_VEGA_ENV set to the environment, use the configuration to select a host
- the current selection process is based on selecting the first node which responds with a 200
- I've put the Apollo client provider inside the network loader, this will allow to support changing the client when a different node is chosen explicitly (e.g. for cases as requested in this ticket: https://github.com/vegaprotocol/token-frontend/issues/931)
- removed the ETHEREUM_CHAIN_ID from the environment and changed to use the config from web3's useEthereumConfig hook


apps/token/.env.stagnet1
```diff
@@ -1,7 +1,8 @@
 # App configuration variables
 NX_VEGA_ENV = "STAGNET"
 NX_VEGA_URL = "https://n03.s.vega.xyz/query"
-NX_VEGA_NETWORKS='{"DEVNET":"https://dev.token.vega.xyz","STAGNET":"https://dev.token.vega.xyz","STAGNET2":"staging2.token.vega.xyz","TESTNET":"token.fairground.wtf","MAINNET":"token.vega.xyz"}'
+NX_VEGA_CONFIG_URL = "https://static.vega.xyz/assets/stagnet1-network.json"
+NET":"https://dev.token.vega.xyz","STAGNET":"https://dev.token.vega.xyz","STAGNET2":"staging2.token.vega.xyz","TESTNET":"token.fairground.wtf","MAINNET":"token.vega.xyz"}'
```
mattrussell36, line 5: chore (blocking): This looks incorrect

apps/trading/pages/_app.page.tsx
```diff
@@ -30,79 +27,73 @@ function AppBody({ Component, pageProps }: AppProps) {
...
+        <VegaManageDialog
+          dialogOpen={store.vegaWalletManageDialog}
+          setDialogOpen={(open) => store.setVegaWalletManageDialog(open)}
+        />
+        <NetworkSwitcherDialog
+          dialogOpen={store.vegaNetworkSwitcherDialog}
+          setDialogOpen={(open) => store.setVegaNetworkSwitcherDialog(open)}
+          onConnect={({ network }) => {
+            if (VEGA_NETWORKS[network]) {
+              window.location.href = VEGA_NETWORKS[network] ?? '';
```
mattrussell36, line 63: chore (blocking): Should probably use nextjs's router.push function here.

libs/environment/src/components/network-loader/network-loader.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+type MessageComponentProps = {
+  children: ReactNode;
+};
+
+type ErrorComponentProps = MessageComponentProps & {
+  showTryAgain?: boolean;
+};
+
+const Error: FC<ErrorComponentProps> = ({
```
mattrussell36, line 35: chore (blocking): Using FC<> here is pretty redundant IMO, our general code style here is to just type props directly and not use ComponentTypes (like you already have on line 38)

```suggestion
const Error = ({
```

libs/environment/src/components/network-loader/network-loader.tsx
```diff
@@ -0,0 +1,163 @@
...
+  children,
+  showTryAgain,
+}: ErrorComponentProps) => (
+  <div>
+    <div className="mb-16">{children}</div>
+    {showTryAgain && (
+      <Button
+        className="mt-8"
+        variant="secondary"
+        onClick={() => window.location.reload()}
```
mattrussell36, line 45: question (blocking): Will this error in our next app? We might need to check for window first `typeof window !== 'undefined'`

libs/environment/src/components/network-loader/network-loader.tsx
```diff
@@ -0,0 +1,163 @@
...
+import { useEnvironment } from '../../hooks';
+import type { ConfigStatus } from '../../types';
+
+type NetworkLoaderProps<T> = {
+  children?: ReactNode;
+  skeleton?: ReactNode;
+  createClient: (url: string) => ApolloClient<T>;
+};
+
+type StatusComponentProps = {
```
mattrussell36, line 22: suggestion (non blocking): Locate next to component, where its used

libs/environment/src/components/network-loader/network-loader.tsx
```diff
@@ -0,0 +1,163 @@
...
+  const { configStatus, VEGA_URL } = useEnvironment();
+
+  const client = useMemo(() => {
+    if (VEGA_URL) {
+      return createClient(VEGA_URL);
+    }
+    return undefined;
+  }, [VEGA_URL, createClient]);
+
+  useEffect(() => {
```
mattrussell36, line 150: question (blocking): Why do we need this effect, can we not just have the initial value set to true?

libs/environment/src/components/network-loader/network-loader.tsx
```diff
@@ -0,0 +1,163 @@
...
+    setShowCallout(true);
+  }, []);
+
+  return !client ? (
```
mattrussell36, line 154: suggestion (non blocking): Early return if no client I think would be clearer
```
if (!client) {
   return canShowCallout
     ? ...
     : null
}
...
```

libs/environment/src/components/network-loader/network-loader.tsx
mattrussell36: Additionally we don't call `setShowCallout` anywhere else? Do we even need that state?

maybe just this would suffice?
```
if (!client) return <StatusComponent>

return <ApolloProvider />
```

tsconfig.base.json
```diff
@@ -21,6 +21,7 @@
       "@vegaprotocol/cypress": ["libs/cypress/src/index.ts"],
       "@vegaprotocol/deal-ticket": ["libs/deal-ticket/src/index.ts"],
       "@vegaprotocol/deposits": ["libs/deposits/src/index.ts"],
+      "@vegaprotocol/environment": ["libs/environment/src/index.ts"],
```
mattrussell36, line 24: chore (blocking): Need to delete network-switcher lib

libs/environment/src/hooks/use-environment.spec.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+beforeEach(() => {
+  // @ts-ignore typscript doesn't recognise the mock implementation
+  global.fetch.mockReset();
+  // @ts-ignore typscript doesn't recognise the mock implementation
+  global.fetch.mockImplementation(mockFetch);
+
+  window.localStorage.clear();
+
+  process.env['NX_VEGA_URL'] = mockEnvironmentState.VEGA_URL;
```
mattrussell36, line 56: question (blocking): I think we should probably tear down our mocked environment after these tests run

mattrussell36: We probably want to clear localstorage too

mattrussell36 (changes requested):

apps/trading/pages/_app.page.tsx
notbot00: totally true, changed this, thanks!

libs/environment/src/components/network-loader/network-loader.tsx
notbot00: true, using both is stupid here. FC does inject the children type for convenience, so i don't have to manually append it to the types i'm using, but happy to add it manually, if that's the preferred way.

libs/environment/src/components/network-loader/network-loader.tsx
notbot00: the ui-toolkit callouts don't behave very nice when rendered open by default, this is a hack to prevent this error:

Screenshot 2022-06-14 at 09 23 23

notbot00: the second part with the return conditionals I've added

libs/environment/src/hooks/use-environment.spec.tsx
notbot00: done!

libs/environment/src/components/network-loader/network-loader.tsx
mattrussell36: Up for discussion, but lets keep to the convention for now. I prefer manually typing children as then you know it should be used within

libs/environment/src/components/network-loader/network-loader.tsx
mattrussell36: makes sense!

tsconfig.base.json
notbot00: right, missed this, removed now!

libs/environment/src/components/network-loader/network-loader.tsx
mattrussell36: So this is happening because the loader component uses Math.random() internally which can result in mismatches  between a server render and a client render. Happy to resolve this for now but will make a follow up ticket to change how Loader is implemented or potentially just wrap its render with a check for a window object.

apps/simple-trading-app/src/app/components/simple-market-list/simple-market-toolbar.spec.tsx
```diff
@@ -31,13 +30,18 @@ describe('SimpleMarketToolbar', () => {
...
+      </MockedProvider>
+    );
+
+    await waitFor(() => {
+      expect(screen.getByTestId('market-products-menu').children).toHaveLength(
+        3
       );
-      await new Promise((resolve) => setTimeout(resolve, 0));
+      expect(screen.getByTestId('market-assets-menu').children).toHaveLength(6);
+      expect(screen.getByRole('combobox').children).toHaveLength(10);
```
sam-keen, line 44: Chore (non-blocking): These length values you're expecting are a bit opaque, perhaps they could be documented through var naming or commenting

libs/environment/src/hooks/use-environment.spec.tsx
```diff
@@ -0,0 +1,200 @@
...
+      VEGA_NETWORKS: {},
+    });
+  });
+
+  it('throws a validation error when NX_VEGA_ENV is not found in the environment', async () => {
+    delete process.env['NX_VEGA_ENV'];
+    const { result } = renderHook(() => useEnvironment(), {
+      wrapper: MockWrapper,
+    });
+    expect(result.error).not.toBe(undefined);
```
sam-keen, line 129: Chore (blocking?): IMO it'd be nice to expect more detail from the error than simply knowing one occurred.

sam-keen: Also on lines 137, 163 and 172

sam-keen (commented):
Couple of small things, but looks great!

libs/environment/src/hooks/use-environment.spec.tsx
notbot00: the reason i did it this way is that the error messages are generated by a third party, we don't really want to have our test failing because they changed an error text, for example. I was thinking if I create enough test cases here, it will be clear enough.

apps/simple-trading-app/src/app/components/simple-market-list/simple-market-toolbar.spec.tsx
notbot00: I'm actually not sure what is happening in these tests, I didn't write them, what happened here was I noticed they were failing, and I fixed them by moving things around and removing some hacks which were not working.

apps/explorer/webpack.config.js
```diff
@@ -11,6 +12,12 @@ module.exports = (config, context) => {
     : [];
   return {
     ...config,
-    plugins: [...additionalPlugins, ...config.plugins],
+    plugins: [
+      ...additionalPlugins,
+      ...config.plugins,
+      new webpack.DefinePlugin({
+        'process.platform': JSON.stringify(process.platform),
```
notbot00, line 19: not sure about this part. turned out it's needed by `better-ajv-errors` (https://github.com/atlassian/better-ajv-errors). 

i wasn't aware that it's not designed to be used in the browser (see: https://github.com/atlassian/better-ajv-errors/issues/40). it improves the way errors are read, and as it stands, this hack makes it work in the browsers, too, but I'm not sure if this is worth doing here. let me know your thoughts.

apps/simple-trading-app/src/app/components/simple-market-list/simple-market-list.spec.tsx
```diff
@@ -50,14 +49,13 @@ describe('SimpleMarketList', () => {
...
     };
-    await act(async () => {
-      render(
-        <MockedProvider mocks={[mocks, filterMock]}>
-          <SimpleMarketList />
-        </MockedProvider>
-      );
-      await new Promise((resolve) => setTimeout(resolve, 0));
-    });
+
```
macqbat, line 52: please don't change anything in these tests (console-lite). Current version in master is valid. Another ongoing PR has next changes there, so it here can cause big conflicts

macqbat (changes requested):
Please revert these changes with unit tests for console-lite. I can do it also at my own if you want...

apps/simple-trading-app/src/app/components/simple-market-list/simple-market-list.spec.tsx
notbot00: FYI, these tests were failing when I edited them. since then, they have been updated on master, yes, but they are still full of ambiguous and sometime redundant hacks imo. i was trying to add some improvements to them while merging the current master.
the thing i don't really agree with is that you're basically saying that I should spend extra time removing my improvements, so it doesn't potentially cause someone else to spend time with resolving conflicts some time in the future. not sure if I agree with this approach. PRs should be valid in isolation, it makes life a lot easier.

macqbat: yeah, this is exactly what i'm asking for here: the isolation for the our PR's.
OK, It doesn't matter, i can resolve all conflicts, np.
By the way, all so called by you "hacks" are wholly taken from apollo client documentation site. What is really a hack can sometimes be deceptive. I would be more careful with such statements.

macqbat (commented):
as above

apps/simple-trading-app/src/app/components/simple-market-list/simple-market-list.spec.tsx
notbot00: i feel like this is heading toward a direction of being a misunderstanding. I am not trying to insult anyone's code here. I just pointed out that the solutions in those tests are not ideal, and I had to spend time trying to figure out why my PR is being affected by them.

if there's a problem with my changes, I would like to understand what is it. it just didn't seem like a well explained reason that this might cause conflicts in some future PRs. 

the thing with setTimeout here is that it pushes a callback to the event loop, but the component being tested does a lot of things and there are a lot of useEffect calls which also add callbacks to the event loop, i don't think the current approach guarantees the execution order at all. this is why i decided to remove the Promise setTimeout, because it felt like a hack. i wouldn't have touched it if the tests passed, tbh. anyway, tests are passing now without my changes (which wasn't the case ), i will go back here and remove this if it's causing you difficulties, it is not worth the argument imo.

macqbat: Idea of Promise with setTimeout comes streightly from documentation https://www.apollographql.com/docs/react/development-testing/testing/ . Tests with this solution went smoothly, that's why they has been commited and merged. Callback with await clause just should "cut off" rendering phase from testing phase. Synchonize these two asychnonous process. I can understand what was the idea of the authors of this solution. And it even worked, however, as it turned out, not always :-(

And about possible conflics in repo, I was thinking more about your situation in case when my current PR would merge first. I made there many changes not only in unit test but also in component, so more problems would be at your side in such a case.

apps/explorer/webpack.config.js
mattrussell36: I hadn't realised ajv was not suitable browsers. As discussed on slack, can we take a look at using zod?

apps/token/src/config/env.ts
```diff
@@ -7,10 +9,49 @@ const windowOrDefault = (key: string) => {
 
 const TRUTHY = ['1', 'true'];
 
+interface VegaContracts {
+  claimAddress: string;
+  lockedAddress: string;
+}
+
+const customClaimAddress = process.env['CUSTOM_CLAIM_ADDRESS'] as string;
+const customLockedAddress = process.env['CUSTOM_LOCKED_ADDRESS'] as string;
```
dexturr, line 18: chore (blocking): need to be prefixed with NX_

dexturr:
```suggestion
const customClaimAddress = process.env['NX_CUSTOM_CLAIM_ADDRESS'] as string;
const customLockedAddress = process.env['NX_CUSTOM_LOCKED_ADDRESS'] as string;
```

apps/explorer/webpack.config.js
notbot00: @mattrussell36 thanks for the suggestion! zod looks pretty neat, had to make some adjustments with the error handling, but refactored my PR so that is uses zod now for validation.

mattrussell36 (approved):

libs/environment/src/hooks/use-environment.spec.tsx
notbot00: ended up changing this, since the third party got removed, updated the tests to check actual error messages

apps/simple-trading-app/src/app/components/simple-market-list/simple-market-toolbar.spec.tsx
notbot00: ended up removing the test changes here, because they got fixed on master meanwhile.

apps/explorer/src/app/app.tsx
```diff
@@ -20,13 +18,11 @@ function App() {
...
-
   return (
     <EnvironmentProvider>
       <ThemeContext.Provider value={theme}>
         <TendermintWebsocketProvider>
-          <ApolloProvider client={client}>
+          <NetworkLoader createClient={createClient}>
             <div
               className={`${
                 menuOpen && 'h-[100vh] overflow-hidden'
```
macqbat, line 28: why not classNames func?

macqbat (approved):

PR #512, opened 2022-06-07
feat(#500): colors fixes

bglownia:
# Related issues 🔗

Closes #[500]

# Description ℹ️

Intent enum was modified
Progress and Help was replaced with None - White/Black
Prompt was mapped to Warning
Primary was added (yellow for dark theme, pink for light theme)

TailwindIntents was removed
Highlight was replaced by Primary

White and black rgba palette was replaced by rgb colors, alpha channel was causing issues in <select />. 

Not used colors were removed from theme.js 
Vega colors were moved to vega namespace (text-red-vega -> text-vega-red)

intent namespace was removed from theme.js colors (text-intent-warning -> text-warning)

strong normal and muted was added to white and black (text-black-normal, text-black-strong, text-black-muted, text-white-normal,  text-white-muted, text-white-strong)

Decimal part in PriceCell is displayed as muted


MadalinaRaicu (approved):

MadalinaRaicu:
(non-blocking):  do we have any `secondary` color that we can add? (there is one used for cancel buttons for orders/positions for example)

mattrussell36 (approved):
Looks like some linting to fix but otherwise all good

libs/tailwindcss-config/src/theme.js
```diff
@@ -16,69 +22,67 @@ module.exports = {
...
-      '02': 'rgba(255, 255, 255, 0.02)',
-      '05': 'rgba(255, 255, 255, 0.05)',
-      10: 'rgba(255, 255, 255, 0.10)',
-      25: 'rgba(255, 255, 255, 0.25)',
-      40: 'rgba(255, 255, 255, 0.40)',
-      60: 'rgba(255, 255, 255, 0.60)',
-      80: 'rgba(255, 255, 255, 0.80)',
-      95: 'rgba(255, 255, 255, 0.95)',
-      100: 'rgba(255, 255, 255, 1.00)',
+      strong: '#FFF',
```
sam-keen, line 41: Chore (blocking): We've begun to rely on opacity with these white and black colours. I think some solid greys are a great idea, but we should keep semi-opaque colours too

libs/tailwindcss-config/src/theme.js
```diff
@@ -16,69 +22,67 @@ module.exports = {
...
-      25: 'rgba(0, 0, 0, 0.25)',
-      40: 'rgba(0, 0, 0, 0.40)',
-      60: 'rgba(0, 0, 0, 0.60)',
-      80: 'rgba(0, 0, 0, 0.80)',
-      95: 'rgba(0, 0, 0, 0.95)',
-      100: 'rgba(0, 0, 0, 1)',
+      strong: '#000',
+      normal: '#000',
+      muted: '#BFCCD6',
+      '02': shadeOfGray(100 - 2),
```
sam-keen, line 59: Chore (blocking): A lot of these colours are shared with the black palette (i.e. white.40 is the same hex as black.60). As per the last comment, I think we should have a palette for solid greys, and we can create a ticket to assess where each palette is used

sam-keen (commented):
Realised we need to approach greys differently or we'll have issues

libs/tailwindcss-config/src/theme.js
bglownia: You can get semi opaque by `className="text-black text-opacity-20"`
Supported opacity levels will be taken from theme setup which already has it fine grained
```
 opacity: {
    0: '0',
    2: '0.02',
    5: '0.05',
    10: '0.1',
    15: '0.15',
    20: '0.2',
    25: '0.25',
    30: '0.3',
    35: '0.35',
    40: '0.4',
    45: '0.45',
    50: '0.5',
    55: '0.55',
    60: '0.6',
    65: '0.65',
    70: '0.7',
    75: '0.75',
    80: '0.8',
    85: '0.85',
    90: '0.9',
    95: '0.95',
    100: '1',
  },
```

libs/tailwindcss-config/src/theme.js
bglownia: It like this for purpose. It's just easier to use when you want to set up colors for dark and light themes
`<span className="text-black-95 dark:text-white-95">Text<span>`
instead of 
`<span className="text-black-95 dark:text-black-05">Text<span>`

It make it easier also to search and replace ie. text-(black|white)-95 -> text-$1-90

PR #513, opened 2022-06-07
fix: increase timeout

jtsang586:
Increase timeout for failing test in CI


MadalinaRaicu (approved):

edd (approved):
👍 Looks good to merge. Is there a backup plan if we keep needing to nudge this up?

jtsang586:
> 👍 Looks good to merge. Is there a backup plan if we keep needing to nudge this up?

Honestly, these tests are taking a bit too long for my liking already. I'm pretty sure its because the hosted machines on Github are pretty low spec so solution might be to upgrade to better VMs which would be pricey...

PR #535, opened 2022-06-09
Fix/411 asset dropdown style

### LINKED ISSUE #411: Selecting an asset to deposit style broken
On macos / firefox 100:
Screenshot 2022-05-17 at 13 31 55

mattrussell36:
# Related issues 🔗

Closes #411 

# Description ℹ️
This PR adds a couple new solid colors which fix the background color of the select dropdown on firefox. It also Adjusts input styles and tailwind conifg so that they can be shared from a sensible place.

# Demo 📺

Screen Shot 2022-06-08 at 6 55 25 PM

# Technical 👨‍🔧
- Adds a shared.ts file to the ui-toolkit so we have somewhere to share base input/select/textarea classes and adds that file to tailwind config
- Also fixes the appended input icon padding
- Also removes inludesPadding functions from usage with inputs as its not required


MadalinaRaicu (approved):

apps/explorer/tailwind.config.js
```diff
@@ -6,6 +6,7 @@ const vegaCustomClasses = require('../../libs/tailwindcss-config/src/vega-custom
 module.exports = {
   content: [
     join(__dirname, 'src/**/*.{js,ts,jsx,tsx}'),
+    'libs/ui-toolkit/src/utils/shared.ts',
```
notbot00, line 9: question (non-blocking): trying to understand why this is needed, but can't wrap my head around it. this is supposed to be just a js array of class names for form elements, right? why do the apps need to know about this file?

notbot00 (approved):

sam-keen (approved):
Noice

apps/explorer/tailwind.config.js
mattrussell36: I initially added this file to only the ui-toolkit tailwind config expecting the shared classes to be bundled in and used by the apps. However it doesn't work unless the app specific tailwind config includes it.

PR #536, opened 2022-06-09
Feat/smart contracts update

### LINKED ISSUE #483: Smart contracts update
Update smart contract addresses and ABIs for the new bridge

- [ ] Add the new ABI as a new abi in the smart-contracts repo
- [ ] Create a new class for the smart contracts (we can't edit existing as there will be a period of time where we will need to support both). Base this on the existing class.
- [ ] Add new withdrawals hook to support additional parameters now needed
- [ ] Add logic to use old or new bridge based on an environment variable

[ref](https://vegaprotocol.slack.com/archives/C026LG6B83E/p1653557426863809)

dexturr:
# Related issues 🔗

Closes #483 

# Description ℹ️

Adds the ability to use the new or old bridge contracts. The old one will need ot be used on the token site for mainnet until after mainnet has the new smart contracts enabled.

### DIFF
--- a/apps/token/src/routes/withdraw/index.tsx
+++ b/apps/token/src/routes/withdraw/index.tsx
@@ -15,6 +15,7 @@ import type {
   WithdrawPageVariables,
 } from './__generated__/WithdrawPage';
 import { WithdrawManager } from '@vegaprotocol/withdraws';
+import { Flags } from '../../config';
 
 const Withdraw = () => {
   const { t } = useTranslation();
@@ -152,7 +153,7 @@ export const WithdrawContainer = ({ currVegaKey }: WithdrawContainerProps) => {
       <WithdrawManager
         assets={data.assets || []}
         accounts={accounts}
-        useNewContract={false}
+        useNewContract={Flags.USE_NEW_CONTRACTS}
       />
     </>
   );
--- a/apps/token/src/routes/withdraw/index.tsx
+++ b/apps/token/src/routes/withdraw/index.tsx
@@ -153,7 +153,7 @@ export const WithdrawContainer = ({ currVegaKey }: WithdrawContainerProps) => {
       <WithdrawManager
         assets={data.assets || []}
         accounts={accounts}
-        useNewContract={Flags.USE_NEW_BRIDGE_CONTRACT}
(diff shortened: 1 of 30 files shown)


apps/token/src/routes/withdraw/index.tsx
```diff
@@ -149,7 +150,11 @@ export const WithdrawContainer = ({ currVegaKey }: WithdrawContainerProps) => {
           </Callout>
         </div>
       )}
-      <WithdrawManager assets={data.assets || []} accounts={accounts} />
+      <WithdrawManager
+        assets={data.assets || []}
+        accounts={accounts}
+        useNewContract={Flags.USE_NEW_CONTRACTS}
```
mattrussell36, line 156: chore (blocking): Can we rename this one too `isNewContract`

mattrussell36 (changes requested):
Couple things to change here. Looks good, though I don't much like how we use collateral bridge and collateral bridge new. I think I would prefer to rename the current one to collateral-bridge-deprecated. Also I wonder if we can look into Jeremy's suggestion on slack

libs/smart-contracts/src/contracts/collateral-bridge.ts
```diff
@@ -4,6 +4,7 @@ import abi from '../abis/erc20_bridge_abi.json';
 
 export class CollateralBridge {
   public contract: ethers.Contract;
+  public isNewContract = false;
 
   constructor(
```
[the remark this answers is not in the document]
dexturr: I think I'd prefer to add a deprecated decorator, which we can use everywhere. Will do this 👍

mattrussell36 (approved):

libs/withdraws/src/lib/use-withdraw.spec.tsx
```diff
@@ -82,7 +82,7 @@ beforeEach(() => {
   };
   mockERC20Approval = {
     request: {
-      query: ERC20_APPROVAL_QUERY,
+      query: ERC20_APPROVAL_QUERY_NEW,
```
sam-keen, line 85: Chore (non-blocking): ideally the tests should cover both contracts

sam-keen (approved):

MadalinaRaicu (approved):

PR #539, opened 2022-06-09
Fix/trading app tests

mattrussell36:
# Related issues 🔗

N/A

# Description ℹ️

Fixes trading app tests by adding mocks where required. Also changes failing tests to use plain cypress rather than cucumber with page objects which were making it hard to track down why tests were failing.


apps/trading-e2e/src/integration/withdrawals.ts
```diff
@@ -0,0 +1,172 @@
...
+      'Withdrawal transaction pending'
+    );
+    cy.getByTestId('dialog-text').should('have.text', 'Awaiting transaction');
+  });
+
+  it.skip('creates a withdrawal on submit'); // Needs capsule
+  it.skip('creates a withdrawal on submit and prompts to complete withdrawal'); // Needs capsule
+});
+
+describe('withdrawals', () => {
```
dexturr, line 94: nitpick (non-blocking): would prefer this in a different file

apps/trading-e2e/src/integration/trading.ts
```diff
@@ -0,0 +1,307 @@
...
+        });
+      cy.getByTestId('tab-orders')
+        .get(`[col-id='${orderCreatedAt}']`)
+        .each(($dateTime) => {
+          cy.wrap($dateTime).invoke('text').should('not.be.empty');
+        });
+    });
+  });
+
+  describe('accounts', () => {
```
dexturr, line 278: nitpick (non-blocking): would prefer each window to have its' own tests

dexturr (approved):

notbot00 (approved):

PR #545, opened 2022-06-13
Feat/358 market list toolbar

### LINKED ISSUE #358: Ability to filter markets by settlement asset
## Task
As a user I would like to filter markets by settlement asset

## ACs
- [x] It should allow me to filter a list of the settlement asset as a typeahead/autocomplete and select one
- [x] It should persist the filtered state via the url, so I can share the link?

macqbat:
# Related issues 🔗

Closes #358

# Description ℹ️

Add a toolbar over simple market list for filtering the results. State of the filters should be saved in the url for deeplinking

# Demo 📺

Screenshot 2022-06-13 at 12 59 17

Screenshot 2022-06-13 at 15 42 17

# Technical 👨‍🔧

useNavigate hook is used for manually create possibly shortest url


MadalinaRaicu (approved):

apps/simple-trading-app-e2e/src/integration/market-list.test.ts
```diff
@@ -0,0 +1,67 @@
+describe('market list', () => {
+  describe('simple url', () => {
+    beforeEach(() => cy.visit('/markets'));
+
+    it('selected menus', () => {
```
notbot00, line 5: chore (not-sure-if-blocking-or-not): i think generally speaking the `it('...')` statements look a lot more readable when we aim to use it expressively, like this could read "it selects menus" when used as `it('selects menus', () => {...})`

apps/simple-trading-app/src/app/components/simple-market-list/simple-market-toolbar.tsx
```diff
@@ -0,0 +1,156 @@
...
+  const params = useParams();
+  const { assets, products, assetsPerProduct } = useMarketFiltersData();
+  const [activeNumber, setActiveNumber] = useState(
+    products?.length ? products.indexOf(params.product || '') + 1 : -1
+  );
+  const [activeAsset, setActiveAsset] = useState(params.asset || 'all');
+  const [activeState, setActiveState] = useState(params.state || 'Active');
+  const [sliderStyles, setSliderStyles] = useState<Record<string, string>>({});
+  const slideContRef = useRef<HTMLUListElement | null>(null);
+  const onMenuClick = useCallback(
```
notbot00, line 21: chore (non-blocking): wrapping this in a useCallback seems redundant here

macqbat: sure, good catch, removed

apps/simple-trading-app-e2e/src/integration/market-list.test.ts
macqbat: As far as it's not commonly accepted convention, I think it's just different approach to the names of test cases. I would't change it here, because, to be consequent, I should change it every other cases, and for such act I'm not convinced enough :-)

PR #583, opened 2022-06-16
Fix/Trading grid layout fixes

mattrussell36:
# Related issues 🔗

N/A

# Description ℹ️

This PR fixes the layout of the trade grid which was creating double scrollbars on the right hand side of the page. It also adjusts the colors and layout of the grid tabs slightly

# Demo 📺

Before:
Notice the scroll bars and the tabs not really looking like tabs:
Screen Shot 2022-06-16 at 3 41 01 PM

After:
Screen Shot 2022-06-16 at 3 36 38 PM

Dark mode:
Screen Shot 2022-06-16 at 3 36 48 PM

# Technical 👨‍🔧

Nothing complicated here just some classes to adjust the positing of the header in the grid and some spacing and color changes.

### DIFF
--- a/apps/trading-e2e/src/integration/home.ts
+++ b/apps/trading-e2e/src/integration/home.ts
@@ -39,8 +39,6 @@ describe('home', () => {
     });
 
     it('redirects to a default market with the landing dialog open', () => {
-      cy.url().should('include', `/markets/${oldestMarket.id}`); // Should redirect to oldest market
-
       // Overlay should be shown
       cy.getByTestId(selectMarketOverlay).should('exist');
       cy.contains('Select a market to get started').should('be.visible');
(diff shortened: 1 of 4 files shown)


dexturr (approved):

MadalinaRaicu (approved):
(non-blocking): have we thought of any alternatives for the auto-sizer? I am asking because on very large screens they don't actually scale (i think they don't work with width and height percentages) and it looks a bit off.

notbot00 (approved):

mattrussell36:
> (non-blocking): have we thought of any alternatives for the auto-sizer? I am asking because on very large screens they don't actually scale (i think they don't work with width and height percentages) and it looks a bit off.

Open to suggestions! Lets chat on this.

PR #595, opened 2022-06-21
test: allow connecting to a local ganache node by signing transactions in browser

dexturr:
# Related issues 🔗

N/A, as discussed in testing meet on monday

# Description ℹ️

Optionally allow the UI to sign transactions in browser and connect to a local ethereum provider to faciliate capsule e2e testing

# Technical 👨‍🔧

Needed to implement my own URL connector class based on the web3 react version and pirated our eip1193 bridge from other tests to faciliate signing. This does not do a very good job of in progress states as it stands, so there is potential for improvement there. I am also unsure howe we can best do negative testing for refusing transactions but this should give us the happy path scenarios for now.


libs/smart-contracts/src/contracts/token-vesting.ts
```diff
@@ -14,13 +14,13 @@ export class TokenVesting {
   }
 
   stakeTokens(amount: string, vegaPublicKey: string) {
-    return this.contract.stake_tokens(amount, vegaPublicKey);
+    return this.contract.stake_tokens(amount, `0x${vegaPublicKey}`);
```
bglownia, line 17: This add 0x repeats in couple place. We should have function for this, opposite of remove0x
ie 
`const add0x = (str:string) = str.replace(/^(?!0x)(.*)/, '0x$1')`
or
`const add0x = (str:string) = !str.startsWith('0x') ? '0x${str}' : str`

bglownia (approved):

apps/token/src/lib/url-connector.ts
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  /** {@inheritdoc Connector.activate} */
+  public async activate(): Promise<void> {
+    this.actions.startActivation();
+
+    await this.isomorphicInitialize();
+
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    return this.provider!.request({ method: 'eth_chainId' })
```
notbot00, line 143: question (non-blocking): any reason for the mix of async and Promises in this file? not sure, i usually find it more readable when it's not mixed up

notbot00 (approved):

apps/token/src/lib/url-connector.ts
```diff
@@ -0,0 +1,150 @@
+import { ethers, Wallet } from 'ethers';
+import { Connector } from '@web3-react/types';
+
+import { Eip1193Bridge } from '@ethersproject/experimental';
+import type { ConnectionInfo } from '@ethersproject/web';
+import type { Actions } from '@web3-react/types';
+import { ENV } from '../config/env';
+
+export class CustomizedBridge extends Eip1193Bridge {
```
mattrussell36, line 9: chore (non-blocking): I feel like this is something we'll want across apps? Perhaps it should live in libs/web3?

apps/token/src/routes/staking/associate/wallet-associate.tsx
```diff
@@ -35,19 +35,20 @@ export const WalletAssociate = ({
     appDispatch,
     appState: { walletBalance, allowance, walletAssociatedBalance, decimals },
   } = useAppState();
+  console.log(allowance.toString());
```
mattrussell36, line 38: chore (blocking): stray console.log here

apps/token/src/routes/staking/associate/wallet-associate.tsx
```diff
@@ -35,19 +35,20 @@ export const WalletAssociate = ({
...
     dispatch: approveDispatch,
-  } = useTransaction(() =>
-    token.approve(
+  } = useTransaction(() => {
+    return token.approve(
       ethereumConfig.staking_bridge_contract.address,
-      Number.MAX_SAFE_INTEGER.toString()
-    )
-  );
+      removeDecimal(Number.MAX_SAFE_INTEGER.toString(), decimals).toString()
```
mattrussell36, line 49: question (non-blocking): I'm not sure why we decided to use Number.MAX_SAFE_INTEGER as the default approval amount. Seems a little weird and it might be nicer to just use a nice rounded number?

mattrussell36 (changes requested):
Just the stray console.log to remove along with some comments/questions

apps/token/src/lib/url-connector.ts
dexturr: I think let's leave it here for now. I imagine that there will be required changes while we get used to using this for the token frontend. I'd prefer to leave it in here until it's stabilised a bit.

apps/token/src/routes/staking/associate/wallet-associate.tsx
dexturr: I think it was back in the days where everything was 5 d.p. we can change it if we want, I have noticed a bug on the deposits screen for this. I'll change it to 1M.

Issue #590, opened 2022-06-17
Convert accordiong to use Radix


mattrussell36:
Accordion implementation could be done using [radix's accordion](https://www.radix-ui.com/docs/primitives/components/accordion)

This will improve a11y for the accordion.

PR #601, opened 2022-06-21
Feat/590 convert accordiong to use radix

MadalinaRaicu:
# Related issues 🔗

Closes [#590]

# Description ℹ️

Accordion implementation is done using [radix's accordion](https://www.radix-ui.com/docs/primitives/components/accordion)

This will improve a11y of the accordion.

# Demo 📺

image

# Technical 👨‍🔧

The small pixelated chevron might be updated by the future update of the primitives.


dexturr (approved):

libs/ui-toolkit/src/components/accordion/accordion.tsx
```diff
@@ -1,73 +1,71 @@
-import React, { useRef, useState } from 'react';
+import React from 'react';
+import * as AccordionRadix from '@radix-ui/react-accordion';
```
sam-keen, line 2: Chore (blocking): Matt's started using the 'XPrimitive' naming convention for Radix primitives, so this should become "AccordionPrimitive"

libs/ui-toolkit/src/components/accordion/accordion.tsx
```diff
@@ -1,73 +1,71 @@
...
+import { keyframes, styled } from '@stitches/react';
+import ChevronDownIcon from './chevron-down';
 
-export interface AccordionProps {
+export interface AccordionProp {
```
sam-keen, line 6: Suggestion (non-blocking): Better name for this might be 'AccordionItemProps'?

package.json
```diff
@@ -19,15 +19,18 @@
...
     "@radix-ui/react-dropdown-menu": "^0.1.6",
+    "@radix-ui/react-icons": "^1.1.1",
     "@radix-ui/react-radio-group": "^0.1.5",
     "@radix-ui/react-tabs": "^0.1.5",
     "@radix-ui/react-tooltip": "^0.1.7",
     "@sentry/nextjs": "^6.19.3",
     "@sentry/react": "^6.19.2",
     "@sentry/tracing": "^6.19.2",
-    "@vegaprotocol/vegawallet-service-api-client": "^0.4.11",
+    "@stitches/react": "^1.2.8",
```
mattrussell36, line 32: question (blocking): I'm not sure about adding this dep. It seems like a pretty full featured css in js solution when we already have tailwind. Wondering if there is a simpler solution which doesn't muddy the waters of how we style things

libs/ui-toolkit/src/components/accordion/accordion.tsx
```diff
@@ -1,73 +1,71 @@
...
-        </div>
-      </div>
-    </div>
+    <AccordionRadix.Root type="multiple" className="flex flex-col">
+      {panels.map(({ title, content }, i) => (
+        <AccordionRadix.Item value={`item-${i}`} key={`item-${i}`}>
+          <AccordionHeader>
+            <AccordionRadix.Trigger
+              data-testid="accordion-toggle"
+              className="w-full py-2 box-border appearance-none cursor-pointer focus:outline-none flex items-center justify-between border-b border-muted"
```
sam-keen, line 48: Suggestion (non-blocking): using 'classnames' and splitting the string into organised chunks would be more readable here

libs/deal-ticket/src/components/info-market.tsx
```diff
@@ -23,97 +23,101 @@ export interface InfoProps {
 export const Info = ({ market }: InfoProps) => {
   const headerClassName =
     'text-h5 font-bold uppercase text-black dark:text-white';
+  const marketDataPanels = [
+    {
+      key: 'fees',
```
notbot00, line 28: question (non-blocking): do we need these key properties? they don't seem to be used anywhere in the code

notbot00 (approved):

PR #605, opened 2022-06-21
feat(orderbook): improve data handling performance

### LINKED ISSUE #604: Orderbook data handling performance improvements
When there is huge amount of data returned from API  (100000 data rows ~ 7MB of data) orderbook hangs the browser

bglownia:
# Related issues 🔗

Closes #604

# Description ℹ️

Changes made to improve data handling performance in orderbook

# Technical 👨‍🔧
 - immer produce removed from generic data provider - it is small on large data sets; it still can be used in update funtion
 - removing immer required to update functions that were mutating data
 - sequence number based data gap recognition and data provide reload was commented out
 - compactRows function optimized to not use concat and reduce 
 - updates applied not immediately but stored in dataRef and throttled


apps/simple-trading-app/src/app/components/simple-market-list/data-provider.ts
```diff
@@ -87,15 +88,16 @@ export const FILTERS_QUERY = gql`
...
   delta: SimpleMarketDataSub_marketData
-) => {
-  const index = draft.findIndex((m) => m.id === delta.market.id);
-  if (index !== -1) {
-    draft[index].data = delta;
-  }
-  // @TODO - else push new market to draft
-};
+) =>
+  produce(data, (draft) => {
```
mattrussell36, line 94: question (non-blocking) Dont we have to use the result of the produce call here?

libs/accounts/src/lib/accounts-data-provider.ts
```diff
@@ -52,17 +53,18 @@ export const getId = (
...
-  const id = getId(delta);
-  const index = draft.findIndex((a) => getId(a) === id);
-  if (index !== -1) {
-    draft[index] = delta;
-  } else {
-    draft.push(delta);
-  }
-};
+) =>
+  produce(data, (draft) => {
```
mattrussell36, line 59: question (non-blocking): Same here, why don't we have to do anything with the result of produce?

mattrussell36 (approved):
Couple questions about the use of produce but LGTM

libs/market-depth/src/lib/orderbook-data.ts
```diff
@@ -121,37 +132,38 @@ export const compactRows = (
...
-  );
+  const orderbookData: OrderbookRowData[] = [];
+  Object.keys(groupedByLevel).forEach((price) => {
+    const row = extendRow(
+      groupedByLevel[price].pop() as PartialOrderbookRowData
+    );
+    row.price = price;
+    let subRow: PartialOrderbookRowData | undefined;
+    // eslint-disable-next-line no-cond-assign
+    while ((subRow = groupedByLevel[price].pop())) {
```
dexturr, line 155: nitpick (non-blocking): could we rewrite this to avoid the eslint disable?

```js
let subrow = groupedByLevel[price].pop();
while (subrow) {
  //logic
  subrow = groupedByLevel[price].pop();
}
whle

```

libs/market-depth/src/lib/orderbook-data.ts
```diff
@@ -75,22 +76,32 @@ const updateRelativeData = (data: OrderbookRowData[]) => {
...
-  cumulativeVol: {
-    ask: dataType === VolumeType.ask ? volume : 0,
-    bid: dataType === VolumeType.bid ? volume : 0,
-  },
-  askByLevel: dataType === VolumeType.ask ? { [price]: volume } : {},
-  bidByLevel: dataType === VolumeType.bid ? { [price]: volume } : {},
 });
 
+export const extendRow = (row: PartialOrderbookRowData): OrderbookRowData =>
+  Object.assign(row, {
```
dexturr, line 90: nitpick (non-blocking): I normally prefer spread over assign

dexturr (approved):

libs/market-depth/src/lib/orderbook-data.ts
bglownia: Spread will create new object , and old one will need to be garbage collected. Object assign adds props to existing object.

apps/simple-trading-app/src/app/components/simple-market-list/data-provider.ts
mattrussell36: I didn't see the implicit return from the arrow function. For the record I prefer to only do this for one liners. Can we add some curly braces and add a explicit return statement?

PR #606, opened 2022-06-21
Fix/602 fix deal ticket component to be reused by both simple trading app and console v2

### LINKED ISSUE #602: Fix deal ticket component to be reused by both simple trading app and console v2

MadalinaRaicu:
# Related issues 🔗

Closes #602 

# Description ℹ️

A bug that has been introduced in a bad merge conflict attempt to resolve deal ticket container updates. 

# Demo 📺

Fixes the issue below: 

Screenshot 2022-06-21 at 12 32 59

# Technical 👨‍🔧

Deal ticket container gets resued in both console v2 and simple trading app.


elmar-vega (approved):

libs/deal-ticket/src/components/info-market.tsx
```diff
@@ -9,17 +9,45 @@ import {
...
 import pick from 'lodash/pick';
 import omit from 'lodash/omit';
-import type { DealTicketQuery_market } from './__generated__/DealTicketQuery';
+import type {
+  DealTicketQuery,
+  DealTicketQuery_market,
+} from './__generated__/DealTicketQuery';
 import BigNumber from 'bignumber.js';
+import { useQuery } from '@apollo/client';
+import { DEAL_TICKET_QUERY } from './deal-ticket-container';
```
mattrussell36, line 24: chore (blocking): Lets not share the query. Instead can we update the deal ticket query to only contain what it needs and add a new market info query for everything it needs.

mattrussell36 (changes requested):
Looks good though I think we shouldn't share the query across tabs. Lets separate them

sam-keen (approved):

libs/deal-ticket/src/components/info-market.tsx
```diff
@@ -8,17 +8,129 @@ import {
...
-import type { DealTicketQuery_market } from './__generated__/DealTicketQuery';
+import type {
+  MarketInfoQuery,
+  MarketInfoQuery_market,
+} from './__generated__/MarketInfoQuery';
 import BigNumber from 'bignumber.js';
+import { gql, useQuery } from '@apollo/client';
+
+const MARKET_INFO_QUERY = gql`
+  query MarketInfoQuery($marketId: ID!) {
```
mattrussell36, line 26: chore (blocking): Can we add a generateMarketInfoQuery mock, and add it to our tests. Otherwise we'll get flakey tests

mattrussell36 (changes requested):

libs/deal-ticket/src/components/info-market.tsx
MadalinaRaicu: added.

mattrussell36 (approved):

Issue #463, opened 2022-05-25
Vega Transaction component (links to block explorer)


gordsport:
In order to make the UX around being able to see and understand transactions made in the console, we need to have links for all transactions to the block explorer.

- [ ] Links take you to the view of that transaction in the block explorer

jtsang586:
Re-opening this ticket as discussed in the stand up.
Expecting the link to also be present in the accepted modal state which it currently is not.

PR #610, opened 2022-06-21
Feat/471 cancel order

### LINKED ISSUE #471: Cancel order
As a user I want to be able to cancel an order.

- [x] Update vegawallet-service-api-client to include orderCancellation transaction in open api schema
- [x] Update to latest version of vegawallet-service-api-client
- [x] Add a cancel button to the order list
- [ ] Add a vega transaction dialog to prompt user to sign cancellation transaction in wallet
- [ ] Ensure order status is updated (This should already be working as the order subscription should already be running)

MadalinaRaicu:
# Related issues 🔗

Closes #471

# Description ℹ️

Add the ability to cancel orders. 

# Demo 📺
Screenshot 2022-06-24 at 21 54 09
Screenshot 2022-06-24 at 21 54 20
Screenshot 2022-06-24 at 21 56 09
Screenshot 2022-06-25 at 20 01 15

# Technical 👨‍🔧
- [x]  Update to latest version of vegawallet-service-api-client
- [x]  Add a cancel button to the order list
- [x]  Add a vega transaction dialog to prompt user to sign cancellation transaction in wallet
- [x]  Ensure order status is updated


libs/wallet/src/vega-transaction-dialog/vega-transaction-dialog.tsx
```diff
@@ -5,18 +5,34 @@ import {
...
 
 interface OrderDialogProps {
   transaction: VegaTxState;
-  finalizedOrder: OrderEvent_busEvents_event_Order | null;
+  finalizedOrder: Order | null;
+  title?: string;
 }
 
-export const OrderDialog = ({
+export const VegaTransactionDialog = ({
```
dexturr, line 32: chore (blocking): not all transactions are orders, so this should probably be VegaOrderTransactionDialog

libs/wallet/src/order-hooks/use-order-cancel.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+  const cancel = useCallback(
+    async (order) => {
+      if (!keypair) {
+        return;
+      }
+
+      setFinalizedOrder(null);
+
+      const res = await send({
```
dexturr, line 52: chore (blocking): we need some form of error handling either here or where cancel is invoked. Probably here for reusability.

libs/wallet/src/order-hooks/use-order-cancel.tsx
```diff
@@ -0,0 +1,82 @@
...
+    variables: { partyId: keypair?.pub || '' },
+    skip: !id,
+    onSubscriptionData: ({ subscriptionData }) => {
+      if (!subscriptionData.data?.busEvents?.length) {
+        return;
+      }
+      // No types available for the subscription result
+      const matchingOrderEvent = subscriptionData.data.busEvents[0].event;
+
+      if (matchingOrderEvent && matchingOrderEvent.__typename === 'Order') {
```
dexturr, line 32: thought (non-blocking): i wonder if it's worth generating type names as enums or keeping a map in VegaTypes to stop all these magic strings

libs/wallet/src/context.ts
```diff
@@ -3,7 +3,7 @@ import type {
   TransactionResponse,
 } from '@vegaprotocol/vegawallet-service-api-client';
 import { createContext } from 'react';
-import type { VegaConnector } from './connectors';
+import type { VegaConnector } from './connectors/vega-connector';
```
dexturr, line 6: chore (blocking): this is still exported from the index, so i don't think this change is needed

libs/wallet/src/connectors/rest-connector.ts
```diff
@@ -85,7 +85,7 @@ export class RestConnector implements VegaConnector {
     }
   }
 
-  async sendTx(body: TransactionSubmission) {
+  async sendTx(body?: TransactionSubmission) {
```
dexturr, line 88: question (blocking): why are we permitting empty transactions? These would immediately get rejected so I think this doesn't make and sense?

libs/order-list/src/setup-tests.ts
```diff
@@ -1 +1,2 @@
 import '@testing-library/jest-dom';
+import 'jest-canvas-mock';
```
dexturr, line 2: question (blocking): why is this needed?

libs/wallet/src/order-hooks/order-event-query.ts
```diff
@@ -0,0 +1,26 @@
+import { gql } from '@apollo/client';
+
+export const ORDER_EVENT_SUB = gql`
```
dexturr, line 3: nitpick (non-blocking): I would prefer this to be in the useOrderCancel file, as it is not shared.

dexturr (changes requested):

libs/wallet/src/order-hooks/order-event-query.ts
MadalinaRaicu: I would use this in useOrderEdit and maybe useOrderSubmit. eventually I would have all the order hooks living here

libs/wallet/src/vega-transaction-dialog/vega-transaction-dialog.tsx
MadalinaRaicu: updated in last commit

libs/order-list/src/setup-tests.ts
MadalinaRaicu: can be removed, updated in last commit.

libs/wallet/src/order-hooks/use-order-cancel.tsx
MadalinaRaicu: i updated this to use BusEventType.Order

dexturr: Can you undo that please. That is the name of the event type which is not nesseacrily the name of the entity in GQL. Leave as is, I just meant something I think we should look into, as this problem is across the whole codebase. Probably to use the introspection query to create an enum of type names or something but not something for now :)

MadalinaRaicu: ok, alright.

libs/wallet/src/order-hooks/order-event-query.ts
dexturr: We should not share queries, the idea behind GQL is to request what you need only, not everything. If we need to wait for an order and useEdit needs to check the price and useCancel does not these should be 2 separate queries. Sharing them gives us more headaches than it solves problems. Otherwise we are throwing away the majority of the value proposition of GQL. If the orders subs happen to for some reason need exactly the same field I guess we could consider sharing them but the default should always be separate queries unless you have a really good reason that they should be shared, e.g. trade list and charts do a whole bunch of data processing, which needs to not be duplicated for performance reasons, which is why those are shared.

Also chore (blocking): there are fields in this query which are unused can you please remove them. e.g. eventId, block.

libs/wallet/src/order-hooks/use-order-cancel.tsx
```diff
@@ -0,0 +1,88 @@
...
+        setFinalizedOrder(matchingOrderEvent);
+      }
+    },
+  });
+
+  useEffect(() => {
+    if (finalizedOrder) {
+      resetTransaction();
```
dexturr, line 41: question (blocking): @mattrussell36 I notice this is also present in the useSubmitOrder function so asking you haha, but this isn't cleaning up the subscription after the event has been fired. I know this is in a hook so once the element in unrendered the subscription will be cancelled but IMO we should lean up the listener here as well. Was there a reason we decided not to do this in useOrderSubmit?

It's not the worst thing in the world as this is filtered by partyId but wanted to double check. I just worry if this is used as part of non-transient UI we could end up with a memory leek, albeit a slow one

libs/wallet/src/order-hooks/use-order-cancel.tsx
```diff
@@ -0,0 +1,88 @@
...
+          },
+        });
+
+        if (res?.signature) {
+          setId(determineId(res.signature));
+        }
+        return res;
+      } catch (e) {
+        Sentry.captureException(e);
+        return;
```
dexturr, line 69:
```suggestion
```

libs/order-list/src/lib/components/cancel-order-dialog/cancel-dialog.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+export const CancelDialog = ({
+  orderDialogOpen,
+  setOrderDialogOpen,
+  finalizedOrder,
+  transaction,
+  reset,
+}: CancelDialogProps) => {
+  const getDialogIntent = () => {
+    if (finalizedOrder) {
```
edd, line 23: **nitpick (non-blocking)**: I found this if logic a little hard to parse because `finalizedOrder` implied to me that we were *placing* an Order, whereas obviously we're waiting for the *updated* order to come in and to confirm that it's actually cancelled.

It's not worth making a change for this.

edd (approved):
:+1: One trivial variable naming comment, and I think this could have been broken up to better contain just the Order Cancellation stuff, but at least one of the split out PRs would have been pretty large anyway.

libs/wallet/src/order-hooks/use-order-cancel.tsx
MadalinaRaicu: this won't build cause not all code paths return a value in the callback. I need to have a return on this catch.

libs/order-list/src/lib/components/cancel-order-dialog/cancel-dialog.tsx
MadalinaRaicu: updatedOrder might be better

libs/order-list/src/lib/components/order-list-manager/order-list-manager.tsx
```diff
@@ -39,17 +38,17 @@ export const OrderListManager = ({ partyId }: OrderListManagerProps) => {
       const rowNode = gridRef.current.api.getRowNode(d.id);
 
       if (rowNode) {
-        if (!isEqual) {
-          update.push(d);
+        if (!isEqual(d, rowNode.data)) {
```
mattrussell36, line 41: comment (non-blocking): Thanks for fixing this one

libs/order-list/src/lib/components/order-list/order-list.tsx
```diff
@@ -1,24 +1,78 @@
...
 
 export const OrderList = forwardRef<AgGridReact, OrderListProps>(
   ({ data }, ref) => {
+    const [orderDialogOpen, setOrderDialogOpen] = useState(false);
+    const { transaction, finalizedOrder, reset, cancel } = useOrderCancel();
+    return (
+      <>
+        <OrderListTable data={data} cancel={cancel} ref={ref} />
+        <CancelDialog
+          {...{
```
mattrussell36, line 37: chore (blocking): Unnecessary spread

libs/order-list/src/lib/components/order-list/order-list.tsx
```diff
@@ -27,20 +81,19 @@ export const OrderList = forwardRef<AgGridReact, OrderListProps>(
         defaultColDef={{ flex: 1, resizable: true }}
         style={{ width: '100%', height: '100%' }}
         getRowId={({ data }) => data.id}
+        rowHeight={40}
       >
         <AgGridColumn
-          headerName={t('Market')}
+          headerName="Market"
```
mattrussell36, line 87: chore (blocking): Can we keep the `t` function?

libs/order-list/src/lib/components/order-list/order-list.tsx
```diff
@@ -27,20 +81,19 @@ export const OrderList = forwardRef<AgGridReact, OrderListProps>(
...
         />
         <AgGridColumn
-          headerName={t('Amount')}
+          headerName="Amount"
           field="size"
           cellClass="font-mono"
           valueFormatter={({ value, data }: ValueFormatterParams) => {
             const prefix = data.side === Side.Buy ? '+' : '-';
-            return (
-              prefix + addDecimal(value, data.market.positionDecimalPlaces)
```
mattrussell36, line 42: chore (blocking): Why are you removing this? Amount still needs formatting with DPs due to fractional orders

libs/order-list/src/lib/components/order-list/order-list.tsx
```diff
@@ -1,24 +1,78 @@
...
+          }}
+        />
+      </>
+    );
+  }
+);
+
+export const OrderListTable = forwardRef<AgGridReact, OrderListTableProps>(
+  ({ data, cancel }, ref) => {
+    const CancelRendererButton = ({ data }: ICellRendererParams) => {
```
mattrussell36, line 52: chore (blocking): I don't see benefit of creating a function here its not used mored than once and just results in you having to scroll up again once you get to this cell in the render logic below.  Can we move it inline?
```
cellRenderer={({ data} : ICellRendererPrarams) => ...}
```

libs/wallet/src/order-hooks/use-order-cancel.tsx
mattrussell36: chore (blocking): Additionally, not sure why I used a hook to reset transaction here it just makes things more confusing. Can we move resetTransaction() to just be called immediately after setUpdatedOrder. The code is more readable that way. Also could you make the same change in the use-order-submit hook.

libs/wallet/src/order-hooks/use-order-edit.tsx
```diff
@@ -0,0 +1,82 @@
...
+  OrderEvent,
+  OrderEventVariables,
+  OrderEvent_busEvents_event_Order,
+} from './__generated__/OrderEvent';
+import { ORDER_EVENT_SUB } from './order-event-query';
+
+export const useOrderEdit = () => {
+  const { keypair } = useVegaWallet();
+  const { send, transaction, reset: resetTransaction } = useVegaTransaction();
+  const [finalizedOrder, setFinalizedOrder] =
```
mattrussell36, line 17: suggestion(non blocking):
```suggestion
  const [editedOrder, setEditedOrder] =
```

libs/wallet/src/order-hooks/use-order-edit.tsx
```diff
@@ -0,0 +1,82 @@
...
+      // No types available for the subscription result
+      const matchingOrderEvent = subscriptionData.data.busEvents[0].event;
+
+      if (matchingOrderEvent && matchingOrderEvent.__typename === 'Order') {
+        setFinalizedOrder(matchingOrderEvent);
+      }
+    },
+  });
+
+  useEffect(() => {
```
mattrussell36, line 38: chore (blocking): As per my other comments. Lets remove the useEffect and just call this underneath `setFinalizedOrder` above

libs/wallet/src/provider.tsx
```diff
@@ -104,7 +104,7 @@ export const VegaWalletProvider = ({ children }: VegaWalletProviderProps) => {
       disconnect,
       connector: connector.current,
       sendTx,
-    };
+    } as VegaWalletContextShape;
```
mattrussell36, line 107: question (blocking): Why the type cast here?

MadalinaRaicu: formatting error

libs/order-list/src/lib/components/cancel-order-dialog/cancel-dialog.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+interface CancelDialogProps {
+  orderDialogOpen: boolean;
+  setOrderDialogOpen: (isOpen: boolean) => void;
+  finalizedOrder: Order | null;
+  transaction: VegaTxState;
+  reset: () => void;
+}
+
+export const CancelDialog = ({
```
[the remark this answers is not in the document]
MadalinaRaicu: I will look into refactoring this in the #463 ticket

libs/wallet/src/order-hooks/use-order-cancel.tsx
```diff
@@ -0,0 +1,88 @@
+import { useCallback, useEffect, useState } from 'react';
```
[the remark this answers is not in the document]
MadalinaRaicu: interesting, I was trying to fit everything in wallet lib. i would love to do this in #463

libs/wallet/src/provider.tsx
mattrussell36: Hm there's a larger problem then. I have a feeling its due to sendTx types?

libs/wallet/src/order-hooks/use-order-cancel.tsx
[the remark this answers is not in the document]
mattrussell36: Lets make a different ticket for this 👍

libs/wallet/src/order-hooks/use-order-edit.tsx
MadalinaRaicu: this needs to come in another PR -  i will remove it for now

libs/deal-ticket/src/components/deal-ticket-manager.tsx
```diff
@@ -49,12 +48,12 @@ export const DealTicketManager = ({
   };
 
   useEffect(() => {
-    if (transaction.status !== VegaTxStatus.Default) {
+    if (transaction.status !== VegaTxStatus.Default || finalizedOrder) {
```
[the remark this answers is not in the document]
MadalinaRaicu: I added this because the `Order placed` modal would not appear otherwise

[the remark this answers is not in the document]
MadalinaRaicu: It would get stuck into `Awaiting network confirmation` as the tx status comes very quickly into Default. I tried to quickly place an additional order and it works fine.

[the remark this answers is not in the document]
MadalinaRaicu: It is actually the same dialog that reopens. 
As per your suggestion, I will refactor this dialog to be used for both submission and cancellation (without any dialog wrappers).
I think I should update this in a separate PR.

libs/wallet/src/order-hooks/use-order-cancel.tsx
MadalinaRaicu: Updated to use resetTransaction immediately after setUpdatedOrder. 
Let me use sth else instead of useSubscription in another PR. I will refactor a bit order-list and move all the order hooks from the wallet.

MadalinaRaicu: https://github.com/vegaprotocol/frontend-monorepo/issues/657

mattrussell36 (changes requested):

libs/order-list/src/lib/components/mocks/generate-orders.ts
```diff
@@ -0,0 +1,167 @@
...
+
+export const marketOrder = generateOrder({
+  id: 'market-order',
+  type: OrderType.Market,
+  status: OrderStatus.Active,
+});
+
+export const generateMockOrders = (): Orders_party_orders[] => {
+  return [
+    {
```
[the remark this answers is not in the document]
MadalinaRaicu: If I import this from the order-list it messes up with my cypress tests. 
image

libs/order-list/src/lib/components/cancel-order-dialog/cancel-dialog.spec.tsx
```diff
@@ -0,0 +1,71 @@
+import { render, screen } from '@testing-library/react';
+import { OrderStatus, OrderType } from '@vegaprotocol/types';
+import type { VegaTxState, Order } from '@vegaprotocol/wallet';
+import { VegaTxStatus } from '@vegaprotocol/wallet';
+import { CancelDialog } from './cancel-dialog';
+
+describe('CancelDialog', () => {
+  it('should render when an order is successfully cancelled', () => {
+    const orderDialogOpen = true;
+    const setOrderDialogOpen = () => false;
```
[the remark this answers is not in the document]
MadalinaRaicu: updated in last commit

mattrussell36 (approved):

libs/order-list/src/lib/components/mocks/generate-orders.ts
```diff
@@ -0,0 +1,156 @@
...
+
+export const marketOrder = generateOrder({
+  id: 'market-order',
+  type: OrderType.Market,
+  status: OrderStatus.Active,
+});
+
+export const generateMockOrders = (): Orders_party_orders[] => {
+  return [
+    generateOrder({
```
dexturr, line 76: nitpick (non-blocking): lot of duplicated fields here, might be nice to have a default object and spread that into these

libs/wallet/src/order-hooks/order-event-query.ts
```diff
@@ -0,0 +1,26 @@
...
+  subscription OrderEvent($partyId: ID!) {
+    busEvents(partyId: $partyId, batchSize: 0, types: [Order]) {
+      eventId
+      block
```
dexturr, line 7:
```suggestion
```

dexturr (approved):

PR #614, opened 2022-06-22
frontend-monorepo-607: Added correct text colours to intent backgrounds

### LINKED ISSUE #607: [explorer] Transaction type labels are unreadable
Screenshot 2022-06-21 at 15 05 22
Screenshot 2022-06-21 at 15 05 29

# Tasks
- [ ] Make text in transaction label a different colour to the label
- [ ] Ensure it works in both light and dark mode

sam-keen:
# Related issues 🔗

Closes #607 

# Description ℹ️

A change to the way intents were set up created an oversight with text colours on different backgrounds. This PR simply adds the correct text colours to be visible on white and black backgrounds respectively

# Demo 📺

Screenshot 2022-06-22 at 12 04 38
Screenshot 2022-06-22 at 12 04 30

### DIFF
--- a/libs/ui-toolkit/src/utils/intent.tsx
+++ b/libs/ui-toolkit/src/utils/intent.tsx
@@ -19,7 +19,8 @@ export const getIntentShadow = (intent?: Intent) => {
 
 export const getVariantBackground = (variant?: Intent) => {
   return {
-    'bg-black dark:bg-white': variant === Intent.None,
+    'bg-black text-white dark:bg-white dark:text-black':
+      variant === Intent.None,
     'bg-vega-pink text-black dark:bg-vega-yellow dark:text-black-normal':
       variant === Intent.Primary,
     'bg-danger text-white': variant === Intent.Danger,


libs/ui-toolkit/src/utils/intent.tsx
```diff
@@ -19,7 +19,8 @@ export const getIntentShadow = (intent?: Intent) => {
 
 export const getVariantBackground = (variant?: Intent) => {
```
edd, line 20: **nitpick (non-blocking)**: Nittiest nitpick - the function name implies it will only control the background, while the classnames actually also now control text colour. I can live with this.

dexturr (approved):

libs/ui-toolkit/src/utils/intent.tsx
sam-keen: Cheers @edd , 'tis a good point. I will leave it as this for now here, but change it in the bigger ui-toolkit updates ticket as that already changes this funtion name.

PR #624, opened 2022-06-23
Fix/623 Order fixes

### LINKED ISSUE #623: Problems with orders
- [x] Fix transaction dialog to show requested state
- [x] Format size and volume appropriately
- [x] Fix fractional size validation
- [x] Fix user rejected error (shouldn't show error state)

mattrussell36:
# Related issues 🔗

Closes #623 

# Description ℹ️

- Formats sizes and volumes in positions table, order table, trades table and order book with positionDecimalPlaces.
- Adds missing requested state for order dialog so user is prompted to confirm tx in wallet.
- Closes order dialog when order is rejected by user.
- Fixes step for input so deal ticket order size validation works

# Technical 👨‍🔧

- 401 response wasn't defined in wallet service api schema so added it, as a result updated vegwallet-service-api-client package to get latest changes.
- Also fixed some incorrect imports in tests


libs/deal-ticket/src/components/order-dialog.tsx
```diff
@@ -72,7 +89,14 @@ export const OrderDialog = ({
         <p>{t(`Market: ${finalizedOrder.market.name}`)}</p>
       )}
       <p>{t(`Type: ${finalizedOrder.type}`)}</p>
-      <p>{t(`Amount: ${finalizedOrder.size}`)}</p>
+      <p>
+        {t(
```
notbot00, line 93: chore (non-blocking): i feel like these translate functions will break a lot of things when / if we switch over to a real translation library. this looks pretty much like what @lingui provides, except that the translation strings can't be interpolated like this, you usually need to provide the variables to it separately, so depending the language it can figure out where it needs to be inserted. 
probably not a big deal, but we could potentially minimise the future friction if we'd start using a real library instead. they usually just return plain english if there's no extra setup. maybe not super important right now tho.

libs/wallet/src/connectors/rest-connector.ts
```diff
@@ -87,16 +87,22 @@ export class RestConnector implements VegaConnector {
...
+      return res;
     } catch (err) {
       return this.handleSendTxError(err);
     }
   }
 
   private handleSendTxError(err: unknown) {
-    if (typeof err === 'object' && err && 'body' in err) {
+    if (typeof err === 'object' && err && 'body' in err && 'code' in err) {
+      // @ts-ignore Not sure why TS can't infer that 'body' and 'code' do indeed exist on object
```
notbot00, line 99: comment (non-blocking-smartassness): yeah, this is pretty annoying in JS, you can pretty much throw anything, doesn't have to be an Error instance, that's why ts enforces you to treat it as 'unknown' or 'any'. another option here could be to do `if (err instanceof Error) {...}`

notbot00 (approved):

bglownia (approved):

libs/wallet/src/connectors/rest-connector.ts
mattrussell36: I've added a type guard to tidy this up. I can't do instanceof Error here because the api client throws a plain object

```
interface ServiceError {
  code: number;
  body: string | undefined;
  headers: object;
}

export const isServiceError = (err: unknown): err is ServiceError => {
  // Some responses don't contain body object
  if (typeof err === 'object' && err !== null && 'code' in err) {
    return true;
  }
  return false;
};
```

PR #635, opened 2022-06-24
Feat/526 consolelite design update market list

### LINKED ISSUE #526: ConsoleLite Design Update: Market List

macqbat:
# Related issues 🔗

Closes #526

# Description ℹ️

Make market list view closer to design outlines

# Demo 📺
image
image

# Technical 👨‍🔧

Use AgGrid component in different way than it was before - omit market-list lib which is useless for us, use toolkit/agGrid library with customizations done by new param for passing css chunks


dexturr (approved):

apps/simple-trading-app/src/app/components/simple-market-list/simple-market-list.spec.tsx
```diff
@@ -130,8 +137,12 @@ describe('SimpleMarketList', () => {
       await new Promise((resolve) => setTimeout(resolve, 0));
     });
     await waitFor(() => {
-      expect(screen.getByTestId('simple-market-list')).toBeInTheDocument();
+      expect(
+        document.querySelector('.ag-center-cols-container')
```
notbot00, line 141: question (non-blocking): this seems to be testing a third party implementation, as this className gets added by AgGrid, if I'm not mistaken. not sure if relying on the third party class name is better here opposed to the previous version of having an actual test id and testing if that is in the document.

notbot00 (approved):

apps/simple-trading-app/src/app/components/simple-market-list/simple-market-list.spec.tsx
macqbat: I got your point, but here we need to wait for and catch a moment when agGrid is rendered, and since it not accept param as `data-testid`, relaying on catching this would be ineffective.
I think if they would change a name of the css class it would be a breaking change, and the correspondingly bumped version number would not let for an automatic uncontrolled upgrade of the package and all the problems come with it.
Other thing, their [blog](https://blog.ag-grid.com/using-react-testing-library-with-ag-grid/) also recommends smth similar..

Issue #641, opened 2022-06-24
[token] `PROPOSAL_ERROR_INSUFFICIENT_EQUITY_LIKE_SHARE` on a proposal breaks governance page


edd:
# To recreate
- Have a party with money
- Have a wallet for that party locally
- Submit a market update proposal like the following:

```
  ./cmd/vegawallet/vegawallet command send --wallet edd --pubkey 7e84f93c7f5595b57c6924e5fb2647bb13fe64e7eb5540fc5436772f2a2c1d8a --network fairground '{
   "proposalSubmission": {
    "rationale": {
     "description": "Update Lorem Ipsum market"
    }, "priceMonitoringParameters": {
    "terms": {ers": [
     "updateMarket": {
      "marketId": "0097a6e72a65321e16c5d0e88ededbb6e7b1d344a9d5d435f2781efe8982cf18",
      "changes": {lity": "0.9999999",
       "instrument": {nsion": "600"
        "code": "ORANGES.24h",
        "future": {
         "quoteName": "tEuro",
         "settlementPriceDecimals": 5,
         "oracleSpecForSettlementPrice": {
          "pubKeys": [arameter": 0.001,
           "0xfCEAdAFab14d46e20144F48824d0C09B1a03F2BC"
          ],": 0,
          "filters": [
           {gma": 0.5
            "key": {
             "name": "prices.BTC.value",
             "type": "TYPE_INTEGER"
            },
            "conditions": [57721401,
             {tTimestamp": 1657807801
              "operator": "OPERATOR_GREATER_THAN",
              "value": "0"
             }
            ]
           }
          ]
         },
         "oracleSpecForTradingTermination": {
          "pubKeys": [],
          "filters": [
           {
            "key": {
             "name": "vegaprotocol.builtin.timestamp",
             "type": "TYPE_TIMESTAMP"
            },
            "conditions": [
             {
              "operator": "OPERATOR_GREATER_THAN_OR_EQUAL",
              "value": "1648684800000000000"
             }
            ]
           }
          ]
         },
         "oracleSpecBinding": {
          "settlementPriceProperty": "prices.BTC.value",
          "tradingTerminationProperty": "vegaprotocol.builtin.timestamp"
         }
        }
       },
       "priceMonitoringParameters": {
        "triggers": [
         {
          "horizon": "43200",
          "probability": "0.9999999",
          "auctionExtension": "600"
         }
        ]
       },
       "logNormal": {
[...]

PR #642, opened 2022-06-24
chore: workaround for #641 - ignoreErrors on governance page

edd:
# Related issues 🔗

Workaround for #641, ignores request errors, which allows us to render the page again

# Description ℹ️

Due to an API issue in the GraphQL API, the proposals request contains an error if a market update proposal is rejected
because the proposer had insufficient stake. This errorPolicy change allows the page to render again, which is more
useful.

This should be reverted when https://github.com/vegaprotocol/data-node/issues/740 is fixed

### DIFF
--- a/apps/token/src/routes/governance/proposals/proposals-container.tsx
+++ b/apps/token/src/routes/governance/proposals/proposals-container.tsx
@@ -24,7 +24,7 @@ export const ProposalsContainer = () => {
   const { t } = useTranslation();
   const { data, loading, error } = useQuery<Proposals, never>(PROPOSALS_QUERY, {
     pollInterval: 5000,
-    errorPolicy: 'ignore',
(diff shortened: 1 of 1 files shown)


apps/token/src/routes/governance/proposals/proposals-container.tsx
```diff
@@ -24,6 +24,7 @@ export const ProposalsContainer = () => {
   const { t } = useTranslation();
   const { data, loading, error } = useQuery<Proposals, never>(PROPOSALS_QUERY, {
     pollInterval: 5000,
+    errorPolicy: 'ignore',
```
mattrussell36, line 27: chore (non-blocking): Can we add a comment here to serve as a reminder that this is to get around some backend issues and should be removed in future.

mattrussell36 (approved):

dexturr (approved):

PR #643, opened 2022-06-24
Test/154 market summary

### LINKED ISSUE #154: Trading page market summary
- Additional market information to be displayed above the chart
  - Change summary
  - Volume
  - Trading mode
  - State

jtsang586:
# Related issues 🔗

Closes #154 

# Description ℹ️

Additional assertions for market summary using mocks to generate the market data required for it, this is added to:

- can select an active market
- can select a suspended market

There was a lint error in apps/trading-e2e/src/support/mocks/generate-market.ts due to module boundaries which I set to ignore. Not sure what the best alternative solution to this is. Happy to make any suggested changes.

### DIFF
--- a/apps/trading-e2e/src/integration/markets.cy.ts
+++ b/apps/trading-e2e/src/integration/markets.cy.ts
@@ -93,25 +93,25 @@ describe('markets table', () => {
 
     verifyMarketSummaryDisplayed('Suspended');
   });
+
+  function verifyMarketSummaryDisplayed(expectedMarketState: string) {
+    const marketSummaryBlock = 'market-summary';
+    const percentageValue = 'price-change-percentage';
+    const priceChangeValue = 'price-change';
+    const tradingVolume = 'trading-volume';
+    const tradingMode = 'trading-mode';
+    const marketState = 'market-state';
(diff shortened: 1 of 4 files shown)


apps/trading-e2e/src/integration/markets.ts
```diff
@@ -88,5 +90,28 @@ describe('markets table', () => {
     cy.wait('@Market');
     cy.contains('SUSPENDED MARKET');
     cy.url().should('include', '/markets/market-1');
+
+    verifyMarketSummaryDisplayed('Suspended');
   });
 });
+
+function verifyMarketSummaryDisplayed(expectedMarketState: string) {
```
mattrussell36, line 98: suggestion (non blocking): If we aren't re using this function prefer to just keep it inline

mattrussell36 (approved):
Looks good to me, suggestion to me

dexturr (approved):

apps/trading-e2e/src/integration/markets.ts
jtsang586: Done...I think 😄

PR #645, opened 2022-06-24
Feat/462 fills table mvp (no pagination)

### LINKED ISSUE #462: Fills (all my trades / fills for specific order)
As a user of the console I would like to see all my fills, ideally this would be for all my trades AND for a specific order.

### Dependancies:

We need to check with the data-node team if data-node has a fill by party id API? (workaround could be to get all then filter out)

mattrussell36:
# Related issues 🔗

Closes #462 

# Description ℹ️

Adds a fills lib with a fills table and renders it into the portfolio page grid

# Demo 📺

Screen Shot 2022-06-24 at 4 19 05 PM
Screen Shot 2022-06-24 at 4 18 58 PM

# Technical 👨‍🔧

- Uses standard data provider pattern
- Needed to use a storybook decorator with a mutation observer watching the document body element in order to have a correctly functioning theme context.
- Updated portfolio page to fix some layout issue which were causing tables not to render (no height set), also removed filters section for now as this will likely need to be handled by the portfolio grid child.
- Added storybook instance for fills lib


libs/fills/tsconfig.spec.json
```diff
@@ -0,0 +1,19 @@
...
+  "include": [
+    "**/*.test.ts",
+    "**/*.spec.ts",
+    "**/*.test.tsx",
+    "**/*.spec.tsx",
+    "**/*.test.js",
+    "**/*.spec.js",
+    "**/*.test.jsx",
+    "**/*.spec.jsx",
+    "**/*.d.ts"
```
MadalinaRaicu, line 17: would be great to exclude all the stories.* files from tsconfig.lib.json as in https://github.com/vegaprotocol/frontend-monorepo/pull/610/files#diff-c4c808cbef23c19ee8c600f01c29a81bab02a57a053fe1bcebcee1ccd173aec7

libs/fills/tsconfig.json
```diff
@@ -0,0 +1,25 @@
...
+  },
+  "files": [],
+  "include": [],
+  "references": [
+    {
+      "path": "./tsconfig.lib.json"
+    },
+    {
+      "path": "./tsconfig.spec.json"
+    }
```
MadalinaRaicu, line 23:
```suggestion
    }, 
    {
      "path": "./.storybook/tsconfig.json"
    }
```

libs/fills/tsconfig.lib.json
```diff
@@ -0,0 +1,22 @@
...
+  ],
+  "exclude": [
+    "**/*.spec.ts",
+    "**/*.test.ts",
+    "**/*.spec.tsx",
+    "**/*.test.tsx",
+    "**/*.spec.js",
+    "**/*.test.js",
+    "**/*.spec.jsx",
+    "**/*.test.jsx"
```
MadalinaRaicu, line 19:
```suggestion
    "**/*.test.jsx",
    "**/*.stories.ts",
    "**/*.stories.js",
    "**/*.stories.jsx",
    "**/*.stories.tsx"
```

MadalinaRaicu (approved):

libs/fills/src/lib/fills-table.tsx
```diff
@@ -0,0 +1,154 @@
...
+  fills: FillFields[];
+}
+
+export const FillsTable = forwardRef<AgGridReact, FillsTableProps>(
+  ({ partyId, fills }, ref) => {
+    return (
+      <AgGrid
+        ref={ref}
+        rowData={fills}
+        overlayNoRowsTemplate="No fills"
```
bglownia, line 28: non-blocking: This label should be wrapped with `t()`

libs/fills/src/lib/fills-table.tsx
```diff
@@ -0,0 +1,154 @@
...
+              data.market.decimalPlaces
+            );
+            return `${valueFormatted} ${asset}`;
+          }}
+        />
+        <AgGridColumn
+          headerName={t('Role')}
+          field="aggressor"
+          valueFormatter={({ value, data }: ValueFormatterParams) => {
+            const taker = t('Taker');
```
bglownia, line 101: Those two can be defined and translated in module scope instead of valueFormatter functon scope

libs/fills/src/lib/fills-table.stories.tsx
```diff
@@ -0,0 +1,18 @@
+import type { Story, Meta } from '@storybook/react';
```
bglownia, line 1: I'm not sure if we need story for FillsTable

bglownia (approved):

Issue #612, opened 2022-06-22
Light theme (and theme toggle) for Token


sam-keen:
Following the same conventions as the other apps in the monorepo, enable a theme switcher and refactor components as necessary to support a light and dark mode.

PR #650, opened 2022-06-27
Feat/629 Clean up Token styling

### LINKED ISSUE #629: Visual cleanup of Token
The transition to the monorepo using ui-toolkit has meant the styling of Token has taken a hit. Some cleanup is required to get it looking presentable, mostly spacing and text styles.

Success criteria - nothing specific. A subjectively 'better' feel that's closer to TFE in terms of spacing and text.

sam-keen:
# Related issues 🔗

Closes #629

# Description ℹ️

An attempt to bring Token styling (in particular, the header, font styles and spacing) in line with the original TFE repo

# Demo 📺

Screenshot 2022-06-27 at 15 02 25


apps/token/src/styles.css
```diff
@@ -1,3 +1,15 @@
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
+
+@layer base {
```
sam-keen, line 5: @mattrussell36 I think this might be controversial with you, but having played with styled components to achieve the same result for these text elements, I think this is a far better solution for the following reasons:

1. Tailwind applies these classes directly as CSS as they form part of the base styling, with no classes coming through into the DOM. This means we can apply Tailwind classes to modify styling without collisions.
2. It significantly reduces clutter in the form of hundreds of classes added to the DOM to achieve the same effect.
3. It's simpler than custom styled components. Fresh devs don't need to understand to import a special component to use, instead relying on the h1, p tags etc they'd expect.

mattrussell36: chore (non blocking): As discussed on slack I'd prefer to add a wrapping class to apply styles to elements, this makes our config diverge less between apps.

```
.text-flow p {
  @apply mb-12
}
```

campbellssource:
Generally looking good. but here one snag:
Screenshot 2022-06-28 at 09 23 28
The associate with Vega keys is back on black
Screenshot 2022-06-28 at 09 24 41
perhaps something to do with the fact that the ETH wallet used to be white.

(BTW I'm :fine: with it being black)

campbellssource:
Blocking (but maybe not to do with this branch, although issues does not happen on token.fairground.wtf):
https://deploy-preview-650--token-dapp-dev.netlify.app/vesting
Gives me "something went wrong"
Screenshot 2022-06-28 at 09 28 55

dexturr (approved):

apps/token/src/components/wallet-card/wallet-card.tsx
```diff
@@ -29,10 +29,14 @@ interface WalletCardProps {
...
 export const WalletCard = ({ dark, children }: WalletCardProps) => {
-  const className = classNames('text-ui border border-white', 'p-8', {
-    'bg-black text-white': dark,
-    'bg-white text-black': !dark,
-  });
+  const className = classNames(
+    'text-ui border border-white',
+    'pt-4 pl-8 pr-12 pb-12',
+    {
+      'bg-black text-white': dark,
```
notbot00, line 36: question (non-blocking): do we need this dark property? normally I see the tailwind prefix "dark:xyz-classname". are there cases where we want to override the theme setting and force e.g. a light appearance on a dark theme, or is the context here something completely different?

sam-keen: @notbot00 this is an older implementation from TFE that's not been fully removed yet. It'll get done when we add a light theme https://github.com/vegaprotocol/frontend-monorepo/issues/612

PR #666, opened 2022-06-28
Feat/463 vega transaction component links to block explorer

MadalinaRaicu:
# Related issues 🔗

Closes [#463]

# Description ℹ️

In order to make the UX around being able to see and understand transactions made in the console, we need to have links for all transactions to the block explorer.

# Demo 📺

image
# Technical 👨‍🔧

- [x]  Links take you to the view of that transaction in the block explorer (opens another tab)


libs/wallet/src/vega-order-transaction-dialog/vega-order-transaction-dialog.tsx
```diff
@@ -74,7 +76,15 @@ export const VegaOrderTransactionDialog = ({
       >
         {transaction.txHash && (
           <p data-testid="tx-hash" className="break-all">
-            {t(`Tx hash: ${transaction.txHash}`)}
+            Tx hash: &nbsp;
+            <a
```
dexturr, line 80: chore (blocking): can we please add a test to verify this?

apps/trading/lib/config/env.ts
```diff
@@ -17,5 +17,9 @@ export const ENV = {
     process.env['NX_TRADING_SENTRY_DSN']
   ),
   envName: windowOrDefault('NX_VEGA_ENV', process.env['NX_VEGA_ENV']),
-  vegaUrl: windowOrDefault('NX_VEGA_URL', process.env['NX_VEGA_URL']),
+  vegaUrl: windowOrDefault('NX_VEGA_URL', process.env['X_VEGA_URL']),
+  explorerUrl: windowOrDefault(
```
notbot00, line 21: chore (blocking): i think it is confusing to have this duplication in both libs/environment and here. looks like you are using the env hook to get the value, so this here can be deleted imo.

dexturr: This was my mistake when we were talking about it. My bad @MadalinaRaicu

libs/wallet/src/vega-order-transaction-dialog/vega-order-transaction-dialog.tsx
MadalinaRaicu: sure, pushing it now

apps/trading/lib/config/env.ts
```diff
@@ -17,5 +17,5 @@ export const ENV = {
     process.env['NX_TRADING_SENTRY_DSN']
   ),
   envName: windowOrDefault('NX_VEGA_ENV', process.env['NX_VEGA_ENV']),
-  vegaUrl: windowOrDefault('NX_VEGA_URL', process.env['NX_VEGA_URL']),
+  vegaUrl: windowOrDefault('NX_VEGA_URL', process.env['X_VEGA_URL']),
```
MadalinaRaicu, line 20:
```suggestion
  vegaUrl: windowOrDefault('NX_VEGA_URL', process.env['NX_VEGA_URL']),
```

notbot00 (approved):

dexturr (approved):

PR #668, opened 2022-06-28
Feat/hosted wallet

### LINKED ISSUE #651: Connect with an hosted (fairground) wallet
In order to help early use of console v2 in incentives we need to allow connection with an existing hosted wallet

to display on fairground only
- [ ] UI to input custom endpoints
- [ ] ensure vega wallet API service client can switch endpoints

dexturr:
# Related issues 🔗

Closes #651

# Description ℹ️

Allow users to specify a wallet URL that is not localhost:1789
Default to localhost:1789 if the user does not interact with this field
This enables support for the hosted wallet
Minor bug fix for disconnecting not clearing up all state required

# Demo 📺

Screenshot 2022-06-28 at 22 52 45

### DIFF
--- a/libs/wallet/src/connect-dialog/connect-dialog.spec.tsx
+++ b/libs/wallet/src/connect-dialog/connect-dialog.spec.tsx
@@ -92,7 +92,40 @@ it('Successful connection using rest auth form', async () => {
     fireEvent.submit(screen.getByTestId('rest-connector-form'));
   });
 
-  expect(spy).toHaveBeenCalledWith(fields);
+  expect(spy).toHaveBeenCalledWith('http://localhost:1789', fields);
+
+  expect(defaultProps.setDialogOpen).toHaveBeenCalledWith(false);
+});
+
+it('Successful connection using custom url', async () => {
+  const spy = jest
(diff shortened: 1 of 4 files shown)


mattrussell36 (approved):

libs/wallet/src/connect-dialog/connect-dialog.spec.tsx
```diff
@@ -117,7 +150,7 @@ it('Unsuccessful connection using rest auth form', async () => {
     fireEvent.submit(screen.getByTestId('rest-connector-form'));
   });
 
-  expect(spy).toHaveBeenCalledWith(fields);
+  expect(spy).toHaveBeenCalledWith('http://localhost:1789/api/v1', fields);
```
sam-keen, line 153: Chore (non-blocking): this url's used a few times, maybe nicer to put in a var

sam-keen (approved):

libs/wallet/src/connect-dialog/connect-dialog.spec.tsx
dexturr: will put in a follow up, want to get this merged

PR #691, opened 2022-06-30
#638 Add pagination support to generic-data-provider

### LINKED ISSUE #638: Add pagination support to generic-data-provider

bglownia:
# Related issues 🔗

Closes #638

# Description ℹ️

Add pagination support to generic-data-provider. Used in Fills.
Markets list table switched to use infinite row model type.


libs/react-helpers/src/lib/generic-data-provider.ts
```diff
@@ -13,9 +13,30 @@ export interface UpdateCallback<Data, Delta> {
...
+  last?: number;
+  before?: string;
+}
+
+export interface PageInfo {
+  startCursor?: string;
+  endCursor?: string;
+  hasNextPage?: boolean;
+  hasPreviousPage?: boolean;
+}
```
macqbat, line 39: since each prop is optional maybe would be little bit better like:

```
export interface Partial<Pagination> {
  first: number;
  after: string;
  last: number;
  before: string;
}

export interface Partial<PageInfo> {
  startCursor: string;
  endCursor: string;
  hasNextPage: boolean;
  hasPreviousPage: boolean;
}
```

macqbat (approved):

apps/simple-trading-app/src/app/components/simple-market-list/data-provider.ts
```diff
@@ -91,14 +91,15 @@ export const FILTERS_QUERY = gql`
 const update = (
   data: SimpleMarkets_markets[],
   delta: SimpleMarketDataSub_marketData
-) =>
-  produce(data, (draft) => {
+) => {
+  return produce(data, (draft) => {
```
notbot00, line 95: chore (non-blocking): this change looks redundant here

libs/accounts/src/lib/accounts-data-provider.ts
```diff
@@ -55,8 +55,8 @@ export const getId = (
 const update = (
   data: Accounts_party_accounts[],
   delta: AccountSubscribe_accounts
-) =>
-  produce(data, (draft) => {
+) => {
```
notbot00, line 58: chore (non-blocking): same here

libs/fills/src/lib/fills-manager.tsx
```diff
@@ -1,80 +1,92 @@
...
+    Fills_party_tradesPaged_edges[],
+    FillsSub_trades[]
+  >({ dataProvider, update, insert, variables });
+  totalCountRef.current = totalCount;
+  dataRef.current = data;
 
-  const fills = useMemo(() => {
-    if (!data?.length) {
-      return [];
+  const getRows = async ({
```
notbot00, line 57: comment (non-blocking): ngl, i find this piece of code extremely hard to understand, and very not reacty. creating a function here to then be able to modify the result of another hook from inside of a child component without passing the data itself in feels just wrong to me, but again, i am not sure i understand the whole context of what this is trying to achieve. will need some time to go over it and unwrap things, i don't have any suggestions just yet on how to potentially improve this, but the general impression is that code just doesn't need to be this hard to read.

libs/accounts/src/lib/accounts-data-provider.ts
bglownia: This was requested by Matt, he prefer `"arrow-body-style": ["error", "always"]`

libs/order-list/src/lib/components/order-data-provider/order-data-provider.ts
```diff
@@ -79,8 +79,8 @@ export const prepareIncomingOrders = (delta: OrderFields[]) => {
   return incoming;
 };
 
-const update = (data: OrderFields[], delta: OrderFields[]) =>
-  produce(data, (draft) => {
+const update = (data: OrderFields[], delta: OrderFields[]) => {
+  return produce(data, (draft) => {
```
notbot00, line 83: chore (non-blocking): any reason for preferring a function body + explicit return? seen a few changes like this in the PR, got me curious :)

libs/positions/src/lib/positions-data-provider.ts
```diff
@@ -78,15 +78,17 @@ export const POSITIONS_SUB = gql`
 const update = (
   data: Positions_party_positions[],
   delta: PositionSubscribe_positions
-) =>
-  produce(data, (draft) => {
+) => {
+  return produce(data, (draft) => {
```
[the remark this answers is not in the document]
bglownia: This was requested by Matt, he prefer `"arrow-body-style": ["error", "always"]`

libs/order-list/src/lib/components/order-data-provider/order-data-provider.ts
bglownia: This was requested by Matt, he prefer `"arrow-body-style": ["error", "always"]`

libs/react-helpers/src/hooks/use-data-provider.ts
```diff
@@ -32,30 +48,48 @@ export function useDataProvider<Data, Delta>(
...
       setLoading(loading);
       if (!error && !loading) {
-        // if update function returns true it means that component handles updates
+        // if update or insert function returns true it means that component handles updates
         // component can use flush() which will call callback without delta and cause data state update
-        if (!initialized.current || !delta || !update || !update(delta)) {
-          initialized.current = true;
-          setData(data);
+        if (initialized.current) {
+          if (delta && update && update({ delta, data })) {
```
notbot00, line 65: nit (non-blocking): one thing this which makes this harder-than-it-has-to-be to digest is that the update and the insert return a boolean. 
- if they send the data to somewhere, they probably shouldn't be called as a condition of an if statement, but in the body of it
- if they don't send the data anywhere, tit would be less confusing if  they would be called `shouldUpdate` and `shouldInsert`, or something similar to indicate their function a bit more clearly

libs/react-helpers/src/hooks/use-data-provider.ts
```diff
@@ -10,17 +10,33 @@ import type { Subscribe } from '../lib/generic-data-provider';
...
+  variables,
+}: {
+  dataProvider: Subscribe<Data, Delta>;
+  update?: ({ delta, data }: { delta: Delta; data: Data }) => boolean;
+  insert?: ({
+    insertData,
+    data,
+    totalCount,
+  }: {
+    insertData: Data;
```
notbot00, line 26: nit (non-blocking): insertData sounds more like a function name. if this is actually data, i'd go with something `dataToInsert`, `insertionData` or `newData`. not super important, and i know we can hover over anything in our IDE and look at their type signature.

libs/react-helpers/src/lib/generic-data-provider.ts
```diff
@@ -70,20 +119,60 @@ function makeDataProviderInternal<QueryData, Data, SubscriptionData, Delta>(
...
   let client: ApolloClient<object> | undefined = undefined;
   let subscription: Subscription | undefined = undefined;
+  let pageInfo: PageInfo | null = null;
+  let totalCount: number | undefined;
 
   // notify single callback about current state, delta is passes optionally only if notify was invoked onNext
-  const notify = (callback: UpdateCallback<Data, Delta>, delta?: Delta) => {
+  const notify = (
+    callback: UpdateCallback<Data, Delta>,
+    updateData?: { delta?: Delta; insertData?: Data | null }
```
notbot00, line 128: nit (non-blocking): `updateData` also sounds callable, `dataUpdate` on the other hand sounds more like a piece of data which is meant to update something with. okay, will try to stop being a variable-name-nazi, i think i've dropped enough comments on the topic :)

libs/trades/src/lib/trades-data-provider.ts
```diff
@@ -53,8 +53,8 @@ export const sortTrades = (trades: TradeFields[]) => {
   );
 };
 
-const update = (data: TradeFields[], delta: TradeFields[]) =>
-  produce(data, (draft) => {
+const update = (data: TradeFields[], delta: TradeFields[]) => {
```
notbot00, line 56: chore (non-blocking): another of these patterns, feel free to mark it resolved, just taking a note in case it turns out to be relevant

libs/positions/src/lib/positions-data-provider.ts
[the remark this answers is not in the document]
notbot00: got it, thanks.

libs/trades/src/lib/trades-data-provider.ts
bglownia: This was requested by Matt, he prefer `"arrow-body-style": ["error", "always"]`

libs/react-helpers/src/hooks/use-data-provider.ts
notbot00: i can't say i understood the points regarding ag-grid, my comment was purely related to naming and separation of concerns :)

libs/react-helpers/src/hooks/use-data-provider.ts
bglownia: Good point. I renamed it to `insertionData`

libs/react-helpers/src/lib/generic-data-provider.ts
bglownia: Good point. I renamed it to `dataUpdate`

libs/react-helpers/src/lib/generic-data-provider.ts
```diff
@@ -35,8 +57,29 @@ export interface Update<Data, Delta> {
   (data: Data, delta: Delta, reload: (forceReset?: boolean) => void): Data;
 }
 
+export interface Append<Data> {
+  (
+    data: Data | null,
+    pageInfo: PageInfo,
+    insert: Data | null,
```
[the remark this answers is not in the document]
bglownia: Good point. I renamed `insert` to `insertionData` and `insertPageInfo` to `insertionPageInfo`

Issue #477, opened 2022-05-25
Add Cypress tests for TFE


gordsport:
Manual testing job first

gordsport:
closing as a duplicate of the tickets here:
 - https://github.com/orgs/vegaprotocol/projects/110/views/6?filterQuery=-is%3Aclosed+no%3Astatus+-milestone%3A%22%F0%9F%A5%B6+Icebox%22+-label%3Aconsole-lite+TFE

PR #697, opened 2022-07-01
test(token): add rewards page validations

radoslaw-szpiech:
# Related issues 🔗

Relates to #477 

# Description ℹ️

UI validation tests for elements on Rewards page without wallets connected.

### DIFF
--- a/apps/token-e2e/src/integration/rewards.cy.js
+++ b/apps/token-e2e/src/integration/rewards.cy.js
@@ -1,0 +1,41 @@
+import navigation from '../locators/navigation.locators';
+import rewards from '../locators/rewards.locators';
+
+context('Rewards Page - verify elements on page', function () {
+  before('navigate to rewards page', function () {
+    cy.visit('/')
+      .get(navigation.section)
+      .within(() => {
+        cy.get(navigation.rewards).click();
+      });
+  });
+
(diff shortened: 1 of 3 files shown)


apps/token-e2e/src/integration/rewards.cy.js
```diff
@@ -0,0 +1,41 @@
...
+        .should('be.visible')
+        .and('have.text', 'Rewards');
+    });
+
+    it('should have epoch warning', function () {
+      cy.get(rewards.warning)
+        .should('be.visible')
+        .and(
+          'have.text',
+          'Rewards are credited 5 minutes after the epoch ends.This delay is set by a network parameter'
```
dexturr, line 31: comment (non-blocking): this text is generated based on the network parameter that is being set. As we are using capsule we have control of this but if the param changes the test breaks. I don't know that it needs a solution but perhaps something worth mentioning.

dexturr (approved):

apps/token-e2e/src/integration/rewards.cy.js
radoslaw-szpiech: Yes, I am aware of that. If we decide to modify this parameter we will just change it to env variable or sth.

PR #700, opened 2022-07-01
Feat: Containerised docker images

### LINKED ISSUE #431: Publish docker containers to docker hub
Related to #296, we need a CI pipeline to generate these containers and publish them to docker hub

These can then be passed to capsule in the post_start functionality to allow people to spin up a whole network with the frontends connected as well.

dexturr:
# Related issues 🔗

Closes #431 

# Description ℹ️

Publish containers for all the frontends to dockerhub. To facilitate usage in capsule


.github/workflows/publish-containers.yml
```diff
@@ -0,0 +1,55 @@
...
+#       - "v[0-9]+.[0-9]+.[0-9]+"
+#       - "v[0-9]+.[0-9]+.[0-9]+-pre[0-9]+"
+
+on:
+  pull_request:
+jobs:
+  master:
+    strategy:
+      matrix:
+        version: ['0.0.1']
```
mkjmdski, line 15: it's better if you track version in git in form of a tag and read it with `git tag` or `git descirbe`

.github/workflows/publish-containers.yml
```diff
@@ -0,0 +1,55 @@
...
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Docker build
+        uses: docker/build-push-action@v3
+        with:
+          context: '.'
+          file: './Dockerfile'
+          build-args: 'APP=${{ matrix.app }}'
+          platforms: ${{ matrix.build-platform }}
+          tags: ${{ steps.tags.outputs.latest }},${{ steps.tags.outputs.current }}
```
mkjmdski, line 55: you need to add `push: true` flag according to the docks -> https://github.com/marketplace/actions/build-and-push-docker-images

daniel1302: But yeah, it will be pushing a new version with the same tag because this pipeline means to run every PR a3d version is hardcoded.

dexturr: yeah this just a draft, I wanted a PoC I will change it so it's all reasonable once I have it working I will make it robust. just wanted to get over the initial hurdle of getting it up 👍

notbot00 (approved):

Dockerfile.cra
```diff
@@ -7,6 +7,7 @@ ENV PATH /app/node_modules/.bin:$PATH
 COPY package.json ./
 COPY yarn.lock ./
 COPY . ./
+RUN apk add python3 make gcc g++
```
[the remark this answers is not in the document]
dexturr: Extra things for gyp when building on arm. They are subtely different, the next apps we need to run `export` not `build` so that is the only difference right now. I am not docker savvy enough to know if we could share these. I suppose we could take a build arg for that the command to run is. Though I think this is fine until we see how much these are needed.

mattrussell36 (approved):

PR #719, opened 2022-07-06
Feat/622 custom node

### LINKED ISSUE #622: Node switcher
Some users will want to know where the website is reading 

Suggest doing this in the footer.

Screenshot 2022-06-22 at 13 55 17

This opens a component that is shared across apps for switching between known nodes.
https://www.figma.com/file/F995ESZ0QMpTBuQWGAgLRz/Design-system?node-id=1386%3A67731

Moved this from https://github.com/vegaprotocol/token-frontend/issues/931

notbot00:
# Related issues 🔗

Closes #622 

# Description ℹ️

Builds on top of https://github.com/vegaprotocol/frontend-monorepo/pull/698

# Demo 📺

Screenshot 2022-07-05 at 16 47 13

# Technical 👨‍🔧

- exposes a new input field for entering a custom node url
- implements validation for the custom node field: the input needs to be a valid url, the node needs to be on the correct network, graphql subscription queries work
- allows the connecting to the custom node if it passes validation


notbot00:
> 
> I'm not sure why (it doesn't look like it's a `git export`), but there is an environment variable `BRANCH` that will get the same info, so use that if set and only run the git command when it's not available and it should work.

thanks for the catch @edd ! patched this up now.

edd:
In this PR I see the edit link now:

Screenshot 2022-07-11 at 17 42 46

But most of the links are broken:
- Mainnet url goes to a subfolder, needs `https://` at the start
- stagnet goes to dev

PR #727, opened 2022-07-08
fix: use netlify git env vars

notbot00:
Netlify doesn't expose git when building, need to fix this in the env setup.

### DIFF
--- a/tools/utils/setup-environment.ts
+++ b/tools/utils/setup-environment.ts
@@ -5,17 +5,25 @@ import * as log from 'npmlog';
 import * as dotenv from 'dotenv';
 import type { ExecutorContext } from '@nrwl/devkit';
 
-process.env['NX_GIT_COMMIT_HASH'] = execSync('git rev-parse HEAD')
-  .toString()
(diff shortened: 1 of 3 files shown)


dexturr (approved):

tools/utils/setup-environment.ts
```diff
@@ -5,17 +5,25 @@ import * as log from 'npmlog';
...
-  .replace(/[\r\n]/gm, '');
-process.env['NX_GIT_ORIGIN_URL'] = execSync('git remote get-url origin')
-  .toString()
-  .replace('ssh://git@', 'https://')
-  .replace('.git', '')
-  .replace(/[\r\n]/gm, '');
+try {
+  process.env['NX_GIT_COMMIT_HASH'] =
+    process.env['COMMIT_REF'] ??
+    execSync('git rev-parse HEAD')
```
dexturr, line 11: chore (non-blocking): if these are only going to be used in a dev environement I would be ok with just simplifying this and setting hash/branch/repo to just "dev" rather than scripts

edd (approved):

PR #728, opened 2022-07-08
Test/694 block explorer refactor

### LINKED ISSUE #694: Refactor block explorer e2e tests
Currently the Block explorer tests are running in BDD using page object model. The aim of the refactor is to change the tests to align with what is currently in Token and trading app.

jtsang586:
# Related issues 🔗

Closes #694 

# Description ℹ️

Refactors existing block explorer tests to align with the new testing framework in Token. 

- All files related to page object model have been removed, replaced with the more standard Cypress test files
- Previous tests set to ignore are now skipped 
- Removed old transaction assertion which was causing noise in the slack channel for no benefit as there are no transactions currently after starting up Capsule


apps/explorer-e2e/src/integration/asset.cy.js
```diff
@@ -0,0 +1,20 @@
+import '../support/common.functions';
+
+context('Asset page', function () {
+  describe('Verify elements on page', function () {
+    const assetsNavigation = 'a[href="/assets"]';
+    const assetHeader = '[data-testid="asset-header"]';
```
radoslaw-szpiech, line 6: comment (non-blocking) Don't know if it makes any change, but consts may be declared even before context block

apps/explorer-e2e/src/integration/asset.cy.js
```diff
@@ -0,0 +1,20 @@
...
+
+    it('Assets page is displayed', function () {
+      cy.visit('/');
```
radoslaw-szpiech, line 9: suggestion: cy.visit should be placed in before hook

apps/explorer-e2e/src/integration/home.cy.js
```diff
@@ -0,0 +1,141 @@
...
+    it('Should show connected environment', function () {
+      const deployedEnv = Cypress.env('environment').toUpperCase();
+      cy.get('[data-testid="stats-environment"]').should(
+        'have.text',
+        `/ ${deployedEnv}`
+      );
+    });
+
+    it('should show connected environment stats', function () {
+      const statTitles = [
```
radoslaw-szpiech, line 18: suggestion (non-blocking): maybe object instead? as it's long AF and easy to make mistake

apps/explorer-e2e/src/integration/transactions.cy.js
```diff
@@ -0,0 +1,87 @@
...
+    it('transactions details page is displayed in mobile', function () {
+      cy.common_switch_to_mobile_and_click_toggle();
+      cy.get(transactionNavigation).click();
+      clickTopTransaction();
+      validateTxDetailsAreDisplayed();
+    });
+
+    function clickTopTransaction() {
+      // eslint-disable-next-line cypress/no-unnecessary-waiting
+      cy.wait(1000); // Wait for transactions to load if there are any
```
radoslaw-szpiech, line 34: suggestion: cy.get with timeout would be better

apps/explorer-e2e/src/integration/asset.cy.js
jtsang586: I didn't put this in a before hook because `cy.common_switch_to_mobile_and_click_toggle();` already goes to homepage so only 1 tests needs cy.visit

apps/explorer-e2e/src/integration/transactions.cy.js
jtsang586: Originally it was done this way but we didn't want tests failing when there were no transactions. The tests using this function are not running at the moment because there are no transactions in Capsule. This will probably be refactored when we get transactions in

apps/explorer-e2e/src/integration/home.cy.js
jtsang586: Done

radoslaw-szpiech (approved):

dexturr (approved):

PR #733, opened 2022-07-08
Feat/671 Proposal listings page

### LINKED ISSUE #671: New proposal listing page layout
To list all of the proposals e.g.

Screenshot 2022-04-05 at 10 39 30
note: consider using a line order background colours instead of a title to show the difference between live and done proposals 

Here are some other 

- Filter, group and sort proposals
Screenshot 2022-04-05 at 10 40 20

- Link to open search/filter proposals
- [ ] Raise ticket in data node to filter/search
Screenshot 2022-04-05 at 10 42 14

## Proposal list item component: Proposal teaser
- Header for each proposal type. https://www.figma.com/file/F995ESZ0QMpTBuQWGAgLRz/Vega-team-library?node-id=1792%3A70076 . each proposal type has its own type of header
- Sub title for each proposal type
Screenshot 2022-04-05 at 10 54 48

## Proposal list item component: Vote summary
- Show the proposal status
- Show the time status (and vote status)
- Show vote tally
Screenshot 2022-04-05 at 10 55 45

[All Designs in Figma](https://www.figma.com/file/F995ESZ0QMpTBuQWGAgLRz/Vega-team-library?node-id=501%3A18730)

sam-keen:
# Related issues 🔗

Closes #671

# Description ℹ️

Creates a new Token proposals listing page. 

# Demo 📺
Screenshot 2022-07-26 at 20 20 19


apps/token/src/routes/governance/components/proposal-detail-header/proposal-header.spec.tsx
```diff
@@ -0,0 +1,243 @@
...
+const proposal = generateProposal();
+
+const renderComponent = (proposal: Proposals_proposals) => (
+  <ProposalHeader proposal={proposal} />
+);
+
+it('New market', () => {
+  render(
+    renderComponent({
+      ...proposal,
```
mattrussell36: chore (blocking): For all these tests it would be better to generateProposal for each test. Then you are safer from mutating the proposal between tests and you also can remove the proposal and proposal.terms spread

apps/token/src/routes/governance/components/proposal-detail-header/proposal-header.tsx
```diff
@@ -0,0 +1,104 @@
...
+      detailsOne = (
+        <>
+          <Lozenge>{change.networkParameter.key}</Lozenge> {t('to')}{' '}
+          <Lozenge>{change.networkParameter.value}</Lozenge>
+        </>
+      );
+      break;
+    }
+    case 'NewFreeform': {
+      const description = proposal.rationale.description.trim();
```
mattrussell36, line 69: question (blocking): Why all this logic here? Might be missing something but is this just to truncate text? I'd prefer to keep it simple and just rely on `text-overflow: ellipsis` if we can

apps/token/src/routes/governance/components/proposal-detail-header/proposal-header.tsx
```diff
@@ -0,0 +1,104 @@
...
+export const ProposalHeader = ({
+  proposal,
+}: {
+  proposal: Proposals_proposals;
+}) => {
+  const { t } = useTranslation();
+  const { change } = proposal.terms;
+
+  let headerText: string;
+  let detailsOne: ReactNode;
```
mattrussell36, line 15: suggestion (non blocking): Perhaps just have a single `content` variable here set to null initially. This will save rendering out empty divs at the end as detailsTwo is only used for freeform proposals

apps/token/src/routes/governance/components/proposals-list/proposals-list.spec.tsx
```diff
@@ -0,0 +1,227 @@
...
+      openProposalClosesNextWeek,
+      openProposalClosesNextMonth,
+      enactedProposalClosedLastWeek,
+      rejectedProposalClosedLastMonth,
+    ])
+  );
+  const openProposals = screen.getByTestId('open-proposals');
+  const closedProposals = screen.getByTestId('closed-proposals');
+  expect(
+    openProposals.querySelectorAll("[data-testid='proposals-list-item']").length
```
mattrussell36, line 114: chore (blocking) 
```
const openProposals = within(screen.getByTestId('open-proposals'))
openProposals.getAllByTestId("proposals-list-item").length
```

mattrussell36: I think a few other areas of the tests could be written like this

mattrussell36 (changes requested):

apps/token/src/routes/governance/components/proposal-detail-header/proposal-header.tsx
sam-keen: Ah good spot, thanks. I think I'll conditionally render detailsTwo as having the content and elements split like that aids testing.

apps/token/src/routes/governance/components/proposals-list-item/proposals-list-item.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { ProposalsListItemDetails } from './proposals-list-item-details';
+import type { Proposals_proposals } from '../../proposals/__generated__/Proposals';
+
+interface ProposalsListItemProps {
+  proposal: Proposals_proposals;
+}
+
+export const ProposalsListItem = ({
+  proposal,
+  ...props
```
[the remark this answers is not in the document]
sam-keen: Good spot again. This was a relic from passing in individual data-testids. Thanks

apps/token/src/routes/governance/components/proposals-list/proposals-list.spec.tsx
sam-keen: Oh nice, thanks

apps/token/src/routes/governance/components/proposal-detail-header/proposal-header.spec.tsx
```diff
@@ -0,0 +1,238 @@
+import { render, screen } from '@testing-library/react';
+import { generateProposal } from '../../test-helpers/generate-proposals';
+import { ProposalHeader } from './proposal-header';
+import type { Proposals_proposals } from '../../proposals/__generated__/Proposals';
+
+const renderComponent = (proposal: Proposals_proposals) => (
+  <ProposalHeader proposal={proposal} />
+);
+
+it('New market', () => {
```
notbot00, line 10: chore (non-blocking): imho it's a lot nicer when tests are wrapped in a describe, noting what functionality / component are they testing, and the `it` field to be more descriptive of the expectation we're testing, something like:

```
describe('Proposal header', () => {
  it('renders new market', () => {...})
  it('updates market', () => {...})
  ...
})
```

mattrussell36 (approved):

apps/token/src/routes/governance/components/proposal-detail-header/proposal-header.spec.tsx
sam-keen: Yeah, good shout, will do.

apps/token/src/routes/governance/components/proposal-detail-header/proposal-header.tsx
```diff
@@ -0,0 +1,104 @@
...
+  let detailsTwo: ReactNode;
+
+  switch (change.__typename) {
```
notbot00, line 18: chore (non-blocking): i'd extract this to a function above the component, tbh, to make it a bit cleaner - you wouldn't need to attach types to variables which are undefined at the time of creation + giving an extra benefit to easily hook it into a memo function:

```
const getChildren = (change) => {
  switch (..) {
    case : return {
      headerText: '...',
      detailOne: <>...</>
      detailTwo: <>...</>
    }
  }
}

export const ProposalHeader = ({ proposal }) => {
  const { change } = proposal.terms;
  
  const { header, detailOne, detailTwo } = useMemo(() => getChildren(change), [change]);
  ...
```

apps/token/src/routes/governance/components/proposals-list-item/proposals-list-item-details.spec.tsx
```diff
@@ -0,0 +1,370 @@
...
+      <AppStateProvider>
+        <VegaWalletContext.Provider value={mockWalletContext}>
+          <ProposalsListItemDetails proposal={proposal} />
+        </VegaWalletContext.Provider>
+      </AppStateProvider>
+    </MockedProvider>
+  </Router>
+);
+
+it('Enacted', () => {
```
notbot00: chore (non-blocking): here, too, a describe wrapper would improve things, but feel free to ignore me, not super important

apps/token/src/routes/governance/components/proposals-list-item/proposals-list-item-details.tsx
```diff
@@ -0,0 +1,192 @@
...
+  const { willPass, majorityMet, participationMet } = useVoteInformation({
+    proposal,
+  });
+  const { t } = useTranslation();
+  const { voteState } = useUserVote(
+    proposal.id,
+    proposal.votes.yes.votes,
+    proposal.votes.no.votes
+  );
+  const MajorityNotReached = (
```
notbot00, line 32: chore (non-blocking): i'd move these two components outside of `ProposalsListItemDetails`, they don't need to be recreated in the scope every time the component rerenders, they can just be defined statically, like:

```
const MajorityNotReached = <>...</>;
const ParticipationNotReached = <>...</>;

const ProposalsListItemDetails () => {...}
```

apps/token/src/routes/governance/components/proposals-list/proposals-list.tsx
```diff
@@ -1,84 +1,73 @@
...
+      )}
+      <hr className="my-20 border-t-2" />
+      {sortedProposals.closed.length > 0 ? (
+        <ul data-testid="closed-proposals">
+          {sortedProposals.closed.filter(filterPredicate).map((proposal) => (
+            <ProposalsListItem key={proposal.id} proposal={proposal} />
+          ))}
+        </ul>
+      ) : (
+        <p data-testid="no-closed-proposals">{t('noClosedProposals')}</p>
```
notbot00, line 69: question (non-blocking): this looks weird at first glance, why are we printing camel case text to the UI? I'd expect some user readable message here, but I don't have much context on this.

notbot00 (approved):

apps/token/src/routes/governance/components/proposal-detail-header/proposal-header.spec.tsx
```diff
@@ -0,0 +1,277 @@
...
+    );
+    expect(screen.getByTestId('proposal-header')).toHaveTextContent(
+      'Network parameter'
+    );
+    expect(screen.getByTestId('proposal-details-one')).toHaveTextContent(
+      'Network key to Network value'
+    );
+  });
+
+  // Skipped until proposals have rationale
```
dexturr, line 140: nitpick (non-blocking): would be good to have a ticket linked here

apps/token/src/routes/governance/components/proposals-list-item/proposals-list-item-details.spec.tsx
```diff
@@ -0,0 +1,374 @@
...
+          state: ProposalState.Open,
+          terms: {
+            closingDatetime: fiveMinutes.toString(),
+          },
+        })
+      )
+    );
+    expect(screen.getByTestId('proposal-status')).toHaveTextContent('Open');
+    expect(screen.getByTestId('vote-details')).toHaveTextContent(
+      '5 minutes left to vote'
```
dexturr, line 117: chore (non-blocking): we may need to use mock timers, this test could potentially time out or be flakey is it takes a long time to render or similar. Let's keep an eye on it.

apps/token/src/routes/governance/test-helpers/mocks.ts
```diff
@@ -0,0 +1,57 @@
...
+};
+
+export const amendToDate = (amendmentInSeconds: number) => {
+  const now = new Date();
+  return new Date(now.getTime() + amendmentInSeconds * 1000);
+};
+
+const oneWeek = 60 * 60 * 24 * 7;
+
+export const lastMonth = amendToDate(-(oneWeek * 4));
```
dexturr, line 54: chore: not the biggest fan of this. I think I'd prefer to set the system time and use static data as this could be flakey. We do this elsewhere.

dexturr (commented):
Little worried about unit tests being flakey. Happy for it to be done as a follow up and not block this.

apps/token/src/routes/governance/components/proposals-list/proposals-list.tsx
mattrussell36: @notbot00 for context. The token app had a proper translations system set before it got migrated to the monorepo. The `t` function here is from react-i18nnext, whereas the `t` function in other apps (from react-helpers) simply returns the string that was passed in.

PR #737, opened 2022-07-11
Chore/UI tweaks

### LINKED ISSUE #665: Orderbook should be shown by default, instead of trades
The trading view opens with the trades table selected by default. The order book is much more useful and should be shown by default.
It is often done this way on other platforms.

PeteWilliams:
# Related issues 🔗

Closes #665

# Description ℹ️
I've been reviewing Console V2 from a designer's perspective and making some visual tweaks to improve things where I can.

# Demo 📺

## Tab background colour and spacing
Before:
image

After:
image

## Prevented tabs from moving when clicked
Previously, when you clicked on a tab, it went bold, which increased its width, and therefore its container, which moved the text around. I've set it to change the stroke width rather than the font weight, which looks almost the same, but doesn't move the panels.

Before:
before

After
after

## Adjusted panel sizes to give chart more room
And added a bit more space between panels.

Before:
image

After:
image

## Fixed alignment when accordion icon rotates
Before:
before1

After:
after1

## Made white theme header black
And made market name link pink.

Before:
image

After:
image

## Tweaked heading font sizes/weights
Before:
image 
image

After:
image
image

## Reordered panel tabs
Put positions first as more important, and be consistent with Portfolio section
Made orderbook default over Trades.
Renamed Accounts to Collateral

Before:
image
image

After:
image
image

## Adjusted dropdown button hover colour
Before: 
image

After:
image

## Improved colour contrast of some text
Before:
image

After:
image


apps/trading/pages/_app.page.tsx
```diff
@@ -28,7 +28,7 @@ function AppBody({ Component, pageProps }: AppProps) {
 
   return (
     <ThemeContext.Provider value={theme}>
-      <div className="h-full dark:bg-black dark:text-white-60 bg-white relative z-0 text-black-60 grid grid-rows-[min-content,1fr]">
+      <div className="h-full bg-black dark:text-white-60 bg-white relative z-0 text-black-60 grid grid-rows-[min-content,1fr]">
```
sam-keen, line 31: Chore (blocking): this now has both classes `bg-black` and `bg-white`. Either one of them needs removing entirely, or one of them is needs a `dark:` prefix. You might also need to look at text colour.

libs/ui-toolkit/src/components/tabs/tabs.tsx
```diff
@@ -18,36 +18,39 @@ export const Tabs = ({ children }: TabsProps) => {
...
+          <div className="gap-4 bg-white dark:bg-black flex border-r-4 border-r-white dark:border-r-black">
+            {Children.map(children, (child) => {
+              if (!isValidElement(child)) return null;
+              const isActive = child.props.id === activeTab;
+              const triggerClass = classNames(
+                'py-4 px-20',
+                'capitalize',
+                'focus-visible:outline-none focus-visible:shadow-inset-vega-pink dark:focus-visible:shadow-inset-vega-yellow',
+                {
+                  'text-vega-pink dark:text-vega-yellow [-webkit-text-stroke-width:1px]':
```
sam-keen, line 35: Chore (blocking): Hmm, webkit-text-stroke-width is non-standard and shouldn't be used in production. We should find another way to highlight the tab instead of using weights/stroke that'll work in all modern browsers. Suggestions: underline, bg colour

sam-keen (commented):
Just a couple of small things.

apps/trading/pages/_app.page.tsx
PeteWilliams: Thanks, have resolved that now.

libs/ui-toolkit/src/components/tabs/tabs.tsx
PeteWilliams: Thanks Sam. I've removed text-stroke and replaced it with another solution - I've added a pseudo-element to the inactive tab state which contains the same text in bold but is set to `visibility: hidden`. Then there's no change in width when it switches to the active state. This achieves the same result with standard CSS - hopefully it's OK!

MadalinaRaicu (approved):

mattrussell36 (approved):

mattrussell36 (approved):

dexturr (approved):

PR #738, opened 2022-07-11
Feat/776 Resizable panes in trading app

### LINKED ISSUE #776: Add resizable panes to Trading UI
Let's try using @johnwalley's Allotment library.

johnwalley:
# Related issues 🔗

Closes #776

# Description ℹ️

Modifies the Trading Grid page such that the sections can be resized.

# Demo 📺

https://user-images.githubusercontent.com/981531/179237218-5c20999b-ac1c-412c-b050-643b92067027.mp4

# Technical 👨‍🔧

I've not spent much effort on the styles. Specifically, I'm thinking about any padding/margins around components now that I've removed the CSS grid layout. Happy to take some guidance.


MadalinaRaicu (approved):
I have checked out this branch and played around with it, and I would be really happy to see this merged. 
Are there any known issues that prevent this from being merged (apart from waiting for review)? @mattrussell36 @edd

johnwalley:
> I have checked out this branch and played around with it, and I would be really happy to see this merged. Are there any known issues that prevent this from being merged (apart from waiting for review)? @mattrussell36 @edd

I've made a few tweaks and I'm happy for it to move to review and merge.

mattrussell36 (approved):

PeteWilliams:
Having had a quick play with this, I think it's great, though have noticed a couple issues.

One is with the bottom panel (positions/orders/collateral). You don't seem to be able to resize it below the default value other than to hide it completely.  Once hidden it's quite fiddly to bring it back. Not impossible, but I can see some users getting stuck.

I'd suggest allowing the user to resize freely below the default height. In terms of bringing it back from being closed completely, a nice solution would be to still show the tabs at the bottom but no panel content, and then have it pop open again if the user clicks the tab. I'm presuming that might be quite a bit of work, so maybe a easier fix would be just to disable the ability to hide it completely, by setting a minimum height or something.

Also, when you resize your chart, the Y-axis label overlaps the high/low data at the top. Might be better if the high/low bar wraps when it's not got room.
image


One other point, is that we don't currently have any local saving of state (eg, what market you'd selected, what tabs you had selected) which is a bit inconvenient, but potentially becomes annoying if we add customisability like this which you have to re-do every time you access the platform.

AndyWhiteVega:
Tested this : following the video for guidance, all together a pleasant experience where i as the user feel in full control of the sizing, very intuitive, some slight ugly wrapping with the middle pane field values when at minimum, but thats to be expected.

Issue #644, opened 2022-06-24
Logging


mattrussell36:
Add a logging module to decouple Sentry from the app and to improve debugging experience generally

- [x] Add logging module
- [x] Replace Sentry.captureExceptions with equivalent
- [x] Replace Sentry.captureMessage with equivalent
- [x] Replace Sentry.addBreadcrumb with equivalent
- [x] Add "report critical" function which proxies to sentry.captureException with a critical label populated as true. This will automatically create an incident if seen so should be used sparingly.

Implementation currently used on desktop wallet app here: https://github.com/vegaprotocol/vegawallet-desktop/blob/develop/frontend/src/lib/logging.ts
We also have another implementation from console v1 here: https://github.com/vegaprotocol/slimline-console/blob/master/src/lib/logging.ts

dexturr:
Related to #2506

PR #742, opened 2022-07-12
Feat/470 edit order

MadalinaRaicu:
# Related issues 🔗

Closes #[Issue number here]

# Description ℹ️

# Demo 📺

# Technical 👨‍🔧


apps/trading/pages/index.page.tsx
```diff
@@ -20,15 +20,18 @@ const MARKETS_QUERY = gql`
   }
 `;
 
-const marketList = ({ markets }: MarketsLanding) =>
-  orderBy(
+const marketList = ({ markets }: MarketsLanding) => {
+  const filteredMarkets =
```
mattrussell36, line 24: suggestion (non blocking): I quite like using lodash's .chain or .flow methods to make things a little more readable.

libs/orders/src/lib/order-hooks/use-order-edit.tsx
```diff
@@ -0,0 +1,126 @@
...
+            orderId: order.id,
+            marketId: order?.market?.id,
+            price: {
+              value: removeDecimal(
+                order.price,
+                order.market.decimalPlaces ?? 0
+              ),
+            },
+            timeInForce:
+              `TIME_IN_FORCE_${order.timeInForce}` as VegaWalletOrderTimeInForce,
```
mattrussell36, line 63: suggestion (non blocking): You might be able to use `VegaWalletTimeInForce[order.timeInForce]` which I think its a bit cleaner

libs/wallet/src/vega-transaction-dialog/vega-transaction-dialog.tsx
```diff
@@ -17,6 +19,8 @@ export interface VegaTransactionDialogProps {
   transaction: VegaTxState;
   reset: () => void;
   title?: string;
+  edit?: (body?: unknown) => Promise<unknown>;
```
mattrussell36, line 22: question (blocking): Why is the body unknown here?

libs/wallet/src/vega-transaction-dialog/vega-transaction-dialog.tsx
```diff
@@ -84,15 +92,119 @@ interface VegaDialogProps {
...
+}: VegaOrderEditDialogProps) => {
+  const headerClassName = 'text-h5 font-bold text-black dark:text-white';
+  const {
+    register,
+    formState: { errors },
+    handleSubmit,
+  } = useForm<FormFields>({
+    defaultValues: {
+      entryPrice: order?.price
+        ? addDecimalsFormatNumber(
```
mattrussell36, line 122: question (blocking): I'm not sure about formatting the number with commas here as we don't account for them when removing decimals for sending to the wallet. I think just addDecimals will be safer

libs/orders/src/lib/utils/get-default-order.ts
```diff
@@ -5,28 +5,25 @@ import {
...
-    }
-  | {
-      size: string;
-      type: VegaWalletOrderType.Limit;
-      timeInForce: VegaWalletOrderTimeInForce;
-      side: VegaWalletOrderSide;
-      price?: string;
-      expiration?: Date;
-    };
+export type Order = {
```
mattrussell36, line 10: question (blocking): Why combine into one type here? I think it reduces type safety between limit and market orders. They were originally separated so that we'd get field discrimination (eg price) between limit and market orders.

mattrussell36 (changes requested):

libs/orders/src/lib/order-hooks/use-order-edit.tsx
```diff
@@ -0,0 +1,126 @@
...
+            sizeDelta: 0,
+            expiresAt: order.expiresAt
+              ? {
+                  value:
+                    // Wallet expects timestamp in nanoseconds,
+                    // we don't have that level of accuracy so just append 6 zeroes
+                    new Date(order.expiresAt).getTime().toString() + '000000',
+                }
+              : undefined,
+          } as unknown as OrderAmendmentBodyOrderAmendment,
```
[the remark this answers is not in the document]
mattrussell36: Possibly either this object is the wrong shape or what was defined in the vegwallet-service-api-client is incorrect?

libs/orders/src/lib/order-hooks/use-order-edit.tsx
```diff
@@ -0,0 +1,126 @@
...
+                ) {
+                  setUpdatedOrder(matchingOrderEvent.event);
+                  subRef.current?.unsubscribe();
+                }
+              });
+          }
+        }
+        return res;
+      } catch (e) {
+        Sentry.captureException(e);
```
notbot00, line 113: question (non-blocking): not sure what happens if the app doesn't have a Sentry handler and tried using this hook. for me it feels like this should be handled on the app level, and just pass in an onQueryError attribute to the hook.

mattrussell36: So I think implementing https://github.com/vegaprotocol/frontend-monorepo/issues/644 should handle this

libs/orders/src/lib/order-hooks/use-order-edit.tsx
```diff
@@ -0,0 +1,126 @@
...
+                  if (e.event.__typename !== 'Order') {
+                    return false;
+                  }
+
+                  return e.event.id === resId;
+                });
+
+                if (
+                  matchingOrderEvent &&
+                  matchingOrderEvent.event.__typename === 'Order'
```
notbot00, line 103: question (non-blocking): not sure if I'm reading this right, but this doesn't look like it can ever be false, as the matchingOrderEvent above is explicitly looking only for events with type name `!== 'Order'`

libs/orders/src/lib/order-hooks/use-order-edit.tsx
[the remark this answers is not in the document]
MadalinaRaicu: so `price` and `expiresAt` has this format `{ value: string; }`. updated https://github.com/vegaprotocol/vegawallet-service-api-client/pull/4/files

libs/orders/src/lib/components/order-list/order-list.tsx
```diff
@@ -39,15 +51,40 @@ export const OrderList = forwardRef<AgGridReact, OrderListProps>(
...
+          transaction={editTransaction}
+          reset={resetEdit}
+          title={
+            editedOrder
+              ? `Order ${
+                  editOrder?.market?.tradableInstrument.instrument.code ?? ''
+                } updated`
+              : `Edit ${
+                  editOrder?.market?.tradableInstrument.instrument.code ?? ''
+                } order`
```
dexturr, line 83: Translations?

libs/orders/src/lib/order-hooks/use-order-edit.tsx
```diff
@@ -0,0 +1,126 @@
...
+                    new Date(order.expiresAt).getTime().toString() + '000000',
+                }
+              : undefined,
+          } as unknown as OrderAmendmentBodyOrderAmendment,
+        } as OrderAmendmentBody);
+
+        if (res?.signature) {
+          const resId = order.id ?? determineId(res.signature);
+          setUpdatedOrder(null);
+          // setId(resId);
```
dexturr, line 79: Commented code?

```suggestion
```

libs/orders/src/lib/order-hooks/use-order-edit.tsx
```diff
@@ -0,0 +1,126 @@
...
+        const res = await send({
+          pubKey: keypair.pub,
+          propagate: true,
+          orderAmendment: {
+            orderId: order.id,
+            marketId: order?.market?.id,
+            price: {
+              value: removeDecimal(
+                order.price,
+                order.market.decimalPlaces ?? 0
```
dexturr, line 59: question (blocking): why is the ?? required? I am interested in the situation where this is null. Could a user perchance place an order at the wrong level accidentally?

libs/orders/src/lib/utils/get-default-order.ts
MadalinaRaicu: oh, I see. updated.

apps/trading/pages/index.page.tsx
```diff
@@ -20,15 +20,18 @@ const MARKETS_QUERY = gql`
   }
 `;
 
-const marketList = ({ markets }: MarketsLanding) =>
-  orderBy(
+const marketList = ({ markets }: MarketsLanding) => {
```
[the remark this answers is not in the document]
MadalinaRaicu: updated

libs/wallet/src/vega-transaction-dialog/vega-transaction-dialog.tsx
```diff
@@ -45,6 +49,8 @@ export const VegaTransactionDialog = ({
   transaction,
   reset,
   title = '',
+  edit,
```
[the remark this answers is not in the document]
MadalinaRaicu: agree, updated

libs/wallet/src/vega-transaction-dialog/vega-transaction-dialog.tsx
MadalinaRaicu: updated to be Order type

mattrussell36 (approved):

mattrussell36 (approved):

notbot00 (approved):

libs/orders/src/lib/order-hooks/use-order-edit.tsx
MadalinaRaicu: Actually, this extra check helps with checking out the bus event type. If I we remove it you will get this warning
```
Argument of type 'OrderEvent_busEvents_event' is not assignable to parameter of type 'SetStateAction<OrderEvent_busEvents_event_Order | null>'.
  Type 'OrderEvent_busEvents_event_TimeUpdate' is not assignable to type 'SetStateAction<OrderEvent_busEvents_event_Order | null>'.
    Type 'OrderEvent_busEvents_event_TimeUpdate' is missing the following properties from type 'OrderEvent_busEvents_event_Order': type, id, status, rejectionReason, and 6 more.
```

PR #750, opened 2022-07-13
test: increase resolution in config files

jtsang586:
Increase resolution for Cypress tests


jtsang586:
Saw some test failures and ended up fixing some tests along the way

radoslaw-szpiech (approved):

PR #754, opened 2022-07-14
Fix/withdrawals incentive

mattrussell36:
# Related issues 🔗

N/A

# Description ℹ️

- Adds a link to the withdraw page
- Handles withdrawal limits in client side validation and shows limits in the UI
- Fixes deposit approval amount
- Handles extremely high deposit/withdraw limits. This is a bit basic but will do for now. If larger than 1m we show `1m+`
- Adds connect vega wallet buttons to windows where required

# Technical 👨‍🔧

Deposit approval amount wasn't being passed as the smallest denomination resulting in approval limits of 0.009 or thereabouts.

### DIFF
--- a/libs/withdraws/src/lib/withdraw-form.spec.tsx
+++ b/libs/withdraws/src/lib/withdraw-form.spec.tsx
@@ -20,6 +20,9 @@ beforeEach(() => {
     assets,
     min: new BigNumber(0.00001),
     max: new BigNumber(100),
+    limits: {
+      max: new BigNumber(200),
+    },
     ethereumAccount: undefined,
     selectedAsset: undefined,
     onSelectAsset: jest.fn(),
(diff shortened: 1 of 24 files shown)


libs/withdraws/src/lib/withdraw-form.spec.tsx
```diff
@@ -20,6 +20,9 @@ beforeEach(() => {
     assets,
```
dexturr, line 20: chore (non-blocking): would be nice for a test with the limits in it here

edd (approved):
Sat with you for most of this - quick check, looks fine. There’s an unused var warning but not worth holding it up for

dexturr (approved):

MadalinaRaicu (approved):
most probably these cypress tests fail because of the testnet issues.

libs/withdraws/src/lib/withdraw-form.spec.tsx
mattrussell36: Definitely. Will make another PR for this to unblock others

Issue #656, opened 2022-06-28
As a trader I want to place an order even when a market is suspended (in auction) due to liquidity or price monitoring


campbellssource:
On Vega, a Market has a Normal trading mode: e.g. continuous, where buys and sells that cross will match and trade instantly and continuously.
But a Market can also go into auction: where Buys and sells will only cross when the auction end conditions are met.

There are at least 3 reasons why a continuous market will go into auction:
- because it is new and has not yet has sufficient criteria to open
- because the market has a movement in price that hit one of the price monitoring bounds (big price moves are dangerous) 
- because the market has insufficient liquidity to trade safely. There needs to be some more volume on the book there there is open volume

The [market lifecycle spec](https://github.com/vegaprotocol/specs-internal/blob/master/protocol/0043-MKTL-market_lifecycle.md) explains how the `market.status` is used, and that the status `Suspended` is used to show that we are in some sort of "Exception auction". either Price or Liquidity monitoring. or an opening auction.

You can look at some of the following information to see more
```GraphQL
query{
  markets{
    name
    state
    tradingMode
    data{
      marketTradingMode
    }
  }
}
```
Returns
```GraphQL
{
      {
        "name": "All cool",
        "state": "Active",
        "tradingMode": "Continuous",
        "data": {
          "marketTradingMode": "Continuous",
          "trigger": "unspecified"
        }
      },
      {
        "name": "Need more orders",
        "state": "Pending",
        "tradingMode": "Continuous",
        "data": {
          "marketTradingMode": "OpeningAuction",
          "trigger": "Opening"
        }
      },
      {
        "name": "needs more liquidity",
        "state": "Suspended",
        "tradingMode": "Continuous",
        "data": {
          "marketTradingMode": "MonitoringAuction",
          "trigger": "Liquidity"
        }
      },
      {
        "name": "Price moved to quick",
        "state": "Suspended",
        "tradingMode": "Continuous",
        "data": {
          "marketTradingMode": "MonitoringAuction",
          "trigger": "Price"
        }
      }
}
```

note there is another type of Normal trading mode called "Batch Auction". This is out of scope for this ticket.

Generally the UX requirements for when market are in auctions are
- It should be obvious that the market is not in a normal mode, and have a way of learning more about what this is
  - Docs
  - What type of auction are we in (opening, liquidity, Price monitoring)
  - What exactly triggered it
  - What are the conditions for the auction ending
- Should still be able to place order
[...]

Issue #609, opened 2022-06-21
[trading] Displaying "trading mode continuous"


davidsiska-vega:
I don't think this "trading mode continuous" is a useful bit of information
<img width="563" alt="image" src="https://user-images.githubusercontent.com/62546419/174879187-f4f4a2ba-fb43-4af6-b41d-aca6cbe15404.png">

This is the default trading mode and for the foreseeable future there will only be one: "continuous trading". To me it's actively confusing as the market is often in auction and it still says "trading mode continous". I'd suggest we remove that field entirely.

PeteWilliams:
@davidsiska-vega The intention here was that this would show whether the market is in auction, or in standard mode.

It looks at though it's using `tradingMode` which, as you say, always shows 'continuous'. It looks like if we used `data.marketTradingMode`, it would return Continuous/OpeningAuction/MonitoringAuction.

If that's right, would you be happy with using this and having 'Standard/Auction' displayed here?
I'm assuming that 'standard' is more intuitive to a user than 'continuous' and that they don't need the granularity of opening/monitoring auction here - though that will still be available in the market data tab.

PR #741, opened 2022-07-11
Feat/718 integrate node validation

notbot00:
# Related issues 🔗

Closes #718 

# Description ℹ️

- adds custom node support to the node switcher dialog
- integrates the node switcher dialog with initial error messages to show when the default node selection does not succeed

# Demo 📺

Screenshot 2022-07-14 at 08 55 36
Screenshot 2022-07-14 at 08 56 15
Screenshot 2022-07-14 at 08 57 12
Screenshot 2022-07-14 at 08 58 50
Screenshot 2022-07-14 at 09 07 50
Screenshot 2022-07-14 at 09 07 26

MOBILE (based on https://github.com/vegaprotocol/frontend-monorepo/issues/622#issuecomment-1167226655):

Screenshot 2022-07-14 at 11 10 48

# Technical 👨‍🔧

- VALID NODE = 
    - it is a valid url (checked by initialising a `new URL`)
    - returns blockHeight and chainId (query statistics)
    - it's on the same network as the app (chainId does include VEGA_ENV from the environment)
    - does have ssl working (checked by one-time subscription to busEvents change)
    
- open the node switcher on startup:
    - there is no VEGA_URL specified in the environment AND:
         - cannot load the network config
         - network config is not valid
         - none of the nodes in the config is a VALID NODE
         - the selected node turns out not to have ssl support (detached ssl validation due to high latency on some networks)
    - there is a VEGA_URL specified in the environment AND:
        - it is not a VALID NODE
        - turns out not to have ssl support (detached ssl validation due to high latency on some networks)
- open the node switcher on user interaction:
    - when clicking the Edit link in the footer network info (token and explorer apps)
    
- when selecting a node:
    - from the default network nodes:
        - only the VALID NODES can be selected
    - a custom one:
        - has to be a VALID NODE
         
Screenshot 2022-07-14 at 08 29 08

PR #761, opened 2022-07-14
fix: filter rejected markets, order when suspended, remove market state from header, show continuous trading

### LINKED ISSUE #603: Filter out rejected market proposals form the list of markets
Image

This markets list contains markets that have been proposed but will never become markets. They should not appear here.
Proposed markets that have a chance of becoming real should go here (because users may wish to provide liquidity.

To determine if a market has a chance of becoming real we could look at the proposal status. In this case the cabbages market has a proposal state of rejected, so should not be returned in the markets list

```GraphQL
newMarketProposals{
    state
    id
    terms{
      change{
        ...on NewMarket{
          instrument{
            name
          }
        }
      }
    }
  }
...
markets{
  id
  state
  name
}
```
```GraphQL
{
  "state": "Rejected",
  "id": "a9ef0c09d88f7873eb0f86538dfaa0e45a4107f311b0b885b67d8bd4da49e289",
  "terms": {
    "change": {
      "instrument": {
        "name": "Cabbage"
      }
    }
  }
}
...
{
  "id": "a9ef0c09d88f7873eb0f86538dfaa0e45a4107f311b0b885b67d8bd4da49e289",
  "state": "Proposed",
  "name": "Cabbage"
},
```

MadalinaRaicu:
# Related issues 🔗

Closes #603 #656 #609

# Description ℹ️ 

- [x] filter rejected markets (data node returns also already rejected markets)
- [x] order when suspended
- [x] remove market state from header
- [x] show continuous trading
- [x] enable trading when market suspended

# Demo 📺

image

image

# Technical 👨‍🔧

- [x] change use order validation hook to display correct warnings and enable trading


libs/orders/src/lib/order-hooks/use-order-validation.tsx
```diff
@@ -28,33 +28,112 @@ export const useOrderValidation = ({
...
 
     if (keypair.tainted) {
-      return t('Selected public key has been tainted');
+      return {
+        isDisabled: true,
+        message: t('Selected public key has been tainted'),
+      };
+    }
+
+    if (
```
MadalinaRaicu, line 43: @campbellssource @PeteWilliams @jtsang586 

it would be great if I can have your review on this file.

apps/trading/pages/markets/trade-grid.tsx
```diff
@@ -87,20 +89,18 @@ export const TradeMarketHeader = ({
...
             <span className={itemClassName}>Trading mode</span>
             <span data-testid="trading-mode" className={itemValueClassName}>
-              {market.tradingMode}
-            </span>
-          </div>
-          <div className={headerItemClassName}>
-            <span className={itemClassName}>State</span>
-            <span data-testid="market-state" className={itemValueClassName}>
-              {market.state}
+              {startCase(
```
notbot00, line 99: chore (potentially-blocking): i think ideally we want to have a translation for this

libs/ui-toolkit/src/components/dialog/dialog.tsx
```diff
@@ -26,7 +26,7 @@ export function Dialog({
 }: DialogProps) {
   const contentClasses = classNames(
     // Positions the modal in the center of screen
-    'z-20 fixed w-full md:w-[520px] lg:w-[1000px] px-28 py-24 top-[50%] left-[50%] translate-x-[-50%] translate-y-[-50%]',
```
notbot00, line 29: question (blocking): i believe this was a requirement agreed by @sam-keen and @PeteWilliams that the modal sizes should be uniform and that particular size to accommodate the node switcher, because it looked pretty cramped under ~850px.
I've ended up changing this to 960px >[here](https://github.com/vegaprotocol/frontend-monorepo/pull/741/files#diff-1a450fe1bce39104bcd5b2c6b764f71945dbbd98f3f5b4bb7aae39138bda92acR29)< to prevent overflows, as it turns out our lg breakpoint os 960px.

apps/trading/pages/markets/trade-grid.tsx
```diff
@@ -87,20 +89,18 @@ export const TradeMarketHeader = ({
             <span className={itemClassName}>Volume</span>
             <span data-testid="trading-volume" className={itemValueClassName}>
               {market.data && market.data.indicativeVolume !== '0'
-                ? market.data.indicativeVolume
+                ? formatNumber(market.data.indicativeVolume)
```
mattrussell36: chore (blocking): Probably want to format with market.positionDecimal places here.

libs/orders/src/lib/order-hooks/use-order-validation.tsx
mattrussell36: Also @candida-d

libs/orders/src/lib/order-hooks/use-order-validation.spec.tsx
```diff
@@ -90,107 +90,146 @@ function setup(
...
+describe(`useOrderValidation`, () => {
+  it('Returns empty string when given valid data', () => {
+    const { result } = setup();
+    expect(result.current).toStrictEqual({ isDisabled: false, message: `` });
+  });
 
-it('Returns an error message when the keypair is tainted', async () => {
-  const { result } = setup(defaultOrder, {
-    keypair: { ...defaultWalletContext.keypair, tainted: true },
+  it('Returns an error message when no keypair found', async () => {
```
mattrussell36, line 99: question (blocking): Do all these tests need to be async. Nothing async is happening in the hook or the tests

libs/orders/src/lib/order-hooks/use-order-validation.tsx
```diff
@@ -28,33 +28,112 @@ export const useOrderValidation = ({
...
+      [
+        MarketState.Settled,
+        MarketState.Rejected,
+        MarketState.TradingTerminated,
+      ].includes(market.state)
+    ) {
+      return {
+        isDisabled: true,
+        message: t(
+          `This market is ${market.state.toLowerCase()} and no longer accepting orders`
```
mattrussell36, line 53: chore (non-blocking): we might want to add a map of market states to translations. Right now this could read "This market is tradingterminated and no longer accepting orders".

Perhaps a map like
```
const marketTranslations: { [S in MarketState]: string } = {
  [MarketState.TradingTerminated]: t('terminated'),
  [MarketState.Rejected]: t('rejected')
  ... // and so on
}
...
message: t(`This market is now ${marketTranslations[market.state]}' and is no longer accepting orders`)
```

mattrussell36: Worth checking the other resulting messages too.

mattrussell36 (changes requested):

apps/trading/pages/markets/trade-grid.tsx
MadalinaRaicu: just updated

libs/orders/src/lib/order-hooks/use-order-validation.spec.tsx
MadalinaRaicu: I did not change this, they were async before but they don't need to be. Updated.

libs/orders/src/lib/order-hooks/use-order-validation.tsx
MadalinaRaicu: updated

notbot00 (approved):

apps/trading/pages/markets/__generated__/Market.ts
```diff
@@ -144,6 +144,12 @@ export interface Market_market {
    * GBX (pence)        1              4       GBP   0.000001 (  0.0001p)
    */
   decimalPlaces: number;
+  /**
+   * positionDecimalPlaces indicated the number of decimal places that an integer must be shifted in order to get a correct size (uint64).
```
candida-d, line 148:
```suggestion
   * positionDecimalPlaces indicate the number of decimal places that an integer must be shifted in order to get a correct size (uint64).
```

libs/orders/src/lib/order-hooks/use-order-validation.spec.tsx
```diff
@@ -90,107 +91,148 @@ function setup(
...
+    ${MarketState.Settled}
+    ${MarketState.Rejected}
+    ${MarketState.TradingTerminated}
+  `(
+    'Returns an error message for market state when no longer accepting orders',
+    ({ state }) => {
+      const { result } = setup({ market: { ...defaultOrder.market, state } });
+      expect(result.current).toStrictEqual({
+        isDisabled: true,
+        message: `This market is ${marketTranslations(
```
candida-d, line 123: If I'm reading this correctly, then I believe it should say "This market is (state) and not accepting orders", because, for example, a market that is pending will not have been able to take orders yet, ever.

libs/orders/src/lib/order-hooks/use-order-validation.spec.tsx
```diff
@@ -90,107 +91,148 @@ function setup(
...
-    market: { ...market, positionDecimalPlaces: 0 },
-    fieldErrors: { size: { type: 'validate', message: ERROR_SIZE_DECIMAL } },
   });
-  expect(result.current).toEqual(ERROR.FIELD_PRICE_STEP_NULL);
-});
 
-it('Returns an error message when the order size has more decimals then allowed', async () => {
-  const { result } = setup({
-    fieldErrors: { size: { type: 'validate', message: ERROR_SIZE_DECIMAL } },
+  it('Returns an error message when the order size has more decimals then allowed', () => {
```
candida-d, line 229:
```suggestion
  it('Returns an error message when the order size has more decimals than allowed', () => {
```

libs/orders/src/lib/order-hooks/use-order-validation.tsx
```diff
@@ -28,33 +37,120 @@ export const useOrderValidation = ({
...
+      [
+        MarketState.Settled,
+        MarketState.Rejected,
+        MarketState.TradingTerminated,
+      ].includes(market.state)
+    ) {
+      return {
+        isDisabled: true,
+        message: t(
+          `This market is ${marketTranslations(
```
candida-d, line 62: I might have left this comment in the wrong place (sorry if I did as now it's in 2 places) but ... 

If I'm reading this correctly, then I believe it should say "This market is (state) and not accepting orders", because, for example, a market that is proposed will not have been able to take orders yet, ever.

libs/orders/src/lib/order-hooks/use-order-validation.tsx
```diff
@@ -28,33 +37,120 @@ export const useOrderValidation = ({
...
         market.state === MarketState.Proposed ||
         market.state === MarketState.Pending
       ) {
-        return t('Market is not active yet');
+        return {
+          isDisabled: false,
+          message: t(
+            `This market is ${marketTranslations(
+              market.state
+            )} and only accepting liquidity orders`
```
candida-d, line 135:
```suggestion
            )} and only accepting liquidity commitment orders`
```

libs/orders/src/lib/order-hooks/use-order-validation.tsx
```diff
@@ -28,33 +37,120 @@ export const useOrderValidation = ({
...
+            };
+          }
+        }
+
+        return {
+          isDisabled: false,
+          message: t(
+            `This market is ${marketTranslations(
+              market.state
+            )} and only accepting liquidity orders`
```
candida-d, line 121:
```suggestion
            )} and only accepting liquidity commitment orders`
```

candida-d (commented):
I added in some suggested changes, and some comments - all on the wording. :)

libs/orders/src/lib/order-hooks/use-order-validation.tsx
```diff
@@ -64,41 +160,62 @@ export const useOrderValidation = ({
...
 
     if (
       fieldErrors?.size?.type === 'validate' &&
       fieldErrors?.size?.message === ERROR_SIZE_DECIMAL
     ) {
       if (market.positionDecimalPlaces === 0) {
-        return t('No decimal amounts allowed for this order');
+        return {
+          isDisabled: true,
+          message: t('No decimal amounts allowed for this order'),
```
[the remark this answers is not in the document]
MadalinaRaicu: updated to the suggested wording

MadalinaRaicu:
> I added in some suggested changes, and some comments - all on the wording. :)

thanks for the review and for using the `suggest a change` feature!  made updating really easy

dexturr (approved):

mattrussell36 (approved):

libs/orders/src/lib/order-hooks/use-order-validation.tsx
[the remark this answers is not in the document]
PeteWilliams: @candida-d I think 'fractional' might be a bit confusing.
Would something like 'Order sizes must be whole numbers for this market' make more sense?
Think it's usually best to tell users what they need to do rather than what they can't do where possible, and might avoid confusing terms in this instance.

PeteWilliams:
Looking at @campbellssource's requirements on https://github.com/vegaprotocol/frontend-monorepo/issues/656 I think there's a couple things we're missing:

> if market.tradingmode = monitoring auction then also show market.data.trigger to show if it is a liquidity or price monitoring auction

That could look like this
image

Can we also change the capitalised cases to sentence case (ie, 'Monitoring Auction' becomes 'Monitoring auction')?

Can we also change 'Continuous Trading' here to just 'Continuous' as we don't need to duplicate the 'trading' that's in the header above?
image

> change the "State" column on the market list (both /markets and the switch market modal) so that it only shows market. market.tradingmode (and then show the trigger if monitoring auction) as above.

image

That could look like this:
image

Note, I've also changed the column heading there to 'Trading  mode' to match the market header.


On the warning below the ticket, the warning icon shrinks when it goes over more than one line, can we fix that please?
Also, the red bar should be the full height of the accompanying text block. It's currently a fixed height. 
image


I'm also not entirely sure what that warning is telling me - am I still able to place a limit order but not a market order? I did try reading the auction spec, but couldn't work this out. And if we're doing away with 'suspended' for the mode, maybe it should say 'in auction' here? @candida-d / @campbellssource 

If it's a warning rather than an error it should probably use the orange 'warning' intent rather than the red one.

candida-d:
> I'm also not entirely sure what that warning is telling me - am I still able to place a limit order but not a market order? I did try reading the auction spec, but couldn't work this out. And if we're doing away with 'suspended' for the mode, maybe it should say 'in auction' here? 

@petewilliams I took it as the market would be in a liquidity monitoring auction and thus only accepts liquidity commitments, or liquidity provision orders, which I believe currently can’t be placed via the UI. https://github.com/vegaprotocol/specs/blob/master/protocol/0038-OLIQ-liquidity_provision_order_type.md#liquidity-provisioning-order-features

libs/orders/src/lib/order-hooks/use-order-validation.tsx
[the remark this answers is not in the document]
candida-d: @PeteWilliams Yeah that makes sense, I agree. I’ll suggest a change.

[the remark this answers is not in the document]
candida-d: @PeteWilliams Yeah that makes sense, I agree. I’ll suggest a change.

libs/orders/src/lib/order-hooks/use-order-validation.tsx
```diff
@@ -64,41 +160,62 @@ export const useOrderValidation = ({
...
 
     if (
       fieldErrors?.size?.type === 'validate' &&
       fieldErrors?.size?.message === ERROR_SIZE_DECIMAL
     ) {
       if (market.positionDecimalPlaces === 0) {
-        return t('No decimal amounts allowed for this order');
+        return {
+          isDisabled: true,
+          message: t('No fractional amounts are allowed for this order'),
```
candida-d, line 207:
```suggestion
          message: t('Order sizes must be in whole numbers for this market'),
```

PR #770, opened 2022-07-14
fix: rename tradesPaged to tradesConnection

mattrussell36:
# Related issues 🔗

Closes N/A

# Description ℹ️

This fixes the fills table

# Demo 📺

N/A

# Technical 👨‍🔧

Query changed from `tradesPaged` to `tradesConnection`. But also some needed to re-add some fields which got removed but are required to calculate various fields in the table. Namely the fee fields/agressor and settlementAsset

### DIFF
--- a/libs/fills/src/lib/fills-data-provider.ts
+++ b/libs/fills/src/lib/fills-data-provider.ts
@@ -5,7 +5,7 @@ import type { PageInfo, Pagination } from '@vegaprotocol/react-helpers';
 import type { FillFields } from './__generated__/FillFields';
 import type {
   Fills,
-  Fills_party_tradesPaged_edges,
+  Fills_party_tradesConnection_edges,
 } from './__generated__/Fills';
 import type { FillsSub } from './__generated__/FillsSub';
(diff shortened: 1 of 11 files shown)


jtsang586 (approved):

libs/fills/src/lib/fills-data-provider.ts
```diff
@@ -17,19 +17,41 @@ const FILL_FRAGMENT = gql`
...
     market {
       id
+      name
       decimalPlaces
+      positionDecimalPlaces
       tradableInstrument {
         instrument {
           id
           code
+          product {
```
dexturr, line 46: question (blocking): why are these new fields needed? Is this conflicts?

mattrussell36: They are needed to format price values with the right decimal places and were accidentally removed. We loose some type safety when passing code through to AgGrid. @bglownia is sureing up the types in one of his PRs atm.

macqbat (approved):

PR #827, opened 2022-07-20
fix: failing simple app test

jtsang586:
# Description ℹ️

Fix for failing simple app test where Cypress was finding more than one element to click

### DIFF
--- a/apps/simple-trading-app-e2e/src/integration/market-selector.test.ts
+++ b/apps/simple-trading-app-e2e/src/integration/market-selector.test.ts
@@ -27,12 +27,13 @@ describe('market selector', () => {
       cy.getByTestId('market-pane')
         .children()
         .find('[role="button"]')
+        .first()
         .should('contain.text', markets[0].name);
       cy.getByTestId('market-pane')
+        .children()
+        .find('[role="button"]')
         .first()
-        .within(() => {
-          cy.get('[role="button"]').click();
-        });
+        .click();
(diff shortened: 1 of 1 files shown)


apps/simple-trading-app-e2e/src/integration/market-selector.test.ts
```diff
@@ -28,7 +28,11 @@ describe('market selector', () => {
         .children()
         .find('[role="button"]')
         .should('contain.text', markets[0].name);
-      cy.getByTestId('market-pane').children().find('[role="button"]').click();
+      cy.getByTestId('market-pane')
+        .first()
```
macqbat, line 32: there should be only one `market-pane` there. Giving method `first()` in chain after `cy.getByTestId('market-pane')` seems to me something is not correct there...

notbot00 (approved):

apps/simple-trading-app-e2e/src/integration/market-selector.test.ts
macqbat: Maybe there is a case that there are two markets with exactly the same market name. So would it be better then:
```
cy.getByTestId('market-pane').children().find('[role="button"]').first().click();
```
?

jtsang586: Fair point, I've changed it to include .first() after finding the button

PR #851, opened 2022-07-25
Task/eth tx dialog in hook

mattrussell36:
# Related issues 🔗

N/A

# Description ℹ️

# Demo 📺

N/A

# Technical 👨‍🔧

TBD

### DIFF
--- a/libs/deposits/src/lib/use-submit-deposit.tsx
+++ b/libs/deposits/src/lib/use-submit-deposit.tsx
@@ -2,7 +2,6 @@ import { gql, useSubscription } from '@apollo/client';
 import type {
   DepositEvent,
   DepositEventVariables,
-  DepositEvent_busEvents_event_Deposit,
 } from './__generated__/DepositEvent';
 import { DepositStatus } from '@vegaprotocol/types';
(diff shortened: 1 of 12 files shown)


libs/deposits/src/lib/use-submit-deposit.tsx
```diff
@@ -72,19 +71,17 @@ export const useSubmitDeposit = () => {
       });
 
       if (matchingDeposit && matchingDeposit.event.__typename === 'Deposit') {
```
dexturr, line 73: nitpick (non-blocking): any way we can get rid of this magic string

dexturr (approved):

edd (approved):
👍 Reviewed in person, so all comments were dealt with directly. I like this pattern.

libs/deposits/src/lib/use-submit-deposit.tsx
mattrussell36: Not without a ts-ignore. We only get Deposit events via the subscription but because the event field is a union type of all possible bus event types this is needed to appease TS

PR #852, opened 2022-07-25
Test/622 node switcher tests

### LINKED ISSUE #622: Node switcher

jtsang586:
# Related issues 🔗

Closes #622

# Description ℹ️

Cypress tests for node switcher on Block Explorer

### DIFF
--- a/apps/explorer-e2e/src/integration/node-switcher.cy.js
+++ b/apps/explorer-e2e/src/integration/node-switcher.cy.js
@@ -75,7 +75,9 @@ context('Node switcher', function () {
     });
 
     function validateNodeError(errortype, errorMsg) {
-      cy.getByTestId(nodeErrorType).should('have.text', errortype);
+      cy.getByTestId(nodeErrorType).should('have.text', errortype, {
+        timeout: 1000,
+      });
       cy.getByTestId(nodeErrorMsg).should('have.text', errorMsg);
     }
   });
--- a/apps/explorer-e2e/src/integration/node-switcher.cy.js
+++ b/apps/explorer-e2e/src/integration/node-switcher.cy.js
@@ -1,24 +1,28 @@
 const nodeErrorType = 'node-error-type';
 const nodeErrorMsg = 'node-error-message';
-const nodeId = 'node-url-1';
-const customNodeBtn = 'node-url-custom';
+const nodeId = 'node-url-0';
+const customNodeBtn = 'custom-node';
 
 context('Node switcher', function () {
   before('visit home page', function () {
+    cy.intercept('GET', 'https://static.vega.xyz/assets/capsule-network.json', {
+      hosts: ['http://localhost:3028/query'],
+    }).as('nodeData');
     cy.visit('/');
+    cy.wait('@nodeData');
   });
 
   describe('form validations and responses', function () {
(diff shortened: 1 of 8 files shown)


apps/explorer-e2e/src/integration/node-switcher.cy.js
```diff
@@ -0,0 +1,82 @@
...
+  before('visit home page', function () {
+    cy.intercept('GET', 'https://static.vega.xyz/assets/capsule-network.json', {
+      hosts: ['http://localhost:3028/query'],
+    }).as('nodeData');
+    cy.visit('/');
+    cy.wait('@nodeData');
+  });
+
+  describe('form validations and responses', function () {
+    this.beforeEach('open node selector', function () {
```
mattrussell36: question (blocking): Why use `this` keyword here?

apps/explorer-e2e/src/integration/node-switcher.cy.js
```diff
@@ -0,0 +1,82 @@
...
+      cy.getByTestId('git-network-data').within(() => {
+        cy.getByTestId('link').click({ force: true });
```
mattrussell36, line 18: question (blocking): Why are you using force here? We should avoid it

apps/explorer-e2e/src/integration/node-switcher.cy.js
```diff
@@ -0,0 +1,82 @@
...
+
+      cy.getByTestId('node-row')
+        .eq(0)
+        .within(() => {
+          cy.getByTestId(nodeId)
+            .should('exist')
+            .and('have.attr', 'aria-checked', 'true');
+          cy.get('label').should('have.text', Cypress.env('networkQueryUrl'));
+          cy.getByTestId('response-time-cell')
+            .find('.font-mono')
```
mattrussell36, line 32: suggestion (non blocking): Do you need the font-mono selector? Seems a bit brittle to rely on that being present. Same for the other assertions

apps/explorer-e2e/src/integration/node-switcher.cy.js
jtsang586: That's a mistake. Removed `this`

apps/explorer-e2e/src/integration/node-switcher.cy.js
jtsang586: force was needed because the modal wasn't closing after each test. I've now removed the `force:true` and added an afterEach step to close the modal.

apps/explorer-e2e/src/integration/node-switcher.cy.js
jtsang586: Agreed. I've changed the data-testid so I don't need to find the font-mono selector anymore

mattrussell36 (approved):

radoslaw-szpiech (approved):

PR #871, opened 2022-07-26
Feat/633: Oracles MVP in block explorer 

### LINKED ISSUE #633: Oracles in block explorer
Oracles are a key concept in Vega they deserve an item in the left hand nav.

I think the page should facilitate a few use cases:
- As a user I want to follow a link from a market description to a specific oracle spec to find out more about the oracle
- ...I want to see the definition of the oracle spec
- I want to see all `Submit oracle` transactions and their contents (regardless of matched spec)
- I want see bindings to events in markets (I imagine this is an "internal transaction"  
- I want to see what oracles are already used (and perhaps can be re-used)

Suggest the page appears after markets in the left hand Nav

Image

# MVP
A single page that lists the results of a single query as raw JSON
```GraphQL
query{
	oracleSpecs{
    status
    id
    createdAt
    updatedAt
    pubKeys
    filters{
      key{
        name
        type
      }
      conditions{
        value
        operator
      }
    }
    data{
      pubKeys
      matchedSpecIds
      broadcastAt
    }
  }
}
```
Warning this returns a GQL error if there is no data on any specs
Issues: this only shows oracle events that did match specs, hence MVP.

# MVP +1
2 pages, the second one to list results of 
```GraphQL
query{
	oracleData{
    pubKeys
    data{
      name
      value
    }
    matchedSpecIds
    broadcastAt
  }
}
```

sam-keen:
# Related issues 🔗

Closes #633

# Description ℹ️

A simple MVP JSON blob of each oracle's specs and data on a new /oracles route. 

# Demo 📺

Screenshot 2022-07-28 at 15 21 43


apps/explorer/src/app/routes/oracles/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+      }
+      data {
+        pubKeys
+      }
+    }
+  }
+`;
+
+const Oracles = () => {
+  const { data } = useQuery<OracleSpecsQuery>(ORACLE_SPECS_QUERY);
```
mattrussell36, line 36: question (non blocking): I notice we aren't error handling here or for other queries in explorer?

mattrussell36 (approved):

apps/explorer/src/app/routes/oracles/index.tsx
sam-keen: Yep, something we need to address!

dexturr (approved):

PR #888, opened 2022-07-27
Chore/Add resizable panels to portfolio page

PeteWilliams:
# Description ℹ️

We had resizeable panels on the trading page but not the portfolio page. Now we have both.
Also made some minor tweaks to trading page padding, and prevented long text on the accordion from switching to centered text. 

# Demo 📺

panels

Accordion before:
image

Accordion after:
image


MadalinaRaicu (approved):

dexturr (approved):

apps/trading/pages/portfolio/index.page.tsx
```diff
@@ -10,76 +10,88 @@ import classNames from 'classnames';
 import type { ReactNode } from 'react';
 import { VegaWalletContainer } from '../../components/vega-wallet-container';
 import { DepositsContainer } from './deposits-container';
+import { Allotment, LayoutPriority } from 'allotment';
+import 'allotment/dist/style.css';
```
mattrussell36, line 14: chore (blocking): I think this might include the styles twice in the css bundle. Can you create a an allotment component in ui-toolkit and import it there once?

mattrussell36 (changes requested):

apps/trading/pages/portfolio/index.page.tsx
PeteWilliams: Have now added a component.

mattrussell36 (approved):

PR #916, opened 2022-07-29
Task/Remove vegawallet service api client

### LINKED ISSUE #861: Bin vegawallet-service-api-client
- Remove dependency on @vegaprotocol/vegawallet-service-api-client
- Expand rest-connector to provide replace required functionality

mattrussell36:
# Related issues 🔗

Closes #861 

# Description ℹ️

Replaces the generated client code with basic fetch calls in the rest connector. Generated client code was just causing additional work for something that is relatively simple. Additionally incoming wallet changes need us to be more flexible in the short term.

# Demo 📺

N/A

# Technical 👨‍🔧

- Adds custom types for wallet for transactions
- Adds zod validation for rest-connector

### DIFF
--- a/libs/ui-toolkit/src/components/theme-switcher/theme-switcher.tsx
+++ b/libs/ui-toolkit/src/components/theme-switcher/theme-switcher.tsx
@@ -13,7 +13,7 @@ export const ThemeSwitcher = ({
   sunClassName?: string;
   moonClassName?: string;
 }) => {
-  const sunClasses = classNames('dark:hidden text-black', sunClassName);
+  const sunClasses = classNames('dark:hidden text-white', sunClassName);
   const moonClasses = classNames(
     'hidden dark:inline text-white',
     moonClassName
--- a/libs/ui-toolkit/src/components/theme-switcher/theme-switcher.tsx
+++ b/libs/ui-toolkit/src/components/theme-switcher/theme-switcher.tsx
@@ -13,7 +13,7 @@ export const ThemeSwitcher = ({
   sunClassName?: string;
   moonClassName?: string;
 }) => {
-  const sunClasses = classNames('dark:hidden text-white', sunClassName);
+  const sunClasses = classNames('dark:hidden text-black', sunClassName);
   const moonClasses = classNames(
     'hidden dark:inline text-white',
     moonClassName
--- a/libs/wallet/src/connectors/rest-connector.ts
+++ b/libs/wallet/src/connectors/rest-connector.ts
@@ -191,7 +191,13 @@ export class RestConnector implements VegaConnector {
   }
 
   private async request(endpoint: Endpoint, options: RequestInit) {
-    const fetchResult = await fetch(`${this.url}/${endpoint}`, options);
+    const fetchResult = await fetch(`${this.url}/${endpoint}`, {
+      ...options,
+      headers: {
+        ...options.headers,
+        'Content-Type': 'application/json',
+      },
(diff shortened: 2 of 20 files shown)


libs/ui-toolkit/src/components/theme-switcher/theme-switcher.tsx
```diff
@@ -13,7 +13,7 @@ export const ThemeSwitcher = ({
   sunClassName?: string;
   moonClassName?: string;
 }) => {
-  const sunClasses = classNames('dark:hidden text-black', sunClassName);
+  const sunClasses = classNames('dark:hidden text-white', sunClassName);
```
sam-keen, line 16: Chore (blocking): This change would mean the icon is invisible on sites with white backgrounds in light mode. I think we need a prop to specify if the icon's always on a dark background.

sam-keen (changes requested):

libs/ui-toolkit/src/components/theme-switcher/theme-switcher.tsx
mattrussell36: My bad not sure why I changed this its unrelated. reverting

sam-keen (approved):

libs/wallet/src/connectors/rest-connector.ts
```diff
@@ -146,18 +189,29 @@ export class RestConnector implements VegaConnector {
...
+        status: fetchResult.status,
+        data: textResult,
+      };
+    } else {
+      const jsonResult = await fetchResult.json();
+      return {
+        status: fetchResult.status,
+        data: jsonResult,
+      };
+    }
```
dexturr, line 215: probably want this all in a catch just in case it sends back html or something

libs/wallet/src/connectors/rest-connector.ts
```diff
@@ -48,26 +85,22 @@ export class RestConnector implements VegaConnector {
...
     try {
-      const service = new DefaultApi(
-        createConfiguration({
-          baseServer: new ServerConfiguration<Record<string, never>>(url, {}),
-        })
-      );
+      this.url = url;
 
-      const res = await service.authTokenPost(params);
+      const res = await this.request('auth/token', {
```
dexturr, line 90: I dislike the number of magic strings for endpoints in here. could we store these in a map?

dexturr (approved):

libs/wallet/src/connectors/rest-connector.ts
mattrussell36: Changed to enum

libs/withdraws/src/lib/withdraw-dialog.tsx
```diff
@@ -69,6 +69,18 @@ const getProps = (
   ethTx: EthTxState,
   ethUrl: string
 ) => {
+  const renderVegaTxError = () => {
+    if (vegaTx.error) {
+      if ('errors' in vegaTx.error) {
+        return vegaTx.error.errors['*'].map((e) => <p>{e}</p>);
```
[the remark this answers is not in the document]
mattrussell36: Yeah this is just what the service returns. Will add a comment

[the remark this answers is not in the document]
mattrussell36: I've ended up ensuring a simple error string is returned by the rest connector, so we don't need this logic in any react components

PR #931, opened 2022-08-02
fix: disallow updating custom node before validation

### LINKED ISSUE #856: Node switcher issues
I'm creating a ticket here with the list of issues/bugs that I've seen with the node switcher.

1.  Link to switch node should be disabled if there is no config url
2. After raising an error i.e. switching to incorrect node. The error stays in the modal after closing and opening it again
3. Input fake url and click check twice freezes the modal with the user unable to do anything else
4. Issue with nodes in `Mainnet` not connecting
5. Version/commit hash has un-clickable link which is underlined?
6. Should the switcher `check` the connection before `connecting`? I seem to be able to click connecting regardless

notbot00:
# Related issues 🔗

Closes #856 

# Description ℹ️

Disallow clicking the "Connect" button before validating a custom node. 

# Demo 📺

Screenshot 2022-08-02 at 11 03 05


mattrussell36 (approved):

libs/environment/src/utils/validate-node.tsx
```diff
@@ -40,24 +40,15 @@ export const getIsNodeDisabled = (env: Networks, data?: NodeData) => {
...
   currentNode: string | undefined,
-  inputText: string,
   env: Networks,
   state: Record<string, NodeData>
 ) => {
   if (!currentNode) {
     return true;
   }
 
-  if (
```
dexturr, line 51: question: should we add/remove/alter tests for this behaviour change? Or is this currently not united tested?

notbot00: the react component itself is not tested, that's where this is being called. most cases regarding functionality are covered in the hook tests, but it might make sense to do some ui tests for the react component, too.

dexturr (approved):

libs/environment/src/utils/validate-node.tsx
notbot00: added a bunch of tests for the node switcher component @dexturr

PR #935, opened 2022-08-02
fix: walletconnect does not work

### LINKED ISSUE #919: WalletConnect does not work. Link non-responsive on TFE
Image

The "WalletConnect" link within the connect eth wallet component is non-responsive for me. Firefox.
True on token.fariground.wtf
and
the mainnet deploy preview https://token-monorepo.netlify.app/

asiaznik:
# Related issues 🔗

Closes #919

# Technical 👨‍🔧

Bug was caused by a `global` and not `globalThis` reference in `@walletconnect/socket-transport/dist/esm/index.js`. That's why there's a dirty `global = globalThis` added to the `polyfills.js`

Fixing `global` issue revealed another with missing `Buffer` - *webpack@5+* doesn't include node's polyfills anymore hence I needed to add it manually to the `polyfills.ts`.

# Demo 📺

image

### DIFF
--- a/apps/token/src/assets/env-config.js
+++ b/apps/token/src/assets/env-config.js
@@ -1,3 +1,1 @@
 window._env_ = {};
-// eslint-disable-next-line no-undef, no-global-assign, no-native-reassign
-global = globalThis;
--- a/apps/token/src/assets/env-config.js
+++ b/apps/token/src/assets/env-config.js
@@ -1,1 +1,3 @@
 window._env_ = {};
+// eslint-disable-next-line no-undef, no-global-assign, no-native-reassign
+global = globalThis;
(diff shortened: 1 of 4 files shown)


apps/token/src/assets/env-config.js
```diff
@@ -1 +1,3 @@
 window._env_ = {};
+// eslint-disable-next-line no-undef, no-global-assign, no-native-reassign
+global = globalThis;
```
dexturr, line 3: chore (blocking): this solution won't work as this file is overrided when running in docker. Can we put this else where please?

dexturr (changes requested):
small issue

dexturr (approved):

mattrussell36 (approved):

AndyWhiteVega:
Tested: Working fine

Issue #884, opened 2022-07-27
Token-e2e: Adjust Staking flow tests to new implementation of validators table


radoslaw-szpiech:
Validators table implementation is changed that highly impacted e2e tests. We need to change the way how we interact with validators table on Staking page. Unfortunately, we cannot put `data-testid` inside the table. One of the possible solution is to add some distinguishable names to validators in vegcapsule - https://github.com/vegaprotocol/vegacapsule/blob/82cb9120245fe807487372fb638c25b449599b98/config/config.go#L106 & https://github.com/vegaprotocol/vegacapsule/blob/df08e1e7187b285130656d4b2cb9d03265cb2fc5/net_confs/config_frontends.hcl#L121-L133 suggest that we should be able to control it through config files. Then we can find them in the validators table, jump to whole row and then validate values using attribute `col-id`.

PR #939, opened 2022-08-03
Test/868 staking ACs

### LINKED ISSUE #868: Add AC references to staking tests
We want to start integrating ACs with Cypress tests.

This may require tests to be rewritten or new tests added to fit with the ACs

ACs can be found in:
https://github.com/vegaprotocol/specs/blob/Add-interface-acceptance-criteria/user-interface/1002-STKE-staking.md

For now we can reference ACs by adding comments next to tests with the AC ids

Tool for measuring coverage:

jtsang586:
# Related issues 🔗

Closes #868 & #884 

# Description ℹ️

Adds some additional tests and ACs. Fix broken flow tests from the validator list changes.

### DIFF
--- a/apps/token-e2e/src/integration/flow/staking-flow.cy.js
+++ b/apps/token-e2e/src/integration/flow/staking-flow.cy.js
@@ -29,7 +29,6 @@ const partValidatorId = '…';
 const txTimeout = Cypress.env('txTimeout');
 const epochTimeout = Cypress.env('epochTimeout');
 
-// Tests skipped because of change of the UI of data nodes
 context('Staking Tab - with eth and vega wallets connected', function () {
   // 1002-STKE-002, 1002-STKE-032
   before('visit staking tab and connect vega wallet', function () {
@@ -108,6 +107,7 @@ context('Staking Tab - with eth and vega wallets connected', function () {
         .should('be.visible');
 
       cy.navigate_to('staking');
+
       cy.validate_validator_list_total_stake_and_share('0', '', '2.00', '100%');
     });
 
@@ -121,6 +121,7 @@ context('Staking Tab - with eth and vega wallets connected', function () {
       );
 
       cy.get('button').contains('Select a validator to nominate').click();
+
       cy.click_on_validator_from_list(0);
 
       cy.staking_validator_page_add_stake('2');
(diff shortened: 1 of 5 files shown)


apps/token-e2e/src/integration/flow/staking-flow.cy.js
```diff
@@ -435,7 +431,7 @@ context.skip('Staking Tab - with eth and vega wallets connected', function () {
 
       cy.get(vegaWalletStakedBalances, txTimeout)
         .should('contain', 2.0, txTimeout)
-        .and('contain', this.validatorName);
+        .and('contain', '');
```
sam-keen, line 434: Question (non-blocking): why check for an empty string here?

sam-keen (approved):

apps/token-e2e/src/integration/flow/staking-flow.cy.js
jtsang586: Nice spot. Corrected!

apps/token-e2e/src/integration/flow/staking-flow.cy.js
```diff
@@ -199,22 +199,25 @@ context.skip('Staking Tab - with eth and vega wallets connected', function () {
...
 
+      // 1002-STKE-049
       cy.get(stakeNextEpochValue, epochTimeout)
         .contains(2.0, epochTimeout)
         .should('be.visible');
 
       cy.get(vegaWalletNextEpochBalances, txTimeout)
         .should('contain', 2.0, txTimeout)
-        .and('contain', this.validatorName)
+        .and('contain', partValidatorId)
```
mattrussell36, line 215: question (non blocking): Seems a odd to be asserting that the ellipsis is displayed in so many places? I might be missing something here.

mattrussell36 (approved):

apps/token-e2e/src/integration/flow/staking-flow.cy.js
jtsang586: Originally there were kind of like placeholders for when we could get validators in Capsule with names but that probably won't happen anytime soon. I can also see there's a lot of repeated assertions for the epoch text. I've tidied that up so it's only asserting the texts once per element in the tests.

PR #942, opened 2022-08-03
Feat/804 simplest proposal form

### LINKED ISSUE #804: Create the simplest make a proposal form
From:
- https://github.com/vegaprotocol/token-frontend/issues/1229

image

- new form at /governance/propose
- text field input that expects full JSON proposal
- no client side validation yet
 - submit button:
 - if proposal fails in wallet: refresh form with previous input, and error message.
 - if proposal is submitted but the status returned is ProposalState.rejected on chain, refresh the form with previous input and 
 - show error details: proposal.errorDetails [String], rejectionReason: ProposalRejectionReason
 - if proposal is submitted and is any other status, the user is redirected to the proposal details page
 
Cannot redirect to proposal after creation due to this issue:
- https://github.com/vegaprotocol/vega/issues/5704 
(Could poll, but that's yet more polling logic in this application that needs to be rewritten yet again)

sam-keen:
# Related issues 🔗

Closes #804

# Description ℹ️

# Demo 📺

https://user-images.githubusercontent.com/6803987/182893796-f1d72148-b027-46db-9213-2cea04585d95.mp4

# Technical 👨‍🔧
- New lib/governance with propose form and use-proposal-submit hook.
- Added proposal transaction types to wallet-types
- Small refactor of error handling in rest connector to handle a different shape of error types returned by wallet


MadalinaRaicu (approved):

libs/governance/src/lib/proposals-hooks/use-proposal-submit.ts
```diff
@@ -0,0 +1,58 @@
...
+import { determineId } from '@vegaprotocol/react-helpers';
+import { useProposalEvent } from './use-proposal-event';
+import type { ProposalSubmission } from '@vegaprotocol/wallet';
+import type { ProposalEvent_busEvents_event_Proposal } from './__generated__/ProposalEvent';
+
+export const useProposalSubmit = () => {
+  const { keypair } = useVegaWallet();
+  const waitForProposalEvent = useProposalEvent();
+
+  const { send, transaction, setComplete, TransactionDialog } =
```
notbot00, line 13: vent (non-blocking): for me it feels very, very strange for a pattern to couple hooks with the UI. i'd expect hooks to return some sort of state or action from which you can easily compute what you want to display on the ui.

libs/governance/src/lib/proposals-hooks/use-proposal-event.ts
```diff
@@ -0,0 +1,75 @@
...
+          errorDetails
+        }
+      }
+    }
+  }
+`;
+
+export const useProposalEvent = () => {
+  const client = useApolloClient();
+  const subRef = useRef<Subscription | null>(null);
```
notbot00, line 29: question (non-blocking): curious, are there any particular reasons we have an overwhelming preference for useRef over useMemo? maybe i'm just more used to useMemo, and that's why it feels more readable to see it.

notbot00 (approved):

libs/governance/src/lib/proposals-hooks/use-proposal-submit.ts
dexturr: cc @mattrussell36 who was a big fan of this. I quite like it myself, as it allows you to either use default UI or take control of it. What do you see as the downsides?

mattrussell36: As discussed on slack :-), the hook does now return the dialogOpen state so this could easily be refactored to not return the components as well. Lets consider this in a different PR though

libs/governance/src/lib/proposals-hooks/use-proposal-event.ts
mattrussell36: For me useMemo doesn't make sense in this use case. We just need to store the subscription instance so we can ensure we stop the subscription when the component unmounts (or when the proposal data comes in) but also manually start it when the function is called. I think a useMemo implementation would be more complex and require more guards to prevent starting it on mount

PR #952, opened 2022-08-04
chore: upgrade to React 18

MadalinaRaicu:
# Related issues 🔗

Closes #951 

# Description ℹ️

Trying to upgrade react only and leave version of nextjs that brings the tailwind module issues. 
Please read: https://betterprogramming.pub/react-v18-why-useeffect-suddenly-go-crazy-db1b42eb2730


apps/token/src/routes/tranches/index.tsx
```diff
@@ -24,7 +24,7 @@ const TrancheRouter = ({ name }: RouteChildProps) => {
   if (error) {
     return (
       <Callout intent={Intent.Danger} title={t('errorLoadingTranches')}>
-        {error}
+        {error.message}
```
MadalinaRaicu, line 27: didn't compile if it was not string | ReactNode

libs/accounts/src/lib/accounts-table.spec.tsx
```diff
@@ -21,41 +21,43 @@ const singleRow: Accounts_party_accounts = {
 };
 const singleRowData = [singleRow];
 
-it('should render successfully', async () => {
-  await act(async () => {
-    const { baseElement } = render(<AccountsTable data={[]} />);
-    expect(baseElement).toBeTruthy();
+describe('AccountsTable', () => {
```
MadalinaRaicu, line 24: jus added `describe(..` ad updated `act(async`  after the testing library upgrade.

libs/environment/src/hooks/use-config.spec.tsx
```diff
@@ -1,4 +1,4 @@
-import { renderHook } from '@testing-library/react-hooks';
+import { renderHook, waitFor } from '@testing-library/react';
```
MadalinaRaicu, line 1: `waitForFirstUpdated` won't be used anymore, we ca use `waitFor` instead

libs/environment/src/hooks/use-environment-errors.spec.tsx
```diff
@@ -0,0 +1,199 @@
...
+  process.env['NX_GIT_BRANCH'] = mockEnvironmentState.GIT_BRANCH;
+  process.env['NX_GIT_ORIGIN_URL'] = mockEnvironmentState.GIT_ORIGIN_URL;
+  process.env['NX_GIT_COMMIT_HASH'] = mockEnvironmentState.GIT_COMMIT_HASH;
+  process.env['NX_GITHUB_FEEDBACK_URL'] =
+    mockEnvironmentState.GITHUB_FEEDBACK_URL;
+});
+
+describe('throws error', () => {
+  beforeEach(() => {
+    // @ts-ignore: typescript doesn't recognize the mock implementation
```
MadalinaRaicu, line 105: just moving some tests in a separate test suite. we are updating`process.env` in each test case and they are async. when I ran them together with the tests from `use-environment` some of them would fail. this fixes this issue.

libs/environment/src/hooks/use-nodes.spec.tsx
```diff
@@ -91,13 +90,13 @@ describe('useNodes hook', () => {
       },
     });
 
-    await waitForNextUpdate();
+    // await waitFor(() => { });
```
MadalinaRaicu, line 93:
```suggestion
```

libs/environment/src/hooks/use-nodes.spec.tsx
```diff
@@ -1,10 +1,11 @@
-import { renderHook, act } from '@testing-library/react-hooks';
+import { renderHook, act } from '@testing-library/react';
```
MadalinaRaicu, line 1: `@testing-library/react` gets replaced by `@testing-library/react`

MadalinaRaicu:
`hydrateRoot being imported by nextjs issue is known`

apps/simple-trading-app/tsconfig.json
```diff
@@ -10,6 +10,7 @@
     "noImplicitOverride": true,
     "noPropertyAccessFromIndexSignature": false,
     "noImplicitReturns": true,
+    "noImplicitAny": false,
```
macqbat, line 13: It's possible to add in a few places some types and we could avoid adding this implicit any

macqbat: Check it please https://github.com/vegaprotocol/frontend-monorepo/pull/970

MadalinaRaicu: yes, wanted to do that but wasn't sure of some of the types in simple-trading-app and didn't want to break anything.

libs/react-helpers/src/hooks/use-data-provider.ts
```diff
@@ -55,7 +56,7 @@ export function useDataProvider<Data, Delta>({
     return Promise.reject();
   }, []);
   const callback = useCallback(
-    ({ data, error, loading, delta, insertionData, totalCount }) => {
+    ({ data, error, loading, delta, insertionData, totalCount }: any) => {
```
macqbat, line 59: non-blocking: this `any` doesn't look nice, maybe it's possible to create from here an issue to fix this in the future?

macqbat (approved):

asiaznik (approved):

libs/react-helpers/src/hooks/use-data-provider.ts
macqbat: or this proposal maybe can handle it: https://github.com/vegaprotocol/frontend-monorepo/pull/977

bglownia: @macqbat It looks like we started working on this as the same time, I pushed fix already

bglownia (approved):

PR #963, opened 2022-08-08
feat: multi open accordion

### LINKED ISSUE #744: Use multi-open accordion for market info
There's a `type` value on the accordion which you can set to `single` or `multiple`. For the market info panel, I think it would be better to allow them to open multiple sections at once.

Probably best to build this as a togglable property of the accordion component than just change it globally. 

I did try changing it to `multiple` which allowed them to open independently, but for some reason it stopped the chevron from rotating on open…

MadalinaRaicu:
# Related issues 🔗

Closes #744

# Description ℹ️

Updated accordion to open multiple expansion panels.
Updated icon so that we are not using the pixelated one anymore. (hard to distinguish small pixelated icons on different resolutions)

# Demo 📺

Screenshot 2022-08-08 at 10 53 44

from: 

Screenshot 2022-08-08 at 10 47 25


libs/ui-toolkit/src/components/accordion/chevron-down.tsx
```diff
@@ -1,31 +1,11 @@
...
-      <rect x="8" y="4" width="2" height="2" />
-      <rect x="6" y="6" width="2" height="2" />
-      <rect x="4" y="4" width="2" height="2" />
-      <rect x="2" y="2" width="2" height="2" />
-      <rect x="0" y="0" width="2" height="2" />
-    </svg>
-  );
-};
+import { styled } from '@stitches/react';
+import { ChevronDownIcon } from '@radix-ui/react-icons';
```
sam-keen, line 2: Question (non-blocking): Why did you choose to use an icon from the radix icon set? Was the blueprint one not appropriate?

package.json
```diff
@@ -29,6 +29,7 @@
     "@sentry/nextjs": "^6.19.3",
     "@sentry/react": "^6.19.2",
     "@sentry/tracing": "^6.19.2",
+    "@stitches/react": "^1.2.8",
```
notbot00, line 32: question (?): this new install doesn't seem to be used anywhere, do we need it?

sam-keen (approved):

package.json
edd: Oh phew! I was going to comment that adding a new (largish) dep for an animation might be the wrong approach - but in that case, it's easy to remove!

MadalinaRaicu: I am using it. In ChevronDownIcon

notbot00: alright, i missed that. in that case, let me rephrase, why add a new library to wrap the icon when you could essentially just wrap it in a div, add some tailwind classes to it to handle the transition and the transform?

MadalinaRaicu: cause in `multiple` type I can't control the active state myself as I did previously. so the Chevron would not rotate anymore when wrapped in a div. I need to control `[data-state=open] &` on the icon (which is added by radix primitives on `multiple`) and it would be great to do this with tailwind but did not get my head around adding the transition when this `[data-state=open] &` is set.  any suggestion?

MadalinaRaicu: https://www.radix-ui.com/docs/primitives/components/accordion `multiple` is quite different than `single`

MadalinaRaicu: this is why I did it as in their example in documentation.

notbot00: haven't really done a deep dive in the docs, but looks like this could potentially be resolved with using a controlled state, as before. the docs suggest that the accordion root takes a `value: string[]` type, i assume that would be for multiples. did you try changing the value state from
```
const [value, setValue] = useState<string>('');
```
to:
```
const [value, setValue] = useState<string[]>([]);
```
?

let me know if that is not working, and happy to approve.

MadalinaRaicu: 'value' isavailable only on 'single' type. that was the issue.

MadalinaRaicu: `onValueChange` is not on `multiple`, so can't set value. 
on `single` I can have only one expansion panel open. I try to open another one and the other one closes. this is why I used this type of chevron (as per their documentation).

MadalinaRaicu: updated with setting up the array of values as in your suggestion.

notbot00 (approved):

dexturr (approved):

AndyWhiteVega:
tested: looks good

PR #964, opened 2022-08-08
fix: provider sometimes not having signer

dexturr:
# Related issues 🔗

N/A bug noticed when testing release

# Description ℹ️

Fix intermittent errors with interacting with Ethereum

# Technical 👨‍🔧

The async use effect was sometimes resolving in the incorrect order, creating a race condition. There is a hack fix here with a todo that cancels it. We should find a better way of doing this in future however I do not want this to push the release out by a further week so on balance I thought it best to go for the hacky fix now and look into a better solution later.

### DIFF
--- a/apps/token/src/contexts/contracts/contracts-provider.tsx
+++ b/apps/token/src/contexts/contracts/contracts-provider.tsx
@@ -30,6 +30,7 @@ export const ContractsProvider = ({ children }: { children: JSX.Element }) => {
   // contracts so that we can sign transactions, otherwise use the provider for just
   // reading data
   React.useEffect(() => {
+    let cancelled = false;
     const run = async () => {
       let signer = null;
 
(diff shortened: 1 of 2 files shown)


sam-keen (approved):

MadalinaRaicu (approved):

apps/token/src/contexts/contracts/contracts-provider.tsx
```diff
@@ -55,23 +55,28 @@ export const ContractsProvider = ({ children }: { children: JSX.Element }) => {
...
+              ),
+              claim: new Claim(ENV.addresses.claimAddress, signer || provider),
+            });
+          }
         }
       }
     };
     run();
+    return () => {
+      //  TODO: hacky quick fix for release to prevent race condition, find a better fix for this.
```
edd, line 77: Yes

PR #967, opened 2022-08-08
chore: show total supply on governance page

dexturr:
# Related issues 🔗

From feedback

# Description ℹ️

Show total supply in vote information breakdown

### DIFF
--- a/apps/token/src/i18n/translations/dev.json
+++ b/apps/token/src/i18n/translations/dev.json
@@ -625,6 +625,6 @@
   "FilterProposalsDescription": "Filter by proposal ID or proposer ID",
   "Freeform proposal": "Freeform proposal",
   "NewProposal": "New proposal",
-  "MinProposalRequirements": "You must have at least 1 VEGA staked to make a proposal",
+  "MinProposalRequirements": "You must have at least 1 VEGA associated to make a proposal",
   "totalSupply": "Total Supply"
 }
--- a/apps/token/src/i18n/translations/dev.json
+++ b/apps/token/src/i18n/translations/dev.json
@@ -625,6 +625,6 @@
   "FilterProposalsDescription": "Filter by proposal ID or proposer ID",
   "Freeform proposal": "Freeform proposal",
   "NewProposal": "New proposal",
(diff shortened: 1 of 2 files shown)


apps/token/src/i18n/translations/dev.json
```diff
@@ -625,5 +625,6 @@
   "FilterProposalsDescription": "Filter by proposal ID or proposer ID",
   "Freeform proposal": "Freeform proposal",
   "NewProposal": "New proposal",
-  "MinProposalRequirements": "You must have at least 1 VEGA to make a proposal"
+  "MinProposalRequirements": "You must have at least 1 VEGA to make a proposal",
```
edd, line 628: *question (blocking)*: It should be clearer what state this 1 vega needs to be in. And also it's not entirely true that it has to be 1 (see https://github.com/vegaprotocol/vega/issues/5786). But ignoring the number, could this be:

```suggestion
  "MinProposalRequirements": "You must have at least 1 VEGA staked to make a proposal",
```

It's also not quite true. It just has to be associated. But it's important because 1 VEGA in your general account won't count. @candida-d & me talked about this a lot on Friday so she'll understand where I'm coming from.

edd (changes requested):
Big review blocking :x: because I want a wording question answered. Code-wise it all looks good.

apps/token/src/i18n/translations/dev.json
```diff
@@ -625,5 +625,6 @@
   "FilterProposalsDescription": "Filter by proposal ID or proposer ID",
   "Freeform proposal": "Freeform proposal",
   "NewProposal": "New proposal",
-  "MinProposalRequirements": "You must have at least 1 VEGA to make a proposal"
+  "MinProposalRequirements": "You must have at least 1 VEGA staked to make a proposal",
```
campbellssource, line 628: `associated` rather than `staked` please

dexturr:
```suggestion
  "MinProposalRequirements": "You must have at least 1 VEGA associated to make a proposal",
```

edd (approved):

sam-keen (approved):

PR #977, opened 2022-08-09
A propose for avoiding no-explicity-any directive on false

macqbat:
# Related issues 🔗

---

# Description ℹ️
A propose for avoiding no-explicity-any directive on false

# Technical 👨‍🔧

Add new type for update callback arguments


asiaznik:
Nice! 👍🏻 ~~`any`~~

MadalinaRaicu (approved):

PR #979, opened 2022-08-09
Test/fix nightly run

jtsang586:
- Add implicit wait for intermittent failing test

- Attempt to fix night run

### DIFF
--- a/apps/explorer-e2e/src/integration/node-switcher.cy.js
+++ b/apps/explorer-e2e/src/integration/node-switcher.cy.js
@@ -71,7 +71,7 @@ context('Node switcher', function () {
     });
 
     afterEach('Close node switcher', function () {
-      cy.getByTestId(closeDialogBtn).click();
+      cy.getByTestId(closeDialogBtn).should('be.be.visible').click();
     });
 
     function validateNodeError(errortype, errorMsg) {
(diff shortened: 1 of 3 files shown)


dexturr (approved):

apps/explorer-e2e/src/integration/node-switcher.cy.js
```diff
@@ -71,7 +71,7 @@ context('Node switcher', function () {
     });
 
     afterEach('Close node switcher', function () {
-      cy.getByTestId(closeDialogBtn).click();
+      cy.getByTestId(closeDialogBtn).should('be.be.visible').click();
```
notbot00, line 74: question (non-blocking): this kinda looks like a typo, but i'm not sure if `be.be.(...)` is any different from just stating `be.(...)`?

notbot00 (approved):

apps/explorer-e2e/src/integration/node-switcher.cy.js
jtsang586: You're right. I've got no idea why nothing has complained and how the tests passed. But it's corrected!

MadalinaRaicu (approved):

PR #992, opened 2022-08-10
feat: [console-lite] - market list - refactor filters query, int tests of long market list

### LINKED ISSUE #778: Console lite - market list - connect pagination
## Task
Although `marketsConnection` query is already exposed,  `makeDataProvider` is still struggling with `totalCount` case, etc. So it's much better wait for final implementation of both data provider and backend endpoint

## ACs
- [ ] connect with marketsConnection query
- [ ] Resolve sorting problem (client or server side ?)
- [ ] Resolve reseting price change from the cell

macqbat:
# Related issues 🔗

Closes #778

# Description ℹ️

Refactor list filtering, cover with test long list view

# Technical 👨‍🔧

Use more fixture for generate long list in int test. Remove redundant query for filters


apps/simple-trading-app-e2e/src/integration/market-list.test.ts
```diff
@@ -93,4 +93,68 @@ describe('market list', () => {
         .should('have.text', 'Future');
     });
   });
+
+  describe('long list of results should be handled properly', () => {
+    it('5000 of markets', () => {
```
sam-keen, line 98:
```suggestion
    it('handles 5000 markets', () => {
```

apps/simple-trading-app-e2e/src/integration/market-list.test.ts
```diff
@@ -93,4 +93,68 @@ describe('market list', () => {
...
+          cy.get('body').realPress('Tab');
+        }
+        cy.focused().parent('.ag-row').should('have.attr', 'row-index', '49');
+        cy.get('.ag-center-cols-container')
+          .find('[role="row"]')
+          .should('have.length', 31);
+      });
+    });
+
+    it('50000 of markets', () => {
```
sam-keen, line 127:
```suggestion
    it('handles 50000 markets', () => {
```

apps/simple-trading-app/src/app/components/simple-market-list/simple-market-toolbar.spec.tsx
```diff
@@ -3,46 +3,57 @@ import { useLocation, useRoutes, BrowserRouter } from 'react-router-dom';
...
-  };
-
   const WrappedCompForTest = () => {
     const routes = useRoutes([
       {
         path: '/',
-        element: <SimpleMarketToolbar />,
+        element: (
+          <SimpleMarketToolbar
+            data={filterData as unknown as SimpleMarkets_markets[]}
```
sam-keen, line 16: Question (non-blocking): why is it necessary to cast as unknown as well?

sam-keen (approved):

apps/simple-trading-app/src/app/components/simple-market-list/simple-market-toolbar.spec.tsx
notbot00: might be because it's imported from a json file, which doesn't have associated types

macqbat: yes, cast to unknown is redundant if just after we cast to proper type. Thanks for the catch. Let's stay with version with plain json mock file, should have better readability.

PR #995, opened 2022-08-10
feat: reselct field so that GTC is default on limit

### LINKED ISSUE #940: Good 'till Canceled should be the default time in force for limit orders. Not IOC
Good `til canceled is a better default time in force than IOC for limit order.
It is a more common default.
It also works both auction and continuous modes.

However, Good till canceled is applicable for market orders. So IOC should remain the default for Market orders

MadalinaRaicu:
# Related issues 🔗

Closes #940 

re-select field so that GTC is default on limit  orders and IOC on market orders.

# Demo 📺

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,7 +21,7 @@
 
 # IDE - VSCode
 .vscode/*
-!.vscode/settings.json
+.vscode/settings.json
 !.vscode/tasks.json
 !.vscode/launch.json
 !.vscode/extensions.json
--- a/libs/deal-ticket/src/components/time-in-force-selector.tsx
+++ b/libs/deal-ticket/src/components/time-in-force-selector.tsx
@@ -4,6 +4,7 @@ import {
   VegaWalletOrderType,
 } from '@vegaprotocol/wallet';
 import { t } from '@vegaprotocol/react-helpers';
+import { useEffect } from 'react';
 
 interface TimeInForceSelectorProps {
   value: VegaWalletOrderTimeInForce;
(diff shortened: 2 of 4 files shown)


libs/deal-ticket/src/components/time-in-force-selector.tsx
```diff
@@ -44,7 +45,13 @@ export const TimeInForceSelector = ({
             timeInForce === VegaWalletOrderTimeInForce.FOK ||
             timeInForce === VegaWalletOrderTimeInForce.IOC
         );
-
+  useEffect(() => {
```
MadalinaRaicu, line 48: TIF field is already selected on Market tab, so whenever order type updates we need to reselect TIF field.

.gitignore
```diff
@@ -21,7 +21,7 @@
 
 # IDE - VSCode
 .vscode/*
-!.vscode/settings.json
+.vscode/settings.json
```
MadalinaRaicu, line 24: settings.json was not ignored.

sam-keen (approved):

libs/deal-ticket/src/components/time-in-force-selector.tsx
mattrussell36, line 48: chore (blocking): I'd prefer not to use useEffect like this? Lets just change the TIF state in the function where order type changes.

mattrussell36 (changes requested):

libs/deal-ticket/src/components/time-in-force-selector.tsx
MadalinaRaicu: updated

mattrussell36 (approved):

AndyWhiteVega:
Tested: works fine

PR #902, opened 2022-07-28
Feature/473 position margin levels

bglownia:
# Related issues 🔗

Closes #473

Issue #473, opened 2022-05-25
Position margin level


gordsport:
As a user of console I want to see how close I am to being liquidated.

#### Acceptance criteria
- [x] Show estimated liquidation price
- [x] Show entry price
- [x] Show progress bar indicating how close you are to the liquidation price
- [x] Color should change using Intent to indicate 'danger' level using @witgaw calculations below

gordsport:
@witgaw - can you please update on the merging of the python notebook regarding margin warnings so we can use this?

campbellssource:
@PeteWilliams and @witgaw to discuss a good time/price point to start warning

PeteWilliams:
Have talked to @witgaw who has proposed the following conditions should trigger a warning:

1.   `margin account balance` < `search level` + (`initial level - search level)/2` [we’re more than half way down between initial level and search level]
AND
  2.  `collateral account balance` < `initial level - search level` [the margin won’t be topped up to the desired level <-- they still don’t get closed out as long as `margin account balance` (after topup) > `search level`, but they are cutting it close so I think it warrants a warning.

campbellssource:
Here is the workbook for calculating liquidation level: https://github.com/vegaprotocol/research/blob/master/notebooks/misc/Misc%20position%20metrics.ipynb

PeteWilliams:
Have added initial designs for this:
https://www.figma.com/file/K49Nx7XfocShhHcwwQzQkm/Console-V2?node-id=2172%3A32829

When the position hits the condition [referenced above](https://github.com/vegaprotocol/frontend-monorepo/issues/473#issuecomment-1164290751), it's liquidation indicator bar should turn from white to orange (the 'warning' colour):
![image](https://user-images.githubusercontent.com/1325606/175934415-6e6e9f30-46ef-4416-a2df-6297927d6132.png)

We should also show a toast that alerts the user that this has happened:
![image](https://user-images.githubusercontent.com/1325606/175934633-623bcefd-ec93-4a8e-abc4-3aee18d876c2.png)

[...]

PR #996, opened 2022-08-10
fix: orderbook decimal formats

### LINKED ISSUE #859: Review all decimal places in Console (except market info)
This to conduct a review all decimal places (asset decimal, market decimal, position decimal, tick size etc). Document how they all relate and build rules into UI's to show them correctly.

There are 4 definitions of the number of decimal places 

- `Asset.decimals`
- `Market.decimalPlaces`
- `Market.positionDecimalPlaces`
- `settlementPriceDecimals`

Generally:

- `Asset.decimals` should alway match the asset on the native chain. e.g. for ERC20 it is often 18.
- `Market.decimalPlaces` is the number of decimal places for the Quote. e.g if it is BTC/USD it'll be how precise you can quote BTC in USD. .e.g 5 (USD may be the asset DAI which has 18 DP)
- `Market.positionDecimalPlaces` is the size field on a order/position and the Volume in the order book. note on Testnet this has only ever been 0. if it we more than zero we'd be using the new fractional order sizes
- `settlementPriceDecimals` used for oracle price (not displayed anywhere in console).

Notes:
- Accounts are always asset.decmals
- the limit price is always `Market.decimalPlaces`
- the size is always `Market.positionDecimalPlaces`
- market decimal places and tick size are very related as it tells you how much there is between each possible price on the market
- oracle price used for settlement in oracle spec = `settlementPriceDecimals` 

Where each should be used:
## Chart
- y axis should use `Market.decimalPlaces`

## Depth chart
- X axis `Market.decimalPlaces`
- Y axis `Market.positionDecimalPlaces`

## Ticket 
- Amount (aka size): use  `Market.positionDecimalPlaces`
- the price next to amount on market order or input on Limit order: `Market.decimalPlaces`

## info
This is going to go into a separate ticket

## Order book 
- bid and ask vol = `Market.positionDecimalPlaces`
- cumulative vol = `Market.positionDecimalPlaces`
- price = `Market.decimalPlaces`

## Trades
- price = `Market.decimalPlaces`
- size = `Market.positionDecimalPlaces`

## Positions 
Screenshot 2022-07-27 at 13 48 52

new table:

image
https://github.com/vegaprotocol/frontend-monorepo/issues/840

- amount aka size: `Market.positionDecimalPlaces`
- Avg entry price: `Market.decimalPlaces`
- mark price: `Market.decimalPlaces`
- liquidation price `Market.decimalPlaces`
- leverage = n/a
- margin allocated = `Asset.decimals`
- PnL = `Asset.decimals`

## Orders
- amount (aka size) : `Market.positionDecimalPlaces`
- Filled =  `Market.positionDecimalPlaces`
- Price = `Market.decimalPlaces`
- Mark price = `Market.decimalPlaces`
- pNL: `asset.decimals` (this assumes PnL is not rounded) ACTION: check rounding.

## Collateral 
- All `asset.decimals`

## Fills 
This table needs a number of changes do have commented decimals on this ticket:  
https://github.com/vegaprotocol/frontend-monorepo/issues/845#issuecomment-1203789356

notbot00:
# Related issues 🔗

Closes #859 

# Description ℹ️

#### Chart
- [x] y axis should use `Market.decimalPlaces`

#### Depth chart
- [x] X axis `Market.decimalPlaces`
- [x] Y axis `Market.positionDecimalPlaces` (changed in this PR, based on: https://github.com/vegaprotocol/pennant/issues/541)

#### Ticket 
- [x] Amount (aka size): use  `Market.positionDecimalPlaces` (already correct: https://github.com/vegaprotocol/frontend-monorepo/blob/master/libs/deal-ticket/src/components/deal-ticket.tsx#L47)
- [x] the price next to amount on market order or input on Limit order: `Market.decimalPlaces` (already correct: https://github.com/vegaprotocol/frontend-monorepo/blob/master/libs/deal-ticket/src/components/deal-ticket.tsx#L84-L97)

#### Order book 
- [x] bid and ask vol = `Market.positionDecimalPlaces` (already correct: https://github.com/vegaprotocol/frontend-monorepo/blob/master/libs/market-depth/src/lib/orderbook-row.tsx#L43-L49)
- [x] cumulative vol = `Market.decimalPlaces` (apparently also already correct)
- [x] price = `Market.positionDecimalPlaces` (apparently also already correct)

#### Trades
- [x] price = `Market.decimalPlaces` (already correct: https://github.com/vegaprotocol/frontend-monorepo/blob/master/libs/trades/src/lib/trades-table.tsx#L60-L76)
- [x] size = `Market.positionDecimalPlaces` (already correct: https://github.com/vegaprotocol/frontend-monorepo/blob/master/libs/trades/src/lib/trades-table.tsx#L77-L93)

#### Positions 
- [x] changes in https://github.com/vegaprotocol/frontend-monorepo/pull/902 seem all correct

#### Orders
- [x] amount (aka size) : `Market.positionDecimalPlaces` (already correct: https://github.com/vegaprotocol/frontend-monorepo/blob/master/libs/orders/src/lib/components/order-list/order-list.tsx#L116-L134)
- [x] filled =  `Market.positionDecimalPlaces` (already correct: https://github.com/vegaprotocol/frontend-monorepo/blob/master/libs/orders/src/lib/components/order-list/order-list.tsx#L155-L177)
- [x] price = `Market.decimalPlaces` (already correct: https://github.com/vegaprotocol/frontend-monorepo/blob/master/libs/orders/src/lib/components/order-list/order-list.tsx#L178-L197)
- [ ] mark price = `Market.decimalPlaces` (not found in orders)
- [ ] PnL: `asset.decimals` (this assumes PnL is not rounded) ACTION: check rounding. (not found in orders)

#### Collateral 
- [x] all `asset.decimals` (already correct - only balance field is the only numeric field to be formatted - https://github.com/vegaprotocol/frontend-monorepo/blob/master/libs/accounts/src/lib/accounts-table.tsx#L131-L141)


libs/positions/src/lib/positions-table.tsx
```diff
@@ -153,9 +153,19 @@ export const PositionsTable = forwardRef<AgGridReact, PositionsTableProps>(
             'color-vega-red': ({ value }: { value: string }) =>
               Number(value) < 0,
           }}
-          valueFormatter={({ value, data }: ValueFormatterParams) =>
```
bglownia, line 156: positions-table is totally changed in #473, PR is waiting for review

libs/market-depth/src/lib/orderbook-row.tsx
```diff
@@ -43,19 +43,19 @@ export const OrderbookRow = React.memo(
         <Vol
           testId={`bid-vol-${price}`}
           value={bid}
-          valueFormatted={addDecimal(bid, positionDecimalPlaces)}
+          valueFormatted={addDecimal(bid, decimalPlaces)}
```
mattrussell36, line 46: question (blocking): Are you sure these three changes are correct? My understanding is that price values should use market.decimalPlaces, and volume/size should use market.positionDecimalPlaces (for fractional order sizes)

campbellssource: That's on me. I had it wrong in the ticket. have corrected it now.

libs/positions/src/lib/positions-table.tsx
notbot00: reverted the position table changes so we don't get conflicts while waiting

libs/market-depth/src/lib/orderbook-row.tsx
notbot00: changed this to align with the criteria

libs/market-depth/src/lib/orderbook-row.tsx
```diff
@@ -50,12 +50,12 @@ export const OrderbookRow = React.memo(
         <PriceCell
           testId={`price-${price}`}
           value={BigInt(price)}
-          valueFormatted={addDecimalsFormatNumber(price, decimalPlaces)}
+          valueFormatted={addDecimalsFormatNumber(price, positionDecimalPlaces)}
```
mattrussell36, line 53: chore (blocking): Should be market.decimalPlaces for price

libs/market-depth/src/lib/orderbook-row.tsx
```diff
@@ -50,12 +50,12 @@ export const OrderbookRow = React.memo(
...
         />
         <Vol
           testId={`ask-vol-${price}`}
           value={ask}
-          valueFormatted={addDecimal(ask, positionDecimalPlaces)}
+          valueFormatted={addDecimal(ask, decimalPlaces)}
```
mattrussell36, line 58: chore (blocking): should be market.positionDecimalPlaces for volume/size

mattrussell36 (changes requested):

libs/market-depth/src/lib/orderbook-row.tsx
notbot00, line 58:
```suggestion
          valueFormatted={addDecimal(ask, positionDecimalPlaces)}
```

libs/market-depth/src/lib/orderbook-row.tsx
```diff
@@ -50,7 +50,7 @@ export const OrderbookRow = React.memo(
         <PriceCell
           testId={`price-${price}`}
           value={BigInt(price)}
-          valueFormatted={addDecimalsFormatNumber(price, decimalPlaces)}
+          valueFormatted={addDecimalsFormatNumber(price, positionDecimalPlaces)}
```
notbot00, line 53:
```suggestion
          valueFormatted={addDecimalsFormatNumber(price, decimalPlaces)}
```

libs/market-depth/src/lib/orderbook-row.tsx
notbot00: changed it back

libs/market-depth/src/lib/orderbook-row.tsx
notbot00: changed it back

libs/market-depth/src/lib/depth-chart.tsx
```diff
@@ -161,7 +162,15 @@ export const DepthChartContainer = ({ marketId }: DepthChartManagerProps) => {
 
   return (
     <AsyncRenderer loading={loading} error={error} data={data}>
-      {depthData && <DepthChart {...depthData} theme={theme} />}
+      {depthData && (
+        <DepthChart
+          {...depthData}
+          theme={theme}
+          volumeFormat={(volume) =>
```
johnwalley, line 169: We should also pass a value for `priceFormat`. The chart defaults to 2dp if we don't provide a value.

notbot00: hmm, there are a lot of `formatLevel` function calls going on a few lines above, which seemingly pre-format the prices going in the chart data. are those redundant? not sure i fully understand their purpose

johnwalley: They seem badly named. I'd probably call them `parseLevel` or similar. It looks like they take the raw API data which are strings and converts it to numbers.

notbot00: okay, removed the decimal formatting from the formatLevel function and renamed it to parseLevel

johnwalley: We are surrounded by strangely named functions!

I think we still need `addDecimal` because it converts the string (integer) we get from the API to a string (decimal floating point style, e.g. NNNN.MMMMM). We can then convert this to a number to pass to the chart.

What we could do with is a function that takes the string from the API and returns the corresponding JS number. We might already have one but I can't see it. There is `toBigNum` but the charts don't support it.

(Off topic for this PR but presumably we work with numbers the way we do because we are concerned about precision and accuracy. This should impact the charts too. I'll put some thought into it) 

Now if volume is encoded in the same way we also need to convert that from an integer string to a decimal string.

All of the above is based on my knowledge of the API data schema from last year. It's possible we have changed how we encode numbers in a way which materially affects the charts but I've not seen anything to suggest that.

notbot00: okay, i think i understand what the problem was, namely that the decimalPlaces and the positionDecimalPlaces are different, so to be accurate along the x and y axes we need to have both numeric values using the correct decimals for each. fixed this. not sure what else we're missing in this PR.

dexturr (approved):

mattrussell36 (approved):

Issue #597, opened 2022-06-21
IPFS deploy


edd:
Following on from #126, which configured the _how_, this ticket involves getting explorer hosted on IPFS.  

# Tasks
- [ ] Create site on ⚡fleek
- [ ] add `.fleek.json` to `token` app
- [ ] Repoint hostname to ⚡fleek
- [ ] Test a deploy, ensure CSS/etc loads properly

elmargasimov:
@edd @dexturr: Is this still blocked or can we move it to the backlog to be prioritised and refined?

Cc: @pennyandrews

Issue #596, opened 2022-06-21
IPFS deploys for Token Frontend


edd:
Following on from #126, which configured the _how_, this ticket involves getting token-frontend hosted on IPFS.  

# Tasks
- [ ] Create site on ⚡fleek
- [ ] add `.fleek.json` to `token` app
- [ ] Repoint hostname to ⚡fleek
- [ ] Test a deploy, ensure CSS/etc loads properly

edd:
Marked as blocked, waiting for a deploy later today

edd:
This was switched over, but is now back on Netlify due to a report from Dan that deeply nested routes do not work:

https://console.fairground.wtf/markets should not show an IPFS folder index

PR #1022, opened 2022-08-11
fix: unbreak nx dependency mapping

edd:
# Description ℹ️

Fix NX dependency mapping with this one simple trick (the nx npmScope did not match the path we used to declare deps). This seemed to cause subtle build failures when building from a clean source.

# Demo
## Before
graph (1)

## After
graph(1)

# To verify
Run the following on `master`:
```bash
rm -rf node_modules && nvm install 16.14.0 && yarn && SKIP_NX_CACHE=true ./node_modules/.bin/nx run token:build:production --source-map=false --skip-nx-cache && cd dist/apps/token && python3 -m http.server 8001
```

and you'll see that it's nearly right, but all the buttons render incorrectly. It seems like `react-helpers` wasn't being built in time, always. Apply the fix, re-run, and it will all work.

Unblocks #596, #597

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -1,5 +1,5 @@
 {
-  "npmScope": "vegaprotocol",
+  "npmScope": "@vegaprotocol",
   "affected": {
     "defaultBase": "master"
   },


asiaznik (approved):
works like a charm 🥇

edd:
Just checking back in. This:
- [x] still works reliably
- [x] is still the least satisfying '@' symbol I have ever typed

macqbat (approved):

mattrussell36 (approved):

PR #1027, opened 2022-08-12
Fix/1008: Tooltip visibility

### LINKED ISSUE #1008: Exit conditions: allow user to click on links
Related to #815 

When hovering over to see market entry/exit conditions. There are links for the user to click on but it's not possible to do so because the modal disappears as soon as the mouse moves away from the trading mode text.

Expected behaviour:
Modal should stay open when mouse is inside it and disappear when mouse moves out of the modal area

https://user-images.githubusercontent.com/30622993/184117706-d0659b16-ce98-495d-ac73-612b1857e18c.mov

notbot00:
# Related issues 🔗

Closes #1008 

# Description ℹ️

Updated radix-ui tooltips to v1.0.0, that fixed the issue of the tooltip contents disappearing on hover.

### DIFF
--- a/libs/ui-toolkit/src/components/tooltip/tooltip.spec.tsx
+++ b/libs/ui-toolkit/src/components/tooltip/tooltip.spec.tsx
@@ -10,7 +10,7 @@ it('Renders a tooltip', async () => {
   // radix applies the data-state attribute
   expect(screen.getByRole('button')).toHaveAttribute('data-state', 'closed');
   fireEvent.mouseOver(screen.getByRole('button'));
-  expect(await screen.findByRole('tooltip')).toBeInTheDocument();
+  expect(screen.queryByRole('tooltip')).not.toBeInTheDocument();
 });
 
 it('Doesnt render a tooltip if no description provided', () => {
--- a/libs/ui-toolkit/src/components/tooltip/tooltip.tsx
+++ b/libs/ui-toolkit/src/components/tooltip/tooltip.tsx
@@ -6,6 +6,7 @@ import {
   Trigger,
   Content,
   Arrow,
+  Portal,
 } from '@radix-ui/react-tooltip';
 
 export interface TooltipProps {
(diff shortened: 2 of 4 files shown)


MadalinaRaicu (approved):

libs/ui-toolkit/src/components/tooltip/tooltip.spec.tsx
```diff
@@ -10,7 +10,7 @@ it('Renders a tooltip', async () => {
   // radix applies the data-state attribute
   expect(screen.getByRole('button')).toHaveAttribute('data-state', 'closed');
   fireEvent.mouseOver(screen.getByRole('button'));
-  expect(await screen.findByRole('tooltip')).toBeInTheDocument();
+  expect(screen.queryByRole('tooltip')).not.toBeInTheDocument();
```
sam-keen, line 13:
```suggestion
  expect(screen.queryByRole('tooltip')).toBeInTheDocument();
```

sam-keen (changes requested):

libs/ui-toolkit/src/components/tooltip/tooltip.spec.tsx
notbot00: since i've changed the tooltip to use portals, it will be never in the rendered document. not sure how to test this properly, in the old react 17 days i could just add a `root: document.body` to the renderer, so the portals would also render on the screen object, but here that doesn't work anymore.

sam-keen: Ah ok, that's interesting. Can you make use of the legacyRoot option? https://testing-library.com/docs/react-testing-library/api/#legacyroot If not, I don't think the test achieves its stated purpose, so perhaps it should be removed for now until we find a way to test portals.

sam-keen: Ah that's interesting. Can you use container option to specify the root correctly? https://testing-library.com/docs/react-testing-library/api/#container If not, I think that this test should probably be removed until we find a way to test portals, as it doesn't fulfil its purpose.

notbot00: tried that, no luck, the portal's missing from the rendered screen :/ didn't want to spend too much time on the tests, but might try this suggestion i found on this thread: https://github.com/facebook/react/issues/11565#issuecomment-1099638877

notbot00: turns out the event handling changed internally in the tooltip component in the new version. tried a few things, wasn't able to test how i can trigger the uncontrolled tooltip to open, so i've left tests for the controlled state only for now.

sam-keen (approved):

PR #1029, opened 2022-08-12
feat: [console-lite] - order review - add info tooltips

### LINKED ISSUE #1004: Tooltips for size and review
- [ ] It should have a tooltip for every row
- [x] No. of contracts tooltip: "The number of contracts determines how many units of the futures contract to buy or sell. For example, this is similar to buying one share of a listed company. The value of 1 contract is equivalent to the price of the contract. For example, if the current price is $50, then one contract is worth $50."
- [x] Notional size: "The notional size represents the position size in the settlement asset of the futures contract. The notional size is calculated by multiplying the number of contracts by the price of the contract. For example, ten contracts traded at a price of $50 has a notional size of $500."
- [x] Est. Margin: "When opening a position on a futures market, you must post margin to cover any potential losses that you may incur. The margin is typically a fraction of the notional position size. For example, for a notional position size of $500, if the margin requirement is 10%, then the estimated margin would be approximately $50."
- [x] Est. Close out: "Because you only need to post a fraction of your position size as margin when trading futures, it is possible to obtain leverage meaning your notional position size exceeds your account balance. In this scenario, if the market moves against your position, it will sometimes be necessary to force close your position due to insufficient funds. The estimated close out tells you the price at which that would happen based on current position and account balance."
- [x] Est. Fees: "When you execute a new buy or sell order, you must pay a small amount of commission to the network for doing so. This fee is used to provide income to the node operates of the network and market makers who make prices on the futures market you are trading."

macqbat:
# Related issues 🔗

Closes #1004

# Description ℹ️

Add info tooltip for order review view

# Demo 📺

Screenshot 2022-08-12 at 16 25 18

# Technical 👨‍🔧

Use Tooltip component with custom css class


apps/simple-trading-app/src/app/components/deal-ticket/deal-ticket-estimates.tsx
```diff
@@ -71,7 +131,21 @@ export const DealTicketEstimates = ({
           &nbsp;
           <small>({quoteName})</small>
         </dt>
-        <dd>{estCloseOut}</dd>
+        <dd className="flex gap-x-5 items-center">
```
sam-keen, line 134: Chore (non-blocking): There's a lot of repetition here, it might be nice to pull these dds into a functional component and pass it the text

sam-keen (approved):

apps/simple-trading-app/src/app/components/deal-ticket/deal-ticket-estimates.tsx
macqbat: yeah, good point, agree, changed :+1:

asiaznik (approved):

Issue #968, opened 2022-08-08
Handle submission failure on IOC/FOK limit orders


MadalinaRaicu:
Created a limit order with IOC time in force on ETHBTC / UNIDAI markets. 

1. In the trading mode is `Continuous`, the order is `stopped` and the dialog pops up with the text `Submission failed` as in this [thread](https://vegaprotocol.slack.com/archives/C037STCKMGC/p1659625539455709).  
![Screenshot 2022-08-10 at 16 18 44](https://user-images.githubusercontent.com/16125548/183945447-5e7f054d-ad9f-4d8a-ae35-e762df9acda7.png)

2. If I try to place the order and the trading mode changes  (during the transition) the transaction gets stucked in pending:  `Awaiting network confirmation` (checking the tx in block explorer does not help, can't verify the status of the tx)
![Screenshot 2022-08-10 at 16 19 27](https://user-images.githubusercontent.com/16125548/183945279-d767cf42-0157-4d14-a59d-8cf1a2daba75.png)

3. If I try to place the same order when it is definitely in trading mode `Monitoring auction - price` I get the `Submission failed` pop up with the rejection reason `Ioc during auction`
![Screenshot 2022-08-10 at 16 20 07](https://user-images.githubusercontent.com/16125548/183945129-5ca548e3-684c-4cdc-8fc3-93a3e802f44f.png)

MadalinaRaicu:
We do show rejection reason in some situations 
<img width="1392" alt="image" src="https://user-images.githubusercontent.com/16125548/183891336-94401db3-6ac7-4bd4-984b-3773a09a68e7.png">

MadalinaRaicu:
Here is the list of orders I tried to place: 
![Screenshot 2022-08-10 at 16 21 05](https://user-images.githubusercontent.com/16125548/183945985-a3a8aa2e-4f87-49ed-a9fd-4e221037b65f.png)

MadalinaRaicu:
IOC or FOK come back immediately stopped, its not a failure, its just that the order wasn’t immediately matched against, and therefore stopped, or it wasn’t immediately filled, and therefore also stopped. But I would expect the order to be cancelled? Maybe we should revisit what can we display to the user?

campbellssource:
FYI here are the Acs for submitting order: https://github.com/vegaprotocol/specs/blob/user-interface-SORD-only/user-interface/6001-SORD-submit_orders.md
[...]

Issue #950, opened 2022-08-04
Edit order should dialog should validate the user's input


campbellssource:

MadalinaRaicu:
the user can input the price only for now. the only validation rule that we already have is that the price is required. is there any minimum or maximum price that the user can input? any upper/lower bounds?

<img width="1392" alt="image" src="https://user-images.githubusercontent.com/16125548/184687519-5b5b870c-6a21-4fbc-95e1-68d85bcd8a96.png">

AndyWhiteVega:
Tested = Looks Good

PR #1037, opened 2022-08-15
fix: order warning messages

### LINKED ISSUE #997: Revisit deal ticket warning messages before placing orders
Revisit the warning message because persistent limit orders are accepted, including GFA, GTT, GTC as well as liquidity commitments. 
Also may be worth including the fact that as the market is suspended, any orders placed now will not trade until the auction ends.

Image

https://vegaprotocol.slack.com/archives/C8WS7M50E/p1660131156926369

MadalinaRaicu:
# Related issues 🔗

Closes #997 & #950 & #968

# Description ℹ️

Revisit the warning message because persistent limit orders are accepted, including GFA, GTT, GTC as well as liquidity commitments.
If the market is suspended, any orders placed now will not trade until the auction ends.

# Demo 📺 & Technical 👨‍🔧
Screenshot 2022-08-15 at 17 09 49
Screenshot 2022-08-15 at 17 10 05
Screenshot 2022-08-15 at 17 10 13
Screenshot 2022-08-15 at 17 10 25
image

Screenshot 2022-08-15 at 19 01 46
Screenshot 2022-08-15 at 19 01 39


libs/orders/src/lib/order-hooks/use-order-validation.tsx
```diff
@@ -88,65 +81,13 @@ export const useOrderValidation = ({
       };
     }
 
-    if (market.state !== MarketState.Active) {
-      if (market.state === MarketState.Suspended) {
-        if (market.tradingMode === MarketTradingMode.Continuous) {
```
MadalinaRaicu, line 93: this was wrong. continuous trading (i.e. not in auction) would keep the market in the active state.

libs/orders/src/lib/components/order-feedback/order-feedback.tsx
```diff
@@ -113,6 +74,27 @@ export const OrderFeedback = ({ transaction, order }: OrderFeedbackProps) => {
...
+const rejectionReason = (
+  order: OrderEvent_busEvents_event_Order
+): string | null => {
+  switch (order.status) {
+    case OrderStatus.Stopped:
+      return t(
+        `The network could not fill the ${order.timeInForce} order and it has been stopped`
+      );
+    case OrderStatus.Rejected:
+      return order.rejectionReason && t(formatLabel(order.rejectionReason));
```
MadalinaRaicu, line 96: reason coming from the backend only when the order is Rejected. we could still add other messages that can appear when the order is Cancelled or Parked.

package.json
```diff
@@ -6,7 +6,8 @@
     "start": "nx serve",
     "build": "nx build",
     "test": "nx test",
-    "postinstall": "husky install && yarn tsc -b tools/executors/next && yarn tsc -b tools/executors/webpack"
+    "postinstall": "husky install && yarn tsc -b tools/executors/next && yarn tsc -b tools/executors/webpack",
+    "test-all": "nx run-many --all --target=test"
```
dexturr, line 10: nitpick: can we make this `test:all` I prefer using colons for commands that have subtle differences

MadalinaRaicu, line 10:
```suggestion
    "test:all": "nx run-many --all --target=test"
```

libs/orders/src/lib/components/order-list/order-edit-dialog.tsx
```diff
@@ -88,15 +88,23 @@ export const OrderEditDialog = ({
         <form onSubmit={handleSubmit(onSubmit)} data-testid="edit-order">
           <FormGroup label={t('Entry price')} labelFor="entryPrice">
             <Input
```
notbot00, line 90: question (non-blocking): not sure if we need to / want to add the market decimalPlaces so that the input `step` can reflect the minimum unit of change

libs/orders/src/lib/components/order-feedback/order-feedback.tsx
```diff
@@ -113,6 +74,27 @@ export const OrderFeedback = ({ transaction, order }: OrderFeedbackProps) => {
...
+        <div>
+          <p className={labelClass}>{t(`Reason`)}</p>
+          <p data-testid="error-reason">{t(orderRejectionReason)}</p>
+        </div>
+      )}
     </div>
   );
 };
+
+const rejectionReason = (
```
notbot00, line 87: nit (non-blocking): i find it a lot more readable when looking at a variable and being able to assume it's a function. usually prefixing it with a verb makes it obvious, like `getRejectionReason`. no big deal tho, pls free free to resolve this without action

libs/orders/src/lib/components/order-feedback/order-feedback.tsx
```diff
@@ -113,6 +74,27 @@ export const OrderFeedback = ({ transaction, order }: OrderFeedbackProps) => {
...
+  order: OrderEvent_busEvents_event_Order
+): string | null => {
+  switch (order.status) {
+    case OrderStatus.Stopped:
+      return t(
+        `The network could not fill the ${order.timeInForce} order and it has been stopped`
```
candida-d, line 93: suggestion (non-blocking): What about the below, which I think sounds friendlier. 

```suggestion
        `Your ${order.timeInForce} order could not be filled and it has been stopped`
```

libs/orders/src/lib/components/order-feedback/order-feedback.tsx
MadalinaRaicu: updated

libs/orders/src/lib/components/order-list/order-edit-dialog.tsx
MadalinaRaicu: this is how it looks like after i pulled latest changes on master

notbot00: this looks more like a side-effect of changing the input type from "text" to "number". number types need a "step" attribute in order to accept decimal values

MadalinaRaicu: updated

libs/orders/src/lib/components/order-list/order-edit-dialog.tsx
```diff
@@ -88,15 +87,24 @@ export const OrderEditDialog = ({
...
-              {...register('entryPrice', { required: t('Required') })}
-              id="entryPrice"
               type="text"
+              step={step}
```
notbot00, line 91: chore (blocking): so, inputs with type="text" cannot have steps, only input type="number". if you make it a text input, the step attribute becomes redundant.

libs/react-helpers/src/lib/format/number.ts
```diff
@@ -4,7 +4,7 @@ import memoize from 'lodash/memoize';
 import { getUserLocale } from './utils';
 
 export function toDecimal(numberOfDecimals: number) {
-  return Math.pow(10, -numberOfDecimals);
```
MadalinaRaicu, line 7: this never worked. because javascript is javascript 
Screenshot 2022-08-16 at 18 36 02

libs/orders/src/lib/components/order-feedback/order-feedback.tsx
```diff
@@ -113,6 +74,27 @@ export const OrderFeedback = ({ transaction, order }: OrderFeedbackProps) => {
...
   );
 };
+
+const getRejectionReason = (
+  order: OrderEvent_busEvents_event_Order
+): string | null => {
+  switch (order.status) {
+    case OrderStatus.Stopped:
+      return t(
+        `Your ${order.timeInForce} order was not filled and it has been stopped`
```
MadalinaRaicu, line 93: This will be revisited as soon as we get a `reason` behind why the order has been stopped by Vega. @campbellssource

libs/orders/src/lib/components/order-list/order-edit-dialog.tsx
```diff
@@ -37,15 +37,15 @@ export const OrderEditDialog = ({
...
-  });
+    setValue,
+  } = useForm<FormFields>();
 
   if (!order) return null;
+  const step = toDecimal(order.market?.decimalPlaces ?? 0);
+  setValue(
+    'entryPrice',
+    addDecimalsFormatNumber(order.price, order.market?.decimalPlaces ?? 0)
+  );
```
notbot00, line 48: chore (blocking?): sorry for holding this up again, i think it would feel a lot less complicated to render this dialog from the parent with a required order object, and leave the defaultValues as it was, so you don't need to return null from the component, and don't need to forcefully reset the field value on every render.

```
{order && < OrderEditDialog  order={order} />}
```

libs/orders/src/lib/components/order-feedback/order-feedback.tsx
```diff
@@ -18,46 +18,7 @@ export const OrderFeedback = ({ transaction, order }: OrderFeedbackProps) => {
...
-              <p className={labelClass}>{t(`Market`)}</p>
-              <p>{t(`${order.market.name}`)}</p>
-            </div>
-          )}
-        </div>
-        <div>
-          {transaction.txHash && (
-            <div>
-              <p className={labelClass}>{t('Transaction')}</p>
-              <a
```
mattrussell36, line 46: question (non blocking): Are we fine not showing the vega tx hash?

libs/orders/src/lib/components/order-list/order-edit-dialog.tsx
mattrussell36: Yup in React if you are setting state in render you are 100% doing something wrong.

libs/orders/src/lib/components/order-list/order-edit-dialog.tsx
```diff
@@ -88,15 +88,24 @@ export const OrderEditDialog = ({
...
+              })}
               id="entryPrice"
-              type="text"
             />
             {errors.entryPrice?.message && (
               <InputError intent="danger" className="mt-4">
                 {errors.entryPrice.message}
               </InputError>
             )}
+            {errors.entryPrice?.type === 'min' && (
```
mattrussell36, line 104: chore (blocking): You should be able to just render the message once here provided you have set up the validation using the register function. Check out the react-hook-form docs..

You should be able to do something like
```
register('entryPrice', {
  validation: {
    min: (value) => {
      if (value is valid) return true
      return 'The price cannot be negative'
    } 
  }
})
```

libs/orders/src/lib/order-hooks/use-order-validation.tsx
```diff
@@ -186,14 +128,17 @@ export const useOrderValidation = ({
       };
     }
 
-    if (fieldErrors?.price?.type === 'required') {
+    if (
+      fieldErrors?.price?.type === 'required' &&
```
mattrussell36, line 132: chore (blocking): You shouldn't have to conditionally change the error message by `type` if using the validate property in the `register` function of the field.

mattrussell36 (changes requested):

libs/orders/src/lib/components/order-feedback/order-feedback.tsx
MadalinaRaicu: we are showing the tx hash on line 72

MadalinaRaicu: cancelled/rejected vs confirmed order were too similar - we would display the same component for both, the title would change based on the order status and if the order is rejected/stopped then we would display a reason

libs/orders/src/lib/components/order-list/order-edit-dialog.tsx
MadalinaRaicu: yeah, my bad. updated.

libs/react-helpers/src/lib/format/number.ts
```diff
@@ -4,7 +4,7 @@ import memoize from 'lodash/memoize';
...
+  return 1 / Math.pow(10, numberOfDecimals);
```
MadalinaRaicu, line 7: and this works till `numberOfDecimals === 6`

libs/orders/src/lib/components/order-list/order-edit-dialog.tsx
MadalinaRaicu: updated

libs/orders/src/lib/order-hooks/use-order-validation.tsx
MadalinaRaicu: the reason why tried to handle most the of the warning messages in 'useOrderValidation' hook is because I wanted to handle better the order in which these messages appear and to do that in one place. 
 
 i.e. here,  if I move these rules to validate, works well if the market is trading continuously but if it is in a monitoring auction then the 'You cannot set a negative price` error gets overridden by `Any orders traded now will not trade until the execution ends`
 
 
image

MadalinaRaicu: I can handle this only in register, if I add the market trading mode as a field in deal ticket limit amount where we register the form and move more rules over there, but is it better?

mattrussell36: No I think this is fine. Its out side the component itself so I think here its better to use the fieldErrors object.

mattrussell36 (approved):

AndyWhiteVega:
Tested each = Looks good

PR #998, opened 2022-08-10
Feat/537: remove old contract for collateral bridge

sam-keen:
# Related issues 🔗

Closes #537

# Description ℹ️

Removes the contract, ABI, files, functions, types and logic needed to run the old mainnet collateral bridge

PR #1045, opened 2022-08-15
Revert "Feat/537: remove old contract for collateral bridge"

edd:
Reverts vegaprotocol/frontend-monorepo#998


MadalinaRaicu (approved):

edd:
This looked like it was required, but wasn't in the end. The deploy was locked to an old deploy and the nameless person trying to deploy a new version didn't noticed. Moving the deploy lock to the latest version made things work as expected.

PR #1047, opened 2022-08-16
Feature/993 derived data provider

### LINKED ISSUE #993: Dependencies handling in data provider
Follow up to #473

Positions table shows metrics which uses party accounts, margins and positions data as arguments.

Multiple source subscription is not supported, so there is a need to create dependencies for data provider which will handle additional subscriptions.

For positions-metrics it should depend on accounts and margins data providers. Or second option positions-margins depend on positions, accounts and margins, without running it's own query

bglownia:
# Related issues 🔗

Closes #993

# Description ℹ️

In trading app positions tab there is need to connect two data sources: party positions and party accounts. While it's possible to run single query to fetch all data at once, it's not possible to subscribe to updates on two kind of objects within single subscription. Derived data provider address this issue. It allows to combine two or more data providers, and deliver data derived from multiple data sources. It uses same interface as data provider.


libs/react-helpers/src/lib/generic-data-provider.ts
```diff
@@ -433,29 +451,137 @@ const memoize = <Data, Delta>(
...
+export type CombineDerivedData<Data> = (
+  parts: Parameters<DependencyUpdateCallback>['0']['data'][]
+) => Data | null;
+
+function makeDerivedDataProviderInternal<Data>(
+  dependencies: DependencySubscribe[],
+  combineData: CombineDerivedData<Data>
+): Subscribe<Data, never> {
+  let subscriptions: ReturnType<DependencySubscribe>[] | undefined;
+  let client: ApolloClient<object> | undefined;
```
asiaznik, line 478: question: does it have to be optional?

libs/react-helpers/src/lib/generic-data-provider.ts
```diff
@@ -433,29 +451,137 @@ const memoize = <Data, Delta>(
...
+        newError = newError || (part && part.error);
+        newLoading = newLoading || !part || part.loading;
+        newLoaded = newLoaded && part && part.loaded;
+      });
+
+    const newData = newLoaded
+      ? combineData(parts.map((part) => part.data))
+      : data;
+    if (
+      newLoading !== loading ||
```
asiaznik, line 520: suggestion (cosmetic): I'd write the conditions and assignments in the same order for better readability

libs/react-helpers/src/lib/generic-data-provider.ts
```diff
@@ -433,29 +451,137 @@ const memoize = <Data, Delta>(
...
+    if (subscriptions || !client) {
+      return;
+    }
+    subscriptions = dependencies.map((dependency, i) =>
+      dependency(
+        (updateData) => {
+          parts[i] = updateData;
+          combine();
+        },
+        client as ApolloClient<object>, // for some reason TS doesn't narrow type base on !client condition
```
asiaznik, line 543: answering the comment: I believe it's because client is optional (AppolloClient | undefined)

asiaznik (approved):

asiaznik (approved):

mattrussell36 (approved):

MadalinaRaicu (approved):

MadalinaRaicu:
can't wait to see this merged and start using it!

PR #1061, opened 2022-08-17
Feat/1033: Market info tooltips

### LINKED ISSUE #1033: Add tooltips to market info
I suspect I'm not the only one who doesn't know what all the data points under Market Info refer to, though there's normally already an explanation in the docs. Makes sense to add tooltips for these in Console.

notbot00:
# Related issues 🔗

Closes #1033 

# Description ℹ️

Add a tooltip to some labels based on the parent issue.

# Demo 📺

Screenshot 2022-08-17 at 12 55 38


libs/deal-ticket/src/components/info-market.tsx
```diff
@@ -371,7 +474,9 @@ const Row = ({
         dtClassName={className}
         ddClassName={className}
       >
-        {startCase(t(field))}
+        <Tooltip description={tooltipMapping[field]} align="start">
```
dexturr, line 477: nitpick (non-blocking): may be nice to have a fallback here if the field is not defined.

dexturr (approved):

libs/deal-ticket/src/components/info-market.tsx
```diff
@@ -371,7 +474,9 @@ const Row = ({
...
+          <div>{startCase(t(field))}</div>
```
macqbat, line 478: nitpick (non-blocking): you can make this div focusable by adding `tabIndex={-1}`. This will allow to be seen a tooltip content on mobile (when no mouse events) after tap on it.

macqbat (approved):

libs/deal-ticket/src/components/info-market.tsx
notbot00: i think it will just not display a tooltip content in that case, because `tooltipMapping[undefined]` is also going to be `undefined`

libs/deal-ticket/src/components/info-market.tsx
notbot00: nice, added

libs/deal-ticket/src/components/info-market.tsx
```diff
@@ -343,6 +346,106 @@ export const Info = ({ market }: InfoProps) => {
...
+    'The aggregated volume being bid at the best bid price on the market.'
+  ),
+  bestOfferVolume: t(
+    'The aggregated volume being offered at the best offer price on the market.'
+  ),
+  bestStaticBidVolume: t(
+    'The aggregated volume being bid at the best static bid price on the market.'
+  ),
+  bestStaticOfferVolume: t(
+    'The aggregated volume being bid at the best static offer price on the market.'
```
candida-d, line 379:
```suggestion
    'The aggregated volume being offered at the best static offer price on the market.'
```

libs/deal-ticket/src/components/info-market.tsx
```diff
@@ -343,6 +346,106 @@ export const Info = ({ market }: InfoProps) => {
...
+  ),
+
+  decimalPlaces: t('The smallest price increment on the book.'),
+  positionDecimalPlaces: t(
+    'How big the smallest order / position on the market can be.'
+  ),
+  tradingMode: t('The trading mode the market is currently running.'),
+  state: t('The current state of the Market'),
```
candida-d, line 387:
```suggestion
  state: t('The current state of the market'),
```

libs/deal-ticket/src/components/info-market.tsx
```diff
@@ -343,6 +346,106 @@ export const Info = ({ market }: InfoProps) => {
...
+    'The first currency in a pair for a currency-based derivatives market.'
+  ),
+  quote: t(
+    'The second currency in a pair for a currency-based derivatives market.'
+  ),
+  class: t(
+    'The classification of the product. Examples: shares, commodities, crypto, FX.'
+  ),
+  sector: t(
+    'Data about the sector Example: "automotive" for a market based on value of Tesla shares.'
```
candida-d, line 399:
```suggestion
    'Data about the sector. Example: "automotive" for a market based on value of Tesla shares.'
```

PR #1065, opened 2022-08-18
feat: [console-lite] - filter markets on market selector

### LINKED ISSUE #1005: Filter for active markets in market selector

macqbat:
# Related issues 🔗

Closes #1005

# Description ℹ️

Add checking state of the market on the results list

# Technical 👨‍🔧

Only active markets are allowed to appear on the result list of the selector

### DIFF
--- a/libs/deal-ticket/src/components/market-selector.tsx
+++ b/libs/deal-ticket/src/components/market-selector.tsx
@@ -32,6 +32,7 @@ export const MARKET_NAMES_QUERY = gql`
     markets {
       id
       name
+      state
       tradableInstrument {
         instrument {
           code
@@ -249,8 +250,10 @@ export const MarketSelector = ({ market, setMarket, ItemRenderer }: Props) => {
 
   useEffect(() => {
     setResults(
-      data?.markets?.filter((item: MarketNames_markets) =>
-        item.name.match(new RegExp(escapeRegExp(lookup), 'i'))
+      data?.markets?.filter(
+        (item: MarketNames_markets) =>
+          item.state === 'Active' &&
+          item.name.match(new RegExp(escapeRegExp(lookup), 'i'))
       ) || []
     );
   }, [data, lookup]);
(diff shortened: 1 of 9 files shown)


libs/deal-ticket/src/components/market-selector.tsx
```diff
@@ -249,8 +250,10 @@ export const MarketSelector = ({ market, setMarket, ItemRenderer }: Props) => {
 
   useEffect(() => {
     setResults(
-      data?.markets?.filter((item: MarketNames_markets) =>
-        item.name.match(new RegExp(escapeRegExp(lookup), 'i'))
+      data?.markets?.filter(
```
MadalinaRaicu, line 253: i did this filtering once and i was wrong. in console v2 you can place limit orders on suspended (not active - continuously trading) markets that are in auction. there were times in testnet when all markets were suspended (in auction). 
if there are no active markets what is the plan to display in the market selector for console-lite?

MadalinaRaicu: what we ended up doing was to filter the markets based on the MarketTradingMode not MarketState

libs/deal-ticket/src/components/market-selector.tsx
```diff
@@ -249,8 +250,10 @@ export const MarketSelector = ({ market, setMarket, ItemRenderer }: Props) => {
...
+        (item: MarketNames_markets) =>
+          item.state === 'Active' &&
```
MadalinaRaicu, line 255: this is our filter `item.tradingMode !== MarketTradingMode.NoTrading`

libs/deal-ticket/src/components/market-selector.tsx
macqbat: Currently we don't have such scenario possible in console lite app. We trade only active markets afaik. Component is displayed in view which is possible to reach for the user only if there is at least one active market. There wasn't such view in design, i could refer to. As component is placed in common library, I could invented something for such case, but from the other hand, I think it would be better to do it based on real scenario, if there were one in the future.

MadalinaRaicu: https://github.com/vegaprotocol/frontend-monorepo/issues/656 check this out. I am working on order warning at the moment changing how `useOrderValidation` hook behaves. you guys are using it too.

madalina237 (approved):

MadalinaRaicu (approved):

asiaznik (approved):

PR #1073, opened 2022-08-18
Feat/722: Network switcher in console2

### LINKED ISSUE #722: Switch to another network from Console
Console should have a way of getting from the network you are on to another network.

for example: you've gone to console.fairground.wtf but now you want to jump to mainnet. There should be some UI that allows you to jump to another network. This to be done by jumping to another URL (as apposed to connecting to a different data node).

Note: the [Node switcher](https://github.com/vegaprotocol/frontend-monorepo/issues/622) exists and should be thought about in context of this.

---

e.g. This is a very rough sketch and needs to be developed, and alternatives considered...

Image

notbot00:
# Related issues 🔗

Closes #722 

# Description ℹ️

Adds a network switcher to the trading app. Also added stagnet3 support to the environment lib.

# Demo 📺

Screenshot 2022-08-18 at 12 53 58


apps/trading/components/navbar/navbar.tsx
```diff
@@ -1,21 +1,25 @@
...
-      <Link href="/" passHref={true}>
-        {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
-        <a className="px-[26px]">
-          <Vega className="fill-white" />
-        </a>
-      </Link>
+      <div className="flex items-center h-full">
+        <Link href="/" passHref={true}>
+          {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
+          <a className="px-[26px]">
```
sam-keen, line 17: Suggestion (non-blocking): absolutely fine if 26px exactly is needed, but if not, you could use either 24px or 28px from the theme

libs/environment/src/components/network-switcher/network-switcher.tsx
```diff
@@ -0,0 +1,177 @@
...
+export const envTriggerMapping: Record<Networks, string> = {
+  ...envNameMapping,
+  [Networks.TESTNET]: t('Fairground'),
+};
+
+export const envDescriptionMapping: Record<Networks, string> = {
+  [Networks.CUSTOM]: '',
+  [Networks.DEVNET]: t('The latest Vega code auto-deployed'),
+  [Networks.STAGNET]: t('A staging environment with trading'),
+  [Networks.STAGNET2]: t('A staging envirnment without trading'),
```
sam-keen, line 34:
```suggestion
  [Networks.STAGNET2]: t('A staging environment without trading'),
```

libs/environment/src/components/network-switcher/network-switcher.tsx
```diff
@@ -0,0 +1,177 @@
...
+  isCurrent: boolean;
+  isAvailable: boolean;
+};
+
+const NetworkLabel = ({
+  isCurrent = false,
+  isAvailable = false,
+}: NetworkLabelProps) => (
+  <span className="text-white-80 dark:text-white-80">
+    {isCurrent
```
sam-keen, line 64: Suggestion (non-blocking): I think nested ternaries are a bit unpleasant to read. I have a small preference for this render logic to sit outside JSX and be done with if/else. Not a strong objection though :)

libs/environment/src/components/network-switcher/network-switcher.tsx
```diff
@@ -0,0 +1,177 @@
...
+    (isOpen: boolean) => {
+      setOpen(isOpen);
+      if (!isOpen) {
+        setAdvancedView(false);
+      }
+    },
+    [setOpen, setAdvancedView]
+  );
+
+  const menuItemClasses = 'pt-12 pb-12 pl-16 pr-16 h-auto';
```
sam-keen, line 87:
```suggestion
  const menuItemClasses = 'py-12 px-16 h-auto';
```

sam-keen (approved):
Couple of minor suggestions, looks good

mattrussell36 (approved):

libs/environment/src/components/network-switcher/network-switcher.tsx
notbot00: this doesn't work, as the dropdown classes override them, that's why i had to hack it in this way. or i could also potentially just not use the dropdown item

libs/environment/src/components/network-switcher/network-switcher.tsx
notbot00: this got removed due to merging master

libs/environment/src/components/network-switcher/network-switcher.tsx
notbot00: sure, refactored it a little so i can use if blocks

apps/trading/components/navbar/navbar.tsx
notbot00: absolutely, changed this to 24

mattrussell36: I think Sam meant to use `p-24` which is set up in our spacing definitions in the theme

apps/trading/components/navbar/navbar.tsx
```diff
@@ -1,21 +1,25 @@
...
-      <Link href="/" passHref={true}>
-        {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
-        <a className="px-[26px]">
-          <Vega className="fill-white" />
-        </a>
-      </Link>
+      <div className="flex items-center h-full">
+        <Link href="/" passHref={true}>
+          {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
+          <a className="px-[24px]">
```
notbot00, line 17:
```suggestion
          <a className="px-24">
```

apps/trading/components/navbar/navbar.tsx
notbot00: oh, i see, i didn't get that part. applied another fix.

PR #1075, opened 2022-08-18
fix: dissasociation without vega wallet (806)

### LINKED ISSUE #806: When visiting disassociate for vesting contract without vega wallet
From:
- https://github.com/vegaprotocol/token-frontend/issues/1249

when visiting disassociate for vesting contract without vega wallet
I see the input allowing me to attempt to disassociate tokens
Even though I do not need to be connected to my vega wallet to do this

We should:

- [ ] Remove the requirement to connect a vega wallet to dissassociate token form. (VegaWalletContainer)
- [ ] Add a select for all keys that the currently connected ethereum wallet has staked to. This state already exists and can be seen under "Associated with Vega Keys" in the Eth wallet card on the RHS].
- [ ] This select should be disabled if only a single key has associated vega. This single key should be selected.
- [ ] A warning message should be shown if the connected eth wallet does not have any vega associated
- [ ] Allow the user to perform a disassociate action, using whichever Vega key the user has connected

image

asiaznik:
# Related issues 🔗

Closes #806

# Description ℹ️

Vega wallet is no longer needed for disassociation of vega tokens. It is possible to disassoc any token via the select element where the options represent available associations.

# Demo 📺

1 key available
Screenshot 2022-08-18 at 23 28 42

2+ keys available
Screenshot 2022-08-18 at 23 30 57

# Technical 👨‍🔧

* It uses `associationBreakdown` from `appState` to determine what's available.
* I've removed `contract-disassociate` and `wallet-disassociate` as the distinction is no longer needed
* A separate page for disassociations without vega wallet is removed - no needed anymore


apps/token/src/lib/format-number.ts
```diff
@@ -1,13 +1,33 @@
...
+
+export const toNumberParts = (
+  value: BigNumber | null | undefined,
+  decimals = 18
+): [integers: string, decimalPlaces: string] => {
+  if (!value) {
+    return ['0', '0'.repeat(decimals)];
+  }
+  // @ts-ignore confident not undefined
+  const separator = BigNumber.config().FORMAT.decimalSeparator as string;
```
asiaznik, line 21: follow up: #1079

mattrussell36 (approved):

apps/token/src/routes/staking/disassociate/disassociate-page.tsx
```diff
@@ -1,65 +1,147 @@
...
+        stakingMethod={payload.stakingMethod}
         dispatch={txDispatch}
       />
     );
   }
 
+  const noKeysMessage = (
+    <div className="disassociate-page__error">
+      {t(
+        'You have no VEGA tokens currently staked through your connected Eth wallet.'
```
candida-d, line 101:
```suggestion
        'You have no VEGA tokens currently associated through your connected Ethereum wallet.'
```

apps/token/src/routes/staking/disassociate/disassociate-page.tsx
```diff
@@ -73,27 +155,9 @@ export const DisassociatePage = ({
           'Any Tokens that have been nominated to a node will sacrifice any Rewards they are due for the current epoch. If you do not wish to sacrifices fees you should remove stake from a node at the end of an epoch before disassocation.'
```
candida-d, line 155:
```suggestion
          'Any tokens that have been nominated to a node will sacrifice rewards they are due for the current epoch. If you do not wish to sacrifice these, you should remove stake from a node at the end of an epoch before disassociation.'
```

MadalinaRaicu (approved):

PR #1094, opened 2022-08-19
Feat/637 Market settlement expiry

### LINKED ISSUE #637: Show show settlement data (if present) in market list
"Future" markets on Vega don't always have a fixed settlement date.
The event that settles a market comes from an oracle and may not come at a particular set time.
However, many markets will have an intended settlement time, and where they do we should show it.
There are two ways

notbot00:
# Related issues 🔗

Closes #637 

# Description ℹ️

Displays an expiry for the market, if available, exposed by the `marketTimestamps` field. If not, it shows a tooltip linking to the oracle in the `oracleSpecForTradingTermination` field. 

# Demo 📺

Screenshot 2022-08-22 at 11 25 42


apps/explorer/src/app/routes/oracles/index.tsx
```diff
@@ -33,15 +34,28 @@ const ORACLE_SPECS_QUERY = gql`
 `;
 
 const Oracles = () => {
-  const { data } = useQuery<OracleSpecsQuery>(ORACLE_SPECS_QUERY);
+  const { hash } = useLocation();
+  const { data, loading } = useQuery<OracleSpecsQuery>(ORACLE_SPECS_QUERY);
+
+  useEffect(() => {
```
mattrussell36, line 40: question (blocking): Annoying that we need this effect. Does linking with a hash not work when using react-router-dom?

apps/trading/pages/markets/trade-grid.tsx
```diff
@@ -53,6 +55,56 @@ const TradingViews = {
 
 type TradingView = keyof typeof TradingViews;
 
+type GetExpiryProps = Pick<TradeMarketHeaderProps, 'market'> & {
```
mattrussell36, line 58: suggestion (non blocking): not sure about the naming and types here. I'd normally reserve naming something XProps for components only. Also I find using `Pick` a bit overkill and hard to read. Additionally would it not be helpful to make getExpiryProps a hook so its visible in dev tools?

I think this would be easier to read
```
const useExpiryProps = ({ market, explorerUrl }: {
  market: TradeMarketHeaderProps['market']
  explorerUrl?: string
}) => { // Note let TS infer return type
  ...
}
```

apps/explorer/src/app/routes/oracles/index.tsx
notbot00: it does, but only when the data is already on the page. when we have a loading state, the router doesn't automatically recheck the contents, so i have to tell it to do so

apps/trading/pages/markets/trade-grid.tsx
notbot00: haven't thought about this fn as a hook, but yeah, totally right, where it's called makes it one. since i'm not a huge fan of returning react components / elements from random functions, did a bit of refactor so make these items individual react components. let me know if this works.

sam-keen (approved):

MadalinaRaicu (approved):

PR #1096, opened 2022-08-19
Feat/375: Risk notice dialog

### LINKED ISSUE #375: Notice about the risks of Trading on Start up (and legal warning)
From: https://github.com/vegaprotocol/console/issues/6664

Derivatives should be used only by appropriate persons. We'd like to ensure that only appropriate persons use the software we create. This ticket to design and implements some sort of message in Console to inform potential users, and potentially discourage non-approporate people from using it

- [ ] Read some of the rationale behind regulations (and talk to legal)
- [ ] draft a message
- [ ] Have legal team review / iterate the message
- [ ] Design for how and where the message appears in console
- [ ] Build it (via UI-toolkit?) into console

Also required:
- [ ] hide this information in testnet (check with Dave Mac - is there other info that could/should be testnet copy)
- [ ] Consider how this interacts with the landing dialog

-----

Reading:
https://3commas.io/blog/the-risks-of-margin-trading-and-how-to-do-it-properly from a trading client (not an exchange) for its users about the risks of trading
https://www.fca.org.uk/news/press-releases/fca-warns-younger-investors-are-taking-big-financial-risks
> The FCA advises consumers to consider five important questions before they invest: 
> Am I comfortable with the level of risk?
> Do I fully understand the investment being offered to me?
> Am I protected if things go wrong?
> Are my investments regulated?
> Should I get financial advice?

IG and CFD providers are required to have something like this on their site:
> Spread bets and CFDs are complex instruments and come with a high risk of losing money rapidly due to leverage. 68% of retail investor accounts lose money when trading spread bets and CFDs with this provider. You should consider whether you understand how spread bets and CFDs work, and whether you can afford to take the high risk of losing your money.

https://www.sec.gov/news/speech/mjw-speech-032114-protecting-retail-investor

----

additional requirement:

make it clear that this is alpha software

----

https://uniswap.org/terms-of-service

notbot00:
# Related issues 🔗

Closes #375 

# Description ℹ️

Display a risk notice on mainnet. Also has some dialog fixes, as the transform styles completely mess up the mobile display for the dialogs

# Demo 📺

Screenshot 2022-08-19 at 17 23 57


mattrussell36 (changes requested):
There are some scroll bars appearing unnecessarily

Screen Shot 2022-08-23 at 3 19 30 PM

notbot00:
thanks, fixed

MadalinaRaicu (commented):
(blocking):  the font size on this one is quite big and different from what we are using around console. I have seen this also in tooltips and environment popover/dropdowns. can we make sure we are using `text-ui` or `text-ui-small` across console?

PeteWilliams:
I notice that this has the close icon in the corner. This could potentially be a legal issue if people can proceed without acknowledging the warning. I'd suggest removing the icon or checking with Dave Mac if we are OK to keep it.

apps/trading/pages/_app.page.tsx
```diff
@@ -7,27 +8,49 @@ import {
...
   const store = useGlobalStore();
+  const { VEGA_ENV } = useEnvironment();
   const {
     isAssetDetailsDialogOpen,
     assetDetailsDialogSymbol,
     setAssetDetailsDialogOpen,
   } = useAssetDetailsDialogStore();
   const [theme, toggleTheme] = useThemeSwitcher();
 
+  useEffect(() => {
```
dexturr, line 41: chore (blocking): can we add some tests for this as well please?

notbot00:
> (blocking): the font size on this one is quite big and different from what we are using around console. I have seen this also in tooltips and environment popover/dropdowns. can we make sure we are using `text-ui` or `text-ui-small` across console?

@MadalinaRaicu i've added text-ui classes, nothing happened. not sure tbh how our typography works, i don't think it does how i'd expect it to 🤔

notbot00:
> I notice that this has the close icon in the corner. This could potentially be a legal issue if people can proceed without acknowledging the warning. I'd suggest removing the icon or checking with Dave Mac if we are OK to keep it.

@PeteWilliams makes sense, removed the closing option

MadalinaRaicu:
> > (blocking): the font size on this one is quite big and different from what we are using around console. I have seen this also in tooltips and environment popover/dropdowns. can we make sure we are using `text-ui` or `text-ui-small` across console?
> 
> @MadalinaRaicu i've added text-ui classes, nothing happened. not sure tbh how our typography works, i don't think it does how i'd expect it to 🤔

maybe with the new design changes it will appear differently. not sure my comment applies anymore.

apps/trading/pages/_app.page.tsx
notbot00: totally hear you.  i gave this a go, the amount of things i need to wire in just to display a page is pretty ridiculous. not sure if this simple if statement is worth the mock party.

mattrussell36: @notbot00 how about moving the useEffect and required imports to the RiskNoticeDialog itself? It should be much easier to unit test then and keeps _app.page.tsx cleaner

notbot00: i really felt repelled by the idea at first, no idea why, my my OCD with the consistency with the other app dialogs. this is actually a super good solution i think, thanks @mattrussell36 🙏  added some tests finally

asiaznik (approved):

PR #1097, opened 2022-08-22
feat: [console-lite] - a bunch of improvements

### LINKED ISSUE #1070: [console-lite] clean up app repo, fixes and improvements
## Task
Find redundant elements and remove or refactor, clean up the application repo

## ACs
- [ ] remove  old drawer component
- [x] make market list as home page
- [ ] increase code test coverage

macqbat:
# Related issues 🔗

Closes #1070

# Description ℹ️

Add home page and some other bunch of improvements

# Technical 👨‍🔧

Move some general states into local value context, add missing home page route


apps/simple-trading-app/src/app/app.tsx
```diff
@@ -18,17 +18,17 @@ import LocalContext from './context/local-context';
...
   const [theme, toggleTheme] = useThemeSwitcher();
-
-  const [menuOpen, setMenuOpen] = useState(false);
-  const onToggle = () => setMenuOpen(!menuOpen);
-
+  const localValues = useLocalValues(toggleTheme);
+  const {
+    vegaWalletDialog,
+    menu: { setMenuOpen },
+  } = localValues;
```
notbot00, line 26:
```suggestion
  const {
    vegaWalletDialog,
    menu: { setMenuOpen },
  } = useLocalValues(toggleTheme);
```

apps/simple-trading-app/src/app/components/nav/nav.tsx
```diff
@@ -11,20 +11,22 @@ interface NavProps {
...
-            `text-h5 block md:mb-40 px-40 md:text-black md:dark:text-white ${
-              isActive && 'text-white md:text-blue md:dark:text-blue'
-            }`
-          }
-        >
-          <NavItem iconName={r.icon} label={r.text} />
-        </NavLink>
-      ))}
+      {routerConfig.map((r) =>
+        r.icon ? (
```
notbot00, line 15: question (blocking?): this reads really strange. if icon is required to render nav links, why not make it a non-optional property? not sure if it makes sense to have route items with missing icons, if they don't render, or is there a specific use-case for those? if that is the case, i'd add a new prop on the route item so it's understandable why they are not rendering, like `isIncludedInNavigation`, or something similar.

apps/simple-trading-app/src/app/app.tsx
macqbat: I need there `localValues` as well

apps/simple-trading-app/src/app/components/nav/nav.tsx
macqbat: home page (with no icon) doesn't need to be displayed in the 'iconic' mobile menu. I wouldn't create new special property for matching it. Come on, it's trivial, clear case, understandable at a glance

notbot00: i understand, what i'm saying is rendering something on the condition whether it has an icon or not doesn't seem to me like the correct abstraction, it works in this case because of a coincidence, and makes future refactor more difficult. i'm happy with it if people agree to do it this way, feel free to resolve this and merge

macqbat: ok, now it's improved according to your indications above

apps/simple-trading-app/src/app/app.tsx
notbot00: missed that

notbot00 (approved):

PR #1098, opened 2022-08-22
Feat/1093: Visual tweaks to the validators form

### LINKED ISSUE #1093: Visual tweaks to validators form
- [ ] Columns to be more tightly spaced
- [ ] Scrollbar to be more visible

sam-keen:
# Related issues 🔗

Closes #1093

# Description ℹ️

Scrollbar more visible
Columns more condensed

### DIFF
--- a/apps/token/src/routes/staking/node-list.tsx
+++ b/apps/token/src/routes/staking/node-list.tsx
@@ -81,7 +81,7 @@ const ValidatorRenderer = ({ data }: ValidatorRendererProps) => {
 // AG Grid places the scrollbar over the bottom validator, which obstructs
 const nodeListGridStyles = `
   .ag-theme-balham-dark .ag-body-horizontal-scroll {
-    opacity: 0.25;
+    opacity: 0.75;
   }
 `;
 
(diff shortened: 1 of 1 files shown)


apps/token/src/routes/staking/node-list.tsx
```diff
@@ -251,11 +260,13 @@ export const NodeList = ({ epoch }: NodeListProps) => {
   return (
     <AsyncRenderer loading={loading} error={error} data={nodes}>
       {epoch && epoch.timestamps.start && epoch.timestamps.expiry && (
-        <EpochCountdown
-          id={epoch.id}
-          startDate={new Date(epoch.timestamps.start)}
-          endDate={new Date(epoch.timestamps.expiry)}
-        />
+        <div className="mb-20">
```
dexturr, line 263: question (non-blocking): could we move this to the root div of the epoch countdown? Does this look reasonable in other places it is rendered?

MadalinaRaicu (approved):

apps/token/src/routes/staking/node-list.tsx
sam-keen: Where it's used elsewhere it's in its own page section with a bit more space around it, but here I don't want quite as much distance, because the epoch data and the validators involved in the epoch are coupled together in a page section, whereas elsewhere the epoch countdown is separated in its own section. I think it's appropriate to have custom spacing with this wrapper div in this instance.

PR #1101, opened 2022-08-22
Feature/840 update positions tab

### LINKED ISSUE #840: Portfolio: Update Positions tab
### Current implementation:

image

### Proposed design:
https://www.figma.com/file/K49Nx7XfocShhHcwwQzQkm/Console-V2?node-id=1616%3A25247

image

### Details

- Group positions by settlement asset (in alphabetical order)
- Show balance below settlement asset heading
- Show amount and notional value (in one cell if possible)
- Add liquidation price
- Add Entry Price -> Liquidation price visualisation
- Show leverage amount
- Show margin allocated
- Show unrealised profit
- Show last traded date
- Add explanation tooltips for liquidation price, realised/unrealised profit and loss (@candida-d feel free to suggest changes to these)
- Add total field for amount, realised/unrealised profit and loss

bglownia:
# Related issues 🔗

Closes #840

# Description ℹ️

Another set of changes in positions tab. Positions are grouped by asset, summary row is added and close position button are added.


libs/positions/src/lib/positions-table.tsx
```diff
@@ -43,12 +51,15 @@ export interface MarketNameCellProps {
...
 export const MarketNameCell = ({ valueFormatted }: MarketNameCellProps) => {
-  return valueFormatted ? (
-    <div className="leading-tight">
-      <div>{valueFormatted[0]}</div>
-      {valueFormatted[1] ? <div>{valueFormatted[1]}</div> : null}
-    </div>
-  ) : null;
+  if (valueFormatted && valueFormatted[1]) {
+    return (
+      <div className="leading-tight">
```
mattrussell36, line 56: chore (non blocking): I could see us needing a cell wrapper component to manage styles for cells with two rows. Perhaps create this and add it to react-helpers?

libs/positions/src/lib/positions-table.tsx
```diff
@@ -113,234 +124,299 @@ export const AmountCell = ({ valueFormatted }: AmountCellProps) => {
...
+          valueFormatter={({
+            value,
+          }: PositionsTableValueFormatterParams & {
+            value: Position['marketName'];
+          }) => {
+            if (!value) {
+              return undefined;
+            }
+            // split market name into two parts, 'Part1 (Part2)'
+            const matches = value.match(/^(.*)\((.*)\)\s*$/);
```
mattrussell36, line 168: chore (blocking): I'm not sure it'll be safe to assume that the market name will be in this format _and_ that the value in the parentheses will always be the expiry date. 

@barnabee any guidance here?

I think we should be just rendering the name as is for now. And revisit once we have a reliable way to retrieve expiry datetimes (from metadata).

Finally if we don't get a match so the second row is omitted, it might make the table ugly as the spacing between the rows will be inconsistent

libs/positions/src/lib/positions-table.tsx
```diff
@@ -113,234 +124,299 @@ export const AmountCell = ({ valueFormatted }: AmountCellProps) => {
...
-        type="rightAligned"
-        cellClassRules={{
-          'text-vega-green-dark dark:text-vega-green': ({
+            );
+          }}
+        />
+        <AgGridColumn
+          headerName={t('Entry price')}
+          field="averageEntryPrice"
+          headerComponentParams={{
```
mattrussell36, line 241: chore (non-blocking): Can we use `headerComponent` here?

libs/positions/src/lib/positions-table.tsx
```diff
@@ -113,234 +124,299 @@ export const AmountCell = ({ valueFormatted }: AmountCellProps) => {
...
+              : addDecimalsFormatNumber(value.toString(), data.decimals)
+          }
+          cellRenderer="PriceFlashCell"
+          headerTooltip={t('P&L excludes any fees paid.')}
+        />
+        <AgGridColumn
+          headerName={t('Unrealised PNL')}
+          field="unrealisedPNL"
+          type="rightAligned"
+          cellClassRules={{
```
mattrussell36, line 366: chore (non blocking): Would it be worthwhile for this be moved to ui-toolkit/ag-grid?

mattrussell36 (changes requested):

libs/positions/src/lib/positions-table.tsx
bglownia: So far market names looks like this
```
Apple Monthly (30 Jun 2022)
AAVEDAI Monthly (30 Jun 2022)
BTCUSD Monthly (30 Jun 2022)
AAVEDAI Monthly (30 Jun 2022)
Tesla Quarterly (30 Jun 2022)
Tesla Quarterly (30 Jun 2022)
OP/USD AUG-SEP22 - Incentive.
UNIDAI Monthly (30 Jun 2022)
AAVEDAI Monthly (30 Jun 2022)
Apple Monthly (30 Jun 2022)
ETHBTC Quarterly (30 Jun 2022)
UNIDAI Monthly (30 Jun 2022)
Tesla Quarterly (30 Jun 2022)
BTCUSD Monthly (30 Jun 2022)
ETHBTC Quarterly (30 Jun 2022)
Apple Monthly (30 Jun 2022)
BTCUSD Monthly (30 Jun 2022)
ETHBTC Quarterly (30 Jun 2022)
UNIDAI Monthly (30 Jun 2022)
CELUSD (June 2022)
UNIUSD (Incentive)
```

If scheme will change we can update regexp. I will modify it to support "ABC (XYZ)" and "ABC - XYZ"

libs/positions/src/lib/positions-table.tsx
bglownia: If we will use header component we will replace default header. We will need to handle sorting and filter menu.

libs/positions/src/lib/positions-table.tsx
bglownia: I'm not sure about it. Creating two rows is rather trivial just add two block elements. It all depends what you want to put in those rows and how this will affect rowHeight

mattrussell36 (approved):

PR #1105, opened 2022-08-22
fix: market dialog query loop, and initial dialog render

mattrussell36:
# Description ℹ️

- Fixes issue where un-memoized since timestamp being used as a variable would fire off a request every render and by the time the request finished the variable had changed causing a fresh request to be made.
- Also fixes the landing dialog only showing when the request for markets had finished. This led to a bad UX because the full page is shown and then a few seconds later the dialog pops up. Now the dialog immediately shows but with a loading state

# Demo 📺

https://user-images.githubusercontent.com/6803987/186046742-4ef9947e-cfc5-4168-ac5e-7de36fa0dca4.mp4

# Technical
- Also adds a useMarkets hook to avoid duplication of market fetching logic

### DIFF
--- a/apps/trading/pages/markets/[marketId].page.tsx
+++ b/apps/trading/pages/markets/[marketId].page.tsx
@@ -129,7 +129,6 @@ const MarketPage = ({ id }: { id?: string }) => {
                   store.setMarketId(marketId);
                 }
               }}
-              title={t('Select a market to get started')}
             />
           </>
         );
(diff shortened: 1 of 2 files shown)


MadalinaRaicu (approved):

MadalinaRaicu:
was about to add a fix in another PR. kind of did but approved anyway, i'll fix any conflicts.

PR #1115, opened 2022-08-23
test: added missing block confirmation coverage

radoslaw-szpiech:
# Description ℹ️

Added coverage for following AC:

must see feedback that the association has been registered by Vega and that it can be used after the number of Ethereum block confirmations required (typically 50) [1000-ASSO-0013](https://github.com/vegaprotocol/specs/blob/62af5e047880327ba39998aed60b786d50030d80/user-interface/1000-ASSO-associate.md#1000-ASSO-0013)


macqbat (approved):

mattrussell36 (approved):
Approved pending PR description update

PR #1122, opened 2022-08-23
Fix/1117: Network stats env

### LINKED ISSUE #1117: [explorer] Network switcher not working as expected
As per https://github.com/vegaprotocol/vega/issues/5971, `n11-testnet` is currently dead. That is the node that https://explorer.fairground.wtf/ points to by default. So I wanted to use the node switcher to select another node, but it didn't work

# Steps to recreate
1. Go to https://explorer.fairground.wtf/
2. Scroll to the bottom
3. Hit 'Edit' at the end of the data node section
4. Enter another, valid node
5. Hit 'Check'
6. Hit 'Connect' after the checks pass
7. Refresh the page

# Expected result
- Polling that is checking for `statistics` on n11 initially should switch to the new node after hitting connect
- Refreshing the page should leave me on the newly selected node

# Actual result
- Polling stays on the old node
- Refreshing the page resets to the original node

# Screenshots

## Step 6 - After hitting connect, network requests still go to n11
image

## Step 7 - after refreshing, it resets back to n11
image (1)

notbot00:
# Related issues 🔗

Closes #1117 

# Description ℹ️

Removed the rest endpoints from explorer and stats apps, so we can use the gql endpoints instead, connecting to the `@vegaprotocol/environment` urls. This fixes the node switching issue.


libs/environment/src/components/network-loader/network-loader.tsx
```diff
@@ -19,6 +19,7 @@ export function NetworkLoader<T>({
 
   const client = useMemo(() => {
     if (VEGA_URL) {
+      console.log(`CREATING ${VEGA_URL}`);
```
mattrussell36, line 22: chore (blocking): Remove

libs/network-stats/src/components/stats-manager/stats-manager.tsx
```diff
@@ -9,75 +11,75 @@ import { Table } from '../table';
...
-        setError(e as Error);
-      }
-    }
+export const StatsManager = ({ className }: StatsManagerProps) => {
+  const { VEGA_ENV } = useEnvironment();
+  const { data, error, startPolling, stopPolling } =
+    useQuery<NetworkStats>(STATS_QUERY);
 
-    const interval = setInterval(getStats, 1000);
+  useEffect(() => {
```
mattrussell36, line 77: chore (blocking): You don't need this useEffect. Instead pass `pollInterval: true` to useQuery options

apps/explorer/src/main.tsx
```diff
@@ -1,24 +1,10 @@
...
 import './styles.css';
 
 import App from './app/app';
-import { ENV } from './app/config/env';
 import { StrictMode } from 'react';
 
-const { dsn } = ENV;
-
-/* istanbul ignore next */
-if (dsn) {
```
mattrussell36, line 14: question (blocking): Why remove this and put it in the useEffect. I don't love it being in an effect

mattrussell36 (changes requested):

apps/explorer/src/main.tsx
notbot00: because we don't know the env here, since it comes from a hook

libs/environment/src/components/network-loader/network-loader.tsx
notbot00, line 22:
```suggestion
```

notbot00: done

libs/network-stats/src/components/stats-manager/stats-manager.tsx
notbot00: unfortunately, if i do it that way, the polling permanently stops once i navigate off the home tab, and go back to it. if i call the start / stop in a useEffect, it retriggers it and it seems to be working fine.

mattrussell36 (approved):

PR #1131, opened 2022-08-24
Fix/1128: Individual proposal pages failing to load

### LINKED ISSUE #1128: Individual proposal pages failing to load
Screenshot 2022-08-24 at 09 46 52

sam-keen:
# Related issues 🔗

Closes #1128

# Description ℹ️

Ensures individual proposal pages are loading, based on the same data as the other proposal pages.

# Technical 👨‍🔧

Individual proposal pages were built early on and relied on the REST endpoint to get the 'terms' data. Fairground was broken for individual proposals because it didn't have an `NX_VEGA_REST` env var, and was (possibly?) using a cached version which pointed at a borked node. The cleanest solution IMO was to remove the unnecessary REST reliance, instead using GQL fully, bringing everything in line.

### DIFF
--- a/apps/token/src/routes/governance/proposal/proposal-container.tsx
+++ b/apps/token/src/routes/governance/proposal/proposal-container.tsx
@@ -1,30 +1,13 @@
 import { gql, useQuery } from '@apollo/client';
-import { Callout, Intent, Splash } from '@vegaprotocol/ui-toolkit';
-import React from 'react';
-import { useTranslation } from 'react-i18next';
+import { AsyncRenderer } from '@vegaprotocol/ui-toolkit';
 import { useParams } from 'react-router-dom';
 
-import { SplashLoader } from '../../../components/splash-loader';
-import { useFetch } from '@vegaprotocol/react-helpers';
 import { Proposal } from '../components/proposal';
 import { PROPOSALS_FRAGMENT } from '../proposal-fragment';
 import type {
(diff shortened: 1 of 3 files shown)


edd (approved):

apps/token/src/routes/governance/proposal/proposal-container.tsx
```diff
@@ -36,16 +19,7 @@ export const PROPOSAL_QUERY = gql`
 `;
 
 export const ProposalContainer = () => {
-  const { t } = useTranslation();
   const params = useParams<{ proposalId: string }>();
-  const proposalUrl = React.useMemo(
-    () => new URL(`governance/proposal/${params.proposalId}`, ENV.restUrl).href,
```
notbot00, line 42: chore (blocking?): i think it would be more future proof (well, less confusing in the future) if we'd remove this restUrl from the ENV object, and remove the whole `NX_VEGA_REST` from all the env files.

asiaznik (approved):

apps/token/src/routes/governance/proposal/proposal-container.tsx
sam-keen: The rest url still appears to be used in some circumstances. It goes beyond the reach of this quick-fix ticket. I'll make a separate ticket to go through and remove the env var where possible.

Issue #626, opened 2022-06-23
Add Oracles info to Console V2


campbellssource:
Follow on from https://github.com/vegaprotocol/frontend-monorepo/issues/625 This to add details about the oracle a market uses.

## Background

When someone proposed a market they also define 2 oracle specifications (or re-use an existing ones).
`oracleSpecForTradingTermination` is used to create an event to [terminate the market](https://github.com/vegaprotocol/specs-internal/blob/master/protocol/0043-MKTL-market_lifecycle.md#trading-terminated),
`oracleSpecForSettlementPrice` is used as the price to [settle](https://github.com/vegaprotocol/specs-internal/blob/master/protocol/0043-MKTL-market_lifecycle.md#settled)  positions that are still open.

```GraphQL
{
  markets{
    tradableInstrument{
      instrument{
        product{
          ...on Future{
            oracleSpecForSettlementPrice{
              id
            }
            oracleSpecForTradingTermination{
              id
            }
          }
        }
      }
    }
  }
}
```

These IDs reference an **oracle Specification**:
```GraphQL
{
  oracleSpecs{
    id
    status
    pubKeys
    createdAt
    updatedAt
    filters{
      key {
        name
        type
      }
      conditions {
        value
        operator
      }
    }
  }
}
```

Each specification is "Listening" to a transaction type **Oracle Submission**
See [WIP docs](https://deploy-preview-121--vega-docusaurus.netlify.app/docs/testnet/tutorials/oracles#3-submit-the-message-to-the-chain-1) Which might already be [live](http://docs.vega.xyz/docs/testnet/tutorials/oracles).

You can see all oracles submissions with
```GraphQL
{
  oracleData{
    data{
      name
      value
    }
    broadcastAt
    matchedSpecIds
	}
}
```

When an oracle submission matches a filter it creates a **Binding**.
You'll see a `broadcastAt` and the specID that it matches.

note: some names are [up for debate](https://github.com/vegaprotocol/vega/issues/5689).

## Design + UX requirements

I suggest that the market info at very least should include the `oracleSpecForTradingTermination` and `oracleSpecForSettlementPrice` IDs. At a minimum it could link to the block explorer that should show all the [raw oracle information](https://github.com/vegaprotocol/frontend-monorepo/issues/633).
But ideally Console would also show this information...

PeteWilliams:
I've added this to the Market info tab under Market spec: 
[...]

PR #1133, opened 2022-08-24
feat: liquidity provisions view

### LINKED ISSUE #491: Liquidity providers view
design from console
Image
- [ ] check if all columns are needed
- [ ] check for other feedback from team/users
- [ ] find a place to put it in Console V2
- [ ] Liquidity data that applies to the whole market (not specific provisions)
- [ ] Consider including the market.account.liquidity balances which is the sum of fees to be distributed in the next liquidity epoch 

See https://github.com/vegaprotocol/frontend-monorepo/issues/625 as there is some liquidity info that was left out of the market info panel. e..g liquidity monitoring target stake etc.
Screenshot 2022-06-23 at 10 25 15
if these don't go into this ticket, please raise another.

MadalinaRaicu:
# Related issues 🔗

Closes #491 & #626

# Description ℹ️

Liquidity provisions view launched from market info. 

# Demo 📺

If user is LP: 
Screenshot 2022-09-01 at 23 16 36

If user is not LP:
Screenshot 2022-09-01 at 23 18 12

If wallet not connected:
image

Open LP view 1:
Screenshot 2022-09-01 at 16 48 14

Open LP view 2:
Screenshot 2022-09-01 at 16 48 27

# Technical 👨‍🔧

- [x] view full oracle details
- [x] view liquidity provisions view


libs/ui-toolkit/src/components/dialog/dialog.tsx
```diff
@@ -37,8 +37,9 @@ export function Dialog({
     {
       'lg:w-[620px] w-full': size === 'small',
       'w-full w-full md:w-[720px] lg:w-[940px]': size === 'medium',
+      'w-full w-full md:w-[80vw] lg:w-[80vw]': size === 'large',
       'left-[0px] top-[99px] h-[calc(100%-99px)] border-0 translate-x-[0] translate-y-[0] border-none overflow-y-auto w-full':
-        size === 'large',
+        size === 'extra-large',
```
MadalinaRaicu, line 42: I needed something in between `medium` and `large`. I am super open to suggestions.

libs/deal-ticket/src/components/market-info/info-market.tsx
```diff
@@ -131,28 +139,31 @@ export const Info = ({ market }: InfoProps) => {
...
         title: t(`Insurance pool`),
         content: (
           <MarketInfoTable
             data={{
-              balance: `${a.balance}
-           ${market.tradableInstrument.instrument.product?.settlementAsset.symbol}`,
+              balance: a.balance,
             }}
+            assetSymbol={assetSymbol}
+            decimalPlaces={market.positionDecimalPlaces} // TODO decimal places needed here?
```
MadalinaRaicu, line 150:
```suggestion
            decimalPlaces={market.positionDecimalPlaces}
```

libs/orders/src/lib/components/order-feedback/order-feedback.tsx
```diff
@@ -52,12 +52,11 @@ export const OrderFeedback = ({ transaction, order }: OrderFeedbackProps) => {
                 : 'text-vega-red-dark dark:text-vega-red'
             }
           >
-            {`${
-              order.side === Side.SIDE_BUY ? '+' : '-'
-            } ${addDecimalsFormatNumber(
-              order.size,
-              order.market?.positionDecimalPlaces ?? 0
-            )}
+            {`${order.side === Side.SIDE_BUY ? '+' : '-'
```
dexturr, line 55: nitpick (non-blocking): seeing this pop up in a few places over the last little while. May be worth a util.

libs/deal-ticket/src/components/market-info/tooltip-mapping.tsx
```diff
@@ -0,0 +1,103 @@
+import { t } from '@vegaprotocol/react-helpers';
+import { Link } from '@vegaprotocol/ui-toolkit';
+import type { ReactNode } from 'react';
+
+export const tooltipMapping: Record<string, ReactNode> = {
+  makerFee: t(
+    'Maker portion of the fee is transferred to the non-aggressive, or passive party in the trade (the maker, as opposed to the taker).'
+  ),
+  liquidityFee: t(
+    'Liquidity portion of the fee is paid to market makers for providing liquidity, and is transferred to the market-maker fee pool for the market.'
```
candida-d, line 10:
```suggestion
    'Liquidity portion of the fee is paid to market makers for providing liquidity, and is transferred to the liquidity fee pool for the market.'
```

libs/liquidity/src/lib/__generated__/MarketLiquidity.ts
```diff
@@ -0,0 +1,252 @@
...
+  /**
+   * The name of the asset (string)
+   */
+  settlementAsset: MarketLiquidity_market_tradableInstrument_instrument_product_settlementAsset;
+}
+
+export interface MarketLiquidity_market_tradableInstrument_instrument {
+  __typename: "Instrument";
+  /**
+   * A short non necessarily unique code used to easily describe the instrument (e.g: FX:BTCUSD/DEC18) (string)
```
candida-d, line 121:
```suggestion
   * A short not necessarily unique code used to easily describe the instrument (e.g: FX:BTCUSD/DEC18) (string)
```

libs/deal-ticket/src/components/market-info/tooltip-mapping.tsx
candida-d: @MadalinaRaicu This is not coming from generated docs, correct?

MadalinaRaicu: correct. it has been manually added in another ticket.

libs/deal-ticket/src/components/market-info/info-market-query.ts
```diff
@@ -8,13 +8,6 @@ export const MARKET_INFO_QUERY = gql`
       decimalPlaces
       positionDecimalPlaces
       state
-      accounts {
```
MadalinaRaicu, line 11: removing these duplicates - might have been added during a merge conflict

libs/deal-ticket/src/components/market-info/info-market.tsx
```diff
@@ -285,172 +336,3 @@ export const Info = ({ market }: InfoProps) => {
     </div>
   );
 };
-
```
MadalinaRaicu, line 288: this file was getting too big. moved the tooltips somewhere else.

libs/deal-ticket/src/components/market-info/info-key-value-table.tsx
```diff
@@ -0,0 +1,106 @@
...
+        inline={isPrimitive}
+        noBorder={true}
+        dtClassName={className}
+        ddClassName={className}
+      >
+        <Tooltip description={tooltipMapping[field]} align="start">
+          <div tabIndex={-1}>{startCase(t(field))}</div>
+        </Tooltip>
+        <span style={{ wordBreak: 'break-word' }}>
+          {isNumber && !unformatted
```
mattrussell36, line 52: chore (blocking): Can we tidy up this huge condition. Preferably set the cell content in the body of the function? This is hard to read
```
let content = null
if (x) {
  content = ...
} else {
  content =
}
return <div>{content}</div>
```

libs/deal-ticket/src/components/market-info/info-key-value-table.tsx
```diff
@@ -0,0 +1,106 @@
...
+} from '@vegaprotocol/ui-toolkit';
+import BigNumber from 'bignumber.js';
+import omit from 'lodash/omit';
+import startCase from 'lodash/startCase';
+import type { ReactNode } from 'react';
+import { tooltipMapping } from './tooltip-mapping';
+
+interface RowProps {
+  field: string;
+  value: any;
```
mattrussell36, line 21: chore (blocking): can we type this properly.

It looks like we aren't using generating types for the MARKET_INFO_QUERY. Can we do so please. Then all the types should flow nicely through this part of the app.

libs/deal-ticket/src/components/market-info/info-key-value-table.tsx
```diff
@@ -0,0 +1,106 @@
...
+            : value}
+        </span>
+      </KeyValueTableRow>
+    );
+  }
+  return null;
+};
+
+export interface MarketInfoTableProps {
+  data: any;
```
mattrussell36, line 70: chore (blocking): any

libs/deal-ticket/src/components/market-info/info-key-value-table.tsx
```diff
@@ -0,0 +1,106 @@
...
+  value,
+  decimalPlaces,
+  asPercentage,
+  unformatted,
+  assetSymbol = '',
+}: RowProps) => {
+  const isNumber = typeof value === 'number' || !isNaN(Number(value));
+  const isPrimitive = typeof value === 'string' || isNumber;
+  const className = 'text-black dark:text-white text-ui !px-0 !font-normal';
+  if (isPrimitive) {
```
mattrussell36, line 39: question (non blocking): Not sure why we need this check. I think it stems from some any types defined in this file.

libs/deal-ticket/src/components/market-info/info-market.tsx
```diff
@@ -51,7 +49,10 @@ export const MarketInfoContainer = ({ marketId }: MarketInfoContainerProps) => {
   }, []);
 
   const { data, loading, error } = useQuery(MARKET_INFO_QUERY, {
```
mattrussell36, line 51: chore (blocking): I realize this was already this way.. but can we add generic type args here please

libs/deal-ticket/src/components/market-info/info-market.tsx
```diff
@@ -51,7 +49,10 @@ export const MarketInfoContainer = ({ marketId }: MarketInfoContainerProps) => {
...
-    variables: { marketId, interval: Interval.INTERVAL_I1H, since: yTimestamp },
+    variables: useMemo(
```
mattrussell36, line 52: chore (suggestion): can we keep hooks used naturally in the body of the function for readability

libs/deal-ticket/src/components/market-info/info-market.tsx
```diff
@@ -128,28 +131,34 @@ export const Info = ({ market }: InfoProps) => {
...
-  );
+  const { VEGA_EXPLORER_URL } = useEnvironment();
+  const keyDetails = {
+    ...pick(
+      market,
+      'name',
+      'decimalPlaces',
+      'positionDecimalPlaces',
+      'tradingMode',
+      'id' as 'marketId'
```
mattrussell36, line 158: question (blocking): Why the typecast, doesnt look like it does anything?

libs/deal-ticket/src/components/market-info/info-market.tsx
```diff
@@ -254,6 +270,32 @@ export const Info = ({ market }: InfoProps) => {
...
+          }}
+          decimalPlaces={
+            market.tradableInstrument.instrument.product.settlementAsset
+              .decimals
+          }
+          assetSymbol={assetSymbol}
+          link={
+            <Link
+              target="_blank"
+              className="text-ui dark:text-white text-black underline hover:underline hover:text-black-60 dark:hover:text-white-80"
```
mattrussell36, line 290: chore (blocking): None of these classes are required, and some are from the old config. Lets just use the raw Link component styles which should handle underline styles

libs/deal-ticket/src/components/market-info/info-market.tsx
```diff
@@ -267,6 +309,15 @@ export const Info = ({ market }: InfoProps) => {
               market.tradableInstrument.instrument.product
                 .oracleSpecForTradingTermination.id,
           }}
+          link={
+            <Link
+              target="_blank"
+              href={`${VEGA_EXPLORER_URL}/oracles#${market.tradableInstrument.instrument.product.oracleSpecForTradingTermination.id}`}
+              className="text-ui dark:text-white text-black underline hover:underline hover:text-black-60 dark:hover:text-white-80"
```
mattrussell36, line 316: chore (blocking): Remove classnames please. As an aside. We should have to set text color everywhere on every element. Those styles should be inherited

libs/liquidity/src/lib/liquidity-data-provider.ts
```diff
@@ -0,0 +1,231 @@
...
+
+export interface LiquidityProvision {
+  party: string;
+  commitmentAmount: string | undefined;
+  fee: string | undefined;
+  equityLikeShare: string;
+  averageEntryValuation: string;
+  obligation: string | null;
+  supplied: string | null;
+  status: LiquidityProvisionStatus | string | undefined;
```
mattrussell36, line 88: question (blocking): Why type this as also string | undefined? The api indicates that this can only every be `LiquidityProvisionStatus`. Some thing is wrong in the types elsewhere if you've had to do this. If its because node can be undefined then you should filter those nodes out first.

libs/liquidity/src/lib/liquidity-manager.tsx
```diff
@@ -0,0 +1,132 @@
...
+  partyId,
+  marketId,
+}: LiquidityManagerProps) => {
+  const gridRef = useRef<AgGridReact | null>(null);
+  const {
+    data: marketLiquidityData,
+    loading,
+    error,
+  } = useQuery<MarketLiquidity>(MARKET_LIQUIDITY_QUERY, {
+    variables: useMemo(() => ({ marketId }), [marketId]),
```
mattrussell36, line 39: chore (blocking): No need for useMemo here

libs/liquidity/src/lib/liquidity-manager.tsx
```diff
@@ -0,0 +1,132 @@
...
+  });
+
+  const { liquidityProviders, suppliedStake, targetStake, code, symbol } =
+    useLiquidityProvision({ data: marketLiquidityData });
```
mattrussell36, line 43: suggestion (non blocking): IMO I'd keep the query and the data manipulation all in one hook. So components that need to use the data can consume it more simply

libs/liquidity/src/lib/liquidity-manager.tsx
```diff
@@ -0,0 +1,132 @@
...
+    'h-full max-h-full',
+    'flex flex-col',
+    'text-ui-small',
+    'gap-8'
+  );
+  const titleClasses =
+    'sm:text-lg md:text-xl lg:text-2xl flex items-center gap-4 whitespace-nowrap';
+
+  const myLpEdges = useMemo(
+    () => (liquidityProviders || []).filter((e) => e.party === partyId),
```
mattrussell36, line 55: suggestion (non blocking): The result of the useLiquidityProvision hook should always be an array so that these falsy checks can be omitted.

libs/liquidity/src/lib/liquidity-data-provider.ts
```diff
@@ -0,0 +1,231 @@
...
+export const formatLiquidityData = ({
+  suppliedStake,
+  assetDecimalPlaces,
+  targetStake,
+  code,
+  symbol,
+  liquidityProviders,
+}: LiquidityData) => {
+  return {
+    suppliedStake: addDecimalsFormatNumber(suppliedStake, assetDecimalPlaces),
```
mattrussell36, line 185: suggestion (non blocking): Its better to return the raw numbers/bignumbers rather than format them here. What if another component wants to format them in a different way?

Utilize AgGrid valueFormatters instead.

libs/liquidity/src/lib/liquidity-manager.tsx
```diff
@@ -0,0 +1,132 @@
...
+    variables: useMemo(() => ({ marketId }), [marketId]),
+  });
+
+  const { liquidityProviders, suppliedStake, targetStake, code, symbol } =
+    useLiquidityProvision({ data: marketLiquidityData });
+
+  const wrapperClasses = classNames(
+    'h-full max-h-full',
+    'flex flex-col',
+    'text-ui-small',
```
mattrussell36, line 48: chore (blocking): Remove this has no effect. use text-xs/sm/md etc

libs/liquidity/src/lib/liquidity-manager.tsx
```diff
@@ -0,0 +1,132 @@
...
+    if (activeEdges?.length) return 'active';
+    else if (inactiveEdges?.length > 0) return 'inactive';
+    return 'active';
+  };
+  const itemClass =
+    'min-w-min w-[120px] whitespace-nowrap pb-3 px-4 border-l border-neutral-300 dark:border-neutral-700';
+  const itemHeading = 'text-neutral-400';
+  return (
+    <AsyncRenderer loading={loading} error={error} data={liquidityProviders}>
+      <header className="w-screen xl:px-4 pt-4 border-b border-neutral-300 dark:border-neutral-700">
```
mattrussell36, line 83: chore (blocking): This UI should be the concern of the app and not the lib as it deals with page layout. Ideally the lib should only export things that are app agnostic. In this case only the table component and the data provider/hook should be exported from the lib. Things like market id and party id should be passed in as props from the app.

libs/liquidity/src/lib/liquidity-manager.tsx
```diff
@@ -0,0 +1,132 @@
...
+              <Tabs activeDefaultId="myLP">
+                <Tab id="myLP" name={t('My liquidity provision')}>
+                  <LiquidityTable ref={gridRef} data={myLpEdges} />
+                </Tab>
+                <Tab children={undefined} id={''} name={''} hidden={true} />
+              </Tabs>
+            </ResizableGridPanel>
+          )}
+          <ResizableGridPanel preferredSize={300} minSize={50}>
+            <Tabs activeDefaultId={getActiveDefaultId()}>
```
mattrussell36, line 119: chore (blocking): can we rename this prop to just `active`

libs/react-helpers/src/lib/format/number.ts
```diff
@@ -19,11 +19,11 @@ export function toBigNum(
 
 export function addDecimal(
   value: string | number | EthersBigNumber,
-  decimals: number,
+  decimals?: number,
```
mattrussell36, line 22: question (blocking): I'm uncomfortable with this change. This function is used a lot to add and remove decimals specifically for price and sizes. Its important that decimals is ALWAYS provided so that this can be done accurately. No someone could easily omit this which could result in transactions being sent with incorrect values.

libs/react-helpers/src/lib/format/number.ts
```diff
@@ -55,19 +55,25 @@ export const formatNumber = (
 };
 
 export const addDecimalsFormatNumber = (
-  rawValue: string | number,
-  decimalPlaces: number,
+  rawValue?: string | number | null,
```
mattrussell36, line 58: question (blocking): I'm also uncomfortable with these changes. These are simple formatting functions. If you need to pass undefined here something is probably wrong with your types and/or code elsewhere

libs/liquidity/src/lib/liquidity-manager.tsx
```diff
@@ -0,0 +1,132 @@
...
+              <div style={{ wordBreak: 'break-word' }}>{marketId}</div>
+            </div>
+          </div>
+        </div>
+      </header>
+      <div className={wrapperClasses}>
+        <ResizableGrid vertical={true}>
+          {partyId && (
+            <ResizableGridPanel preferredSize={300} minSize={50}>
+              <Tabs activeDefaultId="myLP">
```
mattrussell36, line 110: chore (blocking): Why use Tabs component if there is only one tab? I feel like it might be because you want the tab title UI above the table. 

Lets keep it more simple and just render a My LP tab alongside active and inactive

libs/react-helpers/src/lib/format/date.ts
```diff
@@ -31,6 +31,13 @@ export const getDateTimeFormat = once(
     })
 );
 
+export const dateValueFormatter = ({ value }: { value?: string | null }) => {
```
mattrussell36, line 34: suggestion (non blocking):  I'm not sure we really need extra formatter. Just use getDateTimeFormat().format()?

libs/web3/src/lib/use-ethereum-config.ts
```diff
@@ -32,6 +32,19 @@ export const NETWORK_PARAMS_QUERY = gql`
   }
 `;
 
+export const useNetworkParameter = (params: string[]) => {
```
mattrussell36, line 35: chore (blocking): API now allows for getting a single param so lets refactor and use that.

mattrussell36 (changes requested):

libs/deal-ticket/src/components/market-info/info-key-value-table.tsx
MadalinaRaicu: sure, updated

libs/deal-ticket/src/components/market-info/info-market.tsx
MadalinaRaicu: yeah, happy to remove them now. updated

libs/deal-ticket/src/components/market-info/info-market.tsx
MadalinaRaicu: it's not a typecast, it changes the key. Instead of displaying 'ID' in this section, it will display 'Market ID'. 
I followed this approach because I was trying to not miss any field from the market info that the API was giving me, maybe omit some controlled things like `__typename`, but also add new ones. this is why I was checking primitives and formatting keys. 
it makes it **super** easy to update the market info section. 
literally, one line of code is needed to control the fields. 
this seemed necessary when I was working on the market info section and I was going back and forth to update things. the list was getting super large, hundreds of lines of code, formatting seemed to be spread around, and it was repetitive. 
I am happy to discuss how to improve this component. this is why we were not following MarketInfoQuery types fully, .

libs/deal-ticket/src/components/market-info/info-market.tsx
MadalinaRaicu: yeah, happy to remove them now. updated

libs/liquidity/src/lib/liquidity-manager.tsx
MadalinaRaicu: updated.

libs/liquidity/src/lib/liquidity-manager.tsx
MadalinaRaicu: updated.

libs/liquidity/src/lib/liquidity-manager.tsx
MadalinaRaicu: yep, now these can be removed. needed them for the previous design.

libs/liquidity/src/lib/liquidity-manager.tsx
MadalinaRaicu: updated

libs/liquidity/src/lib/liquidity-manager.tsx
MadalinaRaicu: yeah, I am happy to do that. that's what I was proposing yesterday. check this out: 
Screenshot 2022-09-02 at 10 33 49
Screenshot 2022-09-01 at 23 46 54
updated.

libs/react-helpers/src/lib/format/date.ts
MadalinaRaicu: updated.

libs/liquidity/src/lib/liquidity-data-provider.ts
MadalinaRaicu: updated. before using only ag-grid for this, I had to change the formatting in two places. now I don;t need to do this anymore. moved the value formatters back.

libs/liquidity/src/lib/liquidity-manager.tsx
MadalinaRaicu: nice! updated.

libs/deal-ticket/src/components/market-info/info-market.tsx
MadalinaRaicu: updated.

libs/deal-ticket/src/components/market-info/info-market.tsx
MadalinaRaicu: updated.

libs/web3/src/lib/use-ethereum-config.ts
MadalinaRaicu: updated. updated also in token app to use web3 lib. the methods were duplicated. hope that is fine.

libs/deal-ticket/src/components/market-info/info-key-value-table.tsx
MadalinaRaicu: https://github.com/vegaprotocol/frontend-monorepo/pull/1133/files#r961402382

libs/react-helpers/src/lib/format/number.ts
MadalinaRaicu: updated.

libs/react-helpers/src/lib/format/number.ts
MadalinaRaicu: updated.

libs/deal-ticket/src/components/market-info/info-key-value-table.tsx
MadalinaRaicu: https://github.com/vegaprotocol/frontend-monorepo/pull/1133/files#r961402382

libs/deal-ticket/src/components/market-info/info-key-value-table.tsx
MadalinaRaicu: https://github.com/vegaprotocol/frontend-monorepo/pull/1133/files#r961402382

libs/liquidity/src/lib/liquidity-manager.tsx
MadalinaRaicu: updated

libs/liquidity/src/lib/liquidity-data-provider.ts
MadalinaRaicu: for every data.liquidityProviderFeeShare we are trying to find the matching liquidity provision connection node. the API does not match it for us. I tried to filter non-empty connection nodes, however typescript being typescript still complains if I leave the connection as not being possibly null/undefined.

libs/deal-ticket/src/components/market-info/info-market.tsx
mattrussell36: Sorry this should have read. "We should _not_ have to set text color everywhere..."

libs/deal-ticket/src/components/market-info/info-market.tsx
mattrussell36: It is a typecast. All its doing here is typing the last argument you pass to `pick` as the literal string 'marketId' as opposed to it just being inferred as `string`. Its safe to remove here because the label text actually comes from the `marketID` key passed into the MarketInfoTable.

mattrussell36: Overall I think that rather than trying to ensure all the properties on the object are displayed we should more explicitly declare what fields we are showing. This should improve the type safety and simplify things quite a lot as its quite hard to follow all the data and additional props for formatting.

mattrussell36: That said, I think lets merge this and revisit it

mattrussell36 (approved):

apps/trading/components/trading-mode-tooltip/trading-mode-tooltip.tsx
```diff
@@ -65,8 +63,7 @@ const compileGridData = (market: Market_market) => {
       label: isLiquidityMonitoringAuction
         ? t('Est auction end')
         : t('Auction end'),
-      value: getDateTimeFormat().format(new Date(market.data.auctionEnd)),
-      isEstimate: isLiquidityMonitoringAuction ? true : false,
```
notbot00, line 69: question (blocking?): not sure if this was a change in requirements here? are we now displaying all market types prefixed with a `~`?

notbot00 (approved):

libs/web3/src/lib/use-ethereum-config.ts
```diff
@@ -32,6 +32,32 @@ export const NETWORK_PARAMS_QUERY = gql`
   }
 `;
 
+export const useNetworkParam = (param: string) => {
```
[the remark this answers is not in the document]
MadalinaRaicu: updated.

libs/deal-ticket/src/components/market-info/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './__generated__';
```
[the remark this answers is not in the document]
MadalinaRaicu: happy to do this as a follow up.

apps/trading/pages/liquidity/[marketId].page.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const inactiveEdges = useMemo(
+    () =>
+      liquidityProviders.filter(
+        (e) => e.status !== LiquidityProvisionStatus.STATUS_ACTIVE
+      ),
+    [liquidityProviders]
+  );
+
+  const getActiveDefaultId = () => {
+    if (myLpEdges?.length > 0) return 'myLP';
```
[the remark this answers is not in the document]
Comment (MadalinaRaicu) [reply apps/trading/pages/liquidity/[marketId].page.tsx]: updated.

apps/trading/components/trading-mode-tooltip/trading-mode-tooltip.tsx
MadalinaRaicu: @mattrussell36 removed it here https://github.com/vegaprotocol/frontend-monorepo/pull/1133/commits/57bcec59d8e02c6b71ca3ad6bde3a0369113b525  was this in the design? perhaps we should check with @campbellssource

mattrussell36: Good catch. I've amended it to include the tilde when in its a liq monitoring auction

libs/deal-ticket/src/components/market-info/info-market.tsx
```diff
@@ -137,39 +137,45 @@ export const Info = ({ market }: InfoProps) => {
...
+    ...pick(
+      market,
+      'name',
+      'decimalPlaces',
+      'positionDecimalPlaces',
+      'tradingMode'
+    ),
+    state: MarketStateMapping[market.state],
+  };
+  console.log(keyDetails);
```
MadalinaRaicu, line 167:
```suggestion
```

PR #1141, opened 2022-08-24
fix: included origin of token, moved formatters

### LINKED ISSUE #1137: TFE Disassociation - Include origin of token within select dropdown
Currently when dissassociating tokens if both vesting and wallet tokens are associated, the select dropdown doesn't identify  what the select option origin is.

There is a tooltip which takes 3 seconds to appear (which does show this info) see screenshot 
- but from a UX perspective i totally missed it - better to include within the dropdown itself - agreed with Michael

asiaznik:
# Related issues 🔗

Closes #1137

# Description ℹ️

Changed options description to identify what the select option origin is.

# Demo 📺

Screenshot 2022-08-24 at 20 07 47
Screenshot 2022-08-24 at 20 11 06

# Technical 👨‍🔧

Format number housekeeping included.

### DIFF
--- a/apps/token/src/lib/format-number.ts
+++ b/apps/token/src/lib/format-number.ts
@@ -1,33 +1,9 @@
-import React from 'react';
-import { BigNumber } from './bignumber';
+import type { BigNumber } from './bignumber';
+import { formatNumber as format } from '@vegaprotocol/react-helpers';
 
 export const formatNumber = (value: BigNumber, decimals?: number) => {
-  const decimalPlaces =
-    typeof decimals === 'undefined' ? Math.max(value.dp(), 2) : decimals;
-  return value.dp(decimalPlaces).toFormat(decimalPlaces);
-};
-
-export const formatNumberPercentage = (value: BigNumber, decimals?: number) =>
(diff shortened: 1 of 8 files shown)


radoslaw-szpiech (approved):

apps/token/src/lib/format-number.ts
```diff
@@ -1,33 +1,9 @@
...
-    .split(separator);
-  return [integers, decimalsPlaces || ''];
-};
-
-export const useNumberParts = (
-  value: BigNumber | null | undefined,
-  decimals: number
-): [integers: string, decimalPlaces: string] => {
-  return React.useMemo(() => toNumberParts(value, decimals), [decimals, value]);
+  return format(
```
MadalinaRaicu, line 5: (non-blocking): is this very token dApp specific? can't we move it or use directly what we have in react-helpers lib?

MadalinaRaicu: I see we have another formatNumber there. I would've renamed it and moved it in react-helpers to be reused.

MadalinaRaicu (approved):

PR #1164, opened 2022-08-26
chore: remove valid chain check

mattrussell36:
# Related issues 🔗

N/A

# Description ℹ️

Removes check for an invalid chain in the environment setu


libs/environment/src/utils/validate-node.tsx
```diff
@@ -12,10 +12,6 @@ export const getIsNodeLoading = (node?: NodeData): boolean => {
   );
 };
 
-export const getHasInvalidChain = (env: Networks, chain = '') => {
```
notbot00, line 15: not sure really, the implications of removing this will allow any datanode from any chain within the app which we wanted to contain to an environment. did that change?

dexturr: I don't know how we get around this right now. Initial thoughts are we could set the expected chain ID as an env var but I don't really want to have to change that on every site every time we do a single deploy. Feels like we need a network name/id field or similar that we can compare against that remains constant, but that would require core work.

IMO the only thing we can do here right now is remove the feature until we have a better way of detecting the network we are on, unless someone has a 💡 ?

dexturr: But obviously we will need to communicate that trade off.

dexturr (approved):

notbot00 (approved):

PR #1169, opened 2022-08-26
chore: proposal schema change (1057)

### LINKED ISSUE #1057: Proposal title and description schema changes
Related to: https://github.com/vegaprotocol/vega/pull/5912

We need to:
- [ ] Remove getProposalTitle function from everywhere
- [ ] Use title in its' place
- [ ] Remove all references to hash and URL

asiaznik:
# Related issues 🔗

Closes #1057 

# Description ℹ️

The title / description are taken from `rationale` field.

# Demo 📺

Screenshot 2022-08-29 at 15 22 09
Screenshot 2022-08-29 at 15 24 57
Screenshot 2022-08-29 at 15 25 55

# Technical 👨‍🔧

I moved proposal queries to lib/governance so it is now all in one place. Also used new `proposalsConnection` query instead of deprecated `proposals`.


MadalinaRaicu (approved):

libs/governance/src/lib/proposals-queries/proposals-queries.ts
```diff
@@ -0,0 +1,136 @@
+import { gql, useQuery } from '@apollo/client';
+import type { Proposal, ProposalVariables } from './__generated__/Proposal';
+import type {
+  ProposalsConnection,
+  ProposalsConnection_proposalsConnection_edges_node as ProposalNode,
+} from './__generated__/ProposalsConnection';
+
+const PROPOSALS_FRAGMENT = gql`
```
dexturr, line 8: chore (blocking): I don't like the idea of having a shared queries in the app. A lot of the value proposition of GQL is that we request only what we need, rather than everything every time. If we start sharing queries and 1/2 pages need a new field we will be requesting the field for every single page and doing nothing with it. I would rather colocate these queries, so that in each place we request exactly what is needed. Alternatively, if there is a use case for sharing the component that renders the proposals list then a lib that exposes that component would be good, as the data that is used is always guaranteed.

dexturr (changes requested):
Don't think sharing queries is the best idea as it seems a bit leeky to me.

apps/token/src/routes/governance/components/proposal-detail-header/proposal-header.spec.tsx
```diff
@@ -187,29 +203,26 @@ describe('Proposal header', () => {
     );
     // For a rationale over 100 chars, we expect the header to be truncated at
     // 100 chars with ellipsis and the details-one element to contain the rest.
-    expect(screen.getByTestId('proposal-header')).toHaveTextContent(
-      'Class aptent taciti sociosqu ad litora torquent per conubia nostra, per inceptos himenaeos. Aenean…'
-    );
+    expect(screen.getByTestId('proposal-header')).toHaveTextContent('0x0');
     expect(screen.getByTestId('proposal-details-one')).toHaveTextContent(
-      'dolor'
+      'Class aptent taciti sociosqu ad litora torquent per conubia' // shorten to 60 characters
```
sam-keen, line 208: Chore (non-blocking): I think you can remove this comment, the string shortening is happening in proposal-header

sam-keen (changes requested):
Couple of small changes from me. Thanks for picking this up Art

dexturr (approved):

PR #1171, opened 2022-08-26
fix: market ordering

MadalinaRaicu:
# Related issues 🔗

Closes N/A

# Description ℹ️

Fix for ordering markets: In Thursday's demo the default market was one in opening auction: CELUSD. This happened because it was the oldest market added (note: sometimes we don’t have markets trading continuously on fairground, so we were not displaying the oldest market trading continuously). This PR addresses that.

& format network param labels

# Demo 📺

image
image

# Technical 👨‍🔧


apps/explorer/src/app/routes/network-parameters/network-parameters.tsx
```diff
@@ -69,7 +70,7 @@ export const NetworkParameterRow = ({
         'group target:bg-vega-pink target:text-white dark:target:bg-vega-yellow dark:target:text-black'
       }
     >
-      {key}
+      {formatLabel(key)}
```
MadalinaRaicu, line 73: formatting network parameters keys. they are in this format `Governance.Proposal.Asset.MaxClose` and they will be shown as `Governance Proposal Asset Max Close`

MadalinaRaicu:
> LGTM. Though I think @bglownia was saying it would be better to just use the data provider directly here

tried to use it and there was some weird flickering I couldn't get rid of. happy to update to use it in another PR.

sam-keen (approved):

dexturr (changes requested):
Minor thing but important I think

libs/market-list/src/lib/utils/market-utils.ts
```diff
@@ -42,6 +49,12 @@ export const mapDataToMarketList = ({ markets }: MarketList) =>
     ['open', 'id'],
     ['asc', 'asc']
   );
+  return orderedMarkets.sort(
+    (a, b) =>
+      tradingModesOrdering.indexOf(a.tradingMode) -
```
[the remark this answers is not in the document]
MadalinaRaicu: oh, right, I was ordering by id before ordering on tradingMode. do you think it would impact the order in the wrong way?

dexturr (approved):

libs/market-list/src/lib/components/select-market.tsx
```diff
@@ -1,60 +1,64 @@
...
 export const SelectMarketLandingTable = ({
   data,
   onSelect,
 }: {
-  data: MarketList | undefined;
+  data: MarketList_markets[] | undefined;
   onSelect: (id: string) => void;
 }) => {
-  const marketList = data && mapDataToMarketList(data);
+  const textDecorationClassName = `px-8 underline font-sans leading-9 font-bold tracking-tight decoration-solid text-ui light:hover:text-black/80 dark:hover:text-white/80 text-black dark:text-white`;
```
mattrussell36, line 36: question (non-blocking): Do we really need all these links styles here. Why isn't the link component sufficient?

MadalinaRaicu: we don't need this anymore! removed it in recent commit.

apps/trading-e2e/src/integration/home.cy.ts
```diff
@@ -34,11 +28,8 @@ describe('home', () => {
...
       cy.contains('Loading...').should('be.visible');
       cy.contains('Loading...').should('not.exist');
-      cy.wait('@GQL');
 
       cy.get('main[data-testid="market"]', { timeout: 20000 }).should('exist'); // Wait for page to be rendered to before checking url
-
-      cy.url().should('include', `/markets/${oldestMarket.id}`); // Should redirect to oldest market
     });
 
     it('redirects to a default market with the landing dialog open', () => {
```
MadalinaRaicu, line 35:
```suggestion
    it.skip('redirects to a default market with the landing dialog open', () => {
```

PR #1180, opened 2022-08-26
Feat/896 large withdraws

### LINKED ISSUE #896: Handle large withdrawals
We currently prevent large withdrawals with client side validation to get around the fact that the flow of immediately invoking the Ethereum transaction will result in an immediately rejected transaction. We need a pattern to allow creating large withdrawals without auto triggering the Ethereum transaction.

We also need to ensure that the language surrounding this is clear in that _matching_ the withdrawal limit will trigger the enforced withdrawals delay.

mattrussell36:
# Related issues 🔗

Closes #896 

# Description ℹ️

Refactors withdrawals to take account of both withdraw threshold and withdraw default delay. If the withdrawal is over the threshold amount then the confirmation dialog indicates that it must be complete via Ethereum transaction after the delay time has been reached. If the withdrawal is under the threshold then we now provide a button to complete immediately.

# Demo 📺

Screen Shot 2022-08-31 at 3 53 25 PM

# Technical 👨‍🔧

- Added a zustand store for withdrawal state as it was a lot of useMemos and useStates making it hard to read
- Refactored wait functions (for withdrawal and approval) to use promise to avoid nesting multiple callbacks
- Added a new hook, useVerifyWithdrawal which uses the threshold and delay to check if a withdrawal can be completed before executing the Ethereum transaction. This allows us to show the withdrawal availability date and exit early if its not ready yet


libs/web3/src/lib/ethereum-transaction-dialog/ethereum-transaction-dialog.tsx
```diff
@@ -0,0 +1,130 @@
...
+          {t('Error')}: {error.message}
+        </p>
+      );
+    }
+
+    return (
+      <p className={classNames}>
+        {t('Error')}: {t('Unknown error')}
+      </p>
+    );
```
sam-keen, line 77:
```suggestion
    const errorMessage = (isEthereumError(error) && error.reason) || (error instanceof Error && error.message) || t('Unknown error'); 
    
    return (
        <p className={classNames}>
          {t('Error')}: {errorMessage}
        </p>
      );
```

sam-keen (approved):

libs/deposits/src/lib/deposits-table.tsx
```diff
@@ -0,0 +1,75 @@
...
+import type { DepositFields } from './__generated__/DepositFields';
+import { useEnvironment } from '@vegaprotocol/environment';
+import type { DepositStatus } from '@vegaprotocol/types';
+import { DepositStatusMapping } from '@vegaprotocol/types';
+
+export interface DepositsTableProps {
+  deposits: DepositFields[];
+}
+
+export const DepositsTable = ({ deposits }: DepositsTableProps) => {
```
bglownia, line 22: Ag-grid tables can be typed little bit more. This will allow to catch changes in data used by table.

```
import type {
  ICellRendererParams,
  ValueFormatterParams,
  IGetRowsParams,  
  IDatasource
} from 'ag-grid-community';
import type { AgGridReactProps } from 'ag-grid-react';
import type { Deposits_party_depositsConnection_edges_node } from './__generated__/Deposits';


...


export interface GetRowsParams extends Omit<IGetRowsParams, 'successCallback'> {
  successCallback(rowsThisBlock: Deposits_party_depositsConnection_edges_node[], lastRow?: number): void;
}

export interface Datasource extends IDatasource {
  getRows(params: GetRowsParams): void;
}
interface Props extends AgGridReactProps {
  rowData?: Deposits_party_depositsConnection_edges_node[] | null;
  datasource?: Datasource;
}

type PositionsTableValueFormatterParams = Omit<
  ValueFormatterParams,
  'data' | 'value'
> & {
  data: Deposits_party_depositsConnection_edges;
};


export const DepositsTable = (props: Props) => {
  const { ETHERSCAN_URL } = useEnvironment();
  return (
    <AgGrid
      overlayNoRowsTemplate={t('No deposits')}
      defaultColDef={{ flex: 1, resizable: true }}
      suppressCellFocus={true}
      {...props}
    >
    
    
    ....
    
    <AgGridColumn
        headerName="Amount"
        field="amount"
        valueFormatter={({ value, data }: PositionsTableValueFormatterParams & { value: Deposits_party_depositsConnection_edges_node_asset["symbol"] }) => {
          return addDecimalsFormatNumber(value, data.asset.decimals);
        }}
      />
    



```

libs/withdraws/src/lib/use-withdrawals.ts
```diff
@@ -85,41 +98,63 @@ export const useWithdrawals = () => {
     };
   }, [keypair?.pub, subscribeToMore]);
 
-  return queryResult;
+  const withdrawals = useMemo(() => {
+    return orderBy(
+      compact(data?.party?.withdrawalsConnection.edges).map(
+        (edge) => edge.node
+      ),
+      (w) => new Date(w.createdTimestamp).getTime(),
```
bglownia, line 106: `'createdTimestamp'` should be enough. ISO timestamp is safe for string comparsion

libs/withdraws/src/lib/withdrawals-table.tsx
```diff
@@ -98,34 +152,18 @@ export const StatusCell = ({
...
-          <>
-            {t('Open')}
-            <button className="underline" onClick={() => complete(data.id)}>
-              {t('Click to complete')}
-            </button>
-          </>
-        )}
+        {t('Open')}
+        <button className="underline" onClick={() => complete(data)}>
+          {t('Click to complete')}
```
bglownia, line 160: IMHO 'Complete' should be enough

libs/withdraws/src/lib/withdrawals-table.tsx
```diff
@@ -9,19 +9,36 @@ import {
...
+import type { WithdrawalFields } from './__generated__/WithdrawalFields';
+import type { VerifyState } from './use-verify-withdrawal';
+import { ApprovalStatus, useVerifyWithdrawal } from './use-verify-withdrawal';
 
 export interface WithdrawalsTableProps {
-  withdrawals: Withdrawals_party_withdrawals[];
+  withdrawals: WithdrawalFields[];
 }
 
 export const WithdrawalsTable = ({ withdrawals }: WithdrawalsTableProps) => {
```
bglownia, line 30: This table also needs some strong casting (same as deposits table)

libs/withdraws/src/lib/withdrawal-feedback.tsx
```diff
@@ -0,0 +1,94 @@
...
+  submitWithdraw: (withdrawalId: string) => void;
+  availableTimestamp: number | null;
+}) => {
+  const { VEGA_EXPLORER_URL } = useEnvironment();
+  const isAvailable =
+    availableTimestamp === null || Date.now() > availableTimestamp;
+  return (
+    <div>
+      <p className="mb-8">
+        Your funds have been unlocked for withdrawal -{' '}
```
bglownia, line 29: translate  function missing

libs/withdraws/src/lib/withdrawal-feedback.tsx
```diff
@@ -0,0 +1,94 @@
...
+      {isAvailable ? (
+        <Button
+          disabled={withdrawal === null ? true : false}
+          onClick={() => {
+            if (withdrawal) {
+              submitWithdraw(withdrawal.id);
+            }
+          }}
+        >
+          Withdraw funds
```
bglownia, line 81: translate function missing

libs/web3/src/lib/ethereum-transaction-dialog/ethereum-transaction-dialog.tsx
mattrussell36: Good call on tidying this, though I find that condition a little confusing so I've just kept if statements to set an error message so the UI isn't duplicated.

apps/trading/pages/index.page.tsx
```diff
@@ -37,14 +40,15 @@ export function Index() {
       // If a default market is found, go to it with the landing dialog open
       if (marketId) {
         setLandingDialog(true);
+        setMarketId(marketId);
```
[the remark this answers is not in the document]
mattrussell36: This is because the trading link in the nav should navigate to the last used market (see logic in the trading nav). I don't love this UI pattern but its needed for it to work

libs/deposits/src/lib/deposits-table.tsx
mattrussell36: @bglownia I've added some helper types for this in libs/ui-toolkit/src/components/ag-grid/index.tsx, take a look. Perhaps we might want to add them to use in other tables.

libs/withdraws/src/lib/withdrawals-table.tsx
```diff
@@ -148,3 +204,46 @@ const RecipientCell = ({
...
+    return <p>{t('Verifying...')}</p>;
+  }
+
+  if (state.status === ApprovalStatus.Delayed && state.completeTimestamp) {
+    const formattedTime = getDateTimeFormat().format(
+      new Date(state.completeTimestamp)
+    );
+    return (
+      <>
+        <p>{t('Amount over delay threshold')}</p>
```
candida-d, line 242: This error could help a bit more. Maybe something like "The amount you're withdrawing has triggered a time delay"? (Assuming that's correct)

bglownia (approved):

PR #1181, opened 2022-08-29
Chore/theme setup env config

bglownia:
# Description ℹ️
Two issues fixed
- /assets/env-config.js was returning 404 
- when user is using dark theme sometime you could notice that app loaded in light theme and switched to dark


dexturr (approved):

apps/trading/pages/_app.page.tsx
```diff
@@ -93,9 +93,14 @@ function VegaTradingApp(props: AppProps) {
...
           />
           <link rel="stylesheet" href="https://static.vega.xyz/fonts.css" />
+          <script
+            src="https://static.vega.xyz/theme.js"
+            type="text/javascript"
+            async
+          />
           {['1', 'true'].includes(process.env['NX_USE_ENV_OVERRIDES'] || '') ? (
             /* eslint-disable-next-line @next/next/no-sync-scripts */
-            <script src="/assets/env-config.js" type="text/javascript" />
```
dexturr, line 98: chore (blocking): this needs to be this for the docker setup.

dexturr (changes requested):
This will break the current docker setup

apps/static/src/theme.js
```diff
@@ -0,0 +1,9 @@
+(function () {
```
mattrussell36, line 1: question (blocking): I might be missing something here but why cant the useThemeSwitcher hook do this?

bglownia: Theme switcher works with delay. In my browser I can notice that theme is switched after header is rendered.

mattrussell36: Oh yeah I see it. Wondering if in future there is a way we can inline the script to apply the initial class rather than an external script that needs to be fetched.

bglownia: Sure we can inline it. but since we got couple applications I thinks it's just easier to add it as external script loaded async than trying to maintain it in every application separately.
Second thing is that we want to have theme class to appear asap, to avoid any flickering (ie, change body background color). Loading application takes some time, and there is long "no theme applied period" currently.

dexturr (approved):

radoslaw-szpiech (approved):

PR #1182, opened 2022-08-30
test: dissassociation tests reworked to current implementation

radoslaw-szpiech:
# Description ℹ️

Disassociation tests were updated to handle token dropdown.

### DIFF
--- a/apps/token-e2e/src/integration/flow/token-association-flow.cy.js
+++ b/apps/token-e2e/src/integration/flow/token-association-flow.cy.js
@@ -17,6 +17,7 @@ const vestingContractSection = '[data-testid="vega-in-vesting-contract"]';
 const vegaInWalletSection = '[data-testid="vega-in-wallet"]';
 const associatedKey = '[data-test-id="associated-key"]';
 const associatedAmount = '[data-test-id="associated-amount"]';
+const disassocitiationWarning = '[data-testid="disassociation-warning"]';
 
 context(
   'Token association flow - with eth and vega wallets connected',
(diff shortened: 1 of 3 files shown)


asiaznik (approved):

dexturr (approved):

AndyWhiteVega (commented):
It feels to me that we are trying to do too much in one test.
The test: Able to associate & disassociate both wallet and vesting contract tokens
I feel this should be broken down into 4 x separate tests.
This helps with knowing at a glance what is wrong when a test breaks, and having assurance that when a particular test has failed other areas (in this case the other three are working correctly)

radoslaw-szpiech:
> It feels to me that we are trying to do too much in one test. The test: Able to associate & disassociate both wallet and vesting contract tokens I feel this should be broken down into 4 x separate tests. This helps with knowing at a glance what is wrong when a test breaks, and having assurance that when a particular test has failed other areas (in this case the other three are working correctly)

Yes, ideally association should be done completely separately, not a part of test, but it's not possible right now (I'm hoping it would be easier with checkpoints). To check possibility to choose token which I need to disassociate from I need to have it in both vesting & wallet, so two associations are needed & disassociations to check both options.

AndyWhiteVega:
> 

Understood - Maybe we should rethink the test titles in that case?

Since we now have:
Able to associate tokens
Able to disassociate tokens
Able to associate & disassociate both wallet and vesting contract tokens

Could consider:
Able to associate wallet tokens
Able to disassociate wallet tokens
Able to associate vesting contract tokens
Able to disassociate vesting contract tokens
Able to associate & disassociate wallet tokens when vesting contract tokens present
Able to associate & disassociate vesting contract tokens when wallet tokens present

I know this is more work and will take much longer to run - but since they run overnight !!

PR #1192, opened 2022-08-30
chore: filter assets by status (1108)

### LINKED ISSUE #1108: chore: filter assets by status for deposits and withdrawals
Check to see if the deposit and withdraw form shows assets that have not been whitelisted yet (or completed governance)

Asset.status will show is pending if the asset is not ready for depositing. 
the forms should not list assets that are not ready for use. only `STATUS_ENABLED`

```
STATUS_PROPOSED
Asset is proposed to be added to the network

STATUS_REJECTED
Asset has been rejected

STATUS_PENDING_LISTING
Asset is pending listing on the ethereum bridge

STATUS_ENABLED
Asset can be used on the Vega networ
```

asiaznik:
# Related issues 🔗

Closes #1108 

# Description ℹ️

* added filtering of assets tome sure that only ENABLED are available for deposits/withdrawals
* refactored a bit and moved assets queries to `libs/react-helper`


radoslaw-szpiech (approved):

libs/assets/src/lib/asset-details-dialog.tsx
```diff
@@ -76,9 +49,10 @@ export const AssetDetailsDialog = ({
   open,
   onChange,
 }: AssetDetailsDialogProps) => {
-  const { data } = useQuery<AssetsConnection>(ASSETS_CONNECTION_QUERY);
+  const { data } = useAssetsQuery();
   const symbol =
     typeof assetSymbol === 'string' ? assetSymbol : assetSymbol.symbol;
+  console.log(assetSymbol);
```
MadalinaRaicu, line 55: is this left on purpose?

asiaznik: 😨

apps/trading/pages/portfolio/withdraw/withdraw-page-container.tsx
```diff
@@ -59,10 +59,11 @@ export const WithdrawPageContainer = ({
         skip: !keypair?.pub,
       }}
       render={(data) => {
-        const assets = assetsConnectionToAssets(data.assetsConnection);
+        const assets = getEnabledAssets(data.assetsConnection);
         if (!assets.length) {
           return (
             <Splash>
+              chuje
```
asiaznik, line 66:
```suggestion
```

libs/react-helpers/src/lib/assets.ts
```diff
@@ -1,3 +1,5 @@
+import { AssetStatus } from '@vegaprotocol/types';
```
mattrussell36, line 1: question (non blocking): Wondering if this should be moved to libs/assets?

asiaznik: It certainly can be moved here 👍🏻 it would be more sane to have it here

libs/react-helpers/src/lib/assets.ts
```diff
@@ -16,14 +16,15 @@ export interface Asset {
...
+  STATUS_PENDING_LISTING = 'STATUS_PENDING_LISTING',
+  STATUS_PROPOSED = 'STATUS_PROPOSED',
+  STATUS_REJECTED = 'STATUS_REJECTED',
 }
 
-interface AssetsConnection {
-  __typename: 'AssetsConnection';
-  edges: (AssetEdge | null)[] | null;
+export interface AssetWithStatus extends Asset {
+  status: AssetStatus;
```
dexturr, line 27: thought (non-blocking): wondering if we should just be using a basic interface here with status. The asset type requires symbol, decimals ect. none of which are used in `getEnabledAssets` so not sure we need to be extending that for a function that does not require those fields to work.

dexturr (approved):

libs/react-helpers/src/lib/assets.ts
asiaznik: True we do need the "full" type there but because it returns `Asset[]` the conversion from just `{ state: AssetStatus }[]` to `Asset[]` requires middle step of converting it to `unknown` first which imo looks bad (`as unknown as Asset[]`)

PR #1212, opened 2022-08-31
ci: check for draft pr for cypress test run

mattrussell36:
# Description ℹ️

Adds a condition to check if a PR is in draft before triggering the capsule build and test run. Once its changed to ready for review the run will be triggered

### DIFF
--- a/.github/workflows/capsule-cypress.yml
+++ b/.github/workflows/capsule-cypress.yml
@@ -11,7 +11,7 @@ jobs:
   pr:
     name: Run capsule tests - PR
     runs-on: ubuntu-latest
-    if: ${{ github.event_name == 'pull_request' }}
+    if: ${{ github.event_name == 'pull_request' && github.event.pull_request.draft == false }}
     env:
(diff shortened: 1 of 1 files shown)


jtsang586 (approved):

.github/workflows/capsule-cypress.yml
```diff
@@ -6,12 +6,17 @@ on:
       - master
       - develop
   pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - ready_for_review
```
dexturr, line 13: question: why are these now needed? Isn't the check on 19 sifficeint?

mattrussell36: We still need the run to be triggered when the already opened PR switches from draft to ready to review

mattrussell36: I also believe that the other types are required now that `types` are used, though I couldn't quite tell from github documentation

dexturr (approved):

PR #1222, opened 2022-09-01
Feat/707 link to market proposal

### LINKED ISSUE #707: Link to market proposal from market
A active or potential trader of a market may wish to find out all they can about how a market is settled. To do this they will need to look at the oracle. https://github.com/vegaprotocol/frontend-monorepo/issues/626 

However: It is possible that the on-Vega-chain oracle data will not satisfy the users curiosity. More information may be available on the proposal that brought the market into life, or the forum posts where it was discussed.

This ticket is to enable a user journey from A market info in Console: through to the Proposal details (possibly on block explorer or TFE), where ideally someone described and verified the oracle source.

FYI: #633 

# MVP
Under a new Governance section
- List origin proposal ID... link to block explorer
- List all successfully enacted market change proposals for this market... link to block explorer

Consider
- adding text prompts next to the oracle ID's "Check market proposal for detail on oracle"
- pulling in to console relevant data from a proposal

Note: this ticket is probably a minimum before [this](https://github.com/vegaprotocol/frontend-monorepo/issues/581) is taken on properly:

asiaznik:
# Related issues 🔗

Closes #707

# Description ℹ️

Added link to market proposal under market info tab.

# Demo 📺

Screenshot 2022-09-01 at 18 00 01

# Technical 👨‍🔧

* added new `NX_VEGA_TOKEN_URL` to environment configs (needed for proposal page url generation)
* although the proposal id seems to be the same as market id I've chosen to take it from underlying proposal node anyway
* added new component `ExternalLink` which renders with ~~`external`~~ `share` icon.
* applied new `ExternalLink` to the link to governance in token app:
link to governance

### DIFF
--- a/libs/ui-toolkit/src/components/link/link.tsx
+++ b/libs/ui-toolkit/src/components/link/link.tsx
@@ -1,5 +1,7 @@
+import { IconNames } from '@blueprintjs/icons';
 import classNames from 'classnames';
 import type { ReactNode, AnchorHTMLAttributes } from 'react';
+import { Icon } from '../icon';
 
 type LinkProps = AnchorHTMLAttributes<HTMLAnchorElement> & {
   children?: ReactNode;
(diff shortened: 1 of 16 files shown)


notbot00 (approved):

libs/ui-toolkit/src/components/link/link.tsx
```diff
@@ -27,5 +29,27 @@ export const Link = ({ className, children, ...props }: LinkProps) => {
     </a>
   );
 };
-
 Link.displayName = 'Link';
+
+export const ExternalLink = ({ children, className, ...props }: LinkProps) => (
```
sam-keen, line 34: Suggestion (non-blocking): I'm not convinced this component is needed. I do like the use of the icon as a consistent feature of external links, but I'd rather see the Link component extended. Perhaps an optional boolean `external` prop that adds 'target="_blank"' and the icon?

sam-keen (approved):

dexturr (approved):

PR #1254, opened 2022-09-06
Feature/1243 market data providers refactoring

### LINKED ISSUE #1243: Refactor trading app market related data handling
- [ ] Create markets data provider which will hold all static (non ticking) market data
- [ ] Create market data provider which will provide single market from above data provider
- [ ] Create market candles data provider
- [ ] Create market data data provider

bglownia:
# Related issues 🔗

Closes #1243 

# Description ℹ️

Refactoring of data providers. Instead of querying same data many times create data providers which will be used as a source of  any market related data.

# Technical 👨‍🔧

Very important part of this change is to align with changes in API and stop using deprecated gql nodes.
- markets replaced by marketsConnection.edges.nodes
- markets.candles replaced by marketsConnection.edges.nodes.candlesConnection 
- marketData subscription replaced with marketsData
- marketDepthUpdate replaced with marketsDepthUpdate

It's not possible to subscribe for market data changes for all markets it can be done only for selected markets.
All queries for many markets executed via data provider don't subscribe for updates.  Queries for market, market data and market candles in separate providers  so we can have different strategy for cache and update. 
Before we have subscription for real time data (candles, and marketData) for all markets. Which is not great idea.

Providers for single market candles and data has subscription. providers are also separated. Updates in market depth subscription required  to handle separately market depth updates and market data updates for orderbook and market depth chart.


apps/trading/pages/index.page.tsx
```diff
@@ -8,7 +9,10 @@ export function Index() {
   const { replace } = useRouter();
   // The default market selected in the platform behind the overlay
   // should be the oldest market that is currently trading in continuous mode(i.e. not in auction).
-  const { data, error, loading } = useMarketList();
+  const { data, error, loading } = useDataProvider({
+    dataProvider: activeMarketsDataProvider,
+    update: () => true,
```
mattrussell36, line 14: question (non blocking): Does it make sense to provide a sensible default update function so this doesn't have to be passed in everywhere?

libs/market-list/src/lib/market-data-provider.ts
```diff
@@ -0,0 +1,14 @@
+import { makeDerivedDataProvider } from '@vegaprotocol/react-helpers';
+
+import type { Market } from './';
+import { marketsDataProvider } from './';
+
+export const marketDataProvider = makeDerivedDataProvider<Market>(
+  [(callback, client, variables) => marketsDataProvider(callback, client)],
+  ([markets], variables) =>
```
mattrussell36, line 8: chore (non blocking): If find this hard to read. Can we simplify with some nice clear if statements

asiaznik (approved):

PR #1260, opened 2022-09-06
feat: added asset details dialog to market select (1226)

### LINKED ISSUE #1226: Add asset details dialogs to the new market switcher
Add asset details dialogs to the new market switcher as it was done for the old one.

asiaznik:
# Related issues 🔗

Closes #1226 

# Description ℹ️

Added asset details dialog to market select making "Asset" column now clickable.

# Demo 📺

Screenshot 2022-09-06 at 15 49 44

# Technical 👨‍🔧

* added `onCloseAutoFocus` to dialog in order to make sure that focus goes back to the dialog's trigger
* added a handler for popover's `onInteractOutside` which's responsible for not closing itself when a dialog open from within

### DIFF
--- a/libs/assets/src/lib/asset-details-dialog.tsx
+++ b/libs/assets/src/lib/asset-details-dialog.tsx
@@ -16,19 +16,38 @@ import create from 'zustand';
 export type AssetDetailsDialogStore = {
   isAssetDetailsDialogOpen: boolean;
   assetDetailsDialogSymbol: string | Asset;
+  assetDetailsDialogTrigger: HTMLElement | null | undefined;
   setAssetDetailsDialogOpen: (isOpen: boolean) => void;
   setAssetDetailsDialogSymbol: (symbol: string | Asset) => void;
+  setAssetDetailsDialogTrigger: (
+    trigger: HTMLElement | null | undefined
+  ) => void;
+  openAssetDetailsDialog: (
+    symbol: string | Asset,
+    trigger?: HTMLElement | null
+  ) => void;
 };
 
 export const useAssetDetailsDialogStore = create<AssetDetailsDialogStore>(
   (set) => ({
     isAssetDetailsDialogOpen: false,
     assetDetailsDialogSymbol: '',
-    setAssetDetailsDialogOpen: (isOpen: boolean) => {
+    assetDetailsDialogTrigger: null,
+    setAssetDetailsDialogOpen: (isOpen) => {
       set({ isAssetDetailsDialogOpen: isOpen });
     },
-    setAssetDetailsDialogSymbol: (symbol: string | Asset) => {
+    setAssetDetailsDialogSymbol: (symbol) => {
       set({ assetDetailsDialogSymbol: symbol });
+    },
+    setAssetDetailsDialogTrigger: (trigger) => {
+      set({ assetDetailsDialogTrigger: trigger });
+    },
+    openAssetDetailsDialog: (symbol, trigger?) => {
+      set({
+        isAssetDetailsDialogOpen: true,
+        assetDetailsDialogSymbol: symbol,
+        assetDetailsDialogTrigger: trigger,
+      });
     },
   })
 );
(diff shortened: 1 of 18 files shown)


libs/assets/src/lib/asset-details-dialog.tsx
```diff
@@ -19,20 +19,39 @@ import create from 'zustand';
...
+  ) => void;
 };
 
 export const useAssetDetailsDialogStore = create<AssetDetailsDialogStore>(
   (set) => ({
     isAssetDetailsDialogOpen: false,
     assetDetailsDialogSymbol: '',
-    setAssetDetailsDialogOpen: (isOpen: boolean) => {
+    assetDetailsDialogTrigger: null,
+    setAssetDetailsDialogOpen: (isOpen) => {
```
mattrussell36, line 39: chore (non blocking): I refactored the global store to just use a single update function which should be sufficient here too.

libs/assets/src/lib/asset-details-dialog.tsx
```diff
@@ -183,10 +204,25 @@ export const AssetDetailsDialog = ({
       icon={<Icon name="info-sign"></Icon>}
       open={open}
       onChange={(isOpen) => onChange(isOpen)}
+      onCloseAutoFocus={(e) => {
+        /**
+         * This mimics radix's default behaviour that focuses the dialog's
+         * trigger after closing itself
+         */
+        if (trigger) {
```
mattrussell36, line 212: question (blocking): why do we need to recreate radix's behaviour here? Can we not rely on that?

libs/ui-toolkit/src/components/dialog/dialog.tsx
```diff
@@ -59,7 +61,10 @@ export function Dialog({
           className="fixed inset-0 bg-black/50 z-10"
           data-testid="dialog-overlay"
         />
-        <DialogPrimitives.Content className={contentClasses}>
+        <DialogPrimitives.Content
+          className={contentClasses}
+          onCloseAutoFocus={(e) => onCloseAutoFocus?.(e)}
```
mattrussell36, line 66: chore (non blocking): passing undefined here should be fine

```suggestion
          onCloseAutoFocus={onCloseAutoFocus}
```

libs/ui-toolkit/src/components/popover/popover.tsx
```diff
@@ -14,6 +15,12 @@ export const Popover = ({
   open,
   onChange,
 }: PopoverProps) => {
+  const popoverContent = useRef<HTMLDivElement>(null);
+  /**
+   * It's responsible for preventing auto-closing of the popover when
+   * `focusOutside` event is originated from within (e.g. opens dialog)
```
mattrussell36, line 21: question (non blocking): I'm surprised that this isn't the default behavior of the popover component. Are we perhaps doing something wrong somewhere?

mattrussell36 (commented):
Code wise it all looks good to me. I don't love having to add all the focus handling, radix normally does a good job of this out of the box. Why are we having to do so much of that here?

mattrussell36:
@asiaznik I've dug a bit deeper into the code and I have a few thoughts.

- I don't think the focus handling is worth the extra complexity right now and I'd prefer to just let the market popover close when you open the dialog, if the popover _must_ remain open lets use react state to handle it rather than passing elements and symbols around. You might want to combine the asset dialog store into the global store.
- Secondly (and apologies this is less to do with this PR and more with the dialog as a whole), right now all assets are fetched as soon as the page loads, we then find the selected asset by symbol. IMO we can simplify this and prevent some over-fetching by passing the asset id to the dialog and using a slightly different query to only fetch that asset, and we should do this only once the dialog content is rendered.

This query should suffice
```
query Asset($assetId: ID!) {
  asset(assetId: $assetId) {
     ... asset fields
  }
}
```

libs/ui-toolkit/src/components/popover/popover.tsx
asiaznik: It should be like that but because there's not radix's trigger inside the popover it doesn't know it originated from within.

libs/assets/src/lib/asset-details-dialog.tsx
asiaznik: We can't because we don't use radix's trigger 😞 for the asset details dialog.

libs/ui-toolkit/src/components/popover/popover.tsx
asiaznik: But I agree with your comment that it adds complexity that we do not need at the moment. I'm happy to remove that part.

mattrussell36 (approved):

Issue #1083, opened 2022-08-19
Update deprecated type generation


notbot00:
Apollo's built in type generation seems to be soon removed, we might want to switch to the recommended `graphql-code-generator` library instead

<img width="701" alt="Screenshot 2022-08-19 at 11 01 17" src="https://user-images.githubusercontent.com/105208209/185595351-6d8b1039-d0ee-4ad0-a6ea-d6e06831100c.png">

PR #1266, opened 2022-09-07
Feat/migrate libs

notbot00:
# Related issues 🔗

Implements #1083 

# Description ℹ️

- migrates three libs to use the new generated type system (`accounts`, `assets`, `candles-chart`).
- adds a temporary cleanup script to handle the clashing generations in the migrated libs.


macqbat (approved):

asiaznik:
suggestion: How about enforcing the naming convention? Having it set to:

```yml
config:
  preResolveTypes: true
  namingConvention:
    enumValues: keep
    typeNames: 'change-case-all#pascalCase'
```

will deal with inconsistancies such as: `PartyliquidityProvisionsArgs` 👉🏻 `PartyLiquidityProvisionsArgs` or `QueryoracleSpecsConnectionArgs` 👉🏻  `QueryOracleSpecsConnectionArgs`

libs/types/codegen.yml
```diff
@@ -9,21 +9,21 @@ config:
 generates:
   libs/types/src/__generated__/types.ts:
     - typescript
-  libs/**/__generated__:
+  libs/**:
```
asiaznik, line 12: ☝🏻☝🏻☝🏻☝🏻☝🏻 

suggestion: How about enforcing the naming convention? Having it set to:

```yml
config:
  preResolveTypes: true
  namingConvention:
    enumValues: keep
    typeNames: 'change-case-all#pascalCase'
```

will deal with inconsistancies such as: `PartyliquidityProvisionsArgs` 👉🏻 `PartyLiquidityProvisionsArgs` or `QueryoracleSpecsConnectionArgs` 👉🏻  `QueryOracleSpecsConnectionArgs`

notbot00: to me it looks like that is the default setting (https://www.the-guild.dev/graphql/codegen/docs/config-reference/naming-convention):
Screenshot 2022-09-08 at 18 52 14

libs/types/apollo.config.js
```diff
@@ -11,6 +11,12 @@ module.exports = {
       '../../apps/token/client.graphql',
       '../../apps/trading/client.graphql',
     ],
-    excludes: ['**/generic-data-provider.ts', '**/__generated___/*'],
+    excludes: [
+      '**/generic-data-provider.ts',
+      '**/__generated___/*',
+      '../../libs/accounts/**',
```
mattrussell36, line 17: question (non blocking): Why ignore these libs?

mattrussell36 (approved):

MadalinaRaicu (approved):

libs/types/apollo.config.js
notbot00: because they have been migrated, and this is the configuration for the old types gen. we want to stop generating the oldschool types for these. this config file will be removed when all libs and apps have been migrated.

PR #1283, opened 2022-09-07
Feat/1089 override slippage

### LINKED ISSUE #1089: Override slippage percentage
- [x] It should allow you to override slippage amount by clicking a button link next to it (similar to setting size input)
- [x] It should switch to limit order if the user overrides
- [x] It should become a IOC limit order
- [x] It should always be an positive value
- [x] It should set the default amount to the slippage percentage
- [x] It should have a minimum of 0%
- [x] It should have a maximum of 50% for both
- [x] It should update the slippage amount
- [x] It should use the following price for long trades: `best_ask_price * (1 + max_slippage_percentage)`
- [x] It should use the following price for short trades: `best_bid_price * (1 - max_slippage_percentage)`
- [x] - [x] It should change price based on size selection using the VWAP

elmar-vega:
# Related issues 🔗

Closes #1089 

# Description ℹ️

Feature to display and override slippage. Use slippage to display dynamic price changes.

# Demo 📺

Screenshot 2022-09-07 at 16 37 04
Screenshot 2022-09-07 at 16 36 48
Screenshot 2022-09-07 at 16 37 24
Screenshot 2022-09-07 at 16 37 11

# Technical 👨‍🔧

- Refactored deal ticket steps
- Fixed broken tests and added new e2e tests
- Created TrafficLight component to dynamically update colour based on value
- Refactored DealTicketSize to make it more simple

### DIFF
--- a/apps/console-lite/src/app/components/deal-ticket/deal-ticket-estimates.tsx
+++ b/apps/console-lite/src/app/components/deal-ticket/deal-ticket-estimates.tsx
@@ -4,6 +4,7 @@ import { t } from '@vegaprotocol/react-helpers';
 import { Icon, Tooltip } from '@vegaprotocol/ui-toolkit';
 import { IconNames } from '@blueprintjs/icons';
 import * as constants from './constants';
+import { TrafficLight } from '../traffic-light';
 
 interface DealTicketEstimatesProps {
   quoteName?: string;
@@ -13,6 +14,7 @@ interface DealTicketEstimatesProps {
   fees?: string;
   notionalSize?: string;
   size?: string;
+  slippage?: string;
 }
 
 interface DataTitleProps {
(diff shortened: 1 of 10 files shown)


mattrussell36 (approved):

apps/console-lite/src/app/components/deal-ticket/deal-ticket-estimates.tsx
```diff
@@ -13,29 +14,36 @@ interface DealTicketEstimatesProps {
...
-  value: string;
+  value?: string;
+  children?: ReactNode;
   description: string;
   id?: string;
 }
 
-const ValueTooltipRow = ({ value, description, id }: ValueTooltipProps) => (
+export const ValueTooltipRow = ({
+  value,
```
macqbat, line 40: non-blocking: imho instead 
```
value?: string;
children? ReactNode;
```
clearer would be:

`value: string | ReactNode;`

macqbat (approved):

apps/console-lite/src/app/components/deal-ticket/deal-ticket-estimates.tsx
elmar-vega: It's nicer to use children, because it can best nested and looks cleaner and more legible than having to stuff it into value

PR #1294, opened 2022-09-08
Test/626 oracle info

### LINKED ISSUE #626: Add Oracles info to Console V2

jtsang586:
# Related issues 🔗

Closes #626

# Description 

- Updates tests to include additional market info
- Added assertions for new withdrawal delay threshold text
- Change cypress browser to electron to try and improve intermittent failing tests

### DIFF
--- a/apps/trading-e2e/src/integration/market-info.cy.ts
+++ b/apps/trading-e2e/src/integration/market-info.cy.ts
@@ -4,6 +4,7 @@ import { mockTradingPage } from '../support/trading';
 const marketInfoBtn = 'Info';
 const row = 'key-value-table-row';
 const marketTitle = 'accordion-title';
+const externalLink = 'external-link';
 
 describe('market info is displayed', () => {
   before(() => {
@@ -25,9 +26,9 @@ describe('market info is displayed', () => {
 
   it('market price', () => {
     cy.getByTestId(marketTitle).contains('Market price').click();
-    validateMarketDataRow(0, 'Mark Price', '57.49');
-    validateMarketDataRow(1, 'Best Bid Price', '6,817.65');
-    validateMarketDataRow(2, 'Best Offer Price', '6,817.69');
+    validateMarketDataRow(0, 'Mark Price', '0.05749');
+    validateMarketDataRow(1, 'Best Bid Price', '6.81765 ');
+    validateMarketDataRow(2, 'Best Offer Price', '6.81769 ');
   });
 
   it('market volume displayed', () => {
@@ -71,11 +72,7 @@ describe('market info is displayed', () => {
 
     validateMarketDataRow(0, 'Name', 'tBTC TEST');
     validateMarketDataRow(1, 'Symbol', 'tBTC');
-    validateMarketDataRow(
-      2,
-      'Asset ID',
-      '5cfa87844724df6069b94e4c8a6f03af21907d7bc251593d08e4251043ee9f7c'
-    );
+    validateMarketDataRow(2, 'Asset ID', 'market-0');
   });
 
   it('metadata displayed', () => {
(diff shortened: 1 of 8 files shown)


asiaznik (approved):

apps/trading-e2e/src/integration/market-info.cy.ts
```diff
@@ -129,6 +126,21 @@ describe('market info is displayed', () => {
     validateMarketDataRow(2, 'Scaling Factor', '10');
   });
 
+  it('liquidity displayed', () => {
+    // cy.getByTestId(marketTitle).contains('Liquidity').last().click();
```
sam-keen, line 130: Chore (non-blocking): Unless there's a compelling reason, might as well remove this commented-out code

apps/trading-e2e/src/integration/market-info.cy.ts
```diff
@@ -148,6 +160,22 @@ describe('market info is displayed', () => {
       'Termination Oracle',
       'f028fe5ea7de3890962a05a7163fdde562629af649ed81b8c8902fafb6eef04f'
     );
+
+    cy.getByTestId(externalLink)
+      .should('have.attr', 'href')
+      .and('contain', 'https://explorer.fairground.wtf/oracles');
```
sam-keen, line 166: Question (non-blocking): Should these urls be environment-aware? I'm wondering if we'd ever want to run these tests against mainnet

sam-keen (approved):

apps/trading-e2e/src/integration/market-info.cy.ts
jtsang586: Removed

apps/trading-e2e/src/integration/market-info.cy.ts
jtsang586: That's a valid point. I've loosened the assertion so it should still pass when if we ever want to run these tests on another environment.

PR #1307, opened 2022-09-12
chore: env changed to custom

radoslaw-szpiech:
NX_VEGA_ENV changed to CUSTOM in .env.capsule, to avoid network error popup when built locally

### DIFF
--- a/apps/token/.env.capsule
+++ b/apps/token/.env.capsule
@@ -1,5 +1,5 @@
 # App configuration variables
-NX_VEGA_ENV=TESTNET
+NX_VEGA_ENV=CUSTOM
 NX_ETHEREUM_PROVIDER_URL=http://localhost:8545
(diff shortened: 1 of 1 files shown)


apps/token/.env.capsule
```diff
@@ -1,5 +1,5 @@
 # App configuration variables
-NX_VEGA_ENV=TESTNET
+NX_VEGA_ENV=CUSTOM
```
jtsang586, line 2: This should be changed in the token-e2e project

radoslaw-szpiech: Somehow there's no problem with e2e, but when I built just token app locally

jtsang586: We only need the e2e project to build to capsule for the tests. Token should build against testnet by default. Or it might be stagnet 3 soon but it should not be capsule.

jtsang586 (approved):

PR #1312, opened 2022-09-12
chore: rename cell renderers to follow ag name convention

### LINKED ISSUE #1299: chore: Ag naming convention
I have nothing against this. But, I will need to change all cell renderer names to follow this naming convention, for consistency.

_Originally posted by @MadalinaRaicu in https://github.com/vegaprotocol/frontend-monorepo/pull/1267#discussion_r966146503_

MadalinaRaicu:
# Related issues 🔗

Closes #1299

# Description ℹ️ &  Technical 👨‍🔧

- [x] add Ag prefix to cell renderers


dexturr (approved):

libs/react-helpers/src/lib/grid/cumulative-vol-cell.tsx
```diff
@@ -13,7 +13,7 @@ export interface CumulativeVolProps {
   className?: string;
 }
 
-export interface ICumulativeVolCellProps extends ICellRendererParams {
+export interface AgCumulativeVolCellProps extends ICellRendererParams {
```
mattrussell36, line 16: chore (blocking): I think we are crossing concerns here. We need to differentiate what is used in an ag grid and what is not. This particular component is only used in the order book.

Can we
a) not rename it to AgGridCumulative...
b) co locate it with the order book

libs/react-helpers/src/lib/grid/price-cell.tsx
```diff
@@ -1,13 +1,13 @@
...
-export interface IPriceCellProps {
+export interface AgPriceCellProps {
   value: number | bigint | null | undefined;
   valueFormatted: string;
   testId?: string;
 }
 
-export const PriceCell = React.memo(
-  ({ value, valueFormatted, testId }: IPriceCellProps) => {
+export const AgPriceCell = React.memo(
```
mattrussell36, line 9: chore (blocking): This component is not always used within ag grid so renaming here doesn't make sense

libs/react-helpers/src/lib/grid/size.tsx
```diff
@@ -26,11 +26,11 @@ export const Size = ({
   );
 };
 
-export interface ISizeCellProps extends ICellRendererParams {
+export interface AgSizeCellProps extends ICellRendererParams {
   value: number | string;
 }
 
-export const SizeCell = ({ value, data }: ISizeCellProps) => {
+export const AgSizeCell = ({ value, data }: AgSizeCellProps) => {
```
mattrussell36, line 33: chore (blocking): This doesn't appear to be used anywhere. Can we delete

libs/react-helpers/src/lib/grid/vol-cell.tsx
```diff
@@ -42,16 +42,16 @@ export const Vol = React.memo(
...
+        <AgPriceCell value={value} valueFormatted={valueFormatted} />
       </div>
     );
   }
 );
 
 Vol.displayName = 'Vol';
 
-export const VolCell = ({ value, valueFormatted }: IVolCellProps) => (
+export const AgVolume = ({ value, valueFormatted }: AgVolCellProps) => (
```
mattrussell36, line 53: chore (blocking): Doesn't appear that this is used in an ag grid only the order book

libs/ui-toolkit/src/components/price-change/price-change-cell.tsx
```diff
@@ -36,8 +36,8 @@ export const priceChange = (candles: string[]) => {
     : 0;
 };
 
-export const PriceCellChange = React.memo(
-  ({ candles, decimalPlaces }: PriceChangeCellProps) => {
+export const AgPriceCellChange = React.memo(
```
mattrussell36, line 39: chore (blocking): Used in select market tables not ag grid

libs/ui-toolkit/src/components/tooltip/tooltip.tsx
```diff
@@ -58,7 +58,7 @@ export const Tooltip = ({
     children
   );
 
-export const TooltipCellComponent = (props: ITooltipParams) => {
+export const AgTooltip = (props: AgTooltipProps) => {
```
mattrussell36, line 61: chore (non blocking): I don't think we should be mixing our uitoolkit tooltip with something more specific to ag grid

mattrussell36 (changes requested):
I don't love renaming everything to AgXComponent. Most of the components are used outside of ag grid as well and this naming convention confuses things.

Can we go through this again and
A) only rename things to AgX if they are truly only required for AgGrid cells, otherwise they should be left to be used in a generic mannger across different table types
B) If there are any AgGrid specific components required, can we locate them in ui-toolkit/components/ag-grid

libs/positions/src/lib/positions-table.tsx
```diff
@@ -327,7 +329,7 @@ export const PositionsTable = forwardRef<AgGridReact, Props>(
             node,
           }: PositionsTableValueFormatterParams & {
```
mattrussell36, line 330: chore (non blocking): I also added some type helpers to ui-toolkit/components/ag-grid which make defining types here a bit cleaner

libs/react-helpers/src/lib/grid/size.tsx
MadalinaRaicu: my plan was to use it in grids such as order list etc

PR #1313, opened 2022-09-12
chore: use trading header and tidy tailwind usage

mattrussell36:
# Description ℹ️

Follow up from https://github.com/vegaprotocol/frontend-monorepo/pull/1305.

- Removed some tailwind classes that were having no effect.
- Changed trade page to use the Header component.
- Adds custom class `border-default` for border colors so we dont have to repeat `border-neutral-300 dark:border-neutral-600` everywhere.
- Fixes theme switching in orderbook, borders that are rendered conditionally in JS where not using the theme context and therefor not changing value.
- Remove psuedo elements from tabs as they were skewing the text alignment
- Rename fixedBg prop to theme

### DIFF
--- a/apps/trading/components/navbar/navbar.tsx
+++ b/apps/trading/components/navbar/navbar.tsx
@@ -20,7 +20,7 @@ export const Navbar = ({ theme, toggleTheme }: NavbarProps) => {
   }));
   const tradingPath = marketId ? `/markets/${marketId}` : '/markets';
   return (
-    <div className="px-4 flex items-stretch border-b border-neutral-300 dark:border-neutral-400 bg-black">
+    <div className="px-4 flex items-stretch border-b border-default bg-black text-white">
       <div className="flex gap-4 mr-4 items-center h-full">
         <Link href="/" passHref={true}>
           {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
@@ -28,7 +28,7 @@ export const Navbar = ({ theme, toggleTheme }: NavbarProps) => {
             <Vega className="w-13" />
           </a>
         </Link>
-        <NetworkSwitcher fixedBg="dark" />
+        <NetworkSwitcher theme="dark" />
       </div>
       <nav className="flex items-center">
         {[
--- a/libs/environment/src/components/network-switcher/network-switcher.tsx
+++ b/libs/environment/src/components/network-switcher/network-switcher.tsx
@@ -69,11 +69,7 @@ const NetworkLabel = ({
   </span>
 );
 
-export const NetworkSwitcher = ({
-  fixedBg,
-}: {
-  fixedBg?: 'dark' | 'light';
-}) => {
+export const NetworkSwitcher = ({ theme }: { theme?: 'dark' | 'light' }) => {
   const { VEGA_ENV, VEGA_NETWORKS } = useEnvironment();
   const [isOpen, setOpen] = useState(false);
   const [isAdvancedView, setAdvancedView] = useState(false);
@@ -88,10 +84,9 @@ export const NetworkSwitcher = ({
     [setOpen, setAdvancedView]
   );
 
-  const dropdownTriggerClasses = classNames('hover:!bg-neutral-700', {
-    'dark:text-white dark:bg-black text-black bg-white': !fixedBg,
-    'text-black bg-white': fixedBg === 'light',
-    'text-white bg-black': fixedBg === 'dark',
+  const dropdownTriggerClasses = classNames({
+    'text-black hover:!bg-neutral-300': theme === 'light',
+    'text-white hover:!bg-neutral-700': theme === 'dark',
   });
 
   return (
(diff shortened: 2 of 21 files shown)


dexturr (approved):

libs/environment/src/components/network-switcher/network-switcher.tsx
```diff
@@ -88,10 +84,9 @@ export const NetworkSwitcher = ({
     [setOpen, setAdvancedView]
   );
 
-  const dropdownTriggerClasses = classNames('hover:!bg-neutral-700', {
-    'dark:text-white dark:bg-black text-black bg-white': !fixedBg,
-    'text-black bg-white': fixedBg === 'light',
-    'text-white bg-black': fixedBg === 'dark',
+  const dropdownTriggerClasses = classNames({
+    'text-black hover:!bg-neutral-300': theme === 'light',
```
PeteWilliams, line 88: Should be 700 for both themes as the header is black for both. Otherwise it does this on light theme:

image

apps/trading/components/navbar/navbar.tsx
```diff
@@ -20,15 +20,15 @@ export const Navbar = ({ theme, toggleTheme }: NavbarProps) => {
   }));
   const tradingPath = marketId ? `/markets/${marketId}` : '/markets';
   return (
-    <div className="px-4 flex items-stretch border-b border-neutral-300 dark:border-neutral-400 bg-black">
```
PeteWilliams, line 23: This line was deliberately lighter than the dark theme default so the header is clearly separate to the content below:
image

libs/positions/src/lib/positions.tsx
```diff
@@ -82,8 +82,8 @@ export const Positions = memo(
     };
     return (
       <AsyncRenderer loading={loading} error={error} data={data}>
-        <div className="flex justify-between items-center p-3 pb-1">
-          <h4 className="text-m">
+        <div className="flex justify-between items-center px-4 pt-3 pb-1">
```
PeteWilliams, line 85: This was trying to reclaim some space in the positions panel so it can show more positions, but I've got another branch where I'm still fiddling with that.

libs/ui-toolkit/src/components/dropdown-menu/dropdown-menu.tsx
```diff
@@ -29,7 +29,7 @@ export const DropdownMenuTrigger = forwardRef<
     className,
     'text-sm py-1 px-2 rounded bg-transparent border border-neutral-500',
     'focus:border-black dark:focus:border-white whitespace-nowrap',
-    'hover:bg-neutral-100 dark:hover:bg-neutral-700'
```
PeteWilliams, line 32: Think these looked better lighter.

libs/ui-toolkit/src/components/popover/popover.tsx
```diff
@@ -23,7 +23,7 @@ export const Popover = ({
         <PopoverPrimitive.Content
           data-testid="popover-content"
           align="start"
-          className="p-4 rounded bg-neutral-50 dark:bg-neutral-800 dark:text-neutral-200 border border-neutral-400"
+          className="p-4 rounded bg-neutral-100 dark:bg-neutral-800 dark:text-neutral-200 border border-neutral-400"
```
PeteWilliams, line 26: Think these looked better lighter.

libs/market-list/src/lib/components/select-market.tsx
```diff
@@ -76,7 +76,7 @@ export const SelectAllMarketsTableBody = ({
   if (!data) return null;
   return (
     <>
-      <thead className="bg-neutral-50 dark:bg-neutral-800">
```
PeteWilliams, line 79: Think these looked better lighter.

libs/ui-toolkit/src/components/popover/popover.tsx
mattrussell36: The difference is minimal and would rather not introduce an 11th grayscale color right now in the interest of maintenance

libs/positions/src/lib/positions.tsx
mattrussell36: px-4 is required to align the text nicely otherwise it looked a little off.  Top and bottom are the same as before just written slightly clearer

apps/trading/components/navbar/navbar.tsx
mattrussell36: Not sure about this one, for me the difference is so subtle that it doesn't really achieve that and instead just makes it look a little off by introducing a new border color

libs/environment/src/components/network-switcher/network-switcher.tsx
mattrussell36: Not seeing this one

PeteWilliams: Still happening for me:

- Go here https://deploy-preview-1313--vega-trading-app.netlify.app/markets/bb941f84e25b3a06068e33917d01215d56a51e9abff6ff9b9a3f2cf49b495e37
- Switch to light mode
- Connect to wallet
- Hover over dropdown

PR #1317, opened 2022-09-13
Test/governance suite expansion

AndyWhiteVega:
# Related issues 🔗

Closes #[Issue number here]

# Description ℹ️

- Change CI builds to use vega 0.55.0 up from 0.54.0 (only for night run and manual trigger workflows for time being)
- Extra governance tests
- Inclusion of AC's relating to voting and proposals

# Demo 📺

# Technical 👨‍🔧


apps/token-e2e/src/integration/view/wallet-vega.cy.js
```diff
@@ -170,7 +170,7 @@ context('Vega Wallet - verify elements on widget', function () {
       });
     });
 
-    it('should have wallet name visible', function () {
+    it.skip('should have wallet name visible', function () {
```
dexturr, line 173: why do these need to be skipped?

AndyWhiteVega: @dexturr With the recent API changes the wallet_name was no longer supplied and so i believe we stopped the UI from displaying it. There is a ticket to get it re-instated at which point we will unskip these tests

AndyWhiteVega: For now it only shows the key

AndyWhiteVega: Think this will get it back https://github.com/vegaprotocol/vega/issues/6139

PR #1318, opened 2022-09-13
fix: decimal precision out of range

MadalinaRaicu:
# Related issues 🔗

Closes N/A

# Description ℹ️

proposed a new market called Nuggets opening that market in console causes
Application error: a client-side exception has occurred (see the browser console for more information).
this market has 0 decimal places (of any sort)

# Demo 📺

image

# Technical 👨‍🔧

Note: 
image

that is a un-intended consequence. we need to make some improvements to that.
in this situation on other exchanges you can then group by 10s 100s etc


bglownia (approved):

libs/react-helpers/src/lib/format/number.ts
```diff
@@ -32,13 +36,14 @@ export function removeDecimal(value: string, decimals: number): string {
...
 // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat
-export const getNumberFormat = memoize(
-  (digits: number) =>
-    new Intl.NumberFormat(getUserLocale(), {
-      minimumFractionDigits: digits,
-      maximumFractionDigits: digits,
-    })
-);
+export const getNumberFormat = memoize((digits: number) => {
+  if (isNil(digits) || digits < 0)
```
mattrussell36, line 40: chore (blocking):  braces here please

mattrussell36 (changes requested):
One small minor change please

libs/react-helpers/src/lib/format/number.ts
mattrussell36: Added a eslint rule for this https://github.com/vegaprotocol/frontend-monorepo/pull/1322

MadalinaRaicu: updated.

mattrussell36 (approved):

PR #1321, opened 2022-09-13
Stagnet3 API update

### LINKED ISSUE #1213: Update trades query and subscription
Schema is changing so the necessary changes need to be made

https://github.com/vegaprotocol/vega/pull/598

### LINKED ISSUE #1214: Refactor positions subscription
Positions subscription is changing to return multiple updates in in a single payload. We'll need to update the positions data provider.

https://github.com/vegaprotocol/vega/pull/5984

bglownia:
# Related issues 🔗
Closes #1213 
Closes #1214 

# Description ℹ️

This PR updates queries and subscriptions to align with stagnet3. Stagnet3 introduced several breaking api changes, namely the removal of nested market and party fields from subscriptions. Instead subscriptions now return their own bespoke type which means that the frontend must cache additional data (such as decimal place values) from the initial query so we can properly format certain values in the response (often price or size)


macqbat (approved):

libs/trades/src/lib/trades-container.tsx
```diff
@@ -6,26 +6,21 @@ import { AsyncRenderer } from '@vegaprotocol/ui-toolkit';
...
 import { TradesTable } from './trades-table';
-import type { TradeFields } from './__generated__/TradeFields';
 import type {
-  TradesVariables,
-  Trades_market_tradesConnection_edges,
-} from './__generated__/Trades';
+  TradeWithMarket,
+  TradeWithMarketEdge,
+} from './trades-data-provider';
+import type { TradesVariables } from './__generated__/Trades';
```
mattrussell36, line 15: chore (non blocking): This comment is applicable in quite a few places in the PR. We are often importing from the older generated types (`__generated__` rather than `__generated___`). This could certainly be done as a follow up task but it would be good to switch over if possible.

For example here we are using TradesVariables from the old generated types but later in trades-table.tsx we are using types from the new generated types.

mattrussell36 (approved):

libs/market-list/src/lib/market-data-provider.ts
```diff
@@ -62,7 +62,7 @@ const update = (data: MarketData, delta: MarketDataSub_marketsData) => {
 };
 
 const getData = (responseData: MarketDataQuery): MarketData | null =>
-  responseData.marketsConnection.edges[0].node.data || null;
+  responseData?.marketsConnection?.edges[0].node.data || null;
```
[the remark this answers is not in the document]
bglownia: Yes we can, but since we already have full list why we should fetch single market?

apps/trading/components/page-query-container/page-query-container.tsx
```diff
@@ -15,7 +15,10 @@ export const PageQueryContainer = <TData, TVariables = OperationVariables>({
   options,
   render,
 }: PageQueryContainerProps<TData, TVariables>) => {
-  const { data, loading, error } = useQuery<TData, TVariables>(query, options);
+  const { data, loading, error } = useQuery<TData, TVariables>(query, {
+    ...options,
+    errorPolicy: 'ignore',
```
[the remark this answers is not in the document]
bglownia: It was added because errors for candlesConnections. I reverted it.

libs/orders/src/lib/components/order-data-provider/order-data-provider.ts
```diff
@@ -99,23 +98,43 @@ export const update = (
...
+          },
+          cursor: '',
+          __typename: 'OrderEdge',
+        });
       }
     });
   });
 };
 
+// #TODO Order name is in conflict with interface defines in use-order-submit
```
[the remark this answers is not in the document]
bglownia: Will do. I want to look one more time on type names and refactor them. After all changes we made it may be easier to use global types from @vegaprotocol/types than use types generated for per query

libs/trades/src/lib/trades-container.tsx
bglownia: Definitely we want to switch to graphql-codegen style. But we did not wated to complicate this PR adding another change. This will be done postmerge

libs/react-helpers/src/lib/generic-data-provider.spec.ts
```diff
@@ -5,6 +5,7 @@ import {
 } from './generic-data-provider';
 import type {
   CombineDerivedData,
+  CombineDerivedDelta,
```
dexturr, line 8: thought (non-blocking): MIght be nice to have a test in here for the new behaviour introduced in generic data provider

libs/orders/src/lib/components/order-list/order-edit-dialog.tsx
```diff
@@ -69,13 +69,13 @@ export const OrderEditDialog = ({
         <div>
           <p className={headerClassName}>{t(`Size`)}</p>
           <p>
-            {
+            {order.market && (
```
dexturr, line 72: question (non-blocking): if not having a market is a possibility would it be better to throw an error here? Feels like all orders should be tied to a market in some way and if not we should be reporting that so we can look into it

libs/accounts/src/lib/accounts-manager.tsx
```diff
@@ -21,26 +24,29 @@ export const AccountsManager = ({ partyId }: AccountsManagerProps) => {
...
+            }
+          );
+          if (updatedData !== rowNode.data) {
+            update.push(updatedData);
           }
-        );
-        if (updatedData !== rowNode.data) {
-          update.push(updatedData);
+        } else {
+          // #TODO handle new account (or leave it to data provider to handle it)
```
dexturr, line 47: TODO?

dexturr (approved):

libs/accounts/src/lib/accounts-data-provider.ts
```diff
@@ -1,30 +1,47 @@
...
-    } else {
-      draft.push(delta);
-    }
+    deltas.forEach((delta) => {
+      const id = getId(delta);
+      const index = draft.findIndex((a) => getId(a) === id);
+      if (index !== -1) {
+        draft[index].balance = delta.balance;
+      } else {
+        // #TODO handle new account
```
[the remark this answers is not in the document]
bglownia: https://github.com/vegaprotocol/frontend-monorepo/issues/1391

libs/accounts/src/lib/accounts-manager.tsx
bglownia: It's hanging there for a while. I created separate user story for this.
https://github.com/vegaprotocol/frontend-monorepo/issues/1391

libs/react-helpers/src/lib/generic-data-provider.spec.ts
bglownia: Tests are on first place on post merge tasks list.

libs/candles-chart/src/lib/__generated__/Candles.ts
```diff
@@ -24,7 +24,7 @@ export type CandlesEventsSubscription = { __typename?: 'Subscription', candles:
 
 export const CandleFieldsFragmentDoc = gql`
     fragment CandleFields on Candle {
-  timestamp
```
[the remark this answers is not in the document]
bglownia: Not this file but the other one. I deleted duplicates.

Issue #714, opened 2022-07-06
Ethereum chain IDs to have a name: "this app only works on chain ID: 11155111" should mention "Sepolia"


campbellssource:
This on the deposit page on fairground Console, if your eth wallet is contact to a network other than Ropsten.

![Image](https://user-images.githubusercontent.com/13255539/177547826-6ae4702e-cf57-4dcc-a71f-9129594cff35.png)

This message should use the name of the network (perhaps the ID too), and should help the user with a clue about how to fix it.

suggested copy: 
> This app is configured to work with [network name] (id:[3]), please switch your network in your Ethereum wallet.

but I'll defer to a better suggestion in comments

candida-d:
This has tripped up at least one user who thought Console was broken. https://github.com/vegaprotocol/feedback/discussions/80

asiaznik:
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md

PR #1349, opened 2022-09-15
Feat/873 wallet service v2

### LINKED ISSUE #873: Support v2 wallet service
Wallet authentication is changing with the v2 api of the wallet. We need to support this.

Possible split:
- [ ] desktop wallet work to show dialogs (key to expose etc..)
- [ ] 3rd party trading app changes for authentication with the wallet

Note: review the `connect-vega-wallet` labeled issues before starting this work

mattrussell36:
# Related issues 🔗

Closes #873 

# Description ℹ️

- Adds json-rpc-connector
- Updates connect-dialog to contain json rpc connection option
- Changes rest connector to be shown as 'Hosted wallet' connection

# Demo 📺

TBD


libs/wallet/src/connectors/json-rpc-connector.ts
```diff
@@ -0,0 +1,296 @@
+import { z } from 'zod';
```
notbot00, line 1: not sure if interested in anything like this, in the past days i had to do something similar for the wallet desktop, and ended up pulling in a jsonrpc generator. 
this was my template: https://github.com/vegaprotocol/vegawallet-desktop/pull/307/files#diff-f285e942b949fd0bea71f13e8e3814ea2d892a809696ae690a4d4e843cba71d2
it's pretty neat, because it can go and grab a remote openrpc spec and generate all the types for the API you're interacting with, and the `MethodCallValidator` can do runtime checks if needed. you can generate a custom client / connector / anything you might need by writing a template. imho it is worth checking out.

mattrussell36: Cool, thanks for the heads up. I recently got bitten by using a generator for the v1 service, it was a bit premature and it ended up being too much of a manual maintenance job as there was no way to generate a spec. It ended up just being a massive pain, hence pulling it out and doing something more manual. That said I think v2 is a lot more solid and it probably makes more sense to use something like this now. Lets consider it for some future work

libs/wallet/src/provider.tsx
```diff
@@ -69,43 +68,24 @@ export const VegaWalletProvider = ({ children }: VegaWalletProviderProps) => {
     }
   }, []);
 
-  const sendTx = useCallback((body: TransactionSubmission) => {
+  const sendTx = useCallback((pubkey: string, transaction: Transaction) => {
     if (!connector.current) {
-      return null;
+      throw new Error('No connector');
```
sam-keen, line 73: Chore (blocking): translation needed

libs/wallet/src/use-eager-connect.ts
```diff
@@ -11,44 +11,39 @@ export function useEagerConnect(Connectors: {
...
         );
         return;
       }
 
       try {
-        await connect(Connectors[cfgObj.connector]);
+        await connect(Connectors[cfg.connector]);
       } catch {
-        console.warn(`Failed to connect with connector: ${cfgObj.connector}`);
+        console.warn(`Failed to connect with connector: ${cfg.connector}`);
```
sam-keen, line 38: Chore (blocking): translation needed

libs/wallet/src/use-vega-transaction.tsx
```diff
@@ -22,17 +23,19 @@ export enum VegaTxStatus {
...
-  error: string | null;
-  details?: string[] | null;
+  error: WalletError | null;
   txHash: string | null;
   signature: string | null;
   dialogOpen: boolean;
 }
 
+const Errors = {
+  UNKNOWN: new WalletError('Something went wrong', 1, 'Unknown error occurred'),
```
sam-keen, line 33: Chore (blocking): translation needed

libs/wallet/src/vega-transaction-dialog/vega-transaction-dialog.spec.tsx
```diff
@@ -58,7 +59,7 @@ describe('VegaTransactionDialog', () => {
         {...props}
         transaction={{
           ...props.transaction,
-          error: 'rejected',
+          error: new WalletError('rejected', 1),
```
sam-keen, line 62: Chore (blocking): translation needed

sam-keen (changes requested):

libs/wallet/src/connect-dialog/connect-dialog.tsx
```diff
@@ -1,82 +1,264 @@
...
+          </p>
+        )}
+      </ConnectDialogContent>
+      <ConnectDialogFooter />
+    </>
   );
-}
+};
+
+export const HOSTED_WALLET_URL = 'https://wallet.testnet.vega.xyz';
```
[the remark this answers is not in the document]
mattrussell36: Yes, hosted wallet can only run against testnet. This did however remind me of two things, firstly, this should probably be an env var, and secondly hosted wallet should be disabled on mainnet

libs/wallet/src/connectors/injected-connector.ts
```diff
@@ -7,22 +7,15 @@ export class InjectedConnector implements VegaConnector {
...
   }
 
   async disconnect() {
     return;
   }
 
+  // @ts-ignore injected connector is not implemented
   sendTx() {
-    return Promise.resolve(null);
+    throw new Error('Not implemented');
```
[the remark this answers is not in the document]
mattrussell36: I don't think this one needs a translation as this error message never surfaces to the UI

libs/wallet/src/connectors/vega-connector.ts
```diff
@@ -1,21 +1,286 @@
...
+    asset: string;
+    ext: {
+      erc20: {
+        receiverAddress: string;
+      };
+    };
+  };
+}
+
+interface ProposalNewMarketTerms {
```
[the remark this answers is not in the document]
mattrussell36: I've rebased those changes in and moved them here so they are closer to where they are actually used.

libs/wallet/src/use-eager-connect.ts
mattrussell36: Hm not sure if we need translations for console warnings. For me only things that are shown in the UI should need them.

libs/wallet/src/vega-transaction-dialog/vega-transaction-dialog.spec.tsx
mattrussell36: This error is only for test purposes

apps/token/src/components/vega-wallet/vega-wallet.tsx
```diff
@@ -22,40 +22,31 @@ import {
...
 import { DownloadWalletPrompt } from './download-wallet-prompt';
 import { usePollForDelegations } from './hooks';
-import type { VegaKeyExtended } from '@vegaprotocol/wallet';
 import { useVegaWallet } from '@vegaprotocol/wallet';
 import { Button, ButtonLink } from '@vegaprotocol/ui-toolkit';
 
 export const VegaWallet = () => {
   const { t } = useTranslation();
-  const { keypair, keypairs } = useVegaWallet();
+  const { pubKey, pubKeys } = useVegaWallet();
```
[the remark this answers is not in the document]
mattrussell36: Hm this is a big rename task that I already had to do once, I think its clear enough for now, lets change later if we feel it really necessary.

libs/wallet/src/connect-dialog/connect-dialog.tsx
```diff
@@ -1,82 +1,264 @@
...
+            walletUrl={walletUrl}
+            appChainId={appChainId}
+          />
+        </ConnectDialogContent>
+        <ConnectDialogFooter />
+      </>
+    );
+  }
+
+  throw new Error('No connector selected');
```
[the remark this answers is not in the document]
mattrussell36: I'm not sure about the error message itself containing a translation. I think a better solution would be to have a map of translations for error codes. I'd prefer to do this as a follow up so we can consider all errors from the v2 service as well

libs/wallet/src/connect-dialog/json-rpc-connector-form.tsx
```diff
@@ -0,0 +1,252 @@
...
+              {t('Read the docs to troubleshoot')}
+            </Link>
+          </>
+        );
+      } else if (error.code === 0) {
+        title = t('Wrong network');
+        text = (
+          <>
+            {t(`To complete your wallet connection, set your wallet network in your
+            app to ${'TODO: App chain id'} then `)}
```
asiaznik, line 158: This will be addressed by https://github.com/vegaprotocol/frontend-monorepo/issues/714

libs/react-helpers/src/hooks/index.ts
```diff
@@ -6,3 +6,4 @@ export * from './use-outside-click';
 export * from './use-resize';
 export * from './use-screen-dimensions';
 export * from './use-theme-switcher';
+export * from './__generated__';
```
dexturr, line 9: shared types should be in types package no?

dexturr (approved):

libs/wallet/src/connect-dialog/connect-dialog.tsx
```diff
@@ -1,82 +1,274 @@
...
-      selectedConnector !== null &&
-      !(selectedConnector instanceof RestConnector)
-    ) {
-      connectAndClose(selectedConnector);
-    }
-  }, [selectedConnector, connectAndClose]);
+  if (!appChainId) {
+    return (
+      <ConnectDialogContent>
+        <ConnectDialogTitle>{t('Fetching chain ID')}</ConnectDialogTitle>
```
asiaznik, line 67: suggestion: wouldn't it be better rephrased to "Fetching network"?

libs/react-helpers/src/hooks/index.ts
mattrussell36: Hm not sure I follow. This is just types from the use-network-params hook. react-helpers seemed like the most sensible place for this hook and the resulting generated types.

libs/wallet/src/connect-dialog/json-rpc-connector-form.tsx
mattrussell36: This TODO was actually accidentally not complete by me. In this case the incorrect chain message here is for the Vega network, not Ethereum

Screen Shot 2022-09-27 at 8 37 59 PM

libs/wallet/src/connect-dialog/connect-dialog.tsx
mattrussell36: Up for debate. This was suggested by ux

libs/wallet/src/connect-dialog/connect-dialog.tsx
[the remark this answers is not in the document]
sam-keen: Grand :)

libs/wallet/src/connect-dialog/connect-dialog.tsx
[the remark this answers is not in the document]
sam-keen: Fine by me

sam-keen (approved):

PR #1372, opened 2022-09-16
Test/governance suite expansion

AndyWhiteVega:
# Related issues 🔗

Closes #[Issue number here]

# Description ℹ️

- A lot more governance/voting tests
- Teardown functionality - to stop suites bleeding into each other - restarting capsule after flow tests (triggered by workflows)
- Inclusion of AC references (bar a couple)

# Demo 📺

# Technical 👨‍🔧


.github/workflows/capsule-cypress.yml
```diff
@@ -130,7 +130,7 @@ jobs:
         working-directory: frontend-monorepo
 
       - name: Run Cypress tests
-        run: npx nx affected:e2e --record --key ${{ secrets.CYPRESS_RECORD_KEY }}
+        run: npx nx affected:e2e --record --key ${{ secrets.CYPRESS_RECORD_KEY }} --env.teardownNetworkAfterFlows=true
```
jtsang586, line 133: The environment variable can be passed below with the other ones. 
But I think the default should be set to true in cypress.config.js and testers should have a local cypress.env.json file to set it to false if they want to. The default should be the same as CI in my opinion.

AndyWhiteVega: Good call - Rectified!

dexturr (approved):

Issue #1366, opened 2022-09-16
Bug: collateral table overlapping rows


PeteWilliams:
When I click the Collateral tab in the trading view in Console, I see some weirdness:

- There is a blank row where my Vega general account should be shown
- Down in the middle of my BTC accounts, I can see the missing row is overlaying one of the BTC rows
 
![image](https://user-images.githubusercontent.com/1325606/190626313-d840ee6a-67d3-4283-8245-ca4e636e5148.png)

In the DOM, it's positioned in the right place, but it looks like AG grid positions things using CSS translateY transformations, and it just calculates this one wrong somehow. Every other row's value goes up by 34px (the row height), except this one:
<img width="254" alt="image" src="https://user-images.githubusercontent.com/1325606/190626697-20c3f579-486c-4ffd-b26b-b98319fc6ecb.png">

Things to note:

- This happens in Chrome and Safari. Locally and on console.fairground.wtf
- If I click the Asset heading twice (ie, sort Z-A then back to A-Z) it then renders fine - despite this in theory being the same thing as originally shown!
- Sometimes happens to multiple rows

mattrussell36:
Fixed by https://github.com/vegaprotocol/frontend-monorepo/pull/1442

PR #1385, opened 2022-09-19
feat: update collateral tab

### LINKED ISSUE #847: Portfolio: Update Collateral tab
### Current implementation

image

### Proposed design
https://www.figma.com/file/K49Nx7XfocShhHcwwQzQkm/Console-V2?node-id=2537%3A32178

image

image

### Details
- Show 'General' margin as top level, expand to see 'Margin' (position) level margin
- General/Available margin
  - Show total margin balance
  - Show used and available margin levels
  - Add Used -> Available visualisation
  - Add buttons to deposit/withdraw collateral
- Position margin
  - Show market name, linked to market on Trading page
  - Show allocated margin value
  - Add visualisation representing how much margin is allocated to that position
  - Add button to close position
- Add button to deposit a new asset type

MadalinaRaicu:
# Related issues 🔗

Closes #847 & #1366

# Description ℹ️

Update the collaterals tab.

# Demo 📺

After: 
image

image

Screenshot 2022-09-19 at 09 46 24
Screenshot 2022-09-19 at 09 46 53
Screenshot 2022-09-19 at 09 47 01
Before:
Screenshot 2022-09-14 at 20 06 08
Screenshot 2022-09-14 at 20 06 44

190626313-d840ee6a-67d3-4283-8245-ca4e636e5148

# Technical 👨‍🔧


libs/accounts/src/lib/accounts-data-provider.ts
```diff
@@ -50,3 +52,42 @@ export const accountsDataProvider = makeDataProvider<
...
 });
+
+export const getAccountData = (
+  data: AccountFieldsFragment[],
+  assetSymbol: string
+) => {
+  const assetData = data.filter((a) => a.asset.symbol === assetSymbol);
+
+  const deposited = assetData
+    .filter((a) => [AccountType.ACCOUNT_TYPE_GENERAL].includes(a.type))
```
MadalinaRaicu, line 63: Should we count in deposited (GENERAL) or all available balance (i.e. plus REWARDS etc.)?

libs/accounts/src/lib/accounts-data-provider.ts
```diff
@@ -50,3 +52,42 @@ export const accountsDataProvider = makeDataProvider<
...
+    .reduce((acc, a) => acc.plus(a.balance), new BigNumber(0));
+
+  const used = assetData
+    .filter((a) =>
+      [AccountType.ACCOUNT_TYPE_MARGIN, AccountType.ACCOUNT_TYPE_BOND].includes(
```
MadalinaRaicu, line 68: Should we count in another type of accounts too or just MARGIN? (i.e. AccountType.ACCOUNT_TYPE_BOND, FEES, etc.)

MadalinaRaicu: @campbellssource @candida-d

libs/accounts/src/lib/accounts-data-provider.ts
MadalinaRaicu: I am calculating separately `deposited` and `available`.  I have categorized the accounts into `incoming` and `outcoming`.

libs/accounts/src/lib/accounts-manager.tsx
```diff
@@ -60,13 +81,108 @@ export const AccountsManager = ({ partyId }: AccountsManagerProps) => {
...
+              hideHeader={i !== 0}
+            />
+          ))}
+      </AsyncRenderer>
+      <div className="m-auto ml-4">
+        <Button size="sm" onClick={() => setDepositDialog(true)}>
+          {t('Deposit new asset')}
+        </Button>
+      </div>
+      <DepositDialog
```
mattrussell36, line 111: chore (blocking): Lets try and keep lib components/functionality separated. Can you move the deposits dialog and state to be rendered by the app only. 

I suggest moving libs/accounts/src/lib/accounts-container.tsx to apps/trading/pages/portfolio (see apps/trading/pages/portfolio/withdrawals-container)

libs/accounts/src/lib/accounts-manager.tsx
```diff
@@ -60,13 +81,108 @@ export const AccountsManager = ({ partyId }: AccountsManagerProps) => {
...
   >({ dataProvider: accountsDataProvider, update, variables });
+  const [depositDialog, setDepositDialog] = useState(false);
+
+  const symbols = data && getSymbols(data);
+  return (
+    <Web3Container>
+      <AsyncRenderer loading={loading} error={error} data={assetSymbols}>
+        {symbols &&
+          symbols.map((assetSymbol, i) => (
+            <AssetAccountTable
```
mattrussell36, line 97: question (non blocking): I really don't like this approach of rendering multiple data grids for each asset. I know we do a similar thing in the positions table but IMO this is the wrong approach. Lets discuss

libs/accounts/src/lib/accounts-table.tsx
```diff
@@ -1,170 +1,179 @@
...
-  if (data.__summaryRow) {
-    return null;
-  }
-  const sortColumnId = columnApi.getColumnState().find((c) => c.sort)?.colId;
-  switch (sortColumnId) {
-    case 'asset.symbol':
-      return data.asset.id;
-  }
-  return undefined;
+export const ProgressBarCell = ({ valueFormatted }: ValueProps) => {
```
mattrussell36, line 39: chore (blocking): This componet looks like a dupe of that used in the positions table, can we move this to react helpers grid?

libs/withdraws/src/lib/withdraw-form.tsx
```diff
@@ -98,7 +98,11 @@ export const WithdrawForm = ({
           <Controller
             control={control}
             name="asset"
-            rules={{ validate: { required } }}
+            rules={{
+              validate: {
+                required: (value: string) => !!selectedAsset || required(value),
```
mattrussell36, line 103: chore (blocking): the form should use a default asset if one is passed in so we shouldn't need a change here

libs/withdraws/src/lib/withdraw-manager.tsx
```diff
@@ -63,6 +65,10 @@ export const WithdrawManager = ({
     [accounts, assets, update, getThreshold, getDelay]
   );
 
+  if (assetId) {
```
mattrussell36, line 68: chore (blocking): state should not be updated like this in React

libs/withdraws/src/lib/withdrawal-dialogs.tsx
```diff
@@ -24,13 +26,16 @@ export const WithdrawalDialogs = ({
...
         size="small"
       >
-        <WithdrawFormContainer
-          partyId={keypair?.pub}
-          submit={(args) => {
-            setWithdrawDialog(false);
-            createWithdraw.submit(args);
-          }}
-        />
+        {withdrawDialog && (
```
mattrussell36, line 29: question (blocking): Why is this condition required?

libs/deposits/src/lib/deposit-manager.tsx
```diff
@@ -39,6 +41,12 @@ export const DepositManager = ({
     [assets, update]
   );
 
+  useEffect(() => {
+    if (assetId) {
+      handleSelectAsset(assetId);
```
mattrussell36, line 46: chore (blocking): Better to set the initial state than use an effect to update it

libs/deposits/src/lib/deposit-limits.tsx
```diff
@@ -7,6 +7,10 @@ interface DepositLimitsProps {
   balance?: BigNumber;
 }
 
+const format = (value: string) => {
```
mattrussell36, line 10: chore (blocking): If we need this function, then there is a problem elsewhere and its worth fixing. We shouldn't ever have NaN here.

libs/deposits/src/lib/deposit-form.tsx
```diff
@@ -152,7 +151,11 @@ export const DepositForm = ({
         <Controller
           control={control}
           name="asset"
-          rules={{ validate: { required } }}
+          rules={{
+            validate: {
+              required: (value: string) => !!selectedAsset || required(value),
```
mattrussell36, line 156: question (blocking): Why not just let it be validated normally here using defaultValues?

libs/deposits/src/lib/deposit-form.spec.tsx
```diff
@@ -250,7 +250,7 @@ describe('Deposit form', () => {
       screen.getByText('Deposit', { selector: '[type="submit"]' })
     );
 
-    await waitFor(() => {
+    waitFor(() => {
```
mattrussell36, line 253: chore (blocking): Pretty sure you must await this function for it to do anything of value.

See: https://testing-library.com/docs/dom-testing-library/api-async/#waitfor

mattrussell36 (changes requested):

libs/accounts/src/lib/account-deposit.tsx
```diff
@@ -0,0 +1,142 @@
...
+  data: AccountFieldsFragment;
+}
+
+export const assetDecimalsFormatter = ({
+  value,
+  data,
+}: AccountsTableValueFormatterParams) =>
+  addDecimalsFormatNumber(value, data.asset.decimals);
+
+export const AccountDeposit = forwardRef<AgGridReact, AccountsTableProps>(
```
mattrussell36, line 41: chore (blocking): Really not sure about this approach of rendering multiple different grids. Kind of completely defeats the purpose of using ag grid in the first place. Happy to discuss, but I think we should really just leverage what it does best, and not try to hack it so that its also an accordion

MadalinaRaicu: any aggregation I was planning to use was available in the enterprise version. that is how I ended up using it this way, and I guess this is why it has been used similarly with positions

libs/accounts/src/lib/accounts-table.tsx
MadalinaRaicu: there are a few small differences but we can make it more configurable and move it.

mattrussell36:
@MadalinaRaicu @PeteWilliams I don't think we should be making the collapsible rows right now. It requires big work arounds to make it work with AgGrid and the complexity overhead is not worth it. It will make filtering and sorting very difficult if not impossible. Additionally we end up with scroll bars all over the place and multiple ag grid instances rendered (which is quite heavy).

What I propose for now is to just render the rows as they come back from the API and have buttons to open the withdraw and deposit dialogs (general account rows only). If we really deem the progress bar and other changes necessary how about showing that on the asset dialog?

Screen Shot 2022-09-19 at 3 30 20 PM

libs/deposits/src/lib/deposit-manager.tsx
MadalinaRaicu: updated and add an initial state on the asset. (same for above)

libs/withdraws/src/lib/withdrawal-dialogs.tsx
MadalinaRaicu: not anymore

libs/withdraws/src/lib/withdraw-manager.tsx
MadalinaRaicu: i can set an initial state as in the comment below

libs/deposits/src/lib/deposit-form.tsx
MadalinaRaicu: updated

libs/deposits/src/lib/deposit-form.spec.tsx
MadalinaRaicu, line 253:
```suggestion
    await waitFor(() => {
```

PR #1389, opened 2022-09-19
Test/1310 apply staking functions in token e2e tests

### LINKED ISSUE #1310: Apply staking functions in token e2e tests
Apply staking functions in tests preparation steps instead of using UI.

radoslaw-szpiech:
# Related issues 🔗

Closes #1310 

# Description ℹ️

Change adding tokens to vega wallet to use eth bridge.

### DIFF
--- a/apps/token-e2e/src/support/common.functions.js
+++ b/apps/token-e2e/src/support/common.functions.js
@@ -96,7 +96,7 @@ Cypress.Commands.add(
     });
 
     cy.reload();
-    cy.get('header h1', {timeout: 10000}).should('be.visible');
+    cy.get('header h1', { timeout: 10000 }).should('be.visible');
     cy.wait_for_spinner();
     cy.vega_wallet_connect();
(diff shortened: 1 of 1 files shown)


apps/token-e2e/src/support/common.functions.js
```diff
@@ -69,3 +69,29 @@ Cypress.Commands.add('restartVegacapsuleNetwork', () => {
     })
     .should('contain', 'starting network success');
 });
+
+Cypress.Commands.add('set_up_tokens', (wallet = 0, contract = 0) => {
```
jtsang586, line 73: set_up_tokens sounds very vague to me. I would suggest renaming it to make it clear its associating vega tokens.

jtsang586: Same with `wallet` and `contract`. Maybe `vegaWalletAmount` and `ethWalletAmount` would be better?

PR #1396, opened 2022-09-20
Chore/design tweaks

PeteWilliams:
# Description ℹ️

I've been doing some fiddling! Basically manually testing Console and tweaking any UI or usability issues that annoy me.

This includes:

- Adding hover/active states to interactive elements
- Making it more keyboard usable
  - Adding keyboard focus states
  - Adjusting tab order
- Contrast/spacing tweaks
- Made link/button-link styling consistent
- Added implied 'back' button to liquidity page
- Reduced spacing in positions panel
- Fixed chevron not being clickable in dropdown boxes
- Removed blue cell focus border from AG grid tables
- Updated Pennant version to include changes to chart styling
- Updated secondary button styling


libs/deposits/src/lib/deposit-form.tsx
```diff
@@ -340,13 +341,17 @@ const FormButton = ({
 interface UseButtonProps {
   children: ReactNode;
   onClick: () => void;
+  className?: string;
```
mattrussell36, line 344: chore (blocking): This is a pointless className just apply directly

mattrussell36 (changes requested):

libs/ui-toolkit/src/components/tabs/tabs.tsx
```diff
@@ -21,7 +21,7 @@ export const Tabs = ({ children, active: activeDefaultId }: TabsProps) => {
     >
       <div className="border-b border-default">
         <TabsPrimitive.List
-          className="flex flex-nowrap overflow-visible"
+          className="flex flex-nowrap overflow-visible bg-neutral-50 dark:bg-neutral-800"
```
mattrussell36, line 24: chore (blocking): I don't agree with these changes, it deviates off brand and looks worse

libs/deposits/src/lib/deposit-form.tsx
PeteWilliams: Not sure what you mean? I was trying to change the style for the `requestFaucet` button:

`<UseButton onClick={requestFaucet} className="bottom-0 top-auto">`

This was to move it below the dropdown because the action shouldn't be appearing above the field that makes it appear, it should be after it.

Is there a way to add that extra style without adding the `className` property to `UseButtonProps`?

libs/ui-toolkit/src/components/theme-switcher/theme-switcher.tsx
```diff
@@ -17,11 +17,14 @@ export const ThemeSwitcher = ({
   moonClassName?: string;
   fixedBg?: 'light' | 'dark';
 }) => {
-  const sharedClasses = classNames({
-    'text-neutral-800 dark:text-neutral-300': !fixedBg,
-    'text-neutral-800': fixedBg === 'light',
-    'text-neutral-300': fixedBg === 'dark',
-  });
+  const sharedClasses = classNames(
+    'hover:text-white group-focus-visible:text-white',
```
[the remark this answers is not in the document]
PeteWilliams: The defaults weren't fit for purpose. There was no state for most links (some had them added manually but different colours in different places), custom controls like the dropdowns, tabs, toggles, buttons, close buttons, theme switcher, nav items, etc had no focus state. For form elements, it was inconsistent, with some having a blue border, some having a black/white one, some having both, some having none.  On the wallet dropdown, the border disappeared on focus.
It was unusable by keyboard and looked sloppy - can't see any reason to keep it that way?

libs/ui-toolkit/src/components/button/button.tsx
```diff
@@ -21,23 +21,24 @@ const defaultClasses = [
...
-  'text-white',
-  'border-vega-pink',
-  'bg-vega-pink',
-  'enabled:hover:bg-vega-pink-dark enabled:hover:border-vega-pink-dark',
-  'enabled:active:bg-vega-pink-dark enabled:active:border-vega-pink-dark',
+  'text-white dark:text-black',
+  'border-none',
+  'bg-neutral-700 dark:bg-neutral-200',
+  'enabled:focus:bg-black dark:enabled:focus:bg-white enabled:focus-visible:underline',
+  'enabled:hover:bg-black dark:enabled:hover:bg-white enabled:active:font-medium',
```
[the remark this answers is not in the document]
PeteWilliams: The secondary button in the designs wasn't fit for purpose. We had bright yellow for primary, bright pink for secondary. Those don't represent primary/secondary - they are equal prominence, just different colours.

I do think we need a secondary button - for instance I've used it for deposit/withdraw/connect wallet. But using bright pink is too much for those. I've used block black for the light theme and block white for the dark theme  so they are high contrast and looks on-brand but don't dominate - which is what the secondary buttons should do.

apps/trading/components/navbar/navbar.tsx
```diff
@@ -71,6 +71,7 @@ const NavLink = ({ name, path, exact, testId = name }: NavLinkProps) => {
   const linkClasses = classNames('mx-2 py-2 self-end border-b-4', {
     'border-vega-yellow text-white cursor-default': isActive,
     'border-transparent text-neutral-400 hover:text-neutral-300': !isActive,
+    'focus-visible:underline': !isActive,
```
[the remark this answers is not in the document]
PeteWilliams: See below. I don't think there is much variety - borders are used in some places like block elements like input fields and underline for plain text things like nav/tabs. This is the problem with trying to retro-fit styles to components, rather than designing them with all the states in mind up-front - you'll end up with designs where a single style doesn't work for everything.

libs/web3/src/lib/web3-wallet-input.tsx
```diff
@@ -25,7 +25,7 @@ export const Web3WalletInput = ({ inputProps }: Web3WalletInputProps) => {
       <Input
         {...inputProps}
         appendIconName="chevron-down"
-        className="cursor-pointer select-none"
+        className="cursor-pointer select-none rounded-md z-20"
```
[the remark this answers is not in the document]
PeteWilliams: Without it you can't click on the chevron in the dropdown as it's on top of the dropdown and has no action assigned to it.

libs/deposits/src/lib/deposit-form.tsx
PeteWilliams: Resolved as part of switching to ButtonLink.

PeteWilliams (commented):
Have now committed these changes and added comments.

libs/ui-toolkit/src/components/ag-grid/ag-grid-light.tsx
```diff
@@ -22,6 +22,14 @@ const agGridLightVariables = `
     overflow: hidden;
     text-overflow: ellipsis;
   }
+
+  div[data-testid="tab-positions"] .ag-theme-balham .ag-cell {
```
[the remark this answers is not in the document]
mattrussell36: This is still the wrong place for these styles though. Can all grids not just accept this style?

libs/ui-toolkit/src/components/theme-switcher/theme-switcher.tsx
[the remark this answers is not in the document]
mattrussell36: This is still the wrong place for these styles though. Can all grids not just accept this style?

apps/trading/components/vega-wallet-connect-button/vega-wallet-connect-button.tsx
```diff
@@ -30,6 +30,7 @@ export const VegaWalletConnectButton = ({
     return (
       <DropdownMenu open={dropdownOpen}>
         <DropdownMenuTrigger
+          className="text-white hover:border-white hover:!bg-neutral-700 focus:!border-white focus-visible:!bg-neutral-700"
```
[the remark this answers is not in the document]
mattrussell36: There is a better solution for this. You should be able to remove this entire classname prop and just but a dark class on the top level navbar element

libs/ui-toolkit/src/components/tabs/tabs.tsx
mattrussell36: The problem here stems from rendering multiple ag agrids next to each other, which we shouldn't be doing. It creates the additional borders, and if scroll bars appear its even worse. In the second example it looks clean because a) there is only one position open, so there aren't multiple grids in a single tab, and b) the collateral grid is the current one which _is_ still a single grid. As a result it looks much nicer.

If this is a concern lets make sure our designs don't add lots of borders within the tabs where possible

apps/trading/pages/liquidity/[marketId].page.tsx
```diff
@@ -69,9 +69,13 @@ const LiquidityPage = ({ id }: { id?: string }) => {
       <div className="h-full grid grid-rows-[min-content_1fr]">
         <Header
           title={
-            <button onClick={() => push(`/markets/${marketId}`)}>{`${name} ${t(
-              'liquidity provision'
-            )}`}</button>
+            <button
+              className="hover:text-neutral-500 dark:hover:text-neutral-300 sm:text-lg md:text-xl lg:text-2xl "
```
Comment (sam-keen) [inline apps/trading/pages/liquidity/[marketId].page.tsx:73]: Chore (potentially blocking): Is this a one-off button style that won't be repeated? This kind of custom styling on a button element makes consistency and repeatability a challenge as the apps evolve. If there's any chance it'd be repeated, either in the trading app or if you foresee a use case on one of the other apps, a new button variant may be a better option.

apps/trading/pages/markets/trade-grid.tsx
```diff
@@ -199,6 +199,7 @@ export const TradeMarketHeader = ({
         <HeaderStat heading={t('Settlement asset')}>
           <div data-testid="trading-mode">
             <ButtonLink
+              className="inline underline hover:text-black dark:hover:text-white"
```
sam-keen, line 202: Chore (blocking): like above, but this styling seems like something that could definitely be reused, so instead of customising this here, 

1. If you need an element to be inlined, consider using a wrapper tag with inline, rather than adding custom classes to an existing styled component.
2. Consider how to adjust the ButtonLink component

apps/trading/pages/styles.css
```diff
@@ -9,16 +9,30 @@ body,
...
+html.dark body {
   --focus-border: theme('colors.vega.yellow');
   --separator-border: theme('colors.neutral.600');
 }
 
 .border-default {
   @apply border-neutral-300 dark:border-neutral-600;
 }
+
+html.dark ::-webkit-calendar-picker-indicator {
```
sam-keen, line 26: Question (blocking): Why style this here? If a calendar picker needs to be inverted in the trading app, surely it needs it elsewhere too.

libs/deposits/src/lib/deposit-form.tsx
```diff
@@ -176,25 +177,38 @@ export const DepositForm = ({
...
               setAssetDetailsDialogOpen(true);
               setAssetDetailsDialogSymbol(selectedAsset);
             }}
           >
             {t('View asset details')}
-          </button>
+          </UseButton>
+        )}
+        {isFaucetable && selectedAsset && (
+          <span className="absolute bottom-0 right-0">
```
sam-keen, line 192: Chore (blocking): I can't see an element that these `absolutes` are `relative` to. Different browsers can interpret differently, you can get wildly different results based on whether the browser assumes a parent element or the html element. If I'm wrong then apologies and ignore, but if there is no `relative` positioning on the parent element, please set it.

libs/ui-toolkit/src/components/dialog/dialog.tsx
```diff
@@ -63,7 +63,7 @@ export function Dialog({
           <div className={wrapperClasses}>
             {onChange && (
               <DialogPrimitives.Close
-                className="absolute p-2 top-0 right-0 md:top-2 md:right-2"
+                className="absolute p-0 m-1 w-7 h-7 top-0 right-0 md:top-2 md:right-2 text-neutral-500 dark:text-neutral-300 hover:text-black dark:hover:text-white border-[1px] border-transparent focus-visible:border-neutral-300 dark:focus-visible:border-neutral-500 focus-visible:text-black dark:focus-visible:text-white"
```
sam-keen, line 66: Chore (blocking): these really long strings of classes are hard to read. Can you please import the classnames module, and split the styling up the styling onto multiple lines as we do in other files

libs/withdraws/src/lib/withdraw-form.tsx
```diff
@@ -190,15 +191,12 @@ export const WithdrawForm = ({
...
 const UseButton = ({ children, ...rest }: UseButtonProps) => {
   return (
-    <button
-      {...rest}
-      type="button"
-      className="ml-auto text-sm absolute top-0 right-0 underline"
-    >
-      {children}
-    </button>
+    <span className="ml-auto text-sm absolute top-0 right-0">
```
sam-keen, line 196: Chore (blocking): another case of specifying a `relative` on a parent to ensure `absolute` is consistent across browsers

sam-keen (changes requested):

apps/token/src/routes/staking/node-list.tsx
```diff
@@ -248,7 +248,6 @@ export const NodeList = ({ epoch }: NodeListProps) => {
           columnDefs={colDefs}
           defaultColDef={defaultColDef}
           animateRows={true}
-          suppressCellFocus={true}
```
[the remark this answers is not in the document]
PeteWilliams: I've added it in as a default for all AG Grids here, as I agree the cell focus is unnecessary: https://github.com/vegaprotocol/frontend-monorepo/blob/aaf21d40c44b50f64c991e630588730e384c7cf6/libs/ui-toolkit/src/components/ag-grid/ag-grid-dynamic-themed.tsx#L40

libs/deposits/src/lib/deposit-form.tsx
PeteWilliams: They all have `<FormGroup>` as a parent, which are set to `relative` by default:

image

libs/withdraws/src/lib/withdraw-form.tsx
PeteWilliams: As above.

apps/token/src/routes/staking/node-list.tsx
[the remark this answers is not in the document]
sam-keen: Ah cool, I didn't realise it could be applied so broadly. Thanks!

apps/trading/pages/styles.css
sam-keen: Input styling in the ui-toolkit. You can use the tailwind 'invert' class

apps/trading/components/vega-wallet-connect-button/vega-wallet-connect-button.tsx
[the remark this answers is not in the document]
PeteWilliams: It's the other way round sorry - the dark theme header needs to use the light theme styles. The other way round you could add 'dark', but there's no 'light' class to add to overwrite it the other way round, so not sure if there's a way to get that to work?

libs/ui-toolkit/src/components/theme-switcher/theme-switcher.tsx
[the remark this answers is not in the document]
PeteWilliams: Do you mean the grid styling bits above? https://github.com/vegaprotocol/frontend-monorepo/pull/1396/files/20a0be3a4b778b7df1027d7168de8326e8e01cb4#r975846664

The positions grid alignment bits are only really applicable to tables with multi-row cells, like the current positions panel.

The row hover one could be for all grids, but where would that go then if not ag-grid-light/dark.tsx?

libs/ui-toolkit/src/components/ag-grid/ag-grid-light.tsx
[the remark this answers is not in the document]
PeteWilliams: The positions grid alignment one is only applicable to tables with multi-row cells, like the current positions panel. So I put that in apps/trading/pages/styles.scss. Let me know if there's a better place or way to do it.

The footer row hover one could be for all grids, so should that go back in ag-grid-light/dark.tsx?

apps/trading/pages/markets/trade-grid.tsx
PeteWilliams: In this instance, a styling change is required for hover text colour on the light theme because in the market header we use grey text for the values rather than black and the default text is the same grey, so I wanted to change the light theme hover text to black. The other styles (inline, underline etc) are only passed in because it seems the classNames property overwrites the default rather than appending. It only really applies to the market header and this is (currently) the only link there, so not sure how reusable we'd consider that?

If we do want to though, what'd be the best way?
I don't think wrapping would work, because the hover state would be overwritten by the child element.
In terms of adjusting ButtonLink, would I add a property to say whether it's being used in the header?

apps/trading/pages/liquidity/[marketId].page.tsx
Comment (PeteWilliams) [reply apps/trading/pages/liquidity/[marketId].page.tsx]: Yeah I think this could be reused as a back button whenever a market goes to a new page rather than shows something in a panel. For now we only have the liquidity page but there may be more in the future. I think it'd be only used in Console though. So would a new button style in the UI toolkit be the best place for it to go, or somewhere else?

libs/ui-toolkit/src/components/tabs/tabs.tsx
PeteWilliams: Have removed this style for now and created a ticket for David to explore options.
https://github.com/vegaprotocol/ux-and-visual-design/issues/19

libs/ui-toolkit/src/components/theme-switcher/theme-switcher.tsx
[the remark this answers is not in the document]
PeteWilliams: Have removed focus states for now and will create a ticket to resolve remaining issues.

apps/trading/components/navbar/navbar.tsx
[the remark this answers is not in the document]
PeteWilliams: Have removed for now.

apps/trading/pages/liquidity/[marketId].page.tsx
Comment (sam-keen) [reply apps/trading/pages/liquidity/[marketId].page.tsx]: I think a new button style in the UI toolkit's still the best place yeah. The only thing we want to avoid is putting elements in there that we don't currently have a use for, but as this is already being used and sounds like it'll be more useful in future, it should be good :)

apps/trading/pages/markets/trade-grid.tsx
sam-keen: Hmm, so the reason I thought that the 'hover:text-black dark:hover:text-white" bit could be reusable is for the times like you're describing where we want the text to be as vivid as possible. If it's genuinely single use for now and you don't anticipate it being used elsewhere (I can't think of anywhere that needs it either at the minute) then it's ok with me if it's kept as a one-off. 

Regarding using a wrapper - yeah you're correct about hover styles being overridden in the component, but I was just referring to it using the 'inline' class. Positional stuff like block/inline/grid styles/padding and margin etc can be applied to a parent wrapper, this keeps things cleaner and easy to manage moving forward.

apps/trading/pages/liquidity/[marketId].page.tsx
Comment (PeteWilliams) [reply apps/trading/pages/liquidity/[marketId].page.tsx]: Have created a new Link variant rather than button as I think in this instance it's a navigational link, rather than an action button. Let me know if that looks OK! https://github.com/vegaprotocol/frontend-monorepo/pull/1396/commits/c3ab0c322c8c7ca525deebad2e52ac3a5821ffb5

PR #1432, opened 2022-09-22
Feat/1076 proposal not found

### LINKED ISSUE #1076: Proposal not found page
Proposals may not remain on chain between chain upgrades, but the markets they created will.

The plan is to have data-node store these.

Console will link to proposals on token without knowning they are live or not: https://github.com/vegaprotocol/frontend-monorepo/issues/707

This ticket to create a specific proposal not found page. That says: The proposal you are looking for is not here, It may have been enacted before the last chain re-store. You could check forums/discord instead.

dexturr:
# Related issues 🔗

Closes #1076

# Description ℹ️

Add not found page with copy to proposal details page

### DIFF
--- a/apps/token/src/i18n/translations/dev.json
+++ b/apps/token/src/i18n/translations/dev.json
@@ -685,7 +685,6 @@
   "MoreNetParamsInfo": "To see Explorer data on network params visit",
   "MoreMarketsInfo": "To see Explorer data on existing markets visit",
   "MoreAssetsInfo": "To see Explorer data on existing assets visit",
-  "viaContract": "via vesting",
   "ProposalNotFound": "Proposal not found",
   "ProposalNotFoundDetails": "The proposal you are looking for is not here, it may have been enacted before the last chain restore. You could check the Vega forums/discord instead for information about it."
 }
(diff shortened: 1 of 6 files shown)


apps/token/src/i18n/translations/dev.json
```diff
@@ -633,5 +633,7 @@
   "MinProposalRequirements": "You must have at least 1 VEGA associated to make a proposal",
   "totalSupply": "Total Supply",
   "viaWallet": "via wallet",
-  "viaContract": "via vesting"
+  "viaContract": "via vesting",
+  "ProposalNotFound": "Proposal not found",
+  "ProposalNotFoundDetails": "The proposal you are looking for is not here, it may have been enacted before the last chain re-store. You could check forums/discord instead."
```
candida-d, line 638:
```suggestion
  "ProposalNotFoundDetails": "The proposal you are looking for is not here, it may have been enacted before the last chain restore. You could check the Vega forums/discord instead for information about it."
```

candida-d: non-blocking copy change

asiaznik (approved):

Issue #1434, opened 2022-09-22
Finish type migration


notbot00:
## Migrate type generator

Migrate the leftover apps and libs to use the new type system. The easiest way to see which ones need to be migrated is to check for `__generated___` (with 3 underscores on the end), whichever lib/app has those folders, it hasn't been migrated. Also, the `libs/types/apollo.config.js` has a list of the migrated libs in the `excludes` field.

## Tasks

- [ ] Choose a library (or app, if all libs are done) see if it has been migrated (see above)
- [ ] Delete the `__generated__`(2 x underscore) folders
- [ ] Rename the `__generated___`(3 x underscore) folders to `__generated__`(2 x underscore)
- [ ] Add the lib path to the `libs/types/apollo.config.js` file's `excludes` field
- [ ] Search for `gql` in the files, and make sure the queries match what's in the related `.graphql` file
- [ ] Remove the gql definitions
- [ ] Import and use the hook/types/documents from the new generator's output
        - you can directly replace `useQuery` with a generated hook from the generated folder
        - raw graphql documents always end with `...Document`, if you need them - e.g. for the makeDataProvider
- [ ] Fix type and build errors
 
## Additional details / background info
this is a result of https://github.com/vegaprotocol/frontend-monorepo/pull/1219

PR #1439, opened 2022-09-22
Chore/migrate to new types

notbot00:
# Related issues 🔗

Related to #1434

# Description ℹ️

Migrate a few libs:
- `deal-ticket`
- `deposits`
- `environment`
- `fills`


libs/react-helpers/src/lib/get-events.ts
```diff
@@ -0,0 +1,21 @@
+import type { Schema } from '@vegaprotocol/types';
+
+type Event = {
+  __typename?: string;
+};
+
+type BusEvent<T> = {
```
asiaznik, line 7: Question (non-blocking): Wouldn't it be better to constrain `T` here already as `<T extends Event>`?

notbot00:
```suggestion
type BusEvent<T extends Event> = {
```

bglownia (approved):

PR #1459, opened 2022-09-23
Feat/1334 replace json

### LINKED ISSUE #1334: Replace json in Party details
## Story 

As a user
I want to see Party details presented via standard UI elements
So that I don't have to be able to read json to be able to get the information I need

Image

## Acceptance Criteria

- [x] It should display the party id and key on top and make it easy to copy the id
- [x] It should display a list of assets the party owns with the balance as a list.
- [x] It should make it easy to copy the asset id
- [ ] It should be accessible and easy to navigate using a keyboard
- [x] It should work in dark/light mode
- [ ] It should be tested

elmar-vega:
# Related issues 🔗

Closes #1334

# Description ℹ️

This PR is to make the explorer app more user friendly. I added some rendering components to render the JSON data in Party Data. I also changed the layout to be more in line with the current Vega styles. I also made minor adjustments to the testnet env file.

# Demo 📺

Screenshot 2022-09-23 at 13 20 21

# Technical 👨‍🔧

- Added Panel component
- Added InfoPanel component
- Made some network related changes

### DIFF
--- a/apps/explorer/src/app/components/header/header.tsx
+++ b/apps/explorer/src/app/components/header/header.tsx
@@ -1,3 +1,4 @@
+import React from 'react';
 import classnames from 'classnames';
 import { Link } from 'react-router-dom';
 import { ThemeSwitcher, Icon } from '@vegaprotocol/ui-toolkit';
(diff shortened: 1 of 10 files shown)


dexturr (approved):

apps/explorer/src/app/components/header/header.tsx
```diff
@@ -1,3 +1,4 @@
+import React from 'react';
```
dexturr, line 1: nitpick (non-blocking): not used

elmar-vega: Done

PR #1461, opened 2022-09-23
Chore/use cache in GitHub jobs

radoslaw-szpiech:
# Description ℹ️

Two changes in workflows:
- cache node_modules and Cypress runner to be used in other runs, if yarn.lock is not changed
- Download specific binaries version of vega, vegawallet & data-nodes instead of building them each time

It should improve timing in PRs.

### DIFF
--- a/.github/workflows/capsule-cypress-manual-trigger.yml
+++ b/.github/workflows/capsule-cypress-manual-trigger.yml
@@ -79,21 +79,15 @@ jobs:
           token: ${{ secrets.VEGA_CI_BOT_GITHUB_TOKEN }}
           path: './vega'
 
-      - name: Get yarn cache directory path
-        id: yarn-cache-dir-path
-        run: echo "::set-output name=dir::$(yarn cache dir)"
-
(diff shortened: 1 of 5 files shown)


.github/workflows/capsule-cypress-manual-trigger.yml
```diff
@@ -97,17 +97,20 @@ jobs:
 
       - name: Install Vega binaries
         if: ${{ env.RUN_CAPSULE }}
-        run: go install -v ./cmd/vega
-        working-directory: vega
+        run: |
```
dexturr, line 100: follow up (non-blocking): mght be nice to have a shared base action rather than copying this across eventually.

dexturr (approved):
Great stuff dude, nicely done

.github/workflows/capsule-cypress-manual-trigger.yml
radoslaw-szpiech: Yup, probably. There will be some little improvements later, I can add it then,

PR #1493, opened 2022-09-27
fix: 1425 add link on select market table row 

### LINKED ISSUE #1425: market switcher: clicking when there is a hand cursor does nothing
## Description

image

When I hover over the table row (e.g. I've drawn an orange circle) but not the actual text I have a hand cursor indicating that I can click. but when I click nothing happens.
it appears that i need to hit the market name text to open something.

either the hand shouldn't happen unless you are clicking the text. or the whole row should be clickable.

## Steps to Reproduce

Hover where orange circle is... click 

## Device details

- Mac firefox

MadalinaRaicu:
# Related issues 🔗

Closes #1425

# Description ℹ️

Add links to rows in select market components. (visible in both popover & dialog)

# Demo 📺

Screenshot 2022-09-27 at 10 02 47

# Technical 👨‍🔧


libs/market-list/src/lib/components/select-market-table.tsx
```diff
@@ -31,26 +32,44 @@ export const SelectMarketTableHeader = ({
...
+    if (event.key === 'Enter' && onSelect) {
+      return onSelect(id);
+    }
+  };
 
-        return null;
-      })}
-    </tr>
+  return (
+    <Link href={`/markets/${marketId}`} passHref={true}>
```
mattrussell36, line 50: chore (blocking): Two things here.

1. Pretty sure this results in invalid html <a> cannot wrap a <tr>.
2. The imported link tag is the nextjs link tag. This is probably fine for now, but if other apps use it might cause issues. I think the easiest thing is to just call call `router.push('/market/id')` in the onSelect handler in the nextjs page. This way the responsiblity of routing is left to the app only

mattrussell36 (changes requested):

mattrussell36 (approved):

asiaznik (approved):

macqbat (approved):

PR #1505, opened 2022-09-27
test: reintroduce tests that were failing

### LINKED ISSUE #1494: Re-enable disbaled tests after 0.56 release
Some tests were disabled as part of the below PR, we should re-enable these tests.

https://github.com/vegaprotocol/frontend-monorepo/pull/1487/files?file-filters%5B%5D=.tsx&show-deleted-files=true&show-viewed-files=true#diff-4be4af01f451770551210bfb17e4cd65bc4a912caf2e0b94e13537674538cb47

dexturr:
# Related issues 🔗

Closes #1494

# Description ℹ️

Reenable failing tests that were removed in as part of the 56 revert.

### DIFF
--- a/apps/token/.env
+++ b/apps/token/.env
@@ -13,3 +13,4 @@ NX_HOSTED_WALLET_URL=https://wallet.testnet.vega.xyz
 
 #Test configuration variables
 CYPRESS_FAIRGROUND=false
+LC_ALL="en_US.UTF-8"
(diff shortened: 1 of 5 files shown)


mattrussell36 (commented):
These still fail for me due to timezone issues. We can almost fix it by forcing the timezone with `TZ="UTC" nx test token` but it still breaks I think due to native datetime formatting

Perhaps we can change the assertion to parse the date to get around this?
`expect(new Date(screen.getByTestId('voting-date').textContent).toString()).toEqual(...)`

Screen Shot 2022-09-27 at 3 52 41 PM

MadalinaRaicu (approved):

PR #1515, opened 2022-09-28
Feat/1491 insert accounts and positions

### LINKED ISSUE #1391: Handle new account properly
## The Chore

Handle new account in libs/accounts/src/lib/accounts-data-provider.ts
When new account appears we are notified via subscription but there is missing data which we need

```
market {
    id
    tradableInstrument {
      instrument {
        name
      }
    }
  }
  asset {
    id
    symbol
    decimals
  }
```

We got two options here 
- change this provider be derived data provider and fetch assets, markets and accounts separately 
or
- re fetch accounts data when new account will appear

### LINKED ISSUE #1491: Update accounts and positions data providers to handle newly created entries
Right now when a new position or account is created it doesn't get added to the data providers store via the subscription. It used to, but due to recent changes to the API, which flatten the return types of some subscriptions, we can no longer simply insert it as some fields are now missing and must be retrieved from a different data source

## Tasks

- [x] Update the positions data provider to handle newly created positions
- [x] Update the accounts data provider  to handle newly created accounts
- [x] Test it by creating a new position on a market.

- libs/positions/src/lib/positions-data-providers.ts and libs/accounts/src/lib/accounts-data-provider.ts have a todo comment in the update function for the data provider
- You may want to look at the orders data provider which uses a derived data provider, as a similar issue was already fixed there.

macqbat:
# Related issues 🔗

Closes #1491 
Closes #1391 

# Description ℹ️

Handle adding new rows from subscriptions on account and positions views

# Technical 👨‍🔧

Extend dependencies in data providers and merge data from a few sources in one set


MadalinaRaicu:
@macqbat pulled your branch and something is wrong in collateral tab aggregation now, and this problem is not in develop. I was on stagnet3, had no collateral initially, but then topped up with 50,000 fDAI. as I was switching between tabs deposited became 100,000 fDAI. checked out back to develop and this issue was not there anymore. I am not sure why this happens. 
image
as I was trying to debug I saw update method was called with deltas twice

libs/accounts/src/lib/asset-data-provider.ts
```diff
@@ -0,0 +1,24 @@
+import { makeDataProvider } from '@vegaprotocol/react-helpers';
+import { AssetsDocument } from './__generated__/Accounts';
+import type {
+  AssetsQuery,
+  AssetsFieldsFragment,
+} from './__generated__/Accounts';
+
+const getData = (responseData: AssetsQuery) =>
+  responseData.assetsConnection?.edges?.reduce((aggr, edge) => {
```
mattrussell36, line 9: suggestion (non blocking): filter/map might be more readable here? There isn't really any perf reasons yet to prevent the additional loops
```
responseData.assetsConnection.edges
  .filter(e => Boolean(e.node))
  .map(e => e.node)
```

libs/positions/src/lib/positions-data-providers.ts
```diff
@@ -120,27 +114,26 @@ export const POSITIONS_SUBSCRIPTION = gql`
...
-      return account.market?.id === market.id;
+      return account.market?.id === market?.id;
     });
     if (
       !marginAccount ||
       !marginLevel ||
+      !market ||
       !marketData ||
-      position.node.openVolume === '0'
+      position.openVolume === '0'
```
mattrussell36, line 136: question (blocking): Why this condition? I think at this point we should show all positions regardless of whether they have volume or not

mattrussell36 (approved):

libs/positions/src/lib/positions-data-providers.ts
```diff
@@ -277,15 +278,67 @@ export const positionsDataProvider = makeDataProvider<
...
+  if (positions?.positionsConnection?.edges && marketsData && margins) {
+    return positions.positionsConnection.edges.map(
+      (nodes: Positions_party_positionsConnection_edges) => {
+        return {
+          realisedPNL: nodes.node.realisedPNL,
+          openVolume: nodes.node.openVolume,
+          unrealisedPNL: nodes.node.unrealisedPNL,
+          averageEntryPrice: nodes.node.averageEntryPrice,
+          updatedAt: nodes.node.updatedAt,
+          market:
```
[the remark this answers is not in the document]
macqbat: just didn't want to change original types. I would stay with current solution and clean it up in the future maybe..

libs/positions/src/lib/positions-data-providers.ts
```diff
@@ -18,6 +22,26 @@ import type {
...
+import { marginsDataProvider } from './margin-data-provider';
+import type { Margins_party } from './__generated__/Margins';
+
+interface PositionMarginLevel {
+  maintenanceLevel: string;
+  searchLevel: string;
+  initialLevel: string;
+}
+
+interface PositionRejoined {
```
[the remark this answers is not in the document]
macqbat: It would need many changes across trading and console-lite apps

libs/positions/src/lib/positions-data-providers.ts
macqbat: It's basically not my piece of code, and I wouldn't change general flow at this situation. It's rather case for decision of product owner and for separate issue.

PR #1518, opened 2022-09-28
chore: replace Order type with existing OrderSubmissionBody['orderSubission']

### LINKED ISSUE #1564: chore: replace Order type with existing OrderSubmissionBody['orderSubission']
## The Chore

Order type created for order submission is in naming conflict with Order from data provider. It can be replaced by OrderSubmissionBody['orderSubission']

## Tasks

- [x] replace Order type with existing OrderSubmissionBody['orderSubission'] 
- [x] rename OrderWithMarket to Order
- [x] rename TradeWithMarket to Trade
- [x] rename FillsWithMarket to Fills

bglownia:
- Replace Order type which was created to handle order creation and edit by OrderSubmissionBody['orderSubission'] imported from @vegaprotoocol/wallet.  Order was always transformed to OrderSubmissionBody['orderSubission'] and it is almost identical
 
 - drop WithMarket from Order, Trade and Fill types

closes #1564


mattrussell36 (approved):
Nice tidy up. Only thing I would say is its a bit annoying to have to access the property of the type each time. I'm wondering if we should create an alias or perhaps a namespace for all the wallet types

PR #1519, opened 2022-09-28
fix: orderbook sticky behaviour and precision options

### LINKED ISSUE #1251: Putting a limit order way out of normal price removes the aggregation/decimal precision dropdown in orderbook
https://user-images.githubusercontent.com/100769129/188490308-2e3fd2f1-fd20-4a30-98ac-eb73b9590964.mov

When putting a limit order at a price way out of "normal" the dropdown for decimal precision disappears.

1) Go to a market with e.g. 5 decimal points and a price around 4.22455 or a similarly low price
2) Put a limit short of 50 or similarly way off the price

Expected outcome:
Decimal precision dropdown should be available

Outcome: 
Decimal precision dropdown is no longer available.

https://user-images.githubusercontent.com/100769129/188490308-2e3fd2f1-fd20-4a30-98ac-eb73b9590964.mov

### LINKED ISSUE #1262: Decimal point precision on orderbook
On a market with 5 decimal point precision, there is no way to show less than 3 decimal points precision in the orderbook.

**Steps to reproduce:**
1) Enter a market in console V2 that has at least 5 decimal points precision
2) Try to change the orderbook's precision to 0, 1 or 2, precision

**Expected result:**
Being able group/show orderbook's precision with lass than 3 decimalpoints

**Result:** 
Can at most show 3 decimal points precision.

Screenshot 2022-09-06 at 16 45 54

asiaznik:
# Related issues 🔗

Closes #1251 
Closes #1262

# Description ℹ️

1) The orderbook's header and footer are no longer sticky but positioned absolutely because when sticky they were sometimes painted outside of the container or jumping while scrolling or hiding 🥸.
2) The precision dropdown displays now all options from market's decimal places to zero. In the future we may also want to aggregate by 10s or 100s but we've decided with Dan to leave it for now.
 

# Demo 📺

## Precision options:

https://user-images.githubusercontent.com/1980305/192805973-d109f950-a274-44ef-b1ce-56d613309ac7.mov

## Frantic scrolling up and down

https://user-images.githubusercontent.com/1980305/192806535-4e038b92-5152-499c-b16d-b92fa274ea06.mov

## Resizing (y-axis) of the Allotment.Pane

https://user-images.githubusercontent.com/1980305/192830744-9f9cb994-5142-4577-815e-c9d444f86fa1.mov

# Technical 👨‍🔧

Because of the scroll bar I needed to match the widths of the header and footer with the actual width of the grid hence there's a new hook `useResizeObserver` (while at it I've also added `useMutationObserver`).

### DIFF
--- a/apps/trading-e2e/src/support/vega-wallet.ts
+++ b/apps/trading-e2e/src/support/vega-wallet.ts
@@ -3,7 +3,7 @@ export const connectVegaWallet = () => {
   const manageVegaBtn = 'manage-vega-wallet';
   const walletName = Cypress.env('TRADING_TEST_VEGA_WALLET_NAME');
   const walletPassphrase = Cypress.env('TRADING_TEST_VEGA_WALLET_PASSPHRASE');
-
+  cy.log(`WALLET PASSPHRASE: ${walletPassphrase}`);
   cy.getByTestId('connect-vega-wallet').click();
   cy.getByTestId('connectors-list').find('button').click();
   cy.getByTestId(form).find('#wallet').click().type(walletName);
--- a/apps/trading-e2e/src/support/vega-wallet.ts
+++ b/apps/trading-e2e/src/support/vega-wallet.ts
@@ -3,7 +3,6 @@ export const connectVegaWallet = () => {
   const manageVegaBtn = 'manage-vega-wallet';
   const walletName = Cypress.env('TRADING_TEST_VEGA_WALLET_NAME');
   const walletPassphrase = Cypress.env('TRADING_TEST_VEGA_WALLET_PASSPHRASE');
-  cy.log(`WALLET PASSPHRASE: ${walletPassphrase}`);
   cy.getByTestId('connect-vega-wallet').click();
   cy.getByTestId('connectors-list').find('button').click();
   cy.getByTestId(form).find('#wallet').click().type(walletName);
--- a/libs/market-depth/src/lib/orderbook.tsx
+++ b/libs/market-depth/src/lib/orderbook.tsx
@@ -12,7 +12,12 @@ import {
 } from 'react';
 import classNames from 'classnames';
 
-import { formatNumber, t, ThemeContext } from '@vegaprotocol/react-helpers';
+import {
+  formatNumber,
+  t,
+  ThemeContext,
+  useResizeObserver,
+} from '@vegaprotocol/react-helpers';
 import { MarketTradingMode } from '@vegaprotocol/types';
 import { OrderbookRow } from './orderbook-row';
 import { createRow, getPriceLevel } from './orderbook-data';
@@ -109,6 +114,9 @@ export const Orderbook = ({
 }: OrderbookProps) => {
   const theme = useContext(ThemeContext);
   const scrollElement = useRef<HTMLDivElement>(null);
+  const gridElement = useRef<HTMLDivElement>(null);
+  const headerElement = useRef<HTMLDivElement>(null);
+  const footerElement = useRef<HTMLDivElement>(null);
   // scroll offset for which rendered rows are selected, will change after user will scroll to margin of rendered data
   const [scrollOffset, setScrollOffset] = useState(0);
   // actual scrollTop of scrollElement current element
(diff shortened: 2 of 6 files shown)


libs/react-helpers/src/hooks/use-mutation-observer.ts
```diff
@@ -0,0 +1,45 @@
...
+import debounce from 'lodash/debounce';
+import { useEffect, useState } from 'react';
+
+type MutationObserverConfiguration = {
+  debounceTime: number;
+  config: MutationObserverInit;
+};
+
+const DEFAULT_OPTIONS: MutationObserverConfiguration = {
+  debounceTime: 1000 / 16,
```
mattrussell36, line 11: suggestion (non blocking): Presumably this is to get around 60fps? Perhaps use requestionAnimationFrame, though it might be more complex to make debounce time configurable

libs/react-helpers/src/hooks/use-mutation-observer.ts
```diff
@@ -0,0 +1,45 @@
...
+};
+
+export function useMutationObserver(
+  target: Node | null,
+  callback: MutationCallback,
+  options: MutationObserverConfiguration = DEFAULT_OPTIONS
+) {
+  const [observer, setObserver] = useState<MutationObserver | null>(null);
+
+  useEffect(() => {
```
mattrussell36, line 26: suggestion (non blocking): I'm not a fan of setting state inside useEffects that then trigger other useEffects, it makes the code less linear and hard to read. I think the below should suffice? 

```
useEffect(() => {
  if (!target) return
  
  const observer = new MutationObserver(...)
  
  try {
    ...
  } catch() {
    ...
  } 
  
  return () => observer.disconnect()
}, [options, target, callback])
```

libs/react-helpers/src/hooks/use-resize-observer.ts
```diff
@@ -0,0 +1,43 @@
...
+};
+
+export function useResizeObserver(
+  target: Element | null,
+  callback: ResizeObserverCallback,
+  options: ResizeObserverConfiguration = DEFAULT_OPTIONS
+) {
+  const [observer, setObserver] = useState<ResizeObserver | null>(null);
+
+  useEffect(() => {
```
mattrussell36, line 24: suggestion (non blocking): same as with useMutationObserver, this could be combined into a single effect

libs/market-depth/src/lib/orderbook.tsx
```diff
@@ -313,40 +365,78 @@ export const Orderbook = ({
...
       >
-        {tableBody || (
-          <div className="inset-0 absolute">
-            <Splash>{t('No data')}</Splash>
-          </div>
-        )}
+        <div
+          className="relative text-right min-h-full"
+          style={{
+            paddingTop: `${paddingTop}px`,
```
mattrussell36, line 393: chore (blocking): no need to add 'px'

```suggestion
            paddingTop: paddingTop,
```

libs/market-depth/src/lib/orderbook.tsx
```diff
@@ -262,6 +270,50 @@ export const Orderbook = ({
...
+        entries.length === 0
+      ) {
+        return;
+      }
+      const {
+        contentRect: { width, height },
+      } = entries[0];
+      headerElement.current.style.width = `${width}px`;
+      footerElement.current.style.width = `${width}px`;
+      setViewportHeight(height);
```
mattrussell36, line 302: suggestion (non blocking): I realise that this component already uses `viewportHeight` and `setViewportHeight`, but perhaps its worth looking into using `AutoSizer`. We already wrap each trade grid window with it (TradeGridChild) and it should be fairly easy to refactor so that width and height are passed down as props

mattrussell36 (approved):
Looks good! One minor blocking change and a few suggestions

libs/react-helpers/src/hooks/use-mutation-observer.ts
asiaznik: Changed to `useMemo`

libs/react-helpers/src/hooks/use-mutation-observer.ts
asiaznik: Ah that's a leftover from the tests I've done it should be defaulted to 0 in order to cover 60fps because `MutationObserver` fires its changes in a microtask which gets being processed before the next repaint (same behaviour as in raf). Hence the debounce here is only for the operations that don't need to be processed as often.

libs/market-depth/src/lib/orderbook.tsx
asiaznik: I've created a new issue with your suggestion: https://github.com/vegaprotocol/frontend-monorepo/issues/1533
Wrapping the orderbook container quickly with AutoSizer resulted with width and height set to 0 😢. I think there's a clash somewhere or they don't like to be nested.

The change in width here is only needed to accommodate the scrollbar size so the rows are nicely aligned with each other and the scrollbar is not covered.

apps/trading-e2e/src/support/vega-wallet.ts
```diff
@@ -3,7 +3,7 @@ export const connectVegaWallet = () => {
   const manageVegaBtn = 'manage-vega-wallet';
   const walletName = Cypress.env('TRADING_TEST_VEGA_WALLET_NAME');
   const walletPassphrase = Cypress.env('TRADING_TEST_VEGA_WALLET_PASSPHRASE');
-
+  cy.log(`WALLET PASSPHRASE: ${walletPassphrase}`);
```
asiaznik, line 6:
```suggestion
```

PR #1521, opened 2022-09-28
fix: fixed hydration failed issue

asiaznik:
# Description ℹ️

image

The culprits were the theme icons (sun 🌞 and moon 🌚). Now they're rendered both as hidden and then "later" shown depending on the actual theme chosen.


bglownia (changes requested):
I think I found easier way to do this
https://github.com/vegaprotocol/frontend-monorepo/pull/1520/commits/df24c8dcc6bc7e5c6bf9fa6c2bf5a269946b14b3

PR #1523, opened 2022-09-28
Test/governance flows update

### LINKED ISSUE #1511: Refactor Governance proposal e2e tests
## The Chore

There's been some significant UI changes made to the way users submit governance proposals. 
Existing E2E tests need to be refactored to include the changes so tests runs can be green again

jtsang586:
# Related issues 🔗

Closes #1511 

# Description ℹ️

Governance flow tests now passing. 
Some tests have been skipped due to the form validation now no longer allowing failed proposals to be submitted.

### DIFF
--- a/apps/token-e2e/src/integration/flow/governance-flow.cy.js
+++ b/apps/token-e2e/src/integration/flow/governance-flow.cy.js
@@ -10,12 +10,9 @@ const vegaWallet = '[data-testid="vega-wallet"]';
 const vegaWalletName = Cypress.env('vegaWalletName');
 const vegaWalletPassphrase = Cypress.env('vegaWalletPassphrase');
 const connectToVegaWalletButton = '[data-testid="connect-to-vega-wallet-btn"]';
-const newProposalButton = '[data-testid="new-proposal-link"]';
-const newProposalDatabox = '[data-testid="proposal-data"]';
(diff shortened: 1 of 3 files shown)


apps/token-e2e/src/integration/flow/governance-flow.cy.js
```diff
@@ -1103,53 +1008,63 @@ context(
           .should('be.visible');
       });
 
-      it('Unable to create a freeform proposal - when json parent section contains unexpected field', function () {
+      // Json containing unexpected field no longer fails submission
+      it.skip('Unable to create a freeform proposal - when json parent section contains unexpected field', function () {
```
mattrussell36, line 1012: question (non blocking): I don't have full understanding of whats going on with these failed tests, but if this is functionality that no works differently lets just remove the tests

mattrussell36 (approved):

apps/token-e2e/src/integration/flow/governance-flow.cy.js
jtsang586: Normally I would agree but these tests are testing an AC which might not be needed anymore because of the form validations. I think it's best to keep them for now until we get that AC changed/removed or find a different way to test this.

PR #1544, opened 2022-09-29
Feat/1478 better proposal title

### LINKED ISSUE #1478: Better title for update asset governance proposals
Currently update asset proposals just say "update asset" this is becuase there is nothing in the API which will allow us to tell which asset is being updated. We should show the asset being updated in a lozenage as per the designs.

dexturr:
# Related issues 🔗

Closes #1478

# Description ℹ️

Add asset id to proposal title

### DIFF
--- a/apps/token/src/routes/governance/__generated__/ProposalFields.ts
+++ b/apps/token/src/routes/governance/__generated__/ProposalFields.ts
@@ -154,6 +154,10 @@ export interface ProposalFields_terms_change_UpdateAsset {
    */
   quantum: string;
   /**
+   * The asset to update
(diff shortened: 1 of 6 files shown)


mattrussell36 (approved):

mattrussell36:
If we have the assetId, why not query for the asset to show the name?

dexturr:
@mattrussell36 this conforms with the current pattern but something we can definitely look at in future. I am a little sceptical about queries hanging off of queries, maybe we can get the full asset/market/network parameter in the proposals object. That would certainly be my preference

PR #1558, opened 2022-09-29
Feat/1446: Add UpdateAsset proposal form

### LINKED ISSUE #1446: Add UpdateAsset proposal form

sam-keen:
# Related issues 🔗

Closes #1446

# Description ℹ️

Adds the update-asset proposal form

# Demo 📺

Screenshot 2022-09-29 at 19 21 22


apps/token/src/routes/governance/propose/propose.tsx
```diff
@@ -79,6 +79,16 @@ export const Propose = () => {
               </Link>
             </p>
           </li>
+          <li>
+            <p>
+              <Link
+                className="underline"
+                to={`${Routes.GOVERNANCE}/propose/update-asset`}
```
dexturr, line 86: thought (non-blocking): we store the top level routes in a constant I wonder if we want to add a ticket to store child routes names in an object as well to cu down on magic strings. Or maybe we should utilise route names more and just store those in an object.

apps/token/src/routes/governance/propose/update-asset/propose-update-asset.tsx
```diff
@@ -0,0 +1,190 @@
...
+export interface UpdateAssetProposalFormFields {
+  proposalVoteDeadline: string;
+  proposalEnactmentDeadline: string;
+  proposalTitle: string;
+  proposalDescription: string;
+  proposalTerms: string;
+  proposalReference: string;
+}
+
+const docsLink = '/update-asset-proposal';
```
dexturr, line 33: module scope constants should be WHATEVER_THIS_CASE_IS I can't recall the same haha

apps/token/src/routes/governance/propose/update-asset/propose-update-asset.tsx
```diff
@@ -0,0 +1,190 @@
...
+                params.governance_proposal_updateAsset_minProposerBalance
+              }
+              spamProtectionMin={params.spam_protection_proposal_min_tokens}
+            />
+
+            {VEGA_DOCS_URL && (
+              <p className="text-sm" data-testid="proposal-docs-link">
+                <span className="mr-1">{t('ProposalTermsText')}</span>
+                <Link
+                  href={`${VEGA_DOCS_URL}/tutorials/proposals${docsLink}`}
```
dexturr, line 98: can we pull this part out into a const as well please as it is duplicated here and a couple lines down

apps/token/src/routes/governance/propose/update-asset/propose-update-asset.tsx
```diff
@@ -0,0 +1,190 @@
...
+                <ProposalFormSubheader>
+                  {t('UpdateAsset')}
+                </ProposalFormSubheader>
+
+                <ProposalFormTerms
+                  registerField={register('proposalTerms', {
+                    required: t('Required'),
+                    validate: {
+                      validateJson: (value) => {
+                        try {
```
dexturr, line 143: don't we have an isJson util? Or is that only for the block explorer JSON bits?

dexturr (commented):
few minor things and ci pipelines failing. Otherwise all looks good 👍

apps/token/src/routes/governance/propose/update-asset/propose-update-asset.tsx
mattrussell36: We also have libs/react-helpers/src/lib/validate/index.ts could you put whatever function you end up using there? Also update other inputs that use this validation?

apps/token/src/routes/governance/propose/update-asset/propose-update-asset.spec.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+describe('Propose Update Asset', () => {
+  it('should render successfully', async () => {
+    const { baseElement } = renderComponent();
+    await expect(baseElement).toBeTruthy();
+  });
+
+  it('should render the title', async () => {
+    renderComponent();
+    await waitFor(() =>
```
mattrussell36, line 84: suggestion (non blocking): You could also use `expect(await screen.findByTestId('...'))` to prevent some nesting

mattrussell36 (approved):
Nothing blocking to add on top of Dex's comments

apps/token/src/routes/governance/propose/propose.tsx
sam-keen: https://github.com/vegaprotocol/frontend-monorepo/issues/1607

apps/token/src/routes/governance/propose/update-asset/propose-update-asset.tsx
sam-keen: This will be sorted by #1607

apps/token/src/routes/governance/propose/update-asset/propose-update-asset.tsx
sam-keen: @dexturr We used to have a poorly named 'isJson' util, but actually its use was for checking whether data was suitable for being displayed with the syntax highlighter - not only did it check it was valid json, but also checked whether the object had keys. It got renamed to 'suitableForSyntaxHighlighter'.

@mattrussell36 good shout, I'll move 'validateJson' there. Thanks

PR #1573, opened 2022-09-30
fix: 1569 trading tab navigation update

### LINKED ISSUE #1569: Trading tab inconsistent behaviour
The trading tab should let the user navigate to the last viewed market. 

This isn’t working because we only get your last viewed market from local storage when you visit the home page initially or select it from the dropdown (edited) 

If you navigate directly to the portfolio page this never happens so when you click trading there's no cached market.
There is a fallback in place where if that happens you just go to the full list (edited) 

We just need to ensure we set the last viewed market directly in the global store initialization or do it in _app

This can be a fix, though its possible we get hydration errors
```
export const useGlobalStore = create<GlobalStore>((set) => ({
  ...
  marketId: LocalStorage.getItem('the-key'),
  update: (state) => {
    set(state);
  },
}));
```

MadalinaRaicu:
# Related issues 🔗

Closes #1569 

# Description ℹ️

Updating navbar and saving marketId in local storage.


apps/trading/stores/global.ts
```diff
@@ -19,5 +20,8 @@ export const useGlobalStore = create<GlobalStore>((set) => ({
   pageTitle: null,
   update: (state) => {
     set(state);
+    if (state.marketId) {
```
MadalinaRaicu, line 23: navigation breaks if we do this on init. 
```
export const useGlobalStore = create<GlobalStore>((set) => ({
  ...
  marketId: LocalStorage.getItem('marketId'),
})); 
```

apps/trading/stores/global.ts
```diff
@@ -15,9 +16,12 @@ export const useGlobalStore = create<GlobalStore>((set) => ({
   networkSwitcherDialog: false,
   landingDialog: false,
   riskNoticeDialog: false,
-  marketId: null,
+  marketId: LocalStorage.getItem('marketId') || null,
```
MadalinaRaicu, line 19: getting rid of `undefined`

radoslaw-szpiech (approved):

PR #1580, opened 2022-09-30
fix: 1569 navigation link active on market selected refresh

### LINKED ISSUE #1569: Trading tab inconsistent behaviour

MadalinaRaicu:
# Related issues 🔗

Closes #1569 

# Description ℹ️

When refresh after market is selected, Trading tab does not seem to be active/highlighted. 

# Demo 📺 & Technical 👨‍🔧

image


apps/trading/components/navbar/navbar.tsx
```diff
@@ -59,7 +52,7 @@ interface NavLinkProps {
 
 const NavLink = ({ name, path, testId = name }: NavLinkProps) => {
   const router = useRouter();
-  const isActive = router.asPath === path;
+  const isActive = router.asPath.includes(path);
```
MadalinaRaicu, line 55: sth weird happening here. it works for the trading tab but i should not be doing this. fails on the portfolio tab after refresh.

apps/trading/components/navbar/navbar.tsx
```diff
@@ -59,7 +58,11 @@ interface NavLinkProps {
 
 const NavLink = ({ name, path, testId = name }: NavLinkProps) => {
   const router = useRouter();
-  const isActive = router.asPath === path;
+  const [isActive, setIsActive] = useState(false);
```
MadalinaRaicu, line 61: Fixing hydration error. By calling `setIsActive` in `useEffect` a render is triggered after hydrating. https://nextjs.org/docs/messages/react-hydration-error

macqbat (approved):

PR #1581, opened 2022-10-01
ci: separate capsule test run from non capsule test run

mattrussell36:
# Description ℹ️

Separates capsule from non capsule runs so that the wallet service can be started pointing to the right network. E2E projects are targeted using `nx affected` with the `--excludes` flag


radoslaw-szpiech:
(Blocking) Please, share the job where you noticed this behaviour

radoslaw-szpiech:
I would rather have it fixed in this way :) 
https://github.com/vegaprotocol/frontend-monorepo/pull/1583

radoslaw-szpiech (changes requested):
I would rather fix it this way - https://github.com/vegaprotocol/frontend-monorepo/pull/1583

PR #1585, opened 2022-10-03
Chore/migrate types

notbot00:
More type migrations:

- `governance`
- `liquidity`
- `market-depth`
- `market-info`
- `market-list`
- `react-helpers`
- `web3`


notbot00:
i've tried to untangle this, but i just can't. will have to restart once again.

PR #1614, opened 2022-10-04
[trading] Prefix asset path with . (for ipfs)

### LINKED ISSUE #1613: IPFS deploys do not work for subfolder gateways
next.js builds static pages, and those static pages reference static assets as `/_next/...`, which works fine if the
site is served from a domain root (e.g. `console.fairground.wtf/`, `made-up-url.ipfs.io`) but breaks down on IPFS
gateways that are still configured to use subfolders (e.g. `ipfs.io/ipfs/made-up-url`). ~Luckily we can fix this by
using `assetPrefix` to turn `/_next/` in to `./_next/`, which works in both cases.~ Narrator: It doesn't. 

# Tasks
- [ ] Update configuration for trading

Blocks #597

edd:
Configure assetPrefix for trading so that IPFS works with subfolder or subdomains

next.js builds static pages, and those static pages reference static assets as `/_next/...`, which works fine if the
site is served from a domain root (e.g. `console.fairground.wtf/`, `made-up-url.ipfs.io`) but breaks down on IPFS
gateways that are still configured to use subfolders (e.g. `ipfs.io/ipfs/made-up-url`). Luckily we can fix this by
using `assetPrefix` to turn `/_next/` in to `./_next/`, which works in both cases.

- Update next.js config for trading

Closes #1613


edd:
Bah - branched from `main` not `develop`. Fixing.

edd:
See #1615

PR #1615, opened 2022-10-04
[trading] IPFS deploys do not work for subfolder gateways

### LINKED ISSUE #1613: IPFS deploys do not work for subfolder gateways

edd:
Configure assetPrefix for trading so that IPFS works with subfolder or subdomains

next.js builds static pages, and those static pages reference static assets as `/_next/...`, which works fine if the
site is served from a domain root (e.g. `console.fairground.wtf/`, `made-up-url.ipfs.io`) but breaks down on IPFS
gateways that are still configured to use subfolders (e.g. `ipfs.io/ipfs/made-up-url`). Luckily we can fix this by
using `assetPrefix` to turn `/_next/` in to `./_next/`, which works in both cases.

- Update next.js config for trading

Closes #1613

### DIFF
--- a/apps/trading/next.config.js
+++ b/apps/trading/next.config.js
@@ -19,6 +19,7 @@ const nextConfig = {
     // See: https://github.com/gregberge/svgr
     svgr: false,
   },
+  assetPrefix: '.',
   pageExtensions: ['page.tsx', 'page.jsx'],
 };
 


mattrussell36 (approved):

edd:
My mistake - I thought I tested thoroughly enough, but this does indeed break if you navigate directly to a nested URL (https://deploy-preview-1615--vega-trading-app.netlify.app/markets/c7075eebe699920f6aa53caa43aa09e62c1c17c0e051b6ff89fa9b4d875339d5) which I did not check. Thanks for reverting @mattrussell36  - I'll find a way to set this in the build on Fleek but not elsewhere

PR #1634, opened 2022-10-05
fix: market info data provider and oracle spec data - API changes

### LINKED ISSUE #1621: Market info broken in devnet - oracleSpec update
```
"Cannot query field \"oracleSpecForSettlementPrice\" on type \"Future\". Did you mean \"oracleSpecForSettlementData\" or \"oracleSpecForTradingTermination\"?"
"Cannot query field \"settlementPriceProperty\" on type \"OracleSpecToFutureBinding\". Did you mean \"settlementDataProperty\"?"
```

Image

MadalinaRaicu:
# Related issues 🔗

Closes #1621

**Only in devnet until testnet upgrade**
Cypress test is failing because it runs on testnet

This PR 

- [x] fixes market info being broken in console after API changes regarding oracles. 
- [x] add market info data provider. 

# Demo 📺

image


libs/market-info/src/components/market-info/info-market.tsx
```diff
@@ -1,51 +1,43 @@
...
 import Link from 'next/link';
+import { marketInfoDataProvider } from './market-info-data-provider';
 
 const Links = {
   PROPOSAL_PAGE: ':tokenUrl/governance/:proposalId',
 };
 
 export interface InfoProps {
-  market: MarketInfoQuery_market;
+  market: Schema.Market;
```
mattrussell36, line 32: chore (blocking): Why this change? We should be using actual generated query type here as if someone changes a field the component will still expect the entire schema.

Same goes for other uses of types directly from Schema. You may want to create type aliases if the types are super long and make the code hard to read

mattrussell36 (changes requested):

libs/market-info/src/components/market-info/info-market.tsx
MadalinaRaicu: agree, updated

mattrussell36 (approved):

libs/market-info/src/components/market-info/info-market.tsx
```diff
@@ -318,9 +312,9 @@ export const Info = ({ market, onSelect }: InfoProps) => {
           data={market.tradableInstrument.instrument.product.oracleSpecBinding}
         >
           <ExternalLink
-            href={`${VEGA_EXPLORER_URL}/oracles#${market.tradableInstrument.instrument.product.oracleSpecForSettlementPrice.id}`}
+            href={`${VEGA_EXPLORER_URL}/oracles#${market.tradableInstrument.instrument.product.oracleSpecForSettlementData.id}`}
           >
-            {t('View price oracle specification')}
+            {t('View settlement data oracle specification')}
```
MadalinaRaicu, line 317: I will merge this once 0.58 is deployed on testnet

dexturr (approved):

PR #1664, opened 2022-10-06
Feat/lp dashboard markets

### LINKED ISSUE #1354: Health + health tooltip
**User Story**
As a liquidity provider 
I want to understand the current health of the market in more detail, the fee levels and target stake
So that I can make a decision on providing liquidity but also understand better how liquidity on Vega actually behaves 

**Acceptance Criteria**

- There is a way to see the market health based on market status, liquidity status and how close that market is to moving from one fee level to another, and how close it is to the target stake
- We've found a way to progressively show the information e.g. mouseover 
- There is a way to access "key" type information explaining how the market health is calculated  e.g. a tooltip with list of definitions or similar

image
image

Dashboard sketch for reference https://miro.com/app/board/uXjVOs_PBKk=/?moveToWidget=3458764531971937952&cot=14

Visual styles can be used from this early version of a design https://www.figma.com/file/qgCd4r61Z1r3tUkxKehhmu/Website-Design-Updates?node-id=399%3A20530

**Tasks**

- [ ]

### LINKED ISSUE #1357: Benefits + links to docs
**User Story**
As a liquidity provider
I want to understand the benefits of providing liquidity on Vega and how to to do it 
So that I can make an informed decision on what I do next

**Acceptance Criteria**

- There is a short description of the benefit of providing liquidity on Vega
- There are links to [Understand how liquidity fees are calculated](https://docs.vega.xyz/docs/testnet/tutorials/providing-liquidity#resources) ↗ [How to p](https://docs.vega.xyz/docs/testnet/tutorials/providing-liquidity#overview)[rovide liquidity](https://docs.vega.xyz/docs/testnet/tutorials/providing-liquidity#overview) ↗
- [View existing liquidity provisions](https://docs.vega.xyz/docs/testnet/tutorials/providing-liquidity#viewing-existing-liquidity-provisions)↗ [How to amend or remove liquidity](https://docs.vega.xyz/docs/testnet/tutorials/providing-liquidity#amending-a-liquidity-commitment) ↗

image

Dashboard sketch for reference https://miro.com/app/board/uXjVOs_PBKk=/?moveToWidget=3458764531971937952&cot=14

Visual styles (only) can be used from this early version of a design https://www.figma.com/file/qgCd4r61Z1r3tUkxKehhmu/Website-Design-Updates?node-id=399%3A20530

**Tasks**

- [ ]

malinantonsson:
# Related issues 🔗

Closes #1354, closes #1357

# Description ℹ️

- Get and display a list of markets for the Liquidity provision dashboard. This includes market name, asset, volume and 24h - change, committed liquidity, status (trading mode) and the market health.
- Display a health dialog, explaining the different market statuses
- Display an intro section with links to docs

# Demo 📺
Market list
market-list

Health dialog
Screenshot 2022-10-11 at 16 44 02

# Technical 👨‍🔧
- I've created a new provider in the `libs/liquidity` to fetch data + export utility functions
- There's a hook in `apps/liquidity-provision-dashboard` that pulls in different providers (markets liquidity & candles) and combines them


libs/liquidity-provision/src/lib/liquidity-provision-provider.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+import type {
+  LiquidityProvisionMarkets as Markets,
+  LiquidityProvisionMarkets_marketsConnection_edges_node as Market,
+} from './__generated__';
+
+import type { MarketsListData } from './utils';
+import { mapDataToMarketList } from './utils';
+
+export const MARKET_LIST_QUERY = gql`
```
mattrussell36, line 20: chore (blocking): Can you put this query in a .graphql file and utilise the generated hooks

libs/liquidity-provision/README.md
```diff
@@ -0,0 +1,7 @@
+# liquidity-provision
```
mattrussell36, line 1: chore (blocking): We already have a 'liquidity' lib which basically does the same thing. I don't think we need a new lib for this as they are too similar. Can you look into utilising whats in the already existing lib? Feel free to make changes there if required, but please check that the `liquidity/<marketid>` page on the console app still works :-)

apps/liquidity-provision-dashboard/src/app/components/header/header.tsx
```diff
@@ -0,0 +1,13 @@
...
+    <div className="flex items-stretch px-6 py-6" data-testid="header">
+      <h1 className="text-3xl">{t('Top liquidity opportunities')}</h1>
+      <div className="flex items-center gap-2 ml-auto relative z-10">
+        {t('Network switcher')}
+      </div>
+    </div>
+  );
+};
+
+export default Header;
```
mattrussell36, line 13: chore (blocking): Please use named exports and avoid default exports unless default is 100% required (usually only for lazy loading of components)

apps/liquidity-provision-dashboard/src/app/components/market-list/health-bar.scss
```diff
@@ -0,0 +1,143 @@
+.App {
```
mattrussell36, line 1: chore (blocking): Please use tailwind

suggestion (non blocking): We have a ProgressBar component in the toolkit, which looks quite similar but not exactly the same. Have you considered using it so we can keep things visually consistent? Are there any specific features it needs to support for you to use it?

apps/liquidity-provision-dashboard/src/app/app.tsx
```diff
@@ -1,9 +1,24 @@
...
+
 import '../styles.scss';
+import Header from './components/header';
+import Intro from './components/intro';
+import MarketList from './components/market-list';
 
 export function App() {
+  const { data, error, loading } = useLiquidityProvision();
+
+  console.log('error: ', error);
```
mattrussell36, line 11: suggestion (non blocking): Remove stray logs

mattrussell36 (changes requested):

libs/liquidity-provision/src/lib/liquidity-provision-provider.tsx
malinantonsson: Hey @mattrussell36 how do I generate the hooks for this? When I run the command it only creates the types, no hooks. Also, can you point me to any files where the hooks are used? I've seen `. graphql` files in the repo (and the hooks), but I can't find any place where they are actually used.

libs/liquidity-provision/README.md
malinantonsson: I've removed this lib and moved it into the `liquidity` lib. They don't share any code though - the biggest different is that `libs/liquidity/src/lib/liquidity-data-provider.ts` fetches data for only 1 market and I need data for all markets. I also need candles, and some other data + formatting since the dashboard is not _only_ about liquidity. However, there is another ticket to display liquidity providers for 1 market only (similar to `liquidity/<marketid>`) and at that point I'll be able to reuse code from this library.

libs/liquidity/src/lib/utils/liquidity-utils.ts
```diff
@@ -0,0 +1,122 @@
...
+  orderBy(markets, ['marketTimestamps.open', 'id'], ['asc', 'asc']);
+
+const sortMarkets = (markets: Market[]) =>
+  markets.sort(
+    (a, b) =>
+      tradingModesOrdering.indexOf(a.tradingMode) -
+      tradingModesOrdering.indexOf(b.tradingMode)
+  );
+
+// libs/market-list/src/lib/utils/market-utils.ts
```
malinantonsson, line 45: @mattrussell36 some functions in this file (where I've put the path references) are pretty much duplicates of existing functions but I can't reuse them because the types are not _exactly_ the same. That is, some libs pull in more fields than other and the functions uses the generated `types`, which means that I get TS errors when I try to reuse them. 
For example, row 38 here only requires an array of objects with a `tradingMode` key but `Market` is a generated type and has many more required fields. 
Is there a way to handle these things? To me, it would make sense to have a lib with some shared functions and simplified types, let me know if you think it's a good idea or an overkill.

libs/liquidity-provision/src/lib/liquidity-provision-provider.tsx
mattrussell36: useChainIdQuery in connect-dialog is one example. There aren't many examples of using the hooks directly because this a new utility that is only available as a result of migrating from Apollo's now deprecated built in code generation to @graphql-codegen.

That said, I now see that you are setting up a data provider anyway, so rather than using the hook use the generated query document from the same file, have a look at libs/positions/src/lib/positions-data-providers.ts, its likely you will want something similar.

As to why the hooks aren't being generated, I'm not sure, happy to jump on a call to see whats up

libs/liquidity-provision/README.md
mattrussell36: So ideally libs/liquidity would export hooks that can handle both use cases. Perhaps this is a single hook which takes an optional marketId argument or perhaps two separate hooks to keep it simpler.

If there is additional market data required for the page, I'd either split that up into a separate query or possibly use the market-data-provider in libs/market-list.

libs/liquidity/src/lib/utils/liquidity-utils.ts
mattrussell36: So I here you are spot on, where possible we should refactor these functions to declare their own types making them easier to work with across apps.

For example `calcDayVolume` takes an entire Market object when really we only need an array of partial candle objects.

```
const calcDayVolume = (market: Market) => BigNumber

// changes to

const caclDayVolume = (candles: Array<{ volume: string }>) => BigNumber
```

mattrussell36: PS this is something that is a bit messy in the repo as a whole. There is too much sharing of types between libs, ideally each lib will define its own types and its up to the app to fulfill the function's signature.

libs/liquidity-provision/src/lib/liquidity-provision-provider.tsx
malinantonsson: This is now working, my bad - I was using Apollo's built in code generation 🤦‍♀️

libs/liquidity-provision/README.md
malinantonsson: Having a look at `libs/positions/src/lib/positions-data-providers.ts` was really helpful, thanks! :) I've created a second hook and I'm reusing `marketListProvider` & `marketsCandlesProvider` from `libs/market-list`.

libs/liquidity/src/lib/utils/liquidity-utils.ts
malinantonsson: @mattrussell36 Since I'm now reusing some providers some of these functions are no longer needed. The other ones I have refactored to declare their own types and only pass in the values that are actually needed 👍

mattrussell36 (approved):

apps/liquidity-provision-dashboard/src/app/components/market-list/health-bar.tsx
```diff
@@ -0,0 +1,199 @@
...
+  [MarketTradingMode.TRADING_MODE_CONTINUOUS]: '#00a88a',
+  [MarketTradingMode.TRADING_MODE_MONITORING_AUCTION]: '#fb8e7f',
+  [MarketTradingMode.TRADING_MODE_OPENING_AUCTION]: '#68e2e4',
+  [MarketTradingMode.TRADING_MODE_BATCH_AUCTION]: 'batch',
+  [MarketTradingMode.TRADING_MODE_NO_TRADING]: 'none',
+};
+
+const COPY_CLASS = 'text-[8px] leading-[1.2em] font-medium';
+
+const Tooltip = ({
```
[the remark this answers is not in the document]
malinantonsson: There's a ticket for reviewing the designs for this whole component - when I get to it I will try to reuse the tooltip component from ui-toolkit.

PR #1672, opened 2022-10-06
chore: market header live update

### LINKED ISSUE #1595: Market data is not updated in TradeMarketHeader
## Steps to Reproduce

1. Go to any market page
2. Market data (like market trading mode, mark price, trading volume) is not updated

## Expected behavior

TradeMarketHeader show real time market data

## Additional context

Page level MARKET_QUERY should be removed and replaced by fine grained data providers. 
TradeMarketHeader should use makedDataProvider which is subscribed to market data changes.

macqbat:
# Related issues 🔗

Closes #1595 

# Description ℹ️

Make market header live updated

# Demo 📺

Screenshot 2022-10-06 at 16 16 31

# Technical 👨‍🔧

Use data provider with subscription for candles and marketData instead cached useQuery


libs/market-list/src/lib/market.graphql
```diff
@@ -0,0 +1,53 @@
...
+          quoteName
+        }
+      }
+    }
+  }
+  marketTimestamps {
+    open
+    close
+  }
+  depth {
```
bglownia, line 42: market.depth.lastTrade.price should be taken from market depth provider

bglownia (changes requested):
market.depth.lastTrade.price should not be included in marketProvider. lastTrade is real time data and this provider does not have subscriptions. 
Value should be taken from market depth provider (it needs to be modified to get this value)

libs/market-list/src/lib/market.graphql
macqbat: it appears unnecessary in any view so I removed it completely

MadalinaRaicu:
> market.depth.lastTrade.price should not be included in marketProvider. lastTrade is real time data and this provider does not have subscriptions. Value should be taken from market depth provider (it needs to be modified to get this value)

@bglownia  this reminds me that last trade price is shown in deal ticket. I think we are not using a data provider over there just simple query. we may need a ticket

apps/trading/pages/markets/trade-grid.tsx
```diff
@@ -221,95 +221,106 @@ export const TradeMarketHeader = ({
...
+          </Tab>
+          <Tab id="accounts" name={t('Collateral')}>
+            <TradingViews.Collateral />
+          </Tab>
+        </Tabs>
+      </TradeGridChild>
+    </ResizableGridPanel>
+  </ResizableGrid>
+);
+const MainGridWrapped = memo(MainGrid);
```
mattrussell36, line 317: question (non blocking): Was there a specific reason for memo here? Should we be wrapping TradePanels with memo too? It also renders the header.

macqbat: It was added for avoid re-rendering all child components, what was caused by passing marketID down to the children. But now I'm gonna change a little bit the conception, following Bart's advice it can be better to break down the update to all children separately

apps/trading/pages/markets/[marketId].page.tsx
```diff
@@ -130,29 +102,33 @@ const MarketPage = ({ id }: { id?: string }) => {
...
+  const { data, error, loading } = useDataProvider({
+    dataProvider,
+    update: dataUpdate,
+    variables,
+    skip: !marketId,
+  });
+  dataRef.current = data;
+  const marketName = data?.tradableInstrument.instrument.name;
+
   useEffect(() => {
```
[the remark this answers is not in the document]
Comment (macqbat) [reply apps/trading/pages/markets/[marketId].page.tsx]: basically done

apps/trading/pages/markets/[marketId].page.tsx
```diff
@@ -130,29 +102,33 @@ const MarketPage = ({ id }: { id?: string }) => {
...
-    MARKET_QUERY,
-    {
-      variables,
-      fetchPolicy: 'network-only',
-      errorPolicy: 'ignore',
-    }
+  const dataUpdate = useCallback(
+    ({ data }: { data: SingleMarketData }) => {
+      setMarketPrice(calculatePrice(data.data.markPrice, data.decimalPlaces));
+      if (isEqual(data, dataRef.current)) {
```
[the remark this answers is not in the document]
Comment (macqbat) [reply apps/trading/pages/markets/[marketId].page.tsx]: yeah, we went this way, you can check again

apps/trading/pages/markets/trade-grid.tsx
macqbat: So TradePanel also is wrapped with `memo`. It just prevent re-render by restictly checking the props. Without it re-render happens also when `market` object changes in the parent space, what occur quite often

PR #1696, opened 2022-10-10
feat: 1648 add deposit button and show accounts 0 state

### LINKED ISSUE #1648: Accounts table zero assets state
## Story

As a trader
I want to easily deposit from the accounts table
So that if I don't have assets I can easily deposit them

## Acceptance Criteria

- [ ] I'm still prompted to connect my wallet if I haven't already
- [ ] If I have connected, but don't have assets I am prompted to deposit
- [ ] When prompted to deposit I can open the deposit dialog

MadalinaRaicu:
# Related issues 🔗

Closes #1648

# Description ℹ️

Able to deposit any asset from the collateral table. 

# Demo 📺 & Technical 👨‍🔧
Screenshot 2022-10-11 at 18 11 16

Screenshot 2022-10-11 at 18 11 26
Screenshot 2022-10-11 at 18 11 20


libs/accounts/src/lib/accounts-manager.tsx
```diff
@@ -38,37 +36,20 @@ export const AccountManager = ({
...
-    successCallback(rowsThisBlock, lastRow);
-  };
   return (
-    <AsyncRenderer loading={loading} error={error} data={data}>
-      {data && (
-        <AccountTable
-          rowModelType={data?.length ? 'infinite' : 'clientSide'}
-          rowData={data?.length ? undefined : []}
-          ref={gridRef}
-          datasource={{ getRows }}
```
MadalinaRaicu, line 65: I don't think I should use datasource in this scenario.  @bglownia 

Screenshot 2022-10-10 at 13 14 17

mattrussell36 (changes requested):

libs/accounts/src/lib/accounts-manager.tsx
MadalinaRaicu: It is not consistent with Deposits and Withdrawals. The Async renderer would simply display "No data" and no headers. I have the logic to display the headers and `No accounts` overlay within the account table. It is consistent with `Deposits` and `Withdrawals` and it looks quite clean. 

image
Screenshot 2022-10-10 at 13 04 45

apps/trading/pages/portfolio/accounts-container.tsx
```diff
@@ -24,6 +24,12 @@ export const AccountsContainer = () => {
   return (
     <Web3Container>
       <div className="h-full">
+        <header className="flex justify-between items-center p-4">
+          <h4 className="text-lg text-black dark:text-white">
+            {t('Collateral')}
+          </h4>
+          <Button onClick={() => setDepositDialog(true)}>Deposit</Button>
```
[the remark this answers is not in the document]
MadalinaRaicu: I can make the button small but should I change it in `Deposits` too?  I was trying to keep them consistent

Screenshot 2022-10-10 at 16 11 25

libs/accounts/src/lib/accounts-manager.tsx
```diff
@@ -38,39 +36,22 @@ export const AccountManager = ({
...
-          onClickAsset={onClickAsset}
-          onClickDeposit={onClickDeposit}
-          onClickWithdraw={onClickWithdraw}
-        />
-      )}
-    </AsyncRenderer>
+    <AccountTable
```
bglownia, line 48: Data in this table will not be updated. It will always render data returned by initial value of { data } from useDataProvider hook.
What is the reason behind removing rowModelType and datasource ?

bglownia (changes requested):

libs/accounts/src/lib/accounts-manager.tsx
MadalinaRaicu: This is the reason why I removed it. 
Screenshot 2022-10-10 at 13 14 17

And how exactly does it work, because if I add getRows, rowModelType and dataSource back it seems that I have no accounts.  is this because I am doing `rowData={data?.length ? undefined : []}`
Screenshot 2022-10-11 at 16 10 42

mattrussell36 (approved):

libs/accounts/src/lib/accounts-manager.tsx
bglownia:
```
rowModelType={data?.length ? 'infinite' : 'clientSide'}
rowData={data?.length ? undefined : []}
```

We switch between rowModelTypes because
 - infinite is much easier to do update without re rendering 
 - clientSide because ag-grid do not support overlayNoRowsTemplate when its in infinite rowModelType
 🤷

bglownia (approved):

libs/accounts/src/lib/accounts-manager.tsx
```diff
@@ -57,7 +58,7 @@ export const AccountManager = ({
     successCallback(rowsThisBlock, lastRow);
   };
   return (
-    data && (
+    <AsyncRenderer data={data || []} error={error} loading={loading}>
```
MadalinaRaicu, line 61: this is to not show "No data" but rather display the ag-grid headers and show "No accounts"

PR #1701, opened 2022-10-10
Fix/1326: Do not allow proposal voting unless sufficient associated t…

### LINKED ISSUE #1326: Governance - No handling of attempt to vote without enough tokens associated scenario
If you attempt to vote on a freeform proposal and you have some but not enough tokens to vote.
The proposal details form allows you to vote - but then returns with 'Something wen't wrong'
As part of fixing maybe we could consider including the number of tokens required to vote on this form?

Associate a few more tokens to ensure you meet the governance.proposal.freeform.minVoterBalance network parameter.
Re-attempt above steps - and flow is as expected - allows vote.

likely to affect all othert proposal types too - only checked freeform

Image

sam-keen:
…okens

# Related issues 🔗

Closes #1326

# Description ℹ️

Users were able to click on voting buttons and would then receive an error as they didn't have enough associated tokens. This PR checks the proposal for its type, and notifies the user of the required amount of tokens to allow them to vote, rather than showing them the buttons.


apps/token/src/i18n/translations/dev.json
```diff
@@ -199,6 +199,7 @@
   "governanceRequired": "Required",
   "daysLeft": "{{daysLeft}} left to vote.",
   "toVote": "to vote",
+  "youNeed{{amount}}AssociatedVegaTokensToVote": "You need a minimum of {{amount}} associated $VEGA tokens to vote",
```
dexturr, line 202: chore: this isn't how other strings with varaibles are defined, would be nice to have consistency

apps/token/src/routes/governance/components/vote-details/vote-details.tsx
```diff
@@ -30,6 +31,42 @@ export const VoteDetails = ({ proposal }: VoteDetailsProps) => {
     requiredParticipation,
   } = useVoteInformation({ proposal });
 
+  let minVoterBalance = null;
```
dexturr, line 34: question (blocking): How does spam protection play into this? Do we need to take a max?

dexturr (changes requested):
Need to understand how spam protection plays into this

apps/token/src/routes/governance/components/vote-details/vote-buttons.tsx
```diff
@@ -23,6 +23,7 @@ interface VoteButtonsContainerProps {
   castVote: (vote: VoteValue) => void;
```
dexturr, line 23: chore (blocking): can we add some basics tests for this please, and time box some back filling of other functionality in here

apps/token/src/routes/governance/components/shared/proposal-min-requirements.tsx
```diff
@@ -0,0 +1,38 @@
...
+  );
+
+  const larger =
+    minProposalBalanceFormatted > spamProtectionMinFormatted
+      ? minProposalBalanceFormatted
+      : spamProtectionMinFormatted;
+
+  return (
+    <div className="mb-4" data-testid="min-proposal-requirements">
+      {userAction === 'create' &&
```
dexturr, line 32: nitpick (non-blocking): enum would be better over magic strings

dexturr (approved):

PR #1724, opened 2022-10-12
Test/block exp node tests

### LINKED ISSUE #1719: Increase blockExp validators test coverage to include node data

AndyWhiteVega:
# Related issues 🔗

Closes #1719 

# Description ℹ️

- Extra test for node data
- Tweaks to existing checks

# Demo 📺

# Technical 👨‍🔧

### DIFF
--- a/apps/explorer-e2e/src/integration/validator.cy.js
+++ b/apps/explorer-e2e/src/integration/validator.cy.js
@@ -8,6 +8,7 @@ context('Validator page', { tags: '@smoke' }, function () {
     cy.visit('/');
     cy.get(validatorMenuHeading).click();
     cy.get_validators().as('validators');
+    cy.get_nodes().as('nodes');
   });
 
   describe('Verify elements on page', function () {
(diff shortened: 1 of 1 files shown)


apps/explorer-e2e/src/integration/validator.cy.js
```diff
@@ -82,8 +83,98 @@ context('Validator page', { tags: '@smoke' }, function () {
...
+      cy.get(vegaDataHeader)
+        .contains('Vega data')
+        .next()
+        .within(() => {
+          cy.get(jsonSection)
+            .invoke('text')
+            .convert_string_json_to_js_object()
+            .then((nodesInJson) => {
+              this.nodes.forEach((node, index) => {
+                const nodeInJson = nodesInJson.nodes[index];
```
radoslaw-szpiech, line 100: I think it'd be better to just loop through nodeInJson keys & values instead of writing each case explicitly.

AndyWhiteVega: Unfortunately i can't do that, since some of the values are null, some are blank - Limitations of vegacapsule/config
As mentioned in the comments. Core will be working on unlocking more functionality, hopefully we can then loop through each key

jtsang586 (approved):

PR #1726, opened 2022-10-12
market page: break down components to smaller chunks for better performance

### LINKED ISSUE #1720: redundant re-renderings at single market page
## Description

After last changes in single market page in console app we can see blinking of position and order tabs, what is caused perhaps by redundant re-rendering of parent containers  

## Steps to Reproduce

1. Go to market page
2. See blinking position list
3. Click on order tab
4. See blinking order list
5. Fills and collateral list seems to be ok.

## Expected behavior

Re-rendering should occur only when values are changed

macqbat:
# Related issues 🔗

Closes #1720 

# Description ℹ️

Manage components of market page for better performance

# Technical 👨‍🔧

Avoid re-renders caused by `zustand` updates

### DIFF
--- a/libs/wallet/src/connect-dialog/connect-dialog.tsx
+++ b/libs/wallet/src/connect-dialog/connect-dialog.tsx
@@ -1,3 +1,4 @@
+import create from 'zustand';
 import {
   Button,
   Dialog,
@@ -28,15 +29,28 @@ type WalletType = 'gui' | 'cli' | 'hosted';
 
 export interface VegaConnectDialogProps {
   connectors: Connectors;
-  dialogOpen: boolean;
-  setDialogOpen: (isOpen: boolean) => void;
 }
 
-export const VegaConnectDialog = ({
-  connectors,
-  dialogOpen,
-  setDialogOpen,
-}: VegaConnectDialogProps) => {
+interface VegaWalletDialogStore {
+  dialogOpen: boolean;
+  updateDialogOpen: (open: boolean) => void;
+}
+
+export const useVegaWalletDialogStore = create<VegaWalletDialogStore>(
+  (set) => ({
+    dialogOpen: false,
+    updateDialogOpen: (open: boolean) => set({ dialogOpen: open }),
+  })
+);
+
+export const VegaConnectDialog = ({ connectors }: VegaConnectDialogProps) => {
+  const { dialogOpen, updateDialogOpen } = useVegaWalletDialogStore(
+    (store) => ({
+      dialogOpen: store.dialogOpen,
+      updateDialogOpen: store.updateDialogOpen,
+    })
+  );
+
   const { data, error, loading } = useChainIdQuery();
 
   const renderContent = () => {
@@ -66,14 +80,14 @@ export const VegaConnectDialog = ({
     return (
       <ConnectDialogContainer
         connectors={connectors}
-        closeDialog={() => setDialogOpen(false)}
+        closeDialog={() => updateDialogOpen(false)}
         appChainId={data.statistics.chainId}
       />
     );
   };
 
   return (
-    <Dialog open={dialogOpen} size="small" onChange={setDialogOpen}>
+    <Dialog open={dialogOpen} size="small" onChange={updateDialogOpen}>
       {renderContent()}
     </Dialog>
   );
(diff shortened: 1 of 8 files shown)


sam-keen (approved):

libs/wallet/src/connect-dialog/connect-dialog.tsx
```diff
@@ -28,15 +29,37 @@ type WalletType = 'gui' | 'cli' | 'hosted';
...
   dialogOpen: boolean;
-  setDialogOpen: (isOpen: boolean) => void;
+  updateDialogOpen: (open: boolean) => void;
 }
 
 export const VegaConnectDialog = ({
   connectors,
-  dialogOpen,
-  setDialogOpen,
+  onChangeOpen,
```
bglownia, line 49: Why we cannot name it like in radix ui Dialog? 🤔
`open` and `onOpenChange`

libs/wallet/src/connect-dialog/connect-dialog.tsx
```diff
@@ -28,15 +29,37 @@ type WalletType = 'gui' | 'cli' | 'hosted';
...
 }: VegaConnectDialogProps) => {
+  const { dialogOpen, updateDialogOpen } = useVegaWalletDialogStore(
```
bglownia, line 51: As mentioned above. Move useVegaWalletDialogStore to trading app.
We can have component in trading app which will wrap VegaConnectDialog and will use useVegaWalletDialogStore to pass `open` and `onOpenChange` props

libs/accounts/src/lib/accounts-manager.tsx
```diff
@@ -27,7 +27,7 @@ export const AccountManager = ({
   const update = useCallback(
     ({ data }: { data: AccountFields[] | null }) => {
       dataRef.current = data;
-      gridRef.current?.api.refreshInfiniteCache();
+      gridRef.current?.api?.refreshInfiniteCache();
```
bglownia, line 30: Do we really need this? Look like api is always present.
image

bglownia: If it is happening that AgGrid react instance is having undefined api we should fix it also in other places.

bglownia (changes requested):

libs/wallet/src/connect-dialog/connect-dialog.tsx
```diff
@@ -28,15 +29,37 @@ type WalletType = 'gui' | 'cli' | 'hosted';
 
 export interface VegaConnectDialogProps {
   connectors: Connectors;
+  onChangeOpen?: (open: boolean) => void;
+}
+
+export const useVegaWalletDialogStore = create<VegaWalletDialogStore>(
```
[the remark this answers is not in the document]
macqbat: I think you completely didn't understand the purpose of this change. It **was** in trading app, (as well as separately in console-lite and token). And I moved it into lib. Now it's much more clearer and easier to use, because state of dialog comes outside app and don't interference wits global state of the app. Now this what you propose is just removing back this whole improvement.

libs/wallet/src/connect-dialog/connect-dialog.tsx
```diff
@@ -28,15 +29,37 @@ type WalletType = 'gui' | 'cli' | 'hosted';
...
+  (set) => ({
+    dialogOpen: false,
+    updateDialogOpen: (open: boolean) => set({ dialogOpen: open }),
```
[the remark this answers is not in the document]
macqbat: sure, it's vastly easier, I'm gonna do it, however above example is not quite fortunate :-)

libs/wallet/src/connect-dialog/connect-dialog.tsx
macqbat: yeah, I realised the names of properties os the store are not quite good, but I thought rather about names which inform we are dealing with vegaWallet dialog. So what about names like `vegaWalletDialogOpen` and `updateVegaWalletDialog` ? What do you think?

libs/wallet/src/connect-dialog/connect-dialog.tsx
macqbat: So do you really want me to remove my change of vega wallet dialog state from here?

libs/accounts/src/lib/accounts-manager.tsx
macqbat: I just meet a run time crush there. It isn't connected with my changes, but I just wanted to fix it here. Looks like some race condition in ag-grid initialisation.  Sure, I will remove it then

bglownia (approved):

PR #1730, opened 2022-10-12
Fix/1622 proposal vote deadline fixes

### LINKED ISSUE #1622: Unable to submit proposal using min/max deadline
## Description

Error is shown when submitting a valid proposal clicking the min/max buttons for the voting deadline

## Steps to Reproduce

1. Go to submit a freeform proposal
2. Click on max deadlines for all relevant time periods
3. Click submit
4. Error shown
5. Click on min deadline
6. Same as abobe

Error is also seen on some of the other proposal types

sam-keen:
# Related issues 🔗

Closes #1622

# Description ℹ️

Some intermittent proposal subscription errors were being seen when the 'use minimum' and 'use maximum' buttons were clicked. Investigation into this uncovered that the proposal form fields weren't updating the form state when the buttons were used.

Secondarily, we had been adding 2 mins of extra time to the proposal vote deadline when the minimum was set, but had hardcoded that it happen when the vote deadline was set to 1 hour. This PR changes it to be inline with the nearest rounded hour to the vote deadline from the network parameter for the proposal form type. This means proposal forms with min deadlines longer than 1 hour now also get the extra 2 mins added (without which proposal could error if the user spent some time confirming in their wallet).


apps/token/src/routes/governance/components/propose/proposal-form-vote-and-enactment-deadline.tsx
```diff
@@ -1,29 +1,35 @@
...
 
 interface DeadlineDatesProps {
   vote: Date;
-  enactment: Date;
+  enactment: Date | undefined;
   validation: Date;
 }
 
 interface ValidationFormProps {
+  onValidationMinMax:
```
dexturr, line 30: chore (blocking): optional butI think this is always passed in? Or am I missing something

apps/token/src/routes/governance/components/propose/proposal-form-vote-and-enactment-deadline.tsx
```diff
@@ -108,6 +126,9 @@ const ValidationForm = ({
 };
 
 interface EnactmentFormProps {
+  onEnactMinMax:
```
dexturr, line 129: chore (blocking): optional butI think this is always passed in? Or am I missing something

dexturr (approved):

apps/token/src/routes/governance/components/propose/proposal-form-vote-and-enactment-deadline.tsx
sam-keen: No worries, yeah it needs to either be optional here (which actually makes less sense) or become part of the render logic for if the enactmentForm is rendered (I'll do that instead)

sam-keen: Scratch that - they don't always get passed in

PR #1741, opened 2022-10-13
Market page test refactor

radoslaw-szpiech:
# Description ℹ️

This task was about fixing two failing tests but ended up with refactoring market page test, some changes in components, graphql and some other minor changes. Tests are more detailed now.


apps/trading/components/market-volume/market-volume.tsx
```diff
@@ -57,8 +57,8 @@ export const MarketVolume = ({ marketId }: { marketId: string }) => {
   });
 
   return (
-    <HeaderStat heading={t('Volume')}>
-      <div data-testid="trading-volume">{marketVolume}</div>
+    <HeaderStat heading={t('Volume')} testId="market-volume">
```
MadalinaRaicu, line 60: what is the difference between `data-testid` and `testId`?

radoslaw-szpiech: I need to pass somehow data-testid into the HeaderStat component, but I cannot name it `data-testid` because it's variable here. So I just create variable, pass the value into component and then use it as html attribute

apps/trading-e2e/src/integration/market.cy.ts
```diff
@@ -0,0 +1,163 @@
...
+          for (let i = 0; i < 6; i++) {
+            cy.getByTestId(toolTipLabel)
+              .eq(i)
+              .should('have.text', auctionToolTipLabels[i]);
+            cy.getByTestId(toolTipValue).eq(i).should('not.be.empty');
+          }
+        });
+    });
+  });
+});
```
jtsang586, line 163: I would maybe name this file market-summary.cy.ts?  Just so it's bit more differentiated from markets.cy.ts

radoslaw-szpiech: I just followed the way of naming that we have in types and queries (e.g. Market, Markets, MarketData, MarketsData), but I think you're right. Changing.

apps/trading-e2e/src/integration/market.cy.ts
```diff
@@ -0,0 +1,163 @@
...
+        .should(
+          'contain.text',
+          'This market is in auction until it reaches sufficient liquidity.'
+        )
+        .within(() => {
+          cy.getByTestId('external-link')
+            .should('have.attr', 'href')
+            .and(
+              'include',
+              'https://docs.vega.xyz/docs/testnet/concepts/trading-on-vega/trading-modes#auction-type-liquidity-monitoring'
```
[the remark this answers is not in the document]
radoslaw-szpiech: Will do

jtsang586 (approved):

PR #1742, opened 2022-10-13
chore: format limits (547)

### LINKED ISSUE #547: Format deposit and withdraw limits
Currently if a deposit lifetime limit or a withdrawal threshold is greater than a hard coded value of 1 million we just display "1m+". We can improve this by displaying a more accurate value. However, this number can be very large (see last screen cap) and display poorly, either extremely long or in scientific notation depending on how you format. 

We need to format this value more accurately and in a more user friendly way. Perhaps with rounding and words

Image

Old design:
Screen Shot 2022-06-13 at 7 22 01 AM

asiaznik:
# Related issues 🔗

Closes #547

# Description ℹ️

* added new `compactNumber` function which formats big numbers into the human readable form, e.g 123,000,000,000 -> 123B or 123 billion
* changed deposit and withdrawal limits `<table>` into `<KeyValueTable>`

# Demo 📺

Screenshot 2022-10-13 at 17 37 36

### DIFF
--- a/libs/react-helpers/src/lib/format/number.tsx
+++ b/libs/react-helpers/src/lib/format/number.tsx
@@ -100,3 +100,67 @@ export const useNumberParts = (
 export const isNumeric = (
   value?: string | number | BigNumber | null
 ): value is NonNullable<number | string> => /^-?\d*\.?\d+$/.test(String(value));
+
+const INFINITY = '∞';
+const DEFAULT_COMPACT_ABOVE = 1_000_000;
+const DEFAULT_COMPACT_CAP = new BigNumber(1e24);
+/**
+ * Compacts given number to human readable format.
+ * @param number
+ * @param decimals Number of decimal places
+ * @param compactDisplay Display mode; short -> 1e6 == 1M; ling -> 1e6 1 million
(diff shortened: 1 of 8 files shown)


libs/react-helpers/src/lib/format/number.tsx
```diff
@@ -100,3 +100,67 @@ export const useNumberParts = (
...
+    notation: 'scientific',
+  });
+
+  if (number.isGreaterThan(DEFAULT_COMPACT_CAP)) {
+    const r = /E(\d+)$/i;
+    const formatted = scientificNumFormat.format(Number(number));
+    const eNotation = formatted.match(r);
+    if (eNotation && eNotation.length > 1) {
+      const power = eNotation[1];
+      return (
```
mattrussell36, line 153: chore (non blocking): Seeing as this can return jsx I think it would be clearer if this just became a CompactNumber component. Or alternatively if the compactNumber function did not return jsx, and the rendering of html was left up to the component that uses this function.

mattrussell36 (approved):

MadalinaRaicu (approved):

PR #1749, opened 2022-10-14
chore: single position table (1645)

### LINKED ISSUE #1645: Single table for positions
## Tasks

This task is to prepare the positions table to allow leveraging of AgGrid sorting and filtering as well as reduce surface area for bugs (such as values pushing against each other)

- [x] Render a single data grid that fills the positions panel, with all positions _not_ aggregated by asset
- [x] Add settlement asset column
- [x] Split size cell into notional and open volume columns
- [x] Split entry price and liquidation price into separate columns, progress bar can be displayed in liquidation price column
- [x] Margin allocated cell to use single actual value (not percentage)
- [x] Total value across positions gets removed

asiaznik:
# Related issues 🔗

Closes #1645

# Description ℹ️

* Render a single data grid that fills the positions panel, with all positions not aggregated by asset
* Add settlement asset column
* Split size cell into notional and open volume columns
* Split entry price and liquidation price into separate columns, progress bar can be displayed in liquidation price column
* Margin allocated cell to use single actual value (not percentage)
* Total value across positions gets removed

# Demo 📺

Screenshot 2022-10-14 at 15 55 06

### DIFF
--- a/libs/positions/src/lib/use-positions-data.tsx
+++ b/libs/positions/src/lib/use-positions-data.tsx
@@ -38,7 +38,8 @@ const getSummaryRow = (positions: Position[]) => {
 export const usePositionsData = (
   partyId: string,
   gridRef: RefObject<AgGridReact>,
-  assetSymbol?: string
+  assetSymbol?: string,
+  withSummaryRow?: boolean
 ) => {
   const variables = useMemo(() => ({ partyId }), [partyId]);
   const dataRef = useRef<Position[] | null>(null);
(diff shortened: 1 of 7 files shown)


libs/positions/src/lib/use-positions-data.tsx
```diff
@@ -38,7 +38,8 @@ const getSummaryRow = (positions: Position[]) => {
 export const usePositionsData = (
   partyId: string,
   gridRef: RefObject<AgGridReact>,
-  assetSymbol?: string
+  assetSymbol?: string,
+  withSummaryRow?: boolean
```
mattrussell36, line 42: question (blocking): Do we need withSummaryRow? It looks like it never gets passed in

mattrussell36 (approved):

PR #1760, opened 2022-10-17
chore: change tests to use stagnet3 by default

radoslaw-szpiech:
Change default test env to stagnet3, as tesntet became extremely unstable.


jtsang586 (approved):
Lgtm

PR #1762, opened 2022-10-18
Feat/522 close position

### LINKED ISSUE #522: Close a position
## Story
As a user of the console
I want to be able to close positions
So that I can control profits and losses

## Acceptance criteria

- [ ] I can close a position from the positions table
- [ ] All orders are simultaneously cancelled
- [ ] I am shown that orders will be cancelled

## Additional information

But I think it's important that we update the 'confirm in wallet' dialog to show what's being closed.
So for just a position:
image

For a position with orders: 
image

It also feels like if we are linking orders and positions strongly like this, we should show the number of open orders an a column in the position table:
image

mattrussell36:
# Related issues 🔗

Closes #522 

# Description ℹ️

- Adds new transaction type for batchMarketInstructions
- Updates close position hook to send a batchMarketInstruction
- Removes usePositionEvent hook and replaces with useTransactionResult hook in the wallet lib
- Refactors VegaTransactionDialog to be able to take custom UI for different transaction states so that position and order information can be shown in the dialog at the confirmation stage

# Demo 📺

Screen Shot 2022-10-18 at 15 57 53

Screen Shot 2022-10-21 at 15 04 06


libs/orders/src/lib/uitls.ts
```diff
@@ -0,0 +1,22 @@
+import { t } from '@vegaprotocol/react-helpers';
+import { OrderTimeInForce } from '@vegaprotocol/types';
+
+// More detail in https://docs.vega.xyz/docs/mainnet/graphql/enums/order-time-in-force
```
MadalinaRaicu, line 4: praise (non-blocking): 💯 
chore (blocking):  there's a typo in the name. please rename file to utils.ts

libs/positions/src/lib/close-position-dialog/complete.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const reason =
+    closingOrderResult.rejectionReason &&
+    OrderRejectionReasonMapping[closingOrderResult.rejectionReason];
+  return (
+    <div className="text-vega-red">
+      {reason ? (
+        <p>{reason}</p>
+      ) : (
+        <p>
+          {t('Something went wrong')}: {transactionResult.error}
```
MadalinaRaicu, line 103: (non-blocking): Why do we say `Something went wrong` instead of `Transaction failed because ...` or `Transaction failed. Error: ...`
It gives the impression that the environment is not stable like we don't know what went wrong. If we have an error we should know what went wrong.

mattrussell36: I agree. I'll update

MadalinaRaicu (approved):

PR #1771, opened 2022-10-18
feat: 1486 add details of expected fees margin close out to deal ticket

### LINKED ISSUE #1486: Add details of expected fees/margin usage/close out etc to ticket
## Story

As a user 
I want a preview of the impact of my trade before I place it
So that I can understand the likely outcome

## Acceptance Criteria

- [ ] I can see the estimated closeout price
- [ ] I can see the margin required
- [ ] I can see the notional size
- [ ] I can see fees
- [ ] I can see estimated slippage

### Sketch
image

image

I think these would be a big benefit in Console - I've found myself wanting when using it myself. Also helps explain how fees work in Vega, that liquidation price are variable, that these are leveraged products etc, so helps from a user-education point too.

There's already a version on this in Console Lite, so I'm hoping we can use the code as a starting point, or perhaps reuse the component entirely:  
image

MadalinaRaicu:
# Related issues 🔗

Closes #1486 

# Description ℹ️

Add details of expected fees, margin usage, close out etc. to deal ticket

# Demo 📺

image

# Technical 👨‍🔧

- [x] move all hooks from console-lite to deal-ticket lib
- [x] create a hook to retrieve all


apps/console-lite/.env
```diff
@@ -19,7 +19,7 @@ NX_DEPLOY_URL=$DEPLOY_URL
 NX_DEPLOY_PRIME_URL=$DEPLOY_PRIME_URL
 NX_VEGA_CONFIG_URL="https://static.vega.xyz/assets/testnet-network.json"
 NX_VEGA_ENV = 'TESTNET'
-NX_VEGA_URL="https://api.n11.testnet.vega.xyz/graphql"
+NX_VEGA_URL="https://api.n07.testnet.vega.xyz/graphql"
```
MadalinaRaicu, line 22:
```suggestion
NX_VEGA_URL="https://api.n11.testnet.vega.xyz/graphql"
```

mattrussell36 (approved):

libs/deal-ticket/src/components/deal-ticket/deal-ticket.tsx
```diff
@@ -109,8 +117,17 @@ export const DealTicket = ({
             onSelect={(type) => {
               if (type === OrderType.TYPE_LIMIT) {
                 setValue('timeInForce', OrderTimeInForce.TIME_IN_FORCE_GTC);
```
macqbat, line 119: nitpick: these setValue('timeInForce') are not needed anymore since [this](https://github.com/vegaprotocol/frontend-monorepo/pull/1779) has been merged

libs/deal-ticket/src/components/deal-ticket/deal-ticket.tsx
```diff
@@ -54,13 +53,11 @@ export const DealTicket = ({
     mode: 'onChange',
     defaultValues: getDefaultOrder(market),
   });
-
-  const orderType = watch('type');
-  const orderTimeInForce = watch('timeInForce');
+  const order = watch();
```
macqbat, line 56: question, non-blocking: watching whole object instead certain properties wouldn't provide more re-renders?

libs/deal-ticket/src/hooks/use-calculate-slippage.ts
```diff
@@ -22,7 +22,7 @@ const useCalculateSlippage = ({ marketId, order }: Props) => {
     variables,
     throttleMilliseconds: 5000,
   });
-  const { data: market } = useDataProvider<SingleMarketFieldsFragment, never>({
+  const { data: market } = useDataProvider<Market, never>({
```
macqbat, line 25: Don't get it why this change is for? `Market` type is returned by market**s**Provider.
marketProvider returns `SingleMarketFieldsFragment` type. Now these types maybe are similar, but in the future they can get some discrepancies

libs/deal-ticket/src/hooks/use-market-data.ts
```diff
@@ -0,0 +1,21 @@
...
+    pollInterval: 5000,
+    variables: { marketId },
+    skip: !marketId,
+  });
+  return useMemo(() => {
+    if (
+      data &&
+      data.market?.data?.markPrice !== memoRef.current?.market?.data?.markPrice
+    ) {
+      memoRef.current = data;
```
macqbat, line 17: non-blocking, question: wouldn't it be easier to hold in ref only `markPrice` for comparing?

libs/market-list/src/lib/utils/market-utils.ts
```diff
@@ -5,17 +5,20 @@ import orderBy from 'lodash/orderBy';
 import type { Market, Candle } from '../';
 
 export const totalFees = (fees: Market['fees']['factors']) => {
-  if (!fees) {
-    return undefined;
-  }
-  return formatNumberPercentage(
+  return (
+    fees &&
```
macqbat, line 9: non-blocking, question: this change return value form `undefined` to `false` if fees will be falsy. Is this intentional?

macqbat (approved):

libs/deal-ticket/src/components/deal-ticket/deal-ticket.tsx
MadalinaRaicu: it is intentional, I do need all fields inside the order.

PeteWilliams:
@MadalinaRaicu A few suggestions based on what's been built so far:

- Lose 'Est price' - I don't think this is needed as it's shown in the ticket.
- Change 'Notional size' to 'Notional value' which I think it's the standard label and is more self-explanatory
- Rather than prefixing with 'Est' on the label, use the '~' prefix on the value (like we do with price in the ticket). This keeps the focus on the name part of the label.
- Change 'close out to 'Liquidation price (variable)' as we call it liquidation price elsewhere, and I think it's important to point out it isn't an estimate of what it will be, it's a variable number which will change.
- Can we show the values in text-neutral-500 dark:text-neutral-300 (same as the value tables in the market info panel)
- The liquidation price doesn't change when you change size and shows a value even when your wallet isn't connected - I'm not sure this is right as I'd expect the value to be a function of the available margin?

libs/market-list/src/lib/utils/market-utils.ts
MadalinaRaicu: absolutely not, the intention was to return BigNumber or undefined but I am updating the conditions.

libs/deal-ticket/src/hooks/use-market-data.ts
MadalinaRaicu: you're right. again, I haven't changed this hook, just moved it across. But, I think who wrote it, had the intention of returning more than mark price out of the market data.

libs/deal-ticket/src/hooks/use-market-data.ts
```diff
@@ -0,0 +1,21 @@
+import { useMemo, useRef } from 'react';
+import type { MarketMarkPriceQuery } from './__generated__/MarketMarkPrice';
+import { useMarketMarkPriceQuery } from './__generated__/MarketMarkPrice';
+
+export const useMarketData = (marketId: string) => {
```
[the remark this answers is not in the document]
MadalinaRaicu: this is how it was named before. I just moved this across. but makes sense to rename it.

libs/deal-ticket/src/hooks/use-calculate-slippage.ts
MadalinaRaicu: makes sense. I'm reverting to `SingleMarketFieldsFragment`

libs/deal-ticket/src/components/deal-ticket/deal-ticket.tsx
MadalinaRaicu: when I merged with master this whole bit got deleted as in the mentioned PR

libs/market-list/src/lib/utils/market-utils.ts
MadalinaRaicu: updated in last commit.

MadalinaRaicu:
> @MadalinaRaicu A few suggestions based on what's been built so far:
> 
> * Lose 'Est price' - I don't think this is needed as it's shown in the ticket.
> * Change 'Notional size' to 'Notional value' which I think it's the standard label and is more self-explanatory
> * Rather than prefixing with 'Est' on the label, use the '~' prefix on the value (like we do with price in the ticket). This keeps the focus on the name part of the label.
> * Change 'close out' to 'Liquidation price (variable)' as we call it liquidation price elsewhere, and I think it's important to point out it isn't an estimate of what it will be, it's a variable number which will change.
> * Can we show the values in text-neutral-500 dark:text-neutral-300 (same as the value tables in the market info panel)
> * The liquidation price doesn't change when you change size and shows a value even when your wallet isn't connected - I'm not sure this is right as I'd expect the value to be a function of the available margin?

everything updated as you suggested. except last point. 
if party balance is coming as undefined we just consider it 0. this is the formula `(marginMaintenanceLevel - positionAccountBalance - generalAccountBalance) / volume + markPrice`. this is why a value is shown.

libs/deal-ticket/src/hooks/use-market-data.ts
macqbat: yeah, I remember we had in plans some revision of all these hooks, now seems to be a good opportunity, that's why I mentioned it here :wink:

PR #1779, opened 2022-10-19
chore: fix for persist force value selection after change of order type

### LINKED ISSUE #1656: Time in force value resets if toggling order type
## Description

When selecting a time in force value, your selection is lost if you order type and back again.

## Steps to Reproduce

1. Go to the deal ticket
2. Change time in force to Fill or Kill
3. Switch to limit order
4. Switch back to market order

## Expected behavior

Time in force should still be on Fill or Kill

## Actual behavior

Time in force reverts to Immediate or Cancel

## Additional context
Note. Only FOK and IOC are valid market order types. Limit order types include those, along with GTC, GTT, GFA, GFN

macqbat:
# Related issues 🔗

Closes #1656 

# Description ℹ️

Add persistence of force value selection after change of order type in deal ticket


libs/deal-ticket/src/components/deal-ticket/time-in-force-selector.tsx
```diff
@@ -28,11 +29,23 @@ export const timeInForceLabel = (tif: string) => {
...
+type PrevSelectedValue = {
+  [key in PossibleOrderKeys]: Schema.OrderTimeInForce;
+};
+
 export const TimeInForceSelector = ({
   value,
   orderType,
   onSelect,
 }: TimeInForceSelectorProps) => {
+  const [prevValue, setPrevValue] = useState<PrevSelectedValue>({
```
mattrussell36, line 45: suggestion (non blocking): I'm fine with this as is, but I think a better solution would be lift this state into the parent, and add keep the TIF logic in the on click handler of the type selector

mattrussell36 (approved):

libs/deal-ticket/src/components/deal-ticket/time-in-force-selector.tsx
macqbat: We would need to add some setter for ref value or pass ref down to the tif selector. Not quite sure it could be for us a big adventage...

PR #1782, opened 2022-10-19
feat(1675): server side sort and filters of orders

### LINKED ISSUE #1675: Spike: AG Grid sorting / filtering POC
We'd like to achieve this https://github.com/vegaprotocol/frontend-monorepo/issues/1643

bglownia:
# Related issues 🔗

Closes #1675

# Description ℹ️

This is POC of sort and filter of paginated data 

# Technical 👨‍🔧

Order and sort of paginated data needs to be done on server side. To achieve that sort and filter parameters needs to be passed to data provider as a part of variables.
Update function both from data provider and from use data provider hook needs to support sorting and filtering by all supported columns. After received update via Apollo subscription row needs to be checked if it matches filter and if it's index should change accordingly to selected to sort parameter.


bglownia:
This was POC only, and PR created to show possible solution. No need to merge at the moment.

PR #1848, opened 2022-10-25
Feat/1512: equity like share vote

### LINKED ISSUE #1512: Equity like share vote
## Story

As a liquidity provider
I want to be able to take part in and see the results of equity like share votes for change market proposals (to progress the proposal even if the required majority has not been met)
So that I can take part in and see the results of equity like share votes rather than rely on using the API directly

## Acceptance Criteria

- [ ] I can see on market change proposals (proposal form and proposal details page) my equity like share and how that relates to the vote (Equity like share is a network param - governance.proposal.updateMarket.minProposerEquityLikeShare )
- [ ] I can see LP votes on a proposal details page (market change only) - we'll need to work out if this is a second vote, that can only occur if the first fails or if the two run concurrently
- [ ] As an LP I can vote on an equity like share vote 
- [ ] I can understand the difference between the regular vote and equity like share vote i.e. we've explained it
- [ ] I can see which of the vote types is the one that will be applied on the proposal details page (Token or LP is set to be the decisive one).

## Tasks

- [x] Explore and sketch
- [x] Team and stakeholder review - https://vegaprotocol.slack.com/archives/C038EB6V2RE/p1664793351695239
- [x] Visual Design (if needed)
- [x] Team review
- [x] Develop

## Sketch

Image

To be clear: there is only one Vote method. If you have EQ like share it'll count that. There are not different vote methods for Tokens and QL like share.
Here is what I think the minimum change could be.

1. add another tally to show the EQ like vote status (this one doesn't need a count, only %)
2. re-label the existing vote tally as "token vote" not just "votes"
3. add some logic to show what way the vote will pass (if it will)
4. add the detail for the liquidity vote
5. more re-labelling to be clear this is token not liquidity vote detail
6. remove the details of parties and tokens (this is displayed more efficiently in the visualisation
- The re-labelling of "Votes" to "Token Votes" could be done across the board
- remove the countdown and "set to pass/fail" from each tally. There is a time stamp at the top. The set to pass fail is covered by 3.
- add senctence under the two vote tallies 

The way this falls short in that it doesn't tell you what your EQ like share is before or after you vote (it does show your tokens but only in the side bar).

From Michael 

- on market change proposals: show your equity like share (liquidity provider vote, AKA LP) - This is not included in the design above and could be another ticket (done as part of vote re-factor?)
- Find out from @EVODelavega if you submit two votes, one for VEGA and a different for LP? - there is one vote. not
- Figma file  https://www.figma.com/file/F995ESZ0QMpTBuQWGAgLRz/Design-System?node-id=1718%3A70064 (either token or LP) (but this is WPI for a design that require more drastic layout changes. Scope of this ticket is the simplest solution.

sam-keen:
# Related issues 🔗

Closes #1512

# Description ℹ️

Adds LP voting data to update asset proposals. Also improves unit test coverage around voting


apps/token/src/routes/governance/components/proposal-votes-table/proposal-votes-table.tsx
```diff
@@ -44,20 +59,44 @@ export const ProposalVotesTable = ({ proposal }: ProposalVotesTableProps) => {
     >
       <KeyValueTableRow>
         {t('willPass')}
-        {willPass ? '👍' : '👎'}
+        {isUpdateMarket
```
mattrussell36, line 62: suggestion (non blocking): possibly add a Thumbs component? Should make this ui a touch more maintanable as well as make this nested ternary a bit more readable..

This nested ternary could then be replaced with something more readable


something like:
```
isUpdateMarket ? (
  <Thumbs up={updateMarketWillPass} upText={updateMarketVotePassMethod} />
) : (
  <Thumbs up={willPassByTokenVote}
)
```

apps/token/src/routes/governance/hooks/use-vote-information.ts
```diff
@@ -1,162 +1,185 @@
...
+        : new BigNumber(100),
+    [requiredMajorityLP]
+  );
+
+  const noTokens = useMemo(() => {
     return new BigNumber(proposal.votes.no.totalTokens);
   }, [proposal.votes.no.totalTokens]);
 
-  const yesTokens = React.useMemo(() => {
+  const noELSWeight = useMemo(() => {
```
mattrussell36, line 45: suggestion (non blocking): Took me a bit too long to remember ELS is equity like share, can we rename to `noEquityLikeShareWeight`?

apps/token/src/routes/governance/hooks/use-vote-information.ts
```diff
@@ -1,162 +1,185 @@
...
+  const willPassByTokenVote = useMemo(
     () =>
       participationMet &&
       new BigNumber(yesPercentage).isGreaterThanOrEqualTo(
         requiredMajorityPercentage
       ),
     [participationMet, requiredMajorityPercentage, yesPercentage]
   );
+
+  const willPassByLPVote = useMemo(
```
mattrussell36, line 145: suggestion (non blocking): I _really_ dislike this huge chain of useMemos. If possible lets combine this all into a few useMemos (or just one ?!) that read more linearly

mattrussell36 (approved):

apps/token/src/routes/governance/components/proposal-votes-table/proposal-votes-table.tsx
sam-keen: Nice idea!

PR #1852, opened 2022-10-26
feat: add governance link to navbar

### LINKED ISSUE #576: Link to governance
As a console user
I want a way to easily see how Vega is governed and understand how I can contribute
So that I can confirm Vega is governed in a decentralised way and know how to influence it

mattrussell36:
# Related issues 🔗

Closes #576 

# Description ℹ️

Adds a link to the governance page on token

# Demo

Screen Shot 2022-10-25 at 20 34 28

# Technical 👨‍🔧

- Was able to remove the state from NavLink as the useEffect in navbar.tsx handles the client/server rendering mismatch
- Changed yellow border to be rendered with a span so that the text of the links is more easily vertically centered, with governance off to the right it look out of line compared to the pubkey dropdown

### DIFF
--- a/apps/trading/components/navbar/navbar.tsx
+++ b/apps/trading/components/navbar/navbar.tsx
@@ -1,12 +1,13 @@
 import classNames from 'classnames';
 import { useRouter } from 'next/router';
 import Link from 'next/link';
-import { NetworkSwitcher } from '@vegaprotocol/environment';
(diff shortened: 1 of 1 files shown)


macqbat (approved):
by the way: Navbar destroys mobile view. I'm gonna create an issue for it

PR #1860, opened 2022-10-26
Feat/1660 validate available margin

### LINKED ISSUE #1660: Validate that user has sufficient collateral before placing an order
## Story

As a trader
I want to be warned that I don't have enough collateral for an order before submitting a transaction
So that I don't waste time creating orders that won't be placed

## Acceptance Criteria

- [ ] I cannot place an order without sufficient collateral
- [ ] I cannot place an order without sufficient collateral accounting for margin requirements
- [ ] I am shown an insufficient collateral message if the order will not be placed

macqbat:
# Related issues 🔗

Closes #1660 

# Description ℹ️

Validate user asset balance to be enough to satisfy estimated margin

# Demo 📺

image

# Technical 👨‍🔧
Compare estimate margin and user's asset deposit


libs/deal-ticket/src/components/deal-ticket-validation/use-order-margin-validation.tsx
```diff
@@ -0,0 +1,44 @@
...
+    variables: { partyId: pubKey || '' },
+    skip: !pubKey,
+  });
+
+  const settlementAccount = useSettlementAccount(
+    market.tradableInstrument.instrument.product.settlementAsset.id,
+    partyBalance?.party?.accounts || [],
+    AccountType.ACCOUNT_TYPE_GENERAL
+  );
+  if (settlementAccount) {
```
bglownia, line 26: Should we enter this block when we 
settlementAccount.balance is falsy or
settlementAccount.asset.decimals is falsy or
estMargin?.margin is undefined

bglownia (approved):

libs/deal-ticket/src/components/deal-ticket-validation/use-order-margin-validation.tsx
macqbat: Do you know that you're even right ? At least in case when no settlementAccount, we can assume user should be treated as no having enough asset. I've made smth like that.

PR #1861, opened 2022-10-26
feat(958): add cancel all orders button

### LINKED ISSUE #958: Cancel all orders button
The API provides a mechansim to cancel all orders in one transaction. You just submit `orderCancellation` without a `marketId` or `orderID`: https://docs.fairground.vega.xyz/docs/api-howtos/cancel-order/

So, we should add a 'Cancel all' button to the bottom of the orders list:
image

bglownia:
# Related issues 🔗

Closes #958

# Description ℹ️

Add cancel all orders button


apps/trading-e2e/src/integration/trading-orders.cy.ts
```diff
@@ -29,50 +30,49 @@ describe('orders list', { tags: '@smoke' }, () => {
     cy.wait('@Orders').then(() => {
       expect(subscriptionMocks.OrdersUpdate).to.be.calledOnce;
     });
+    cy.wait('@Markets');
   });
-  it('renders orders', () => {
+  it.only('renders orders', () => {
```
mattrussell36, line 35: chore (blocking): Theres only one test in this describe block but probably worth removing the `.only`

libs/orders/src/lib/components/order-list/order-list.tsx
```diff
@@ -259,18 +323,35 @@ export const OrderListTable = forwardRef<AgGridReact, OrderListTableProps>(
...
             return value ? getDateTimeFormat().format(new Date(value)) : '-';
           }}
         />
         <AgGridColumn
           colId="amend"
           headerName=""
           field="status"
-          cellRenderer={({ data }: VegaICellRendererParams<Order>) => {
+          cellRenderer={({ data, node }: VegaICellRendererParams<Order>) => {
+            if (node?.rowPinned) {
```
mattrussell36, line 339: chore (non blocking): the bottom pinned row has a background hover color, can we remove it?

Also having tested it out, I don't think its clear enough that this will close_all_ orders and not just orders for the given market. I think we should make the orders on the market page be only for that market. The orders tab on the portfolio page can continue to show all orders across all markets.

bglownia: We are showing all orders on market page not filtered by market Id
We can filter them out and show only orders for the market when filters will be available on API.

PR #1865, opened 2022-10-26
feat(1655): persist deal ticket

### LINKED ISSUE #1655: Persist deal ticket state for each market
## Story

As a trader
I want the deal ticket for a market to keep its state even if I navigate to other markets and pages
So that I don't have to re-input values and settings for the deal ticket every time

## Acceptance Criteria

- [x] Order type is persisted
- [x] Order direction is persisted
- [x] Order size is persisted
- [x] Order price is persisted, if no price has been stored, market mark price should be the default
- [x] Time in force is persisted
- [x] Deal ticket state is persisted _per_ market

## Tasks

- [x] Figure out most suitable method for data persistence. Local storage vs indexedDB

### LINKED ISSUE #1694: It is possible to create GTT limit order with expiry date in past
## Description

User can submit order with Expiry date and time set to past. Date picker does not allow to do it, but one can modify input field with their own date.

## Steps to Reproduce

1. Go to Console app
2. Choose limit order.
3. Choose Good 'til Time (GTT)
4. Modify the date input field to set it to past.
5. Submit order.

## Expected behavior

It should not be possible.

## Screenshots

image

asiaznik:
# Related issues 🔗

Closes #1655
Closes #1694 

# Description ℹ️

All deal ticket properties are now persisted in the local storage. It makes use of the new `usePersistedOrder` ➡️  `useLocalStorage` hook.

# Demo 📺

https://user-images.githubusercontent.com/1980305/198352968-c6b1146c-2a4e-4bda-b68d-5b05e48154c9.mov

Screenshot 2022-10-27 at 18 58 00


apps/trading-e2e/src/integration/trading-deal-ticket.cy.ts
```diff
@@ -327,7 +330,7 @@ describe('limit order validations', { tags: '@smoke' }, () => {
       });
     });
 
-    it('selections should be remembered', () => {
+    it.skip('selections should be remembered', () => {
```
mattrussell36, line 333: question (non blocking): This intentional?

libs/deal-ticket/src/components/deal-ticket/expiry-selector.tsx
```diff
@@ -22,6 +22,11 @@ export const ExpirySelector = ({ value, onSelect }: ExpirySelectorProps) => {
         onChange={(e) => onSelect(e.target.value)}
         min={minDate}
       />
+      {new Date() > date && (
```
mattrussell36, line 25: suggestion (non blocking): I'd probably pass a validation rule to the Controller component so that. See the rules prop here https://react-hook-form.com/api/usecontroller/controller/#main

mattrussell36 (approved):

PR #1866, opened 2022-10-26
fix: 1842 set expiry to date time format

### LINKED ISSUE #1842: Show expiry time in trade page header
We currently only show the date but not the time. We should show the time as well.

Current:
Screenshot 2022-10-24 at 15 58 20

## Tasks

- [ ] Adjust date format to include time

MadalinaRaicu:
# Related issues 🔗

Closes #1842 

# Description ℹ️

set expiry to date time format

# Demo 📺 
Screenshot 2022-10-26 at 16 12 30

### DIFF
--- a/libs/market-info/src/components/market-expires/market-expires.tsx
+++ b/libs/market-info/src/components/market-expires/market-expires.tsx
@@ -1,8 +1,7 @@
 import { getDateTimeFormat, t } from '@vegaprotocol/react-helpers';
-import { format, isValid, parseISO } from 'date-fns';
+import { isValid, parseISO } from 'date-fns';
 
 import type { SingleMarketFieldsFragment } from '@vegaprotocol/market-list';
-export const EXPIRE_DATE_FORMAT = 'MMM dd';
 
 export const getMarketExpiryDate = (
   tags?: ReadonlyArray<string> | null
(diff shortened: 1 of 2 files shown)


libs/market-info/src/components/market-expires/market-expires.tsx
```diff
@@ -1,7 +1,7 @@
-import type { SingleMarketFieldsFragment } from '@vegaprotocol/market-list';
-import { getDateFormat, t } from '@vegaprotocol/react-helpers';
+import { getDateTimeFormat, t } from '@vegaprotocol/react-helpers';
 import { format, isValid, parseISO } from 'date-fns';
 
+import type { SingleMarketFieldsFragment } from '@vegaprotocol/market-list';
 export const EXPIRE_DATE_FORMAT = 'MMM dd';
```
mattrussell36, line 5: question (blocking): do we not need to update this format string? Looks like there is a mix of using date-fns for formatting as well the Intl object. Can we tidy this up to use the later?

MadalinaRaicu: I kept this format because it was used in console-lite. I don't know the reqs/designs on that one and I didn't want this PR to affect anything else except the console.

dexturr (approved):

PR #1871, opened 2022-10-26
fix(1758): data node being down

### LINKED ISSUE #1758: The default data node URL being down should not crash the site
On Sunday the mainnet observer data node went down. This is the one hosted and run by Vega as is the one that is the default to connect to on the token site. This resulted in the site being unable to load for minutes at a time.

This should not be the case and the node switcher dialog should show up as quickly as possible to allow the user to select or enter a different data node as soon as it is apparent the default is not working.

dexturr:
# Related issues 🔗

Closes #1758

# Description ℹ️

The data node being down caused the site to be inoperable. This was due to the request hanging and there being no timeout on the frontends, this is true for all requests. This means that any request hanging could potentially cause the whole site to become inoperable.

# Technical

I needed to add the timeout link for data nodes, as requests hanging would bring any site to a stop. This made me realise that we were duplicating a LOT of code in the apollo client files. I have inverted the cache policy dependency so the consumers can specify it themselves.


libs/apollo-client/src/lib/apollo-client.ts
```diff
@@ -11,18 +12,19 @@ import { getMainDefinition } from '@apollo/client/utilities';
...
+export function createClient(base?: string, cacheConfig?: InMemoryCacheConfig) {
   if (!base) {
     throw new Error('Base must be passed into createClient!');
   }
   const urlHTTP = new URL(base);
   const urlWS = new URL(base);
   // Replace http with ws, preserving if its a secure connection eg. https => wss
   urlWS.protocol = urlWS.protocol.replace('http', 'ws');
-
+  const timeoutLink = new ApolloLinkTimeout(5000);
```
dexturr, line 27: question: What should this be? I am not married to this, but felt "good enough"

dexturr, line 27:
```suggestion
  const timeoutLink = new ApolloLinkTimeout(10000);
```

apps/token/src/app.tsx
```diff
@@ -26,9 +26,148 @@ import {
...
+    Query: {
+      fields: {
+        nodes: {
+          // Merge function to make the validator list random but remain consistent
+          // as the user navigates around the site. If the user refreshes the list
+          // will be randomised.
+          merge: (existing = [], incoming) => {
+            // uniqBy will take the first of any matches
+            const uniq = uniqBy([...incoming, ...existing], 'id');
+            // sort result so that the input is consistent
```
sam-keen, line 71: Question (non-blocking): why is it important the the input is consistent?

sam-keen (approved):

apps/token/src/app.tsx
```diff
@@ -26,9 +26,148 @@ import {
...
+  InMemoryCacheConfig,
+  Reference,
+} from '@apollo/client';
+import sortBy from 'lodash/sortBy';
+import uniqBy from 'lodash/uniqBy';
+
+import { deterministicShuffle } from './lib/deterministic-shuffle';
+import { addDecimal } from '@vegaprotocol/react-helpers';
+
+const formatUintToNumber = (amount: string, decimals = 18) =>
```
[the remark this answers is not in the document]
dexturr: Yeah it is, I want to get rid of this anyway so I will create a ticket for that :+1:

apps/token/src/app.tsx
dexturr: So this was a random requirement, the valdiators wanted the list to be randomised before the scoring mechanism was introduced. But we wanted to persist that across a user session so the table didn't jump around on update. This is the fix for that, when we do the validator table improvements my hope is we can kill this 🙏

dexturr: Or possibly have we already sorted the table and this can go now?

apps/token/src/app.tsx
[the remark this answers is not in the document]
dexturr: https://github.com/vegaprotocol/frontend-monorepo/issues/2015

PR #1900, opened 2022-10-28
Feat/1657 show deal ticket errors contextually

### LINKED ISSUE #1657: Show deal ticket errors contextually
## Story

As a trader
I want order errors to show up close to where that error exists
So that I can more quickly get the deal ticket into a more valid state

## Acceptance Criteria

- [x] Errors regarding type appear under the type toggle
- [x] Errors regarding size show under the size field
- [x] Errors regarding price appear under the price field
- [x] Errors regarding time in force appear under the time in force dropdown
- [x] Errors regarding expiry time appear under the expiry date picker
- [x] Client side validation errors don't show until the place order button is clicked
- [x] If a client side error is present the transaction is not attempted and an error is shown

macqbat:
# Related issues 🔗

Closes #1657

# Description ℹ️

Show deal tickets errors separately under each section
# Demo 📺

image


libs/deal-ticket/src/components/deal-ticket-validation/use-order-validation.tsx
```diff
@@ -308,5 +338,5 @@ export const useOrderValidation = ({
     isInvalidOrderMargin,
   ]);
 
-  return { message, isDisabled };
+  return { message, isDisabled, section: section as DealTicketSection };
```
mattrussell36, line 341: suggestion (non blocking): The type cast here hides the fact that useMemo can return `string`. I think a better solution would be to have a specific type for 'none' state. IE:. `constants.DEAL_TICKET_SECTION_TYPE_NONE`

libs/deal-ticket/src/components/deal-ticket-validation/use-order-validation.tsx
```diff
@@ -45,6 +46,15 @@ export const marketTranslations = (marketState: MarketState) => {
   }
 };
 
+export type DealTicketSection =
```
mattrussell36, line 49: suggestion (non blocking): I think its a touch cleaner to export a constant object then you could dynamically create this type based on the object values

libs/deal-ticket/src/components/deal-ticket/deal-ticket-error.tsx
```diff
@@ -0,0 +1,52 @@
...
+  'data-testid'?: string;
+}
+
+export const DealTicketError = ({
+  errorMessage,
+  'data-testid': dataTestId = 'deal-ticket-error-message',
+}: Props) => {
+  const [message, setMessage] = useState<DealTicketErrorMessage['message']>('');
+  const ts = useRef<null | number>(null);
+  useEffect(() => {
```
mattrussell36, line 22: chore (blocking): I think it would be better to leave this extra complexity out for now. Once the design team has given us some animation guidelines we can come back and animate consistently with a suitable library

mattrussell36 (changes requested):

libs/deal-ticket/src/components/deal-ticket/deal-ticket-error.tsx
macqbat: ok, as you wish, I'm gonna remove animation completely for now. I just wanted to avoid really ugly effect of "jumping" sections when some error messages appear and hide. These few additional lines (of useState and useEffects) seemed to me as not too complicated and worked nice as well  :-( . But ok, I'm gonna remove it ofc, especially if we have in plans to do  a consistent solution.

libs/deal-ticket/src/components/deal-ticket-validation/use-order-validation.tsx
macqbat: casting could be removed only in case when useMemo got explicitly returned types

libs/deal-ticket/src/components/deal-ticket-validation/use-order-validation.tsx
macqbat: not quite get it how such an object could look like, and at least what would be an advantage of creating an object instead a list of constants. So if it's non blocking I leave it as it is

bglownia (approved):

PR #1917, opened 2022-10-31
change label

baldator:
# Related issues 🔗

I didn't create an issue for this :) but the demo image is self-explanatory 

# Description ℹ️

Fix the label in validators table

# Demo 📺

image

# Technical 👨‍🔧


dexturr:
Thanks for contributing

PR #1921, opened 2022-11-01
chore(#1873): trading hash router

### LINKED ISSUE #1873: [trading] switch next router for react router
# Problem
Deeply nested urls (e.g. `/markets/12345`) rely on server-side routing. Netlify supports next.js, so this works fine. Fleek does not*. IPFS doesn't have something that can handle routing this to a .html file**

# Solution
Sadly, it's hash based routing. It used to be more common, back before [History.pushState](https://developer.mozilla.org/en-US/docs/Web/API/History/pushState) was widely supported. It's... pretend routing, effectively. But given we can't rely on a server to do the redirect for us, and we don't know (for example) the market IDs in advance, it's the best tool for the job.

# Tasks
- [x] Add react router dep
- [x] Switch all navigation to [HashRouter](https://v5.reactrouter.com/web/api/HashRouter)
   - [x] `console.fairground.wtf/#/markets/1234`
   - [x] `console.fairground.wtf/#/portfolio`
   - [x] `console.fairground.wtf/`
- [x] ESLint rules to prevent usage of `next/link` or `next/router` accidentally
- [x] Update Cypress tests to navigate directly to the hash links
- [ ] Remove nextjs plugin from netlify
- [ ] Update build command to use `next export`

```
* It is on the roadmap! Date: unknown
** Actually it *does* as of September 23rd (https://github.com/ipfs/kubo/pull/8890), but the appropriate version is not widely deployed, so we are opting not to rely on it. We can roll back this change when it is.
```

mattrussell36:
# Related issues 🔗

Closes #1873 

# Description ℹ️

- Puts react router in the the nextjs trading app 😱 
- Adds eslint rule to prevent importing and using next/link, next/router


libs/deal-ticket/src/components/trading-mode-tooltip/compile-grid-data.tsx
```diff
@@ -60,10 +60,11 @@ export const compileGridData = (
   if (isLiquidityMonitoringAuction && market.data?.suppliedStake) {
     grid.push({
       label: (
-        <Link href={`/liquidity/${market.id}`} passHref={true}>
-          <UiToolkitLink onClick={() => onSelect && onSelect(market.id)}>
-            {t('Current liquidity')}
-          </UiToolkitLink>
+        <Link
+          to={`/liquidity/${market.id}`}
```
MadalinaRaicu, line 64: shouldn't this one be '#/' too?

libs/market-info/src/components/market-info/info-market.tsx
```diff
@@ -319,10 +318,11 @@ export const Info = ({ market, onSelect }: InfoProps) => {
...
         >
-          <Link passHref={true} href={`/liquidity/${market.id}`}>
-            <UiToolkitLink onClick={() => onSelect(market.id)}>
-              {t('View liquidity provision table')}
-            </UiToolkitLink>
+          <Link
+            to={`/liquidity/${market.id}`}
+            onClick={() => onSelect(market.id)}
+          >
+            <UILink>{t('View liquidity provision table')}</UILink>
```
MadalinaRaicu, line 325: same here?

libs/ui-toolkit/src/components/link/link.tsx
```diff
@@ -16,15 +16,25 @@ export const Link = ({ className, children, ...props }: LinkProps) => {
...
+      <span {...shared} {...props}>
+        {children}
+      </span>
+    );
+  }
 
   return (
-    <a
-      role="link"
-      data-testid="link"
```
MadalinaRaicu, line 23: finally. each link should have a different data-testid

libs/deal-ticket/src/components/trading-mode-tooltip/compile-grid-data.tsx
mattrussell36: No react router handles the hash character as we are using the HashRouter provider

libs/market-info/src/components/market-info/info-market.tsx
mattrussell36: No react router handles the hash character as we are using the HashRouter provider

libs/ui-toolkit/src/components/link/link.tsx
mattrussell36: This is how it was before, so I didn't want to make any changes to data-testid in this PR. Also you can already change the value as  `props` are spread into the compoment after `shared` which will merge them.

apps/trading/pages/_app.page.tsx
```diff
@@ -31,31 +32,28 @@ const Title = () => {
     if (networkName) return `${pageTitle} [${networkName}]`;
     return pageTitle;
   }, [pageTitle, networkName]);
-  return (
-    <Head>
-      <title>{title}</title>
-    </Head>
-  );
+  return <title>{title}</title>;
```
macqbat, line 35: Title managed like that won't work. It has to be **[direct](https://nextjs.org/docs/api-reference/next/head)** children of <Head /> element

macqbat (changes requested):
lgtm, only title in _app.page file I do not like :sob:. I moved title to separate component to avoid rerender of whole page when title is changing according to market price change, what means it happened very often

macqbat (approved):

PR #1943, opened 2022-11-02
Feat/1513: Vote buttons using subscription and dialog for tx

### LINKED ISSUE #1513: For VEGA transactions replace embedded status alerts(?) with dialogue / modal / console pattern
## The Chore

Status notifications currently appear embedded in the token UI rather than a dialogue that interrupts the whole application. For consistency and to draw the users attention to these status, they should be handled the same as console. Attached screenshot example. 

Vgea transactions:

[ ] Stake
[ ] Remove stake
[ ] Vote
[ ] Create proposal
[ ] Create withdrawal (likely already done in the withdrawal lib)

E.g.
"Awaiting action in Eth wallet"
"Your connected Ethereum address does not have any $VEGA to associate" 
"√ Done Vega key c44... can now participate in governance and nominate a validator with your associated $VEGA"
"Confirm transaction in wallet"
" √ Vega tokens are approved for staking" 

## Tasks

- [ ] Align design of Vega transaction dialogs and Ethereum transaction dialogs (update to use web3 Ethereum transaction hooks from libs/web3 and delete hooks defined in token app). Perhaps into a shared transaction dialog which can be used in both cases 
- [ ] and then what?
- [ ] Etc.

We have some duplicated logic between apps/token and libs/react-helpers. Both use their own useTransaction hook. Lets convert the token app to use the react-helper version.

image

sam-keen:
# Related issues 🔗

Closes #1513

# Description ℹ️

For the vega transactions using embedded status alerts (voting, staking) convert to using the Dialog modal component for consistency. Also subscribed to voting events rather than polling.

# Demo 📺

Screenshot 2022-11-02 at 16 39 04


dexturr:
I think we want the content behind the dialog to be persisted. After staking the form disapears compeltely and is inoperable as a page

Screenshot 2022-11-03 at 15 15 07
Screenshot 2022-11-03 at 15 15 01

PR #1962, opened 2022-11-04
chore: fix update results from dataProvider when variable change

### LINKED ISSUE #1947: Collateral asset amounts does not update after switching key
## Description

When on https://console.fairground.wtf/ on the trading page and positions tab and you change the public key, the collateral tab refreshes but the amounts are stuck on the previous key.

## Steps to Reproduce

1. Go to https://console.fairground.wtf/
2. Watch the Collateral tab
3. Switch to another public key on the same wallet
4. Watch the collateral tab in the bottom left before and after the switch, amounts do not change to the latest wallet.

## Expected behavior

The selected wallets asset amounts should be shown.

## Screenshots
See video

## Device details

- OS: MacOS
- Browser Chrome

https://user-images.githubusercontent.com/100769129/199694087-6a04204d-8164-4d6b-9890-d6ffdc904aeb.mov

macqbat:
# Related issues 🔗

Closes #1947 

# Description ℹ️

Use `updateOnInit` directive instead of taking data property from `useDataProvider` in list views

### DIFF
--- a/libs/orders/src/lib/components/order-list-manager/use-order-list-data.ts
+++ b/libs/orders/src/lib/components/order-list-manager/use-order-list-data.ts
@@ -77,11 +77,9 @@ export const useOrderListData = ({
     update,
     insert,
     variables,
+    updateOnInit: true,
   });
-  if (!dataRef.current && data) {
-    totalCountRef.current = totalCount;
-    dataRef.current = data;
-  }
+  totalCountRef.current = totalCount;
 
   const getRows = makeInfiniteScrollGetRows<OrderEdge>(
     newRows,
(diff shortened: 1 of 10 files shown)


libs/orders/src/lib/components/order-list-manager/use-order-list-data.ts
```diff
@@ -77,11 +77,9 @@ export const useOrderListData = ({
     update,
     insert,
     variables,
+    updateOnInit: true,
   });
-  if (!dataRef.current && data) {
-    totalCountRef.current = totalCount;
-    dataRef.current = data;
-  }
+  totalCountRef.current = totalCount;
```
bglownia, line 82: chore: This could be also moved to update function, just for consistency

bglownia (approved):

libs/orders/src/lib/components/order-list-manager/use-order-list-data.ts
macqbat: yeah, but it's rather outside scope of this issue..

PR #1964, opened 2022-11-07
Chore/list asset

### LINKED ISSUE #1444: Link to whitelist asset on bridge

dexturr:
# Related issues 🔗

Closes #1444 

# Description ℹ️

Add button to proposals that are pending listing to allow users to list the asset on the bridge.

# Demo 📺

Screenshot 2022-11-07 at 10 34 16

### DIFF
--- a/apps/token/src/routes/governance/components/list-asset/list-asset.tsx
+++ b/apps/token/src/routes/governance/components/list-asset/list-asset.tsx
@@ -60,7 +60,7 @@ export const ListAsset = ({
 }: ListAssetProps) => {
   const { t } = useTranslation();
   const {
-    transaction: { perform, Dialog },
+    transaction,
     data,
     loadingAsset,
     errorAsset,
@@ -68,6 +68,7 @@ export const ListAsset = ({
     loadingBundle,
     errorBundle,
   } = useListAsset(assetId);
+  const { perform, Dialog } = transaction;
 
   if (
     !assetData?.erc20ListAssetBundle ||
@@ -89,6 +90,7 @@ export const ListAsset = ({
       <p className="pr-8">{t('ListAssetDescription')}</p>
       <EthWalletContainer>
         <Button
+          data-testid="list-asset"
           onClick={() =>
             perform(
               assetSource,
--- a/apps/token/src/routes/governance/components/list-asset/list-asset.tsx
+++ b/apps/token/src/routes/governance/components/list-asset/list-asset.tsx
@@ -1,0 +1,109 @@
+import type { CollateralBridge } from '@vegaprotocol/smart-contracts';
+import { AssetStatus } from '@vegaprotocol/types';
+import { Button } from '@vegaprotocol/ui-toolkit';
+import { useBridgeContract, useEthereumTransaction } from '@vegaprotocol/web3';
+import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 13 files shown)


apps/token/src/routes/governance/components/list-asset/list-asset.tsx
```diff
@@ -0,0 +1,111 @@
...
+    !assetData.erc20ListAssetBundle ||
+    !data?.asset ||
+    loadingAsset ||
+    loadingBundle
+  ) {
+    return null;
+  }
+  if (data.asset.source.__typename !== 'ERC20') return null;
+  if (data.asset.status !== AssetStatus.STATUS_PENDING_LISTING) return null;
+  if (errorAsset || errorBundle) return null;
```
sam-keen, line 84: Chore (non-blocking) looks like at least some of this custom logic can be simplified by using the AsyncRenderer component

apps/token/src/routes/governance/components/proposal/proposal.spec.tsx
```diff
@@ -49,4 +52,26 @@ it('renders each section', async () => {
   expect(screen.getByTestId('proposal-terms-json')).toBeInTheDocument();
   expect(screen.getByTestId('proposal-votes-table')).toBeInTheDocument();
   expect(screen.getByTestId('proposal-vote-details')).toBeInTheDocument();
+  expect(screen.queryByTestId('proposal-list-asset')).not.toBeInTheDocument();
+});
+
+it('renders whitelsit section if proposal is new asset and source is erc20', async () => {
```
sam-keen, line 58:
```suggestion
it('renders whitelist section if proposal is new asset and source is erc20', async () => {
```

apps/token/src/routes/governance/components/proposal/proposal.tsx
```diff
@@ -77,6 +77,15 @@ export const Proposal = ({ proposal }: ProposalProps) => {
         <div className="mb-8">
           <ProposalChangeTable proposal={proposal} />
         </div>
+        {proposal.terms.change.__typename === 'NewAsset' &&
```
sam-keen, line 80: Chore (blocking) there's a `proposalType` in the file here, you can use that for this check

sam-keen (changes requested):
Couple of small bits :)

apps/token/src/routes/governance/components/list-asset/list-asset.tsx
dexturr: I deliberately am avoiding async renderer here because it's a minute use case (like 99% of proposal will not render this component) so I didn't want a loading spinner that cause the content to jump around or delay the loading of the rest of the page until this was done. So I thought it best that this just adds itself, if the component is in that state. Happy to move it different behaviour if you think something would be better though?

sam-keen: Ooh, that's a good point. That makes sense to me!

apps/token/src/routes/governance/components/proposal/proposal.tsx
sam-keen: Boo, ok fair dos

sam-keen (approved):

PR #1969, opened 2022-11-07
chore(#1897): create separate e2e workflows to be run in parallel

### LINKED ISSUE #1897: Create separate e2e workflows to be run in parallel
## The Chore

We need to create separate e2e test workflows to make them fully independent and parallel.

## Tasks

- [ ] Crate workflows for each app.
- [ ] Modify current cypress test workflow to choose and run affected tests using created workflows.

This work is also necessary to apply vegawallet-dummy in tests, as we need to define test wallet at the very beginning and we are not currently able to switch between wallets during the flow.

radoslaw-szpiech:
# Related issues 🔗

Closes #1897 

# Description ℹ️

Current workflows has been modified to run specific apps and tests on hosted runners in parallel. Each project has separate flow now, and three workflows that we use in e2e - PR, Manual and Nightly - trigger these flows that are needed. It makes them much faster and completely independent from each other.


.github/workflows/cypress-token-e2e.yml
```diff
@@ -10,13 +10,15 @@ on:
...
+        type: string
       tags:
         required: false
         type: string
 
 jobs:
   token-e2e:
-    name: Run Cypress tests - PR
     runs-on: self-hosted
     timeout-minutes: 30
```
jtsang586: Is the 30 minute timeout including the night run tests? 30 minutes is cutting it a bit too close, I think 60 minutes would be safer

radoslaw-szpiech: Just increased it :)

mattrussell36 (approved):

PR #1971, opened 2022-11-07
chore(#1931): create single location for all external links

### LINKED ISSUE #1931: Create single location for all external links
## The Chore

Find all external links (docs, wallet download page, governance etc) and move them to a single place (in react-helpers) so that they can be defined in a single place across all apps.

## Tasks

- [ ] Find all links in all apps
- [ ] Combine them all into a single object in react-helpers
- [ ] Update imports

MadalinaRaicu:
# Related issues 🔗

Closes #1931

# Description ℹ️ & # Technical 👨‍🔧

All external links (docs, wallet download page, governance etc.) are moved to a single place (in react-helpers) so that they can be defined in a single place across all apps.

- [x] All links in all apps
- [x] Combined them all into a single object in react-helpers
- [x] Updated imports


libs/react-helpers/src/lib/constants/links.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export const DocsLinks = (docsUrl = 'https://docs.vega.xyz/testnet') => ({
+  AUCTION_TYPE_OPENING: `${docsUrl}/concepts/trading-on-vega/trading-modes#auction-type-opening`,
+  AUCTION_TYPE_LIQUIDITY_MONITORING: `${docsUrl}/concepts/trading-on-vega/trading-modes#auction-type-liquidity-monitoring`,
+  AUCTION_TYPE_PRICE_MONITORING: `${docsUrl}/concepts/trading-on-vega/trading-modes#auction-type-price-monitoring`,
+  AUCTION_TYPE_CLOSING: `${docsUrl}/concepts/trading-on-vega/trading-modes#auction-type-closing`,
+});
+
+export const ExternalLinks = {
+  SUSHI_PAIRS: 'https://analytics.sushi.com/pairs/',
```
mattrussell36, line 13: chore (blocking): Lets delete these sushi links. They are no longer used

libs/react-helpers/src/lib/constants/links.ts
```diff
@@ -0,0 +1,32 @@
+/** ExternalLinks returns external documentation links that are specific to an environment
```
mattrussell36, line 1: chore (blocking): Can we make this file just `lib/links`, or something similar? Because of the DocsLinks function the values here are no longer just constants.

libs/react-helpers/src/lib/constants/links.ts
```diff
@@ -0,0 +1,32 @@
...
+ * @param {string} docsUrl - the documentation URL for the environment ex. https://docs.vega.xyz/testnet or https://docs.vega.xyz (NX_VEGA_DOCS_URL)
+ */
+
+export const DocsLinks = (docsUrl = 'https://docs.vega.xyz/testnet') => ({
```
mattrussell36, line 5: chore (blocking): 2 things here:

A) I think it would be better to not hard code this default parameter and instead explicitly require an argument to be passed in. This value normally comes from an .env config. If the link changes we now have two places to update making this harder to maintain

B) `DocsLink` is neither a class nor constructor function nor compoment, can it just be lowercase and perhaps more obviously a function. Perhaps `createDosLinks`?

mattrussell36 (changes requested):

libs/react-helpers/src/lib/constants/links.ts
MadalinaRaicu: done. updated in last commit.

libs/react-helpers/src/lib/constants/links.ts
MadalinaRaicu: done. updated in last commit.

libs/react-helpers/src/lib/constants/links.ts
MadalinaRaicu: updated in last commit.

libs/react-helpers/src/lib/links/links.ts
```diff
@@ -0,0 +1,27 @@
...
+  AUCTION_TYPE_LIQUIDITY_MONITORING: `${docsUrl}/concepts/trading-on-vega/trading-modes#auction-type-liquidity-monitoring`,
+  AUCTION_TYPE_PRICE_MONITORING: `${docsUrl}/concepts/trading-on-vega/trading-modes#auction-type-price-monitoring`,
+  AUCTION_TYPE_CLOSING: `${docsUrl}/concepts/trading-on-vega/trading-modes#auction-type-closing`,
+});
+
+export const ExternalLinks = {
+  FEEDBACK: 'https://github.com/vegaprotocol/feedback/discussions',
+  GITHUB: 'https://github.com/vegaprotocol/token-frontend',
+  DISCORD: 'https://vega.xyz/discord',
+  STAKING_GUIDE:
```
dexturr, line 16: shouldn't this (and anything else with an environment hard coded) be in the generate docsLink function?

dexturr (changes requested):
Issue with environements

libs/react-helpers/src/lib/links/links.ts
MadalinaRaicu: updated in last commit.

libs/react-helpers/src/lib/links/links.ts
```diff
@@ -0,0 +1,25 @@
+/** createDocsLinks returns external documentation links that are specific to an environment
+ * @param {string} docsUrl - the documentation URL for the environment ex. https://docs.vega.xyz/testnet or https://docs.vega.xyz/mainnet (NX_VEGA_DOCS_URL)
+ */
+
+export const createDocsLinks = (docsUrl = 'https://docs.vega.xyz/mainnet') => ({
```
sam-keen, line 5: Chore (non-blocking): It'd be nice to have the url for the testnet docs in code too. Perhaps docsUrl could be a union type of the two possible urls

sam-keen (changes requested):

apps/token/src/routes/staking/home/staking-intro.tsx
```diff
@@ -32,7 +34,7 @@ export const StakingIntro = () => {
           <li>
             {t('stakingDescription2')}{' '}
             <UTLink
-              href={Links.VALIDATOR_FORUM}
+              href={ExternalLinks.VALIDATOR_FORUM}
```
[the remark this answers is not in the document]
MadalinaRaicu: this has been updated

libs/react-helpers/src/lib/links/links.ts
MadalinaRaicu: this has been updated too.

dexturr (approved):

libs/react-helpers/src/lib/links/links.ts
sam-keen: I meant something like this:
```suggestion
export const createDocsLinks = (docsUrl: 'https://docs.vega.xyz/mainnet' | 'https://docs.vega.xyz/testnet' = 'https://docs.vega.xyz/mainnet') => ({
```

apps/token/src/routes/staking/home/staking-intro.tsx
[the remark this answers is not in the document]
MadalinaRaicu: missed those files. updated in last commit.

libs/react-helpers/src/lib/links/links.ts
MadalinaRaicu: this is how I did it in the initial version. in this review, @mattrussell36 asked me to not make this parameter default and optional because it would be harder to maintain. I can restrict it to these two hardcoded URLs but do we really want to do that? how much does it help? are we going to have docs for more environments, would the URL change?

MadalinaRaicu: hm, I agree with restricting the URL to be either `'https://docs.vega.xyz/mainnet' | 'https://docs.vega.xyz/testnet'` but let's not pass a default, as @mattrussell36 suggested

MadalinaRaicu: I updated in last commit

sam-keen (approved):

mattrussell36 (approved):

PR #1974, opened 2022-11-08
chore: filter assets to withdraw

### LINKED ISSUE #1957: Remove assets from withdraw form that I don't have
## Chore 

In the withdrawal component, I can currently see a list of all supported assets to withdraw even when I don't have any / some of those assets in my wallet. Instead, I should see in the drop down the choices based on the assets I _do_ have in my wallet.

macqbat:
# Related issues 🔗

Closes #1957 

# Description ℹ️

Filter assets to owned by the user only. Display no asset message when no assets possible to withdraw


libs/withdraws/src/lib/withdraw-form-container.tsx
```diff
@@ -20,24 +19,22 @@ export const WithdrawFormContainer = ({
...
   });
-
-  const {
-    data: assets,
-    loading: assetsLoading,
-    error: assetsError,
-  } = useDataProvider({
-    dataProvider: enabledAssetsProvider,
-  });
+  const assets = data
```
dexturr, line 26: I think this needs to be more rigourous. A user can only withdraw an asset if their balance in their general account is non-zero so I think we need to add that filter logic in here as well

macqbat: I get this point, I was considering this also, but I think it's too tricky because balance of the account is still changing and for this reason we should provide there additional logic for refreshing balances etc..

dexturr: Sure that makes sense :+1: does the withdrawal window not already do this? Just conscious there may be race condition type bugs where a user could try and withdraw more than they have? Obviously not blocking this PR

dexturr (approved):

edd (approved):

PR #1988, opened 2022-11-08
chore: handle overlapping text in price cell 

### LINKED ISSUE #1849: Handle overlapping text in order book
We currently get over lapping text in the order book when its set to a high level of precision. We should add some css to handle overflow with an ellipsis

Seen on: https://stagnet1.console.vega.xyz/markets/18d2ff3b7aa9ac6403804487015a2cc8eefdb91a367ca8213e7ae81836576f45

Screen Shot 2022-10-25 at 08 55 39

## Tasks

- [ ] Truncate text with an ellipsis
- [ ] Ensure resizing the order book can show the full value

macqbat:
# Related issues 🔗

Closes #1849 

# Description ℹ️

Add css for handling overlapping text in priceCell and orderbook components

# Demo 📺

# Technical 👨‍🔧


libs/tailwindcss-config/src/vega-custom-classes.js
```diff
@@ -40,6 +40,9 @@ const vegaCustomClasses = plugin(function ({ addUtilities }) {
     '.color-scheme-dark': {
       colorScheme: 'dark',
     },
+    '.rtl-dir': {
```
bglownia, line 43: Maybe it will be easier to add all css rules (whitespace-nowrap overflow-hidden text-ellipsis text-right) to this class and rename it to something like .truncate-number

bglownia (approved):

libs/tailwindcss-config/src/vega-custom-classes.js
macqbat: imho this rtl is so uniq and not covered by tailwind, so let's not break the pattern of avoiding custom classes if it not necessary

PR #1995, opened 2022-11-08
chore(token tests): update failing tests 

jtsang586:
# Related issues 🔗

Fix some failing tests from some new changes. 

Some governance proposal tests are still failing until https://github.com/vegaprotocol/frontend-monorepo/pull/1919 gets merged


apps/token-e2e/.env
```diff
@@ -14,4 +14,4 @@ NX_VEGA_WALLET_URL=http://localhost:1789
 
 #Test configuration variables
 CYPRESS_FAIRGROUND=false
-CYPRESS_TEARDOWN_NETWORK_AFTER_FLOWS=true
+TEARDOWN_NETWORK_AFTER_FLOWS=true
```
radoslaw-szpiech, line 17: If it's used as `cy.env('TEARDOWN_NETWORK_AFTER_FLOWS')` it should stay defined in .env as `CYPRESS_TEARDOWN_NETWORK_AFTER_FLOWS`

jtsang586: Locally this doesn't get picked up if the variable includes `CYPRESS_` so I think removing the cypress is correct? 
I'd prefer removing this completely and only include it in cypress.config.js tbh. Is there a need for it to be in `.env`?

radoslaw-szpiech (approved):

PR #1997, opened 2022-11-08
chore(1987): persist ui behind dialog when staking

### LINKED ISSUE #1987: Persist UI behind dialog when staking
Currently, some UI is hidden whilst the staking transaction dialog is visible, and when the process is complete, the page is no longer usable as buttons have disappeared.

Ensure UI persists as much as possible throughout the staking process.

sam-keen:
# Related issues 🔗

Closes #1987

# Description ℹ️

Ensuring staking UI persists behind staking transaction dialog. Also added dialog close.

# Demo 📺

Screenshot 2022-11-08 at 16 24 14


apps/token/src/routes/staking/node/staking-form.tsx
```diff
@@ -331,6 +324,36 @@ export const StakingForm = ({
...
+      {formState === FormState.Success && (
+        <StakeSuccess
+          action={action}
+          amount={amount}
+          nodeName={nodeName}
+          removeType={removeType}
+          isDialogVisible={isDialogVisible}
+          toggleDialog={toggleDialog}
+        />
+      )}
```
dexturr, line 356: chore (non-blocking): this feels a bit messy could we throw this into it's own component?

apps/token/src/routes/staking/node/staking-form.tsx
```diff
@@ -196,43 +197,35 @@ export const StakingForm = ({
...
+        availableStakeToAdd.isEqualTo(0) &&
+        availableStakeToRemove.isEqualTo(0) && (
+          <div>
+            {appState.lien.isGreaterThan(0) ? (
+              <span className="text-red">{t('stakeNodeWrongVegaKey')}</span>
+            ) : (
+              <span className="text-red">{t('stakeNodeNone')}</span>
+            )}
+          </div>
+        )}
```
dexturr, line 228: chore (blocking): this needs to be outside this render, in this case the user should not be able to see the form as they have no stake that is addable.

apps/token/src/routes/staking/node/staking-form.tsx
sam-keen: Can do :)

apps/token/src/routes/staking/node/staking-form.tsx
sam-keen: @dexturr The form button doesn't allow the form to be usable unless there's sufficient stake, I thought it was more useful for users to see what they will be able to do, once these messages are resolved.

dexturr: Kewl

apps/token/src/routes/staking/node/staking-form.tsx
```diff
@@ -196,43 +197,35 @@ export const StakingForm = ({
...
-  ) {
-    if (appState.lien.isGreaterThan(0)) {
-      return <span className="text-red">{t('stakeNodeWrongVegaKey')}</span>;
-    } else {
-      return <span className="text-red">{t('stakeNodeNone')}</span>;
+  const toggleDialog = useCallback(() => {
+    setIsDialogVisible(!isDialogVisible);
+  }, [isDialogVisible]);
+
+  useEffect(() => {
```
mattrussell36, line 204: suggestion (non blocking): its not a good pattern to use useEffect to set state that simply relies on other state, it makes the code harder to follow. In this case can you not just call `setIsDialogVisible` at the start of the `onSubmit` function?

sam-keen: great shout, thanks

dexturr (approved):

PR #1998, opened 2022-11-08
feat(1677): toast component

### LINKED ISSUE #1677: Toast component
## The Chore

We need a design for a generic toast component added to the UI toolkit. This will initially be used for showing transaction statuses, but it should be flexible enough for other use cases too.

## Tasks

- [x] Brief/discuss use cases with UX/design team
- [x] Design
- [ ] Review
- [ ] Implement

Note: Radix now has a toast component which we should probably leverage to ensure a11y and interaction concerns are covered. 

We may need both a Toast component and possibly a Toaster component to handle display of multiple toasts (probably stacking them as it will be possible to execute multiple tx simultaneously)

asiaznik:
# Related issues 🔗

Closes #1677

# Description ℹ️

🥪🥪🥪

# Demo 📺

Screenshot 2022-11-09 at 14 59 05
Screenshot 2022-11-09 at 14 59 12
Screenshot 2022-11-09 at 14 59 17
Screenshot 2022-11-09 at 14 59 23
Screenshot 2022-11-09 at 14 59 29

https://user-images.githubusercontent.com/1980305/200849061-571076b8-9e9e-4384-be9f-88339799b08d.mov

### DIFF
--- a/libs/ui-toolkit/src/components/toast/toast.tsx
+++ b/libs/ui-toolkit/src/components/toast/toast.tsx
@@ -1,0 +1,132 @@
+import styles from './toast.module.css';
+
+import type { IconName } from '@blueprintjs/icons';
+import { IconNames } from '@blueprintjs/icons';
+import classNames from 'classnames';
+import { useEffect } from 'react';
+import { useCallback } from 'react';
+import { useLayoutEffect } from 'react';
+import { useRef } from 'react';
+import { Intent } from '../../utils/intent';
+import { Icon } from '../icon';
+
+type ToastContentProps = { id: string };
+type ToastContent = (props: ToastContentProps) => JSX.Element;
+
+type ToastState = 'initial' | 'showing' | 'expired';
+
+export type Toast = {
+  id: string;
+  intent: Intent;
+  render: ToastContent;
+  closeAfter?: number;
+  signal?: 'close';
+};
+
+type ToastProps = Toast & {
+  state?: ToastState;
+  onClose?: (id: string) => void;
+};
+
+const toastIconMapping: { [i in Intent]: IconName } = {
+  [Intent.None]: IconNames.HELP,
+  [Intent.Primary]: IconNames.INFO_SIGN,
+  [Intent.Success]: IconNames.TICK_CIRCLE,
+  [Intent.Warning]: IconNames.ERROR,
+  [Intent.Danger]: IconNames.ERROR,
+};
+
+const getToastAccent = (intent: Intent) => ({
+  // strip
+  'bg-gray-200 text-black text-opacity-70': intent === Intent.None,
+  'bg-vega-blue text-white text-opacity-70': intent === Intent.Primary,
+  'bg-success text-white text-opacity-70': intent === Intent.Success,
+  'bg-warning text-white text-opacity-70': intent === Intent.Warning,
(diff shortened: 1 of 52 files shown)


libs/ui-toolkit/src/components/toast/toast.css
```diff
@@ -0,0 +1,20 @@
...
+  top: 20px;
+  opacity: 0;
+  max-height: 0;
+  border: 0;
+  margin-bottom: 0;
+}
+
+.toast-showing {
+  opacity: 1;
+  transition: all 0.3s;
```
mattrussell36, line 11: fix (blocking): There are some scroll bars popping in and out as the toast animates off screen can we fix that.

TBH I'd prefer to leave the animation out for now as Joelle would like to set out some animation guidelines for us as part of the design system.

libs/ui-toolkit/src/components/toast/toast.tsx
```diff
@@ -0,0 +1,165 @@
...
+    });
+    return () => cancelAnimationFrame(req);
+  }, [id]);
+
+  useEffect(() => {
+    const onClick = (e: MouseEvent) => {
+      if (
+        toastRef.current &&
+        closeAfterAction &&
+        (e.target instanceof HTMLButtonElement ||
```
mattrussell36, line 98: suggestion (non blocking): What if there are multiple buttons/anchors to do different things within the toast content? I guess that might unlikely.. but at a glance it might be better to make `children` a render prop and pass the toast id down to the toast content. Then the toast content can use the toast store to close itself on click of whatever element it wants.

This would solve that problem and keep us a little more in the react paradigm

libs/ui-toolkit/src/components/toast/use-toasts.ts
```diff
@@ -0,0 +1,66 @@
...
+      toasts: [...state.toasts, toast],
+    })),
+  update: (id, newToast) =>
+    set((state) => {
+      const toasts = [...state.toasts];
+      const toastIdx = toasts.findIndex((t) => t.id === id);
+      if (toastIdx > -1) toasts[toastIdx] = newToast;
+      return { toasts };
+    }),
+  close: (id) =>
```
mattrussell36, line 48: question (non blocking): Do we want to remove that toast from the toasts array completely. I guess this is problematic with the animation. Perhaps we should reach for react-spring or similar to make this easier to handle.

libs/ui-toolkit/src/components/toast/toast.tsx
```diff
@@ -0,0 +1,165 @@
...
+        if (toastRef.current.contains(target)) {
+          setTimeout(() => closeToast(), closeAfterAction);
+        }
+      }
+    };
+    document.addEventListener('click', onClick);
+    return () => document.removeEventListener('click', onClick);
+  });
+
+  useEffect(() => {
```
mattrussell36, line 111: suggestion (non blocking): I think these useEffects could be combined

MadalinaRaicu (approved):

libs/ui-toolkit/src/components/toast/use-toasts.ts
asiaznik: I'd leave it as is and change it later once we'll have the animation guideline ready.

PR #2000, opened 2022-11-09
feat(#1643): orders table filtering

### LINKED ISSUE #1643: Orders table filtering
Depends on #1675 

## Story

As a trader
I want to be able to sort and filter the orders table
So that I can more easily find order information

## Acceptance Criteria

- [ ] I can filter by date range
- [ ] I can filter by status
- [ ] I can filter by type.
- [ ] I can filter by TIF
- [ ] Applied filters are editable
- [ ] Default sorting and filter is applied. 
  - [ ] status filtered by active
  - [ ] primary sort by limit -> market -> liqudity provision
  - [ ] secondary sort by price high -> low

bglownia:
# Related issues 🔗

Closes #1643 
Depends on [#6716](https://github.com/vegaprotocol/vega/issues/6716)

# Description ℹ️

Add filters to order table 

# Technical 👨‍🔧

Two custom ag-grid filters added: SetFilter (to support type status and time in force columns ) and DateRangeFilter.
Filtering is done on server side. After filter value is changed variables passed to use data provider are updated and new data provider instance is initialized.
There were issues in handling empty results and showing no data available label because it required to change rowDataModel in ag-grid which caused errors thrown from it. rowDataModel needs to be set to infinite all time, no data label is handled by async renderer.


libs/orders/src/lib/components/order-list-manager/order-list-manager.tsx
```diff
@@ -30,16 +41,49 @@ export const OrderListManager = ({ partyId }: OrderListManagerProps) => {
...
-        rowData={data?.length ? undefined : []}
+        rowModelType="infinite"
         datasource={{ getRows }}
         onBodyScrollEnd={onBodyScrollEnd}
         onBodyScroll={onBodyScroll}
+        onFilterChanged={onFilterChanged}
+        onSortChanged={onSortChange}
       />
-    </AsyncRenderer>
+      <div className="pointer-events-none absolute inset-0 top-5">
```
mattrussell36, line 78: question (non blocking): This seems a bit weird. Why is OrderList not rendered within the AsyncRenderer?

mattrussell36 (approved):

PR #2029, opened 2022-11-11
fix: estimate order margin query asset decimal places fix

### LINKED ISSUE #2002: Incorrect decimal used on estimate order response
## Description

The wrong decimal is used on the response from estimate order. This endpoint use market decimal for price and position market decimal for size, then returns all amounts in Asset decimal.

The front end needs to apply the asset decimals on the returned results (fees and margins)

## Steps to Reproduce

1. Go to stagnet1 console
2. Click on deal ticket
3. enter price and size
4. See the returned margins and fee, it's wrong

## Expected behavior

The amount displayed should be scaled using asset decimal

MadalinaRaicu:
# Related issues 🔗

Closes #2002


asiaznik (approved):

asiaznik (approved):

libs/deal-ticket/src/hooks/use-order-margin.ts
```diff
@@ -78,13 +82,18 @@ export const useOrderMargin = ({
     ).toString();
     const { makerFee, liquidityFee, infrastructureFee } =
       data.estimateOrder.fee;
+    const assetDecimals =
+      market.tradableInstrument.instrument.product.settlementAsset.decimals;
     return {
-      margin: addDecimal(margin, market.decimalPlaces),
-      totalFees: addDecimal(fees, market.decimalPlaces),
+      margin: addDecimalsNormalizeNumber(margin, assetDecimals),
```
bglownia, line 88: I looks into the flow of this value in code and it looks like it's changed back and forth from string to bigNum
It will go through `addDecimalsNormalizeNumber(margin, assetDecimals)` in here
then passed to `useOrderValidation ` in `DealTicketSteps` as `estMargin`
then passed to `useOrderMarginValidation` in `useOrderValidation` as `estMargin`
changed to big num - `const margin = toBigNum(estMargin?.margin || 0, 0)`
than changed to string - `const marginString = margin.toString();`
returned value is passed to `ValidateMargin`
and then it will go through normalizeFormatNumber

It looks like we divide by Math.pow(10, decimals) two times

bglownia (changes requested):
Return values from useOrderMargin as BigInt string 

```
return {
  margin:,
  totalFees: fees,
  fees: {
    makerFee,
    liquidityFee,
    infrastructureFee: infrastructureFee,
  },
};
```

libs/deal-ticket/src/hooks/use-order-margin.ts
MadalinaRaicu: image

MadalinaRaicu: it is converted toBigNum in use-order-margin-validation `toBigNum(estMargin?.margin || 0, 0)` but with 0 decimals.

MadalinaRaicu: ok. time for cleanup

MadalinaRaicu: agree. that is what I am trying to do

bglownia (approved):

PR #2041, opened 2022-11-11
fix: remove useDataProviderHook updateOnInit

bglownia:


libs/market-list/src/lib/market-data-provider.ts
```diff
@@ -20,8 +20,11 @@ const update = (data: MarketData, delta: MarketDataUpdateFieldsFragment) => {
   });
 };
 
-const getData = (responseData: MarketDataQuery): MarketData | null =>
-  responseData?.marketsConnection?.edges[0].node.data || null;
+const getData = (responseData: MarketDataQuery): MarketData | null => {
+  const data = responseData?.marketsConnection?.edges[0].node.data || null;
+  console.log({ data });
```
macqbat, line 25: don't like it :-)

Issue #1909, opened 2022-10-31
Verify readiness for data-node V1 API deprecation


pennyandrews:
## The Chore

This core sprint (next 2 weeks) the core team would like to deprecate the data-node V1 APIs, with the aim to remove before the mainnet release.
We need to make sure that we have migrated all our queries, scripts, tests, bots, sims etc.. to the V2 APIs. Gordon has been notifying in the Release overview notes of the deprecation for the last 3 releases[ here](https://docs.vega.xyz/testnet/releases/overview#:~:text=now%20in%20testing.-,API%20DEPRECATIONS,-Data%20node%3A%20The) for reference

## Tasks

Updated queries can be pulled from here: 

https://github.com/vegaprotocol/system-tests/blob/7732072d81eb25103752974c2e32f3ab3fc3a877/tests/API/queries.graphql

**Console (Matt)**
- [x] Check and verify that we have successfully migrated all our queries, scripts, tests, etc. to the V2 APIs 
- [x] Create new / prioritise existing issues for any additional work required to complete this migration 
- [x] Notify Gordon of the readiness status for your dapp(s) in this thread (with links to issues as applicable) https://vegaprotocol.slack.com/archives/CD4AKTFQT/p1667218910037579

**Governance & Token Utility (Dex)**
- [x] Check and verify that we have successfully migrated all our queries, scripts, tests, etc. to the V2 APIs 
- [x] Create new / prioritise existing issues for any additional work required to complete this migration 
- [x] Notify Gordon of the readiness status for your dapp(s) in this thread (with links to issues as applicable) https://vegaprotocol.slack.com/archives/CD4AKTFQT/p1667218910037579

**Desktop Wallet (Botond)**
- [x] Check and verify that we have successfully migrated all our queries, scripts, tests, etc. to the V2 APIs 
- [x] Create new / prioritise existing issues for any additional work required to complete this migration 
- [x] Notify Gordon of the readiness status for your dapp(s) in this thread (with links to issues as applicable) https://vegaprotocol.slack.com/archives/CD4AKTFQT/p1667218910037579

**Block Explorer (Edd)**
- [x] Check and verify that we have successfully migrated all our queries, scripts, tests, etc. to the V2 APIs 
- [x] Create new / prioritise existing issues for any additional work required to complete this migration 
[...]

PR #2049, opened 2022-11-14
chore: update explorer to use v2 queries

### LINKED ISSUE #1797: Token migration of generated types
Currently we are midway through migrating the apollo type generatation across the entire repository. We should catch up with @notbot00 in regards to this and complete the migration on the token app side so we can remove all the duplicated types.

edd:
Addresses two Explorer graphql query related tasks:
- Update all graphql queries from variables to external files
- Switch apollo `useQuery` for generated query providers
- Update all graphql queries from `@deprecated` queries
- Refactor all uses of types that are now within `Schema.` (`MarketState`, `MarketAccountType`...)

# Related issues 🔗

Relates to #1909. Closes #1797.

# Description ℹ️

Explorer is relying on APIs that are being deprecated. This PR moves it to the `xConnection` queries. No other changes
should be too visible.


apps/explorer-e2e/src/integration/validator.cy.js
```diff
@@ -87,7 +87,7 @@ context('Validator page', { tags: '@smoke' }, function () {
         });
     });
 
-    it('should be able to see relevant node information in vega data section', function () {
+    xit('should be able to see relevant node information in vega data section', function () {
```
notbot00, line 90: question (non-blocking): if this test is breaking, i think leaving a comment would help for future reference explaining what's going on

apps/explorer/src/app/routes/assets/index.tsx
```diff
@@ -1,63 +1,38 @@
...
 
 const Assets = () => {
-  const { data } = useQuery<AssetsQuery>(ASSETS_QUERY);
+  const { data } = useExplorerAssetsQuery();
 
-  const assets =
-    data?.assetsConnection?.edges
-      ?.filter((e) => e && e?.node)
-      .map((e) => e?.node as AssetsQuery_assetsConnection_edges_node) || [];
+  const assets = data?.assetsConnection?.edges?.map((n) => {
```
notbot00, line 11: comment (non-blocking): added a convenience function to grab nodes from these paginated structures, it lives in the react helpers:

```
import {getNodes} from '@vegaprotocol/react-helpers' 

const nodes = getNodes(data?.assetsConnection)
```

not sure if it will work everywhere, might need to take a type parameter for the actual node type, which is usually not generated out of the box, unless you define it as a fragment. probably not super important, just in the future if you get bored with mapping over edges, feel free to give it a go :)

edd: Thanks - I will use that in future, but given how many files this PR touches I'm not going to make the change in this PR. Ticketing for later.

apps/token/src/components/vega-wallet/hooks.ts
```diff
@@ -90,7 +90,9 @@ export const usePollForDelegations = () => {
             ).map((e) => e.node);
             setAccounts(
               accounts
-                .filter((a) => a.type === AccountType.ACCOUNT_TYPE_GENERAL)
+                .filter(
```
notbot00, line 93: nit (non-blocking): i know it's not a change in the implementation, usually i think it's nicer to do a reduce over a map + filter, to reduce the number of iterations. not sure how big these arrays get, if >100, probably doesn't matter too much.

apps/explorer/src/app/routes/assets/index.tsx
notbot00: makes sense

notbot00 (approved):

apps/token/src/components/vega-wallet/hooks.ts
edd: Also noted. It *shouldn't* get that big, but over time it is possible that a party could have that many accounts.

PR #2054, opened 2022-11-14
feat(#2050): same wallet connect flow for both cli and desktop wallet

### LINKED ISSUE #2050: Connect dialog should use wallet v2 after v0.4.0 desktop wallet release
- Create a single button for connecting via JSON RPC for both desktop and cli wallet apps

MadalinaRaicu:
# Related issues 🔗

Closes #2050 

# Description ℹ️

Same wallet connect flow for both CLI and desktop wallet.

# Demo 📺

image

# Technical 👨‍🔧


libs/wallet/src/connect-dialog/connect-dialog.tsx
```diff
@@ -203,18 +197,11 @@ const ConnectorList = ({
         <ConnectDialogTitle>{t('Connect')}</ConnectDialogTitle>
         <CustomUrlInput walletUrl={walletUrl} setWalletUrl={setWalletUrl} />
         <ul data-testid="connectors-list" className="mb-6">
-          <li className="mb-4">
```
mattrussell36, line 206: chore (blocking): Ux team would prefer explicit separate buttons for desktop/cli, even though under the hood they will do the same thing.

```
<ConnectionOption
              type="gui"
              text={t('Desktop wallet app')}
              onClick={() => onSelect('local')}
            />
```

libs/wallet/src/connect-dialog/connect-dialog.tsx
```diff
@@ -251,39 +238,6 @@ const SelectedForm = ({
   reset: () => void;
   onConnect: () => void;
 }) => {
-  if (connector instanceof RestConnector) {
```
mattrussell36, line 254: chore (blocking): unfortunately we still need the rest connector form to support hosted wallet

libs/wallet/src/connect-dialog/connect-dialog.tsx
```diff
@@ -148,10 +145,7 @@ const ConnectDialogContainer = ({
   const { connect, ...jsonRpcState } = useJsonRpcConnect(delayedOnConnect);
 
   const handleSelect = (type: WalletType) => {
-    // Only cli is currently uses JsonRpc, this will need to be updated
-    // when gui does too
-    const connector =
```
mattrussell36, line 153: chore (blocking): still need this condition to support hosted wallet

mattrussell36 (changes requested):

libs/wallet/src/connect-dialog/connect-dialog.tsx
mattrussell36: As discussed lets keep the one button but update the text to 'Connect Vega wallet'

mattrussell36 (changes requested):

PR #2067, opened 2022-11-14
feat(1872): validator info table now using better status query

### LINKED ISSUE #1872: TFE: validator info table could use better status query
## Description

There are 2 node statuses that come out of the API. I believe the league table on the staking page shows the status type that defines consensus vs pending. The validator info page’s table shows the other status, which (I think) describes if a node is validator/non-validator. The validator page should have the same type of status showing as the league table does.

You can see below that on the staking page league table, “best validator” has a status of pending but on their profile page, it says “node_status_validator”
Screenshot 2022-10-26 at 17 09 17
Screenshot 2022-10-26 at 17 09 10

sam-keen:
# Related issues 🔗

Closes #1872


edd (approved):

apps/token/src/routes/staking/home/node-list.tsx
```diff
@@ -47,6 +47,18 @@ interface CanonisedNodeProps {
   [VOTING_POWER]: string;
 }
 
+export const translateStatus = (status: Schema.ValidatorStatus) =>
+  t(
```
dexturr, line 51: nitpick (non-blocking): I thnk this would be nicer as a map IMO

dexturr (approved):

PR #2069, opened 2022-11-14
fix(#2063): console liquidity tab keeps adding duplicate rows

### LINKED ISSUE #2063: Console Liquidity Tab keeps adding duplicate rows
## Description

If you access the liquidity tab, being logged in to a wallet that has added liquidity or not, then wait, the rows of e.g. inactive liquidity keeps on adding on but they are duplicates of what is already there so it just keeps "refreshing" and adding on duplicate rows.

## Steps to Reproduce

1. Go to https://console.fairground.wtf/#/liquidity/07549b2cf93abbf6fb8ad976af3f24876e5946df85a3b6f0a6338672c7453bfa
2. Wait until rows of liquidity rejects or cancellations are added

- OS: MacOS
- Browser Chrome

Uploading Screen Recording 2022-11-14 at 16.13.15.mov…

MadalinaRaicu:
# Related issues 🔗

Closes #2063 

# Description ℹ️

Liquidity provision entries are unique for any partyId and marketId only if the status is not rejected. 

# Demo 📺

image

# Technical 👨‍🔧

### DIFF
--- a/libs/liquidity/src/lib/liquidity-data-provider.ts
+++ b/libs/liquidity/src/lib/liquidity-data-provider.ts
@@ -78,7 +78,7 @@ export const liquidityProvisionsDataProvider = makeDataProvider<
   },
 });
 
-function isLpProvision(
+function isLpFragment(
   entry:
     | LiquidityProvisionFieldsFragment
     | IterableElement<
@@ -99,7 +99,7 @@ export const getId = (
         LiquidityProvisionsUpdateSubscription['liquidityProvisions']
       >
 ) =>
-  isLpProvision(entry)
+  isLpFragment(entry)
     ? `${entry.party.id}${entry.status}${entry.createdAt}`
     : `${entry.partyID}${entry.status}${entry.createdAt}`;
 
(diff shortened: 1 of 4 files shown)


libs/liquidity/src/lib/liquidity-data-provider.ts
```diff
@@ -76,6 +78,31 @@ export const liquidityProvisionsDataProvider = makeDataProvider<
   },
 });
 
+function isLpProvision(
```
mattrussell36, line 81: suggestion (non blocking): I think the naming here is off. They are both LPs but one is a subscription update and one is a query result. Perhaps `isLPFragment` would be better?

mattrussell36: Secondly the function body is super hard to read, can we simplify to the following. The type discrimination on the __typename field should give us all the type safety we need

```
function isLPFragment(
  entry:
    | LiquidityProvisionFieldsFragment
    | IterableElement<
        LiquidityProvisionsUpdateSubscription['liquidityProvisions']
      >
): entry is LiquidityProvisionFieldsFragment {
  if (entry.__typename === 'LiquidityProvision') {
    return true;
  }
  return false;
}
```

mattrussell36 (approved):

libs/liquidity/src/lib/liquidity-data-provider.ts
MadalinaRaicu: agree, updated in last commit.

PR #2081, opened 2022-11-15
chore(#1909): remove deprecated apis

### LINKED ISSUE #2080: Update `Accounts` queries
@vegaprotocol/frontend Update `Accounts` queries

_Originally posted by @asiaznik in https://github.com/vegaprotocol/vega/issues/6817#issuecomment-1315523268_

mattrussell36:
# Related issues 🔗

Related to #1909 
Closes #2080 

# Description ℹ️

Updates `accounts` queries to use `accountsConnection`


libs/market-info/src/components/market-info/info-market.tsx
```diff
@@ -97,7 +98,13 @@ export const Info = ({ market, onSelect }: InfoProps) => {
     [market]
   );
   const { data: asset } = useAssetDataProvider(assetId ?? '');
+
   if (!market) return null;
+
+  const marketAccounts = compact(market.accountsConnection?.edges).map(
```
dexturr, line 104: thought (non-blockng): This is becoming something we do so often would it be worth adding a helper function for it?

dexturr (approved):

MadalinaRaicu (approved):

libs/market-info/src/components/market-info/info-market.tsx
mattrussell36: That would make sense, but multiple accounts queries with different but overlapping resulting types makes this not really possible, at least not without additional code to handle different query shapes. Going to resolve but LMK if you have a good solution.

PR #2095, opened 2022-11-16
chore: separate running vegawallet service out of vegacapsule

radoslaw-szpiech:
# Description ℹ️

Vegawallet needs to be started separately, not in vegacapsule network bootstrap process, in order to use dummy wallet correctly.

### DIFF
--- a/.github/workflows/cypress-explorer-e2e.yml
+++ b/.github/workflows/cypress-explorer-e2e.yml
@@ -79,12 +79,10 @@ jobs:
       ## Setup a Vega wallet for our user
       ######
 
-      - name: Set up Vegawallet
-        uses: ./frontend-monorepo/.github/actions/setup-vegawallet
+      - name: Set up Vegawallet for docker
+        uses: ./frontend-monorepo/.github/actions/setup-vegawallet-docker
         with:
-          recovery: ${{ secrets.TRADING_TEST_VEGA_WALLET_RECOVERY }}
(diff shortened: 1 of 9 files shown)


.github/workflows/cypress-explorer-e2e.yml
```diff
@@ -79,12 +79,10 @@ jobs:
       ## Setup a Vega wallet for our user
       ######
 
-      - name: Set up Vegawallet
-        uses: ./frontend-monorepo/.github/actions/setup-vegawallet
+      - name: Set up Vegawallet for docker
+        uses: ./frontend-monorepo/.github/actions/setup-vegawallet-docker
         with:
-          recovery: ${{ secrets.TRADING_TEST_VEGA_WALLET_RECOVERY }}
           passphrase: ${{ secrets.CYPRESS_TRADING_TEST_VEGA_WALLET_PASSPHRASE }}
```
jtsang586, line 85: Non blocking: Since the wallet mnemonic is exposed there's not much point in having the passphrase in a secret. We can include it as a file in the same location as the recovery file and there won't be a need to create it in the workflow

jtsang586 (approved):
Lgtm

.github/workflows/cypress-explorer-e2e.yml
radoslaw-szpiech: Yeah, basically you're right, as when we know mnemonic, hiding password is pointless as you can set it to whatever you want. But just for "security purpose" I don't want to share it here now.

PR #2152, opened 2022-11-17
feat(explorer): add specific link types, start specific tx views

### LINKED ISSUE #2127: ValidatorHeartbeat transaction (validator)
## Story

As a
I want
So that

## Acceptance Criteria

- [ ] I can
- [ ] I can
- [ ] I can

## Tasks

- [ ] Explore and sketch
- [ ] Visual Design
- [ ] Team review
- [ ] Etc.

edd:
# Related issues 🔗

Rejigs the Validator Heartbeat transaction details page, because that is the easiest transaction that also appears frequently on testnet. It's not a high value transaction. But it's enough to validate the approach and how it'll roll out to all the different transaction types.

- Add `PartyLink`, `NodeLink`, `AssetLink`, `BlockLink` components
- Add `TxDetailsWrapper` to display a generic set of data per tx, plus extra rows for known types
- Add stub `TxGeneric` for transactions without specific components
- Add `TxHeartbeat` view
- Add basic `TxChainEvent` view (shows the most basic details)
- Add basic `TxBatch` view (simply shows the counts of changes in the batch)
- Add basic `TxLpAmend` view (simply shows the market for the order)
- Add basic `TxOrder` view (simply shows the market of the order)
- Update TX detail view to hide decoded tx by default

Closes #2127

# Description ℹ️

Adds more detail to the Validator Heartbeat transaction page, as a basis for adding more transaction specific detail to all of the transaction details page. Most fields also link to another page (e.g. block height is now a link).

# Demo 📺
Screenshot 2022-11-17 at 16 14 59

### DIFF
--- a/apps/explorer/src/app/components/txs/details/tx-details-wrapper.tsx
+++ b/apps/explorer/src/app/components/txs/details/tx-details-wrapper.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 import { DATA_SOURCES } from '../../../config';
 import { t, useFetch } from '@vegaprotocol/react-helpers';
 import { TxDetailsOrder } from './tx-order';
@@ -11,8 +11,10 @@ import { TxDetailsBatch } from './tx-batch';
 import { TxDetailsChainEvent } from './tx-chain-event';
 import { TxContent } from '../../../routes/txs/id/tx-content';
 
+type resultOrNull = BlockExplorerTransactionResult | undefined;
+
 interface TxDetailsWrapperProps {
-  txData: BlockExplorerTransactionResult | undefined;
+  txData: resultOrNull;
   pubKey: string | undefined;
   height: string;
 }
@@ -28,53 +30,15 @@ export const TxDetailsWrapper = ({
     `${DATA_SOURCES.tendermintUrl}/block?height=${height}`
   );
 
+  const child = useMemo(() => getTransactionComponent(txData), [txData]);
+
   if (!txData) {
     return <>{t('Awaiting Block Explorer transaction details')}</>;
   }
 
-  let child;
-
-  if (txData.type === 'Submit Order') {
-    child = (
-      <TxDetailsOrder txData={txData} blockData={blockData} pubKey={pubKey} />
-    );
-  } else if (txData.type === 'Validator Heartbeat') {
-    child = (
-      <TxDetailsHeartbeat
-        txData={txData}
-        blockData={blockData}
-        pubKey={pubKey}
-      />
-    );
-  } else if (txData.type === 'Amend LiquidityProvision Order') {
-    child = (
-      <TxDetailsLPAmend txData={txData} blockData={blockData} pubKey={pubKey} />
-    );
-  } else if (txData.type === 'Batch Market Instructions') {
-    child = (
-      <TxDetailsBatch txData={txData} blockData={blockData} pubKey={pubKey} />
-    );
-  } else if (txData.type === 'Chain Event') {
-    child = (
-      <TxDetailsChainEvent
-        txData={txData}
-        blockData={blockData}
-        pubKey={pubKey}
-      />
-    );
-  } else {
-    child = (
-      <TxDetailsGeneric txData={txData} blockData={blockData} pubKey={pubKey} />
-    );
-  }
-
-  if (!child) {
-    return null;
-  }
-
   return (
     <>
-      <section>{child}</section>
+      <section>{child({ txData, pubKey, blockData })}</section>
 
       <details title={t('Decoded transaction')} className="mt-3">
         <summary className="cursor-pointer">{t('Decoded transaction')}</summary>
@@ -90,3 +54,30 @@ export const TxDetailsWrapper = ({
     </>
   );
 };
+
+/**
+ * Chooses the appropriate component to render the full details of a transaction
+ *
+ * @param txData
+ * @returns JSX.Element
+ */
+function getTransactionComponent(txData: resultOrNull) {
+  if (!txData) {
+    return TxDetailsGeneric;
+  }
+
+  switch (txData.type) {
+    case 'Submit Order':
+      return TxDetailsOrder;
+    case 'Validator Heartbeat':
+      return TxDetailsHeartbeat;
+    case 'Amend LiquidityProvision Order':
+      return TxDetailsLPAmend;
+    case 'Batch Market Instructions':
+      return TxDetailsBatch;
+    case 'Chain Event':
+      return TxDetailsChainEvent;
+    default:
+      return TxDetailsGeneric;
+  }
+}


apps/explorer/.env.testnet
```diff
@@ -1,11 +1,12 @@
...
+NX_TENDERMINT_URL=https://tm.be.testnet.vega.xyz
 NX_BLOCK_EXPLORER=https://be.testnet.vega.xyz/rest
-NX_TENDERMINT_WEBSOCKET_URL=wss://be.testnet.vega.xyz/
+NX_TENDERMINT_WEBSOCKET_URL=wss://tm.be.testnet.vega.xyz/
 NX_VEGA_CONFIG_URL=https://static.vega.xyz/assets/testnet-network.json
 NX_VEGA_NETWORKS={\"MAINNET"\:\"https://explorer.vega.xyz"\,\"TESTNET\":\"https://explorer.fairground.wtf\"}
 NX_VEGA_ENV=TESTNET
 NX_GITHUB_FEEDBACK_URL=https://github.com/vegaprotocol/feedback/discussions
-NX_VEGA_URL=https://api.n09.testnet.vega.xyz/graphql
+NX_VEGA_URL=https://api.n07.testnet.vega.xyz/graphql
```
notbot00, line 10: suggestion (non-blocking): actually, i think you can remove the NX_VEGA_URLs from the env files, that would result in the node switcher finding you a healthy node to connect to, which results in a smoother experience when nodes are down. i think console did this recently.

apps/explorer/src/app/components/links/asset-link/asset-link.tsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import { Routes } from '../../../routes/route-names';
+import { useExplorerAssetQuery } from './__generated__/Asset';
+import { Link } from 'react-router-dom';
+
+export type AssetLinkProps = {
```
notbot00, line 6: suggestion ():  to make sure no unintended prop goes in to the Link component, you can merge this with the Props of the Link, e.g:

```
import { ComponentProps } from 'react'
import { Link } from 'react-router-dom'

type AssetLinkProps = ComponentProps<typeof Link> & {
  id: string
}
```

note: react router might export the props for the link type, not sure if does, `ComponentProps` is a good way to get around not having access to certain component's props type in case.

apps/explorer/src/app/components/links/block-link/block-link.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import { Routes } from '../../../routes/route-names';
+import { Link } from 'react-router-dom';
+
+export type BlockLinkProps = {
```
notbot00, line 5: suggestion (?): same for the BlockLinkProps, would be nicer to merge with the Link component's props type to ensure you can only pass in valid things

apps/explorer/src/app/components/txs/details/tx-details-wrapper.tsx
```diff
@@ -0,0 +1,75 @@
...
+export const TxDetailsWrapper = ({
+  txData,
+  pubKey,
+  height,
+}: TxDetailsWrapperProps) => {
+  const {
+    state: { data: blockData },
+  } = useFetch<TendermintBlocksResponse>(
+    `${DATA_SOURCES.tendermintUrl}/block?height=${height}`
+  );
```
edd, line 28: **nitpick (non-blocking)**: This should have a cache as right now it triggers a refetch every render

apps/explorer/.env.testnet
edd: 👍 Removed

apps/explorer/src/app/components/txs/details/tx-details-wrapper.tsx
edd: Moved to #2172

apps/explorer/src/app/components/links/block-link/block-link.tsx
edd: 👍 As above - this worked nicely. Applied it across all the new Link components

apps/explorer/src/app/components/txs/details/tx-batch.tsx
```diff
@@ -0,0 +1,75 @@
...
+export const TxDetailsBatch = ({
+  txData,
+  pubKey,
+  blockData,
+}: TxDetailsBatchProps) => {
+  if (!txData) {
+    return <>{t('Awaiting Block Explorer transaction details')}</>;
+  }
+
+  const cmd = txData.command as BatchMarketInstructions;
```
notbot00, line 33: question (not sure): what's the type you get when you try to access the command prop on the txData? usually type casting like this with `as` indicates a potential leak

apps/explorer/src/app/components/txs/details/tx-details-wrapper.tsx
notbot00: first time i see someone nitpick on their own code :)

apps/explorer/src/app/components/txs/details/tx-details-wrapper.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  if (!txData) {
+    return <>{t('Awaiting Block Explorer transaction details')}</>;
+  }
+
+  let child;
+
+  if (txData.type === 'Submit Order') {
```
notbot00, line 37: nitpicky suggestion (non-blocking): i think a switch statement would look more neat here, intead of the if/else chain, but this is also subjective.

apps/explorer/src/app/components/txs/details/tx-lp-amend.tsx
```diff
@@ -0,0 +1,45 @@
...
+export const TxDetailsLPAmend = ({
+  txData,
+  pubKey,
+  blockData,
+}: TxDetailsOrderProps) => {
+  if (!txData) {
+    return <>{t('Awaiting Block Explorer transaction details')}</>;
+  }
+
+  const cmd = txData.command as AmendLiquidityProvisionOrder;
```
notbot00, line 32: question (not sure): here also, why do the types break without the casting i wonder?

notbot00: actually found the root, resolving this, commenting below instead.

apps/explorer/src/app/components/txs/details/tx-batch.tsx
notbot00: dropping a comment below on the type interface, that seems to be the source i think

apps/explorer/src/app/routes/types/block-explorer-response.d.ts
```diff
@@ -1,3 +1,5 @@
+import type { UnknownObject } from '../../components/nested-data-list';
+
 export interface BlockExplorerTransactionResult {
```
notbot00, line 3: suggestion (non-blocking): realised this is causing the necessity for the type casting in each component. the way you could potentially resolve this is by making this type a generic, by doing something like:

```
type TransactionResultCommand = 
    | ValidatorHeartbeat
    | SubmitOrder
    | StateVariableProposal
    | AmendLiquidityProvisionOrder
    | BatchMarketInstructions
    | ChainEvent;
    
export type BlockExplorerTransactionResult<T extends TransactionResultCommand> = {
  ...
  command: T
}
```

and the way you'd use it on the component props would be like:

```
const ComponentProps1 = {
   transaction: BlockExplorerTransactionResult<ValidatorHeartbeat>
} 

const ComponentProps2 = {
   transaction: BlockExplorerTransactionResult<SubmitOrder>
} 
```

apps/explorer/src/app/components/txs/details/tx-batch.tsx
edd: Yeah the types in this PR are a little bit lame as they're manually written. I'm going to switch to types generated from the swagger file (https://raw.githubusercontent.com/vegaprotocol/documentation/main/specs/v0.62.1/blockexplorer.openapi.json) - I'll make a ticket for that ⌛

apps/explorer/src/app/components/txs/details/tx-lp-amend.tsx
edd: Will resolve in https://github.com/vegaprotocol/frontend-monorepo/issues/2176 - but the comment below telling me how to do it better is great regardless

apps/explorer/src/app/routes/types/block-explorer-response.d.ts
edd: Thanks for teaching me to typescript better.

I continued the pattern that was there, but realise while doing it that really generating the types from the source would be better, so I've created https://github.com/vegaprotocol/frontend-monorepo/issues/2176. Thanks for telling me how to do this better regardless!

apps/explorer/src/app/components/txs/details/tx-details-wrapper.tsx
edd: I agree and have done this

MadalinaRaicu (approved):

PR #2155, opened 2022-11-17
fix(#2144): fix positions table notional value and remove total row

MadalinaRaicu:


libs/positions/src/lib/positions-table.tsx
```diff
@@ -293,40 +267,45 @@ export const PositionsTable = forwardRef<AgGridReact, Props>(
...
         />
         <AgGridColumn
           headerName={t('Leverage')}
           field="currentLeverage"
           type="rightAligned"
           filter="agNumberColumnFilter"
-          cellRendererSelector={(
-            params: ICellRendererParams
-          ): CellRendererSelectorResult => {
+          cellRendererSelector={(): CellRendererSelectorResult => {
```
mattrussell36, line 306: suggestion (non blocking): we can just use the regula cellRenderer as we always render the PriceFlashCell. Same elsewhere here..

mattrussell36 (approved):

PR #2169, opened 2022-11-18
chore(2071): deal ticket data update fixing

### LINKED ISSUE #2071: Deal ticket doesn't receive latest market and account data
## Description

The deal ticket relies on some market and account data. For example market states and trading modes for validation, and account data for showing your balance and estimated margin requirements. However, it doesn't stay up to date via a subscription or otherwise.

## Steps to Reproduce

1. Go to the deal ticket with a public key that doesn't have any collateral
2. Try to submit and order and you will get a validation message saying 0 balance
3. Deposit some of the required asset
4. Try to place the order again and the you will still get the 0 balance validation message

## Additional context

This is likely because the deal ticket makes its own query for market state and doesn't use the market provider. And equally it uses a hook `usePartyBalanceQuery` rather than an accounts provider.

macqbat:
# Related issues 🔗

Closes #2071 

# Description ℹ️

Refactor deal ticket for less re-render in case when live accounts data update will affect its components

# Demo 📺

# Technical 👨‍🔧

Improve hooks to returning only possibly only simple types, move queries down to children for avoiding drilling down data, etc


libs/deal-ticket/src/components/deal-ticket/deal-ticket.spec.tsx
```diff
@@ -50,7 +50,7 @@ const market: DealTicketMarketFragment = {
       price: '100',
     },
   },
-};
+} as MarketDealTicket;
```
mattrussell36, line 53: suggestion (non blocking): why not just add a type to `market`? IE `const market: MarketDealTicket` that way we can be sure we aren't missing any fields in the mock.

apps/console-lite/src/app/components/deal-ticket/deal-ticket-balance.spec.tsx
```diff
@@ -1,13 +1,11 @@
...
-  AccountFragment,
-  DealTicketMarketFragment,
-} from '@vegaprotocol/deal-ticket';
+import type { AccountFragment } from '@vegaprotocol/deal-ticket';
 import { DealTicketBalance } from './deal-ticket-balance';
 import { Schema } from '@vegaprotocol/types';
+import type { MarketDealTicket } from '@vegaprotocol/market-list';
 
-const tDAI: DealTicketMarketFragment['tradableInstrument']['instrument']['product']['settlementAsset'] =
+const tDAI: MarketDealTicket['tradableInstrument']['instrument']['product']['settlementAsset'] =
```
mattrussell36, line 8: suggestion (non blocking): I think an type alias here would make this more readable, its could also be used elsewhere

 `type SettlementAsset = MarketDealTicket['tradableInstrument']['instrument']['product']['settlementAsset']`

libs/deal-ticket/src/hooks/use-order-margin-validation.ts
```diff
@@ -1,53 +1,43 @@
...
+  const { accountBalance, accountDecimals } = useAccountBalance(assetId);
+  const balance =
+    accountBalance && accountDecimals !== null
+      ? toBigNum(accountBalance, accountDecimals)
+      : toBigNum('0', assetDecimals);
   const margin = toBigNum(estMargin?.margin || 0, assetDecimals);
-  const asset = market.tradableInstrument.instrument.product.settlementAsset;
 
-  const memoizedValue = useMemo(() => {
+  // return only simple types (bool, string) for make memo sensible, no BigNumber please :-)
```
mattrussell36, line 32: good catch. My mistake 🙌

mattrussell36 (approved):

libs/deal-ticket/src/components/deal-ticket/deal-ticket.spec.tsx
mattrussell36: Oh presumably this is because lots of fields in the test are not required, so it keeps the test a bit more succinct

apps/console-lite/src/app/components/deal-ticket/deal-ticket-balance.spec.tsx
macqbat: ok, I created some

PR #2175, opened 2022-11-18
test: 2059 migrate trading e2e tests to use vegawallet v2

### LINKED ISSUE #2059: Migrate trading e2e tests to use vegawallet v2
## The Chore

We need to switch to vegawallet v2, tests need to be updated and verified if no issues are caused by using the new wallet

radoslaw-szpiech:
# Related issues 🔗

Closes #2059 #2195 #2196 #2202 .

# Description ℹ️

Migrating trading e2e tests to support vegawallet v2. All apps use shared connectVegaWallet function, that uses CLI connection with v2 vegawallet. Requires vegawallet-dummy to run, to avoid interaction with terminal.


apps/trading-e2e/src/support/deal-ticket-transaction.ts
```diff
@@ -0,0 +1,82 @@
...
+export type Order = {
+  marketId: string;
+  reference?: string;
+  type: Schema.OrderType;
+  side: Schema.Side;
+  timeInForce: Schema.OrderTimeInForce;
+  size: string;
+  price?: string;
+  expiresAt?: string;
+};
```
dexturr, line 21: Can we not import this from @vegaprotocol/types because of the JSX issue you ran into?

radoslaw-szpiech: No, more like there is no type like that. It is part of one of the interfaces.

MadalinaRaicu (approved):

mattrussell36 (approved):

PR #2178, opened 2022-11-18
fix: fixes around formatting numbers and arrange edit order dialog

MadalinaRaicu:
# Related issues 🔗

Closes #[Issue number here]

# Description ℹ️

- [x] normalize numbers (remove trailing 0s and fix wrong numbers in market info liquidity)
- [x] increase edit order col span dialog
- [x] rename "Propose a change to this market" as the current link does not direct you to a market-specific form

image
image


libs/market-info/src/components/market-info/info-key-value-table.tsx
```diff
@@ -40,8 +40,9 @@ const Row = ({
       return value;
     }
     if (decimalPlaces) {
-      return `${parseFloat(
```
MadalinaRaicu, line 43: this breaks market-info target stake and supplied stake values.

MadalinaRaicu: NOTE: I will handle the follow up on ALL number formatters :D https://github.com/vegaprotocol/frontend-monorepo/issues/2179

apps/trading/components/select-market/select-market-columns.tsx
```diff
@@ -554,7 +554,14 @@ export const columnsPositionMarkets = (
     {
       kind: ColumnKind.Position,
       value: (
-        <p className={signedNumberCssClass(openVolume || '')}>{openVolume}</p>
+        // TODO add decimals to open volume
```
mattrussell36, line 557: chore (blocking): this comment seems irrelevant now

mattrussell36 (changes requested):

apps/trading/components/select-market/select-market-columns.tsx
MadalinaRaicu: removed.

mattrussell36 (approved):

PR #2187, opened 2022-11-21
fix(#2154): duplicate position created

### LINKED ISSUE #2153: Incorrect massive liquidation estimate

### LINKED ISSUE #2154: [console] Duplicate position created for a few seconds
When you create an order on console it is creating a duplicate position row for a few seconds.

First pic is just after placing

Image

Second pic if a few seconds later

Image

If you navigate away and back again it fixes itself too.

bglownia:
# Related issues 🔗

Closes #2154


libs/react-helpers/src/lib/generic-data-provider.ts
```diff
@@ -646,7 +646,7 @@ function makeDerivedDataProviderInternal<
         const updatedPart = parts[updatedPartIndex];
         if (updatedPart.isUpdate) {
           isUpdate = true;
-          if (updatedPart.delta && combineDelta && data) {
```
edd, line 649: **nitpick(non-blocking):** this part is hard to evaluate without a test covering it, and obviously the change hasn't broken a test. It feels quite fragile.

I'll hit Approve but I feel like I'm taking your word for it :crossed_fingers:

bglownia: Condition with check if delta exists was not tested. It's removed so don't need to check it,

PR #2204, opened 2022-11-23
fix(#2200): handle NotFound GraphQL errors in async renderer

### LINKED ISSUE #2200: [console] NotFound error handling
Core has made API changes where not found resources return an error object (with type `NotFound`). Our error handling is often crudely just checking for the presence of an error object when we should also be checking the error type and handling it differently.

## Tasks

- [x] Update AsyncRenderer to handle NotFound errors
- [x] Review usage of errors returned from data providers, useQuery hooks, and useGeneratedQuery hooks and update error handling accordingly
- [x] Update apollo client error link to not report not found errors

MadalinaRaicu:
# Related issues 🔗

Closes #2200 

# Description ℹ️ &  Technical 👨‍🔧

- [x] handle NotFound GraphQL errors in the async renderer
- [x] handle NotFound GraphQL errors in certain queries / data providers
- [x] add method in apollo-client to spot NotFound errors but not filter them

# Demo 📺

Handles this issue observed in empty stagnet3: 

Screenshot 2022-11-22 at 10 02 00

Screenshot 2022-11-23 at 08 26 22

### DIFF
--- a/libs/apollo-client/src/lib/apollo-client.ts
+++ b/libs/apollo-client/src/lib/apollo-client.ts
@@ -64,7 +64,7 @@ export function createClient(base?: string, cacheConfig?: InMemoryCacheConfig) {
     : httpLink;
 
   const errorLink = onError(({ graphQLErrors, networkError }) => {
-    if (graphQLErrors && hasNotFoundGraphQLErrors(graphQLErrors)) {
+    if (graphQLErrors && !hasNotFoundGraphQLErrors(graphQLErrors)) {
       console.log(graphQLErrors);
     }
     if (networkError) {
--- a/libs/apollo-client/src/lib/apollo-client.ts
+++ b/libs/apollo-client/src/lib/apollo-client.ts
@@ -1,4 +1,4 @@
-import type { InMemoryCacheConfig } from '@apollo/client';
+import type { ApolloError, InMemoryCacheConfig } from '@apollo/client';
 import {
   ApolloClient,
   from,
@@ -15,6 +15,8 @@ import { RetryLink } from '@apollo/client/link/retry';
 import ApolloLinkTimeout from 'apollo-link-timeout';
 
 const isBrowser = typeof window !== 'undefined';
+
+const NOT_FOUND = 'NotFound';
 
 export function createClient(base?: string, cacheConfig?: InMemoryCacheConfig) {
   if (!base) {
@@ -61,8 +63,12 @@ export function createClient(base?: string, cacheConfig?: InMemoryCacheConfig) {
     : httpLink;
 
   const errorLink = onError(({ graphQLErrors, networkError }) => {
-    if (graphQLErrors) console.log(graphQLErrors);
-    if (networkError) console.log(networkError);
+    if (graphQLErrors) {
+      console.log(graphQLErrors);
+    }
+    if (networkError) {
+      console.log(networkError);
+    }
   });
 
   return new ApolloClient({
@@ -70,3 +76,16 @@ export function createClient(base?: string, cacheConfig?: InMemoryCacheConfig) {
     cache: new InMemoryCache(cacheConfig),
   });
 }
+
+export const isApolloGraphQLError = (
+  error: ApolloError | Error
+): error is ApolloError => {
+  return error && !!(error as ApolloError).graphQLErrors;
+};
+
+export const isNotFoundGraphQLError = (error: Error | ApolloError) => {
+  return (
+    isApolloGraphQLError(error) &&
+    error.graphQLErrors.some((e) => e.extensions?.type === NOT_FOUND)
+  );
+};
--- a/libs/apollo-client/src/lib/apollo-client.ts
+++ b/libs/apollo-client/src/lib/apollo-client.ts
@@ -13,6 +13,7 @@ import { createClient as createWSClient } from 'graphql-ws';
 import { onError } from '@apollo/client/link/error';
 import { RetryLink } from '@apollo/client/link/retry';
 import ApolloLinkTimeout from 'apollo-link-timeout';
+import type { GraphQLErrors } from '@apollo/client/errors';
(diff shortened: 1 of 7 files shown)


libs/positions/src/lib/positions-manager.tsx
```diff
@@ -25,7 +25,13 @@ export const PositionsManager = ({ partyId }: PositionsManagerProps) => {
 
   return (
     <>
-      <AsyncRenderer loading={loading} error={error} data={data}>
+      <AsyncRenderer
```
MadalinaRaicu, line 28: it was displaying `No data` and it was inconsistent with the other tabs. (regression)

MadalinaRaicu, line 28: it was displaying `No data` and it was inconsistent with the other tabs. (regression)

libs/ui-toolkit/src/components/splash/splash.tsx
```diff
@@ -7,7 +7,7 @@ export interface SplashProps {
 
 export const Splash = ({ children }: SplashProps) => {
   const splashClasses = classNames(
-    'w-full h-full',
+    'w-full h-full text-xs text-center text-gray-200',
```
MadalinaRaicu, line 10: aligning style of ag-grid overlay and splash classes

libs/fills/src/lib/fills-manager.tsx
```diff
@@ -29,7 +29,12 @@ export const FillsManager = ({ partyId }: FillsManagerProps) => {
   };
 
   return (
-    <AsyncRenderer loading={loading} error={error} data={data}>
+    <AsyncRenderer
```
MadalinaRaicu, line 32: allow overlay of ag-grid to be displayed.

libs/deal-ticket/src/components/deal-ticket/deal-ticket-market-amount.tsx
```diff
@@ -46,7 +46,7 @@ export const DealTicketMarketAmount = ({
       <div className="flex items-end gap-4 mb-2">
         <div className="flex-1 text-sm">Size</div>
         <div />
-        <div className="flex-1 text-sm text-right">
+        <div className="flex-2 text-sm text-right">
```
MadalinaRaicu, line 49: unwrap `Estimated uncrossing price`

libs/apollo-client/src/lib/apollo-client.ts
```diff
@@ -61,12 +64,34 @@ export function createClient(base?: string, cacheConfig?: InMemoryCacheConfig) {
...
+    }
   });
 
   return new ApolloClient({
     link: from([errorLink, timeoutLink, retryLink, splitLink]),
     cache: new InMemoryCache(cacheConfig),
   });
 }
+
+const isApolloGraphQLError = (
```
mattrussell36, line 81: question (non blocking): Why do we need this type guard?  `graphqlErrors` is already typed as `GraphqlErrors | undefined` so shouldn't a simple undefined check should suffice?

libs/apollo-client/src/lib/apollo-client.ts
```diff
@@ -61,12 +64,34 @@ export function createClient(base?: string, cacheConfig?: InMemoryCacheConfig) {
     : httpLink;
 
   const errorLink = onError(({ graphQLErrors, networkError }) => {
-    if (graphQLErrors) console.log(graphQLErrors);
-    if (networkError) console.log(networkError);
+    if (graphQLErrors && !hasNotFoundGraphQLErrors(graphQLErrors)) {
```
mattrussell36, line 67: chore (blocking): If graphqlErrors contains a mix of NotFound errors and legitimate errors this check will now prevent the legitimate errors from being logged. Can we change this so we filter out any NotFound errors and still log any remaining errors?

libs/react-helpers/src/hooks/use-data-provider.ts
```diff
@@ -85,7 +86,7 @@ export const useDataProvider = <
         isInsert,
         isUpdate,
       } = args;
-      setError(error);
+      setError(isNotFoundGraphQLError(error) ? undefined : error);
```
mattrussell36, line 89: chore (blocking): I'd revert this, the data provider should still return any error it sees, it should be up to the application/component to decide how to handle it

libs/ui-toolkit/src/components/splash/splash.tsx
mattrussell36: Have you checked both light and dark themes here?

libs/deal-ticket/src/components/deal-ticket/deal-ticket-market-amount.tsx
mattrussell36: To me it looks nicer if the text is contained to an equal portion of the flex container, even if it wraps. If there is enough space it will unwrap.

Below the text flowing outside of the 'grid' and into the left sides space looks off to me as it disrupts the alignment the inputs below

Screen Shot 2022-11-23 at 19 21 11

mattrussell36 (changes requested):

libs/apollo-client/src/lib/apollo-client.ts
MadalinaRaicu: won't compile unless `error is ApolloError`

libs/apollo-client/src/lib/apollo-client.ts
MadalinaRaicu: sure, updated in last commit.

libs/react-helpers/src/hooks/use-data-provider.ts
MadalinaRaicu: agree!!! updated in last commit.

mattrussell36 (approved):

PR #2214, opened 2022-11-23
feat: transactions store

### LINKED ISSUE #1679: Vega transaction toasts
## Story

**As a** user
**I want** vega transactions to be shown as toasts
**So that** I can continue to use the app whilst waiting for a tx to complete

## Acceptance Criteria

- [x] I am prompted via dialog to complete the transaction in wallet
- [ ] I am immediately shown a rejected transaction (when rejected by wallet immediately)
- [x] If the transaction was accepted by the wallet, I can still use the application and am not blocked by an overlay
- [x] If the transaction is accepted I am shown a toast with the tx hash, indicating its pending
- [ ] I can execute multiple transactions one after the other
- [ ] When a transaction result is received, notify state by toast

## Tasks

- [ ] Create transaction store
- [ ] Migrate to using toasts
  - [ ] https://github.com/vegaprotocol/frontend-monorepo/issues/2271
  - [ ] https://github.com/vegaprotocol/frontend-monorepo/issues/2270
  - [ ] Withdrawal creation issue TBD

## Additional information

- We may want to create a new hook/store for this. Say a useVegaTransactionV2 to make development and migration to the new pattern easier and more incremental

bglownia:
# Related issues 🔗

Closes #1679 

# Demo 📺

https://user-images.githubusercontent.com/1980305/206734301-b4b953e1-2254-43b2-a1ca-5d731eb84ce6.mov


libs/wallet/src/TransactionUpdate.graphql
```diff
@@ -0,0 +1,93 @@
...
+}
+
+subscription TransactionUpdate($partyId: ID!) {
+  busEvents(
+    partyId: $partyId
+    batchSize: 0
+    types: [Withdrawal, Deposit, TransactionResult, Order]
+  ) {
+    event {
+      ... on Deposit {
```
mattrussell36, line 79: There are other transactions that other apps will depend on (eg. proposals on the governance app). It seems a little heavy handed to force an app to subscribe to all types of updates, even if it doesn't care about them.

Can we decouple this in someway that allows the app to specify which types of transaction to listen for?

libs/web3/src/lib/use-ethereum-transaction-store.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+export const useEthTransactionManager = () => {
+  const update = useEthTransactionStore((state) => state.update);
+  const transaction = useEthTransactionStore((state) =>
+    state.transactions.find(
+      (transaction) => transaction?.status === EthTxStatus.Default
+    )
+  );
+
+  if (transaction) {
```
mattrussell36, line 95: I mentioned before, but I don't love triggering a tx off the back of a state change. We should be able to trigger it from an event within the app (IE onClick)

libs/wallet/src/TransactionUpdate.graphql
```diff
@@ -0,0 +1,93 @@
...
+    positionDecimalPlaces
+    tradableInstrument {
+      instrument {
+        name
+      }
+    }
+  }
+}
+
+fragment DepositEvent on Deposit {
```
mattrussell36, line 58: Deposit events will be emitted off the back of an Ethereum transaction. Are we intending to share these for both the Ethereum transaction store and the Vega transaction store?

asiaznik:
NEW PR: https://github.com/vegaprotocol/frontend-monorepo/pull/2382

Issue #2190, opened 2022-11-22
Link to markets in listing to Console's liquidity view for secondary information


pennyandrews:
## Story

As a potential Liquidity Provider
I want to be able to see the current spread of liquidity, fee levels, margin requirement and LP share for a specific market listed in the dashboard 
So that I can understand the spread of liquidity provided right now and where I might fit in

## Acceptance Criteria

- [ ] I can see some sort of visual indication that it is possible to click on a market listed in the dashboard to see further details
- [ ] When I click on a row for a market listed in the dashboard, I am taken to the Console Market Liquidity view for that market
- [ ] The Console Market Liquidity view opens in a new tab
- [ ] The Console Market Liquidity view that opens needs to be relative to the network you are viewing in the dashboard i.e. If you're on Fairground testnet then the link needs to take you to Console on Fairground testnet, if you are on mainnet then the link needs to take you to Console on mainnet*
 
_*This A/C needs some discussion as we haven't implemented the network switcher yet for LP dashboard AND Console isn't currently available on mainnet as trading hasn't been enabled yet._

_So we may need to kick this A/C into its own ticket or add it to the existing Network switcher ticket. However, I wanted to include it so that we are thinking about it when we investigate the technical implementation._

## Tasks

- [ ] Investigate options for how we can achieve this linking to Console's Market Liquidity view for each market we list in the dashboard
- [ ] Discuss and agree approach with the team and stakeholder group
- [ ] Implement
- [ ] Test

## Additional details / background info

This is an example Liquidity page for a market in Console https://console.fairground.wtf/#/liquidity/acff1b95980c3355e65c7162c35917388702525c0475ba753f03d0efc9c6d6a3

<img width="2560" alt="Screenshot 2022-11-22 at 10 09 24" src="https://user-images.githubusercontent.com/47143694/203286789-01e79c86-d361-409e-b98e-889204a66654.png">

In case you need to go and find more examples, you can get to this view from the market view here:

<img width="1266" alt="Screenshot 2022-11-22 at 10 00 18" src="https://user-images.githubusercontent.com/47143694/203285010-66a25fc1-8a20-438d-9c76-e6f05b720735.png">

ciaran-:
For network switching, I manually changed the network config locally for the lp-dashboard build to `stagnet3` and verified that the link generated was stagnet 3 and the correct market id. 

I think once we implement a switcher this should 'just work' 🤞 

[...]

PR #2225, opened 2022-11-24
feat(#2190): replace lp dashboard details links with console links 

ciaran-:
# Related issues 🔗

Part of  #2190

# Description ℹ️

Replace links to the lp-dashboard details page with links to the Console liquidity page for a market. This already has a bunch of the information we're trying to display.

# Demo 📺

https://www.loom.com/share/21c744227eab420db23772b7d25d71df

# Technical 👨‍🔧

I'm re-using the `useLinks` hook from the environment library, which was super useful.

### DIFF
--- a/apps/liquidity-provision-dashboard/src/app/components/dashboard/market-list/market-list.tsx
+++ b/apps/liquidity-provision-dashboard/src/app/components/dashboard/market-list/market-list.tsx
@@ -1,5 +1,4 @@
 import { useCallback, useState } from 'react';
-import { useNavigate } from 'react-router-dom';
 import { AgGridColumn } from 'ag-grid-react';
 import type {
   ValueFormatterParams,
@@ -26,6 +25,7 @@ import {
   displayChange,
 } from '@vegaprotocol/liquidity';
 import type { Schema } from '@vegaprotocol/types';
+import { DApp, useLinks } from '@vegaprotocol/environment';
 
 import { HealthBar } from '../../health-bar';
 import { Grid } from '../../grid';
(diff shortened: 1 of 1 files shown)


apps/liquidity-provision-dashboard/src/app/components/dashboard/market-list/market-list.tsx
```diff
@@ -50,7 +50,11 @@ export const MarketList = () => {
         <Grid
           gridOptions={{
             onRowClicked: ({ data }: RowClickedEvent) => {
-              navigate(`/markets/${data.id}`);
+              window.open(
```
MadalinaRaicu, line 53: why can't we use `navigate` ?

ciaran-: Good question!

My understanding is that `useRouter` and thus `navigate` are designed for _internal_ routing changes, which is why I previously added them here to handle navigating to the details page. Now that we're replacing the _internal_ details page with a link to the _external_ console liquidity details page - it didn't make sense to me to try and use `useRouter`/`navigate` - my understanding was that external links were unsupported behaviour (even if they might work) ?

ciaran-: @MadalinaRaicu does this make sense? What do you reckon - stick with this commit or try and make it work with `navigate`/`useRouter` ?

MadalinaRaicu (approved):

PR #2228, opened 2022-11-24
fix: deal ticket set market price

### LINKED ISSUE #2071: Deal ticket doesn't receive latest market and account data

### LINKED ISSUE #2137: Deal ticket does not update limit and market price anymore
It is only on develop branch now.

Deal ticket does not update the price anymore when switching from Limit to Market. 

1. Switch to Limit
2. Populate with a different price than the market price
3. Switch back to Market and you can see the estimates do not update
4. If you refresh, limit price is saved in local storage so estimates still don't update. 

This would be a regression bug. 

Image

Image

MadalinaRaicu:
# Related issues 🔗

Closes #2137 

# Description ℹ️

Deak ticket price and estimates do not update on the order type switch. 

# Demo 📺

image

# Technical 👨‍🔧


libs/deal-ticket/src/components/deal-ticket/deal-ticket.tsx
```diff
@@ -63,6 +65,14 @@ export const DealTicket = ({
   const order = watch();
   // When order state changes persist it in local storage
   useEffect(() => setPersistedOrder(order), [order, setPersistedOrder]);
+
+  // When switch back to market order, set price to market price
+  useEffect(() => {
```
mattrussell36, line 70: chore (blocking): can you show me why we need this? The order estimate query now gets the correct price directly from market.data.markPrice or market.data.indicativePrice (or order.price if limit order).

This also results in loss of state of your user input limit order price see below:

Uploading Nov-24-2022 16-13-07.mp4…

mattrussell36 (changes requested):

libs/deal-ticket/src/components/deal-ticket/deal-ticket.tsx
mattrussell36: I see where the issue is. Will push a fix

mattrussell36 (approved):

PR #2235, opened 2022-11-25
fix: orderbook decimal places issue 

### LINKED ISSUE #2232: Zero selector on order book bug
The zero selector on the order book seems broken. 

Just shows 0.00 many times even though selecting them does change the precision of the orders on the page.

This is new ... sure it was working not long ago.

Image

MadalinaRaicu:
# Related issues 🔗

Closes #2232 

# Description ℹ️

Fix for orderbook dp that was using formatters.

# Demo 📺

image

# Technical 👨‍🔧


libs/positions/src/lib/positions-table.spec.tsx
```diff
@@ -162,7 +162,7 @@ it('displays allocated margin', async () => {
   });
   const cells = screen.getAllByRole('gridcell');
   const cell = cells[8];
-  expect(cell.textContent).toEqual('123,456.00');
+  expect(cell.textContent).toEqual('123,456');
```
MadalinaRaicu, line 165:
```suggestion
  expect(cell.textContent).toEqual('123,456.00');
```

PR #2244, opened 2022-11-28
fix(explorer): remove old chain api

edd:
Removes env vars containing the old chain explorer API and switches
the one component & call to the old explorer with a new BE endpoint.

Blocked by https://github.com/vegaprotocol/vega/pull/6958/files being
unmerged & undeployed.

# Related issues 🔗

Closes #[Issue number here]

# Description ℹ️

# Demo 📺

# Technical 👨‍🔧


apps/explorer/src/app/components/txs/txs-per-block.tsx
```diff
@@ -1,37 +1,32 @@
...
-      block_height: parseInt(blockHeight!),
-      node_url: `${DATA_SOURCES.tendermintUrl}/`,
-    }),
-  });
-
+    state: { data, loading, error },
+  } = useFetch<BlockExplorerTransactions>(
+    `${
+      DATA_SOURCES.blockExplorerUrl
+    }/transactions/before=${blockHeight.toString()}.0&limit=${txCount}`
```
notbot00, line 25: question (non-blocking): out of curiosity, what's that `.0` at the end of the before param's value?

apps/explorer/src/app/hooks/use-txs-data.ts
```diff
@@ -29,11 +29,13 @@ export const getTxsDataUrl = ({ limit, filters }: IGetTxsDataUrl) => {
     url.searchParams.append('limit', limit);
   }
 
+  // Hacky fix for param as array
+  let urlAsString = url.toString();
```
notbot00, line 33: question (non-blocking): again, just out of curiosity, what do you get when you do `url.searchParams.append('filters', filters)` and then do `url.toSting()`?

notbot00 (approved):

apps/explorer/src/app/components/txs/txs-per-block.tsx
edd: Proper answer: What's being passed *looks* like a block height, but it isn't - it's a cursor. `.0` is the index of the first item in the block.

Better answer: It won't work without it

apps/explorer/src/app/hooks/use-txs-data.ts
edd: If you do that, you get `...transactions?limit=10&filters=filters%5Btx.submitter%5D%3Dbd6339d2428c79ac3bc9011771236d17bac92bcb1806423388d52fb440043aef`

What I lazily did not try is removing the `filters=` from the value and seeing if it was happy to append the `[tx.submitter=blah]` array item as is. I will try it out in a follow up.

apps/explorer/src/app/components/txs/txs-per-block.tsx
notbot00: makes more sense, thank you!

PR #2267, opened 2022-11-29
feat(#1853): ledger entries date filtering

### LINKED ISSUE #1853: Ledger entries date filtering
## Story

**As a** trader
**I want** to view ledger entriese for a specific time period
**So that** I can inspect how my collateral was used more closely

## Acceptance Criteria

- [ ] I can filter by start date
- [ ] I can filter by end date

## Additional info

We might be able to leverage AgGrid's built in agDateColumnFilter

Screen Shot 2022-10-25 at 22 50 57

MadalinaRaicu:
# Related issues 🔗

Closes #1853

# Description ℹ️

Add date range filter and update ledger entries on update.

# Demo 📺

image

# Technical 👨‍🔧


libs/react-helpers/src/lib/grid/date-range-filter.tsx
```diff
@@ -84,16 +85,18 @@ export const DateRangeFilter = forwardRef((props: IFilterParams, ref) => {
   };
 
   useEffect(() => {
-    props.filterChangedCallback();
-  }, [value]); //eslint-disable-line react-hooks/exhaustive-deps
+    if (props && props.filterChangedCallback) {
```
mattrussell36, line 88: suggestion (non blocking): call this directly in the onChange callback? I see no value in using useEffect here

libs/react-helpers/src/lib/grid/date-range-filter.tsx
```diff
@@ -84,16 +85,18 @@ export const DateRangeFilter = forwardRef((props: IFilterParams, ref) => {
...
+      props?.filterChangedCallback();
+    }
+  }, [value, props]);
 
   const start = (value.start && toInputValue(value.start)) || '';
```
mattrussell36, line 93: chore (non blocking): I know this was already there but it looks like the `toInputValue` function has the same purpose as `formatForInput` in format/date.ts. Can we use that?

mattrussell36 (changes requested):
Everytime the filter changes the fetch request for ledger entries is made twice. I've also left some further comments

Screenshot 2022-12-01 at 15 55 37
Screenshot 2022-12-01 at 15 55 29

libs/ledger/src/lib/ledger-entries-data-provider.ts
```diff
@@ -51,28 +110,119 @@ const ledgerEntriesOnlyProvider = makeDataProvider<
...
+}: Props) => {
+  const dataRef = useRef<(AggregatedLedgerEntriesEdge | null)[] | null>(null);
+  const totalCountRef = useRef<number>();
+  const newRows = useRef(0);
+
+  const variables = useMemo<LedgerEntriesQueryVariables>(
+    () => ({
+      partyId,
+      dateRange: filter?.vegaTime?.value,
+      pagination: { first: !filter?.vegaTime?.value ? 100 : undefined },
```
mattrussell36, line 154: suggestion (non blocking): I think it would make more sense to have a default filter applied, say a range of one week ago to present, rather than an arbitrary number. This is because you are essentially applying a filter to the user but it won't be represented in the UI anywhere

MadalinaRaicu: I would rather leave the pagination filter unset for now. Even if we add a date range we would still be applying a filter. 
also in orders provider we are displaying first:100 and I am not sure what was the reasoning behind that. perhaps @bglownia knows more on how this works on reload

libs/react-helpers/src/lib/grid/date-range-filter.tsx
MadalinaRaicu: this has been taken from the `ag-grid` documentation example. I can easily remove it and put it in onChange, but there are some errors happening on HMR and dev server that prevent me from properly testing how removing this would work.

MadalinaRaicu: updated.

MadalinaRaicu:
> Everytime the filter changes the fetch request for ledger entries is made twice. I've also left some further comments
> 
> Screenshot 2022-12-01 at 15 55 37 Screenshot 2022-12-01 at 15 55 29

 thank you @bglownia for fixing the double fetch request

libs/react-helpers/src/lib/grid/date-range-filter.tsx
MadalinaRaicu: actually, I need to put this back in the useEffect. if this is called only onChange it will break the reset functionality. filterChangedCallback should be called also when props get updated (not only value)

MadalinaRaicu: we are just following ag-grid documentation example

mattrussell36 (approved):

PR #2289, opened 2022-11-30
feat(explorer): cancel and order transaction view

### LINKED ISSUE #2010: OrderSubmission transaction
# Tasks

- [ ] sha-256 signature to get id
- [ ] fetch order details
- [ ] render order details

Acceptance Criteria defined in the EPIC https://github.com/vegaprotocol/frontend-monorepo/issues/2082

### LINKED ISSUE #2108: OrderCancellation transaction
Acceptance Criteria defined in the EPIC https://github.com/vegaprotocol/frontend-monorepo/issues/2082

edd:
# Related issues 🔗

Closes #2010. Closes #2108.

# Description ℹ️

Adds an order details view for Order related transactions. Which also meant:
- [x] Adding a `code` component that renders a ✅ for non-rejected TXs and ❌ for others
- [x] Adding a `PriceForMarket` component that fetches the correct asset & decimal places
- [x] Rendering Order Details view (show latest version) on TX: Cancel Order page
- [x] Rendering Order Details view (show first version) on TX: Order Submit page

And that tangentially meant:
- [x] Adding block hash and a few other details to the block ID page
- [x] Updating the transaction list to show the `code` in place of the Index
- [x] Updating most hash fields to use `<code></code>` to get monospaced
- [x] Moving transaction type in to the Shared Details header

# Demo 📺

## Submit order
Screenshot 2022-11-30 at 17 19 28

Avoids implying details that may not have been true at creation (remaining/status for example)

## Cancel order
Screenshot 2022-11-30 at 17 19 19

Shows the last known details (remaining/status/version)

## Failed order
Screenshot 2022-11-30 at 17 19 11

Shows the order as it would have been, without implying it was created

### DIFF
--- a/apps/explorer/src/app/components/deterministic-order-details/deterministic-order-details.tsx
+++ b/apps/explorer/src/app/components/deterministic-order-details/deterministic-order-details.tsx
@@ -2,6 +2,8 @@ import { t } from '@vegaprotocol/react-helpers';
 import { useExplorerDeterministicOrderQuery } from './__generated__/Order';
 import type { Schema } from '@vegaprotocol/types';
 import { MarketLink } from '../links';
+import PriceInMarket from '../price-in-market/price-in-market';
+import { Time } from '../time';
 
 export interface DeterministicOrderDetailsProps {
   id: string;
@@ -24,6 +26,23 @@ const sideText: Record<Schema.Side, string> = {
   SIDE_SELL: 'Sell',
 };
 
+const tifShort: Record<Schema.OrderTimeInForce, string> = {
+  TIME_IN_FORCE_FOK: 'FOK',
+  TIME_IN_FORCE_GFA: 'GFA',
+  TIME_IN_FORCE_GFN: 'GFN',
+  TIME_IN_FORCE_GTC: 'GTC',
+  TIME_IN_FORCE_GTT: 'GTT',
+  TIME_IN_FORCE_IOC: 'IOC',
+};
+
+const tifFull: Record<Schema.OrderTimeInForce, string> = {
+  TIME_IN_FORCE_FOK: 'Fill or Kill',
+  TIME_IN_FORCE_GFA: 'Good for Auction',
+  TIME_IN_FORCE_GFN: 'Good for Normal',
+  TIME_IN_FORCE_GTC: "Good 'til Cancel",
+  TIME_IN_FORCE_GTT: "Good 'til Time",
+  TIME_IN_FORCE_IOC: 'Immediate or Cancel',
+};
 const wrapperClasses =
   'grid lg:grid-cols-1 flex items-center max-w-xl border border-zinc-200 dark:border-zinc-800 rounded-md pv-2 ph-5 mb-5';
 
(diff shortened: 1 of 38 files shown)


apps/explorer/src/app/components/deterministic-order-details/deterministic-order-details.tsx
```diff
@@ -0,0 +1,157 @@
...
+  TIME_IN_FORCE_FOK: 'FOK',
+  TIME_IN_FORCE_GFA: 'GFA',
+  TIME_IN_FORCE_GFN: 'GFN',
+  TIME_IN_FORCE_GTC: 'GTC',
+  TIME_IN_FORCE_GTT: 'GTT',
+  TIME_IN_FORCE_IOC: 'IOC',
+};
+
+const tifFull: Record<Schema.OrderTimeInForce, string> = {
+  TIME_IN_FORCE_FOK: 'Fill or Kill',
```
notbot00, line 41: comment (non-blocking i think?): i think we ideally want these values wrapped into the dummy translate function

edd: Good point!

apps/explorer/src/app/components/price-in-market/price-in-market.spec.tsx
```diff
@@ -0,0 +1,92 @@
+import { MemoryRouter } from 'react-router-dom';
+import { MockedProvider } from '@apollo/client/testing';
+import type { MockedResponse } from '@apollo/client/testing';
+import { render } from '@testing-library/react';
+import PriceInMarket from './price-in-market';
+import { ExplorerMarketDocument } from '../links/market-link/__generated__/Market';
+
+function renderComponent(
```
notbot00, line 8: comment (non-blocking): i think this can just be a react component, if you adjust the props to be one object and use an upper case naming

apps/explorer/src/app/components/txs/details/chain-response-code/chain-reponse.code.tsx
```diff
@@ -0,0 +1,44 @@
...
+ * have no value outside the Vega chain and should appear only
+ * on Test networks.
+ */
+export const ChainResponseCode = ({
+  code,
+  hideLabel = false,
+}: ChainResponseCodeProps) => {
+  const isError = successCodes.has(code);
+
+  const icon = isError ? '✅' : '❌';
```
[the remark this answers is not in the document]
edd: Urgh, yes - I flipped it. Good catch 👍

apps/explorer/src/app/components/txs/details/tx-order-cancel.tsx
```diff
@@ -0,0 +1,49 @@
...
+        />
+        <TableRow modifier="bordered">
+          <TableCell>{t('Market')}</TableCell>
+          <TableCell>
+            <MarketLink id={marketId} />
+          </TableCell>
+        </TableRow>
+      </TableWithTbody>
+
+      {orderId.length > 0 ? <DeterministicOrderDetails id={orderId} /> : null}
```
notbot00, line 46: question (non-blocking): hmm, if the orderId defaults to `'-'`, do we need to check for its length? not sure if you intentionally wanted to pass in the dash here

apps/explorer/src/app/components/txs/lib/deterministic-ids.ts
```diff
@@ -0,0 +1,39 @@
...
+ * Encodes a string as bytes
+ * @param hex
+ * @returns number[]
+ */
+export function hexToString(hex: string) {
+  if (!hex.match(/^[0-9a-fA-F]+$/)) {
+    throw new Error('is not a hex string.');
+  }
+  if (hex.length % 2 !== 0) {
+    hex = '0' + hex;
```
notbot00, line 13: comment (blocking?): i think it's generally considered a bad practice to silently modify a function's input parameter. maybe you could create a new variable, something like: 
```
const normalizedHex = hex.length % 2 !== 0 ? `0${hex}` : hex;
```

edd: Yes - I will update this too

apps/explorer/src/app/components/txs/tx-order-type.tsx
```diff
@@ -36,6 +37,31 @@ const displayString: StringMap = {
...
 
+/**
+ * Given a proposal, will return a specific label
+ * @param chainEvent
+ * @returns
+ */
+export function getLabelForProposal(
+  proposal: components['schemas']['v1ProposalSubmission']
+): string {
+  if (proposal.terms?.newAsset) {
```
notbot00, line 48: comment (non-blocking): as a real sucker for `switch`, i'd like to take the opportunity to promote my favourite, 🥁  the switch statement! it's real and it's free, not a pyramid scheme! (couldn't make it rhyme properly 😢 )

notbot00: yeah, actually as pointed out by you in the chat, the if checks are targeting different properties of the same object. false marketing from my side.

edd: Thanks for coming back here. I asked directly because I thought you were going to pull out some ingenious switch usage that I had missed. I'm sure it's possible 😄

notbot00 (approved):

PR #2354, opened 2022-12-07
Feat 2089 second level validator table mouseovers

### LINKED ISSUE #2089: Second level data for validators (mouseovers)
## The Chore

## Tasks

- [ ] and then what?
- [ ] Etc.

sam-keen:
# Related issues 🔗

Closes #2089

# Description ℹ️

Provides extra validator data displayed in tooltips. Brings tooltip colouring in line with recent designs

# Demo 📺

Screenshot 2022-12-07 at 17 15 34


apps/token/src/routes/staking/shared.ts
```diff
@@ -13,6 +14,67 @@ export const rawValidatorScore = (
...
+  totalStake: string
+) => {
+  const totalPenaltiesCalc =
+    rawValidatorScore !== null
+      ? 100 *
+        Math.max(
+          0,
+          1 -
+            (Number(performanceScore) * Number(rawValidatorScore)) /
+              (Number(stakedOnNode) / Number(totalStake || 0))
```
dexturr, line 75: possible division by 0 error here no?

apps/token/src/routes/staking/shared.ts
```diff
@@ -13,6 +14,67 @@ export const rawValidatorScore = (
...
+    : null;
+
+export const getFormattedPerformanceScore = (performanceScore: string) =>
+  new BigNumber(performanceScore).dp(2);
+
+export const getPerformancePenalty = (performanceScore: string) =>
+  new BigNumber(1)
+    .minus(getFormattedPerformanceScore(performanceScore))
+    .times(100)
+    .toString() + '%';
```
dexturr, line 35: instead of adding a percent to strings we should be using the formatPercentage helper (which I think exists. If it doesn't then it should)

apps/token/src/routes/staking/home/validator-tables/standby-pending-validators-table.tsx
```diff
@@ -116,12 +168,16 @@ export const StandbyPendingValidatorsTable = ({
...
           headerTooltip: t('StakeDescription').toString(),
+          cellRenderer: TotalStakeRenderer,
           width: 120,
         },
         {
           field: ValidatorFields.STAKE_NEEDED_FOR_PROMOTION,
           headerName: t(ValidatorFields.STAKE_NEEDED_FOR_PROMOTION).toString(),
-          headerTooltip: stakeNeededForPromotionDescription,
+          headerTooltip: t(stakeNeededForPromotionDescription, {
+            prefix: 'The',
```
dexturr, line 178: ?
```suggestion
            prefix: t('The'),
```

dexturr (approved):

PR #2362, opened 2022-12-08
chore: cleanup type gen

### LINKED ISSUE #1434: Finish type migration

notbot00:
# Related issues 🔗

Closes #1434 

# Description ℹ️

Clean up the remaining generator config, global types, etc from the deprecated apollo type generator.


libs/types/src/global-types-mappings.ts
```diff
@@ -432,6 +433,6 @@ export const DescriptionTransferTypeMapping: TransferTypeMap = {
   TRANSFER_TYPE_TRANSFER_FUNDS_DISTRIBUTE: `Funds added to your general account to fulfil a transfer`,
   TRANSFER_TYPE_CLEAR_ACCOUNT: `Market-related accounts emptied, and balances moved, because the market has closed`,
   TRANSFER_TYPE_UNSPECIFIED: 'Default value, always invalid',
-  TRANSFER_TYPE_CHECKPOINT_BALANCE_RESTORE:
-    'When the network is restored from a checkpoint this sets the balances of parties',
+  TRANSFER_TYPE_CLOSE: '???',
```
notbot00, line 436: `TRANSFER_TYPE_CHECKPOINT_BALANCE_RESTORE` got removed from the gql schema, but these two new ones instead. not sure what the description should be here, please leave some tips here.

edd (approved):
🤞 Have to take your word on this one. Thanks for doing the boring & tedious task.

libs/types/src/global-types-mappings.ts
notbot00: makes sense, thank you! stagnet3 was apparently not up to date, regenerated against stagnet1 and got the correct result 🙏

PR #2365, opened 2022-12-08
feat(2356): use validator performance score from previous epoch

### LINKED ISSUE #2356: Display previous epoch's validator performance score instead of current epoch
We display the current epoch's validator performance score at the moment. This isn't particularly useful, as it is the last epoch's performance score that influences the penalties affecting the current epoch. This leads to confusing scenarios, for example, a validator may be showing that it has penalties, but in this epoch, its performance score is perfect, so it's not clear where the penalties are coming from. If we display the previous epoch's performance score instead, the penalty amount becomes clearer.

- [ ] Display previous epoch performance data in the validator table
- [ ] Display previous epoch performance data in the validators table tooltip

sam-keen:
# Related issues 🔗

Closes #2356

### DIFF
--- a/apps/token/src/routes/staking/shared.ts
+++ b/apps/token/src/routes/staking/shared.ts
@@ -73,7 +73,9 @@ export const getTotalPenalties = (
   totalStake: string
 ) => {
   const calc =
-    rawValidatorScore && performanceScore
+    rawValidatorScore &&
+    performanceScore &&
+    new BigNumber(totalStake).isGreaterThan(0)
       ? new BigNumber(1).minus(
           new BigNumber(performanceScore)
             .times(new BigNumber(rawValidatorScore))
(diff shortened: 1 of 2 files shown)


apps/token/src/routes/staking/shared.ts
```diff
@@ -64,19 +68,20 @@ export const getOverstakingPenalty = (
...
-          )
-      )
-    : new BigNumber(0);
+  const calc =
+    rawValidatorScore && performanceScore
+      ? new BigNumber(1).minus(
+          new BigNumber(performanceScore)
+            .times(new BigNumber(rawValidatorScore))
+            .dividedBy(
+              new BigNumber(stakedOnNode).dividedBy(new BigNumber(totalStake))
```
dexturr, line 81: nitpick: possible division by zero error, i'd add another condition to line 76

dexturr (approved):

PR #2382, opened 2022-12-09
feat: transaction store and toasts

### LINKED ISSUE #1679: Vega transaction toasts

asiaznik:
# Related issues 🔗

Closes #1679 

# Demo 📺

https://user-images.githubusercontent.com/1980305/206734301-b4b953e1-2254-43b2-a1ca-5d731eb84ce6.mov


MadalinaRaicu (approved):

libs/wallet/src/use-vega-transaction-store.tsx
```diff
@@ -0,0 +1,168 @@
...
+      const transaction: VegaStoredTxState = {
+        id: transactions.length,
+        createdAt: now,
+        updatedAt: now,
+        body,
+        error: null,
+        txHash: null,
+        signature: null,
+        status: VegaTxStatus.Requested,
+        dialogOpen: true,
```
mattrussell36, line 67: question (non blocking): why do we need to track dialogOpen state?

libs/withdraws/src/lib/create-withdrawal-dialog.tsx
```diff
@@ -0,0 +1,45 @@
...
+    (state) => state.isOpen
+  );
+  return (
+    <Dialog
+      title={t('Withdraw')}
+      open={isOpen && !connectWalletDialogIsOpen}
+      onChange={(isOpen) => (isOpen ? open() : close())}
+      size="small"
+    >
+      <Web3Container connectEagerly>
```
mattrussell36, line 23: chore (blocking): No need to use Web3Container here anymore

libs/withdraws/src/lib/pending-withdrawals-table.tsx
```diff
@@ -13,138 +13,107 @@ import type {
...
+        >) => {
+          return value ? getDateTimeFormat().format(new Date(value)) : '';
+        }}
+      />
+      <AgGridColumn
+        headerName=""
+        field="status"
+        flex={2}
+        cellRendererParams={{
+          complete: async (withdrawal: WithdrawalFieldsFragment) => {
```
mattrussell36, line 110: chore (non blocking): unnecessary async

libs/withdraws/src/lib/withdrawal-feedback.tsx
```diff
@@ -15,6 +15,25 @@ import { useWeb3React } from '@web3-react/core';
 import { formatDistanceToNow } from 'date-fns';
 import type { WithdrawalFieldsFragment } from './__generated__/Withdrawal';
 
+/*
```
mattrussell36, line 18: chore (non blocking): can we clear this out if its not needed

mattrussell36 (changes requested):
Big chunk of work here! My main blocking chain is the use of Web3Containers everywhere, which shouldn't be required anymore as we now have the Web3ReactProvider wrapping the entire app

mattrussell36:
Just noticed this z-index issue with the order book header

Screenshot 2022-12-14 at 21 47 01

libs/wallet/src/use-vega-transaction-store.tsx
bglownia: dialogOpen indicates it dialog connected with stored transaction should be visible to the user. It controls with toast notification is open. Will be set to false if user will dismiss it. 
It was not renamed to keep a consistency.

mattrussell36 (approved):

PR #2388, opened 2022-12-12
feat(2326): style Token validators

### LINKED ISSUE #2326: Implement validator table designs
Validator tables have been implemented with the minimum of visual change, just adhering to the UX wireframes. Now that the designs are completed, revisit these tables and make them more consistent. 

- [ ]  Ensure anything reusable is implemented via the ui-toolkit

sam-keen:
# Related issues 🔗

Closes #2326

# Description ℹ️

Makes validator list and individual validator pages adhere to latest designs. Includes pulling in the 'reveal top validators' cell into the validators AG-grid table, and reusable visual components that have also been applied to various places across the app.

# Demo 📺

Screenshot 2022-12-12 at 22 11 31
Screenshot 2022-12-12 at 22 11 40
Screenshot 2022-12-12 at 22 11 52
Screenshot 2022-12-12 at 22 12 03
Screenshot 2022-12-12 at 22 46 59
Screenshot 2022-12-12 at 22 52 33


libs/react-helpers/src/lib/format/number.tsx
```diff
@@ -106,6 +106,9 @@ export const formatNumberPercentage = (value: BigNumber, decimals?: number) => {
   return `${formatNumber(value, decimalPlaces)}%`;
 };
 
+export const formatPercentageToDecimal = (percentage: string) =>
```
dexturr, line 109: don't think this is used?

libs/ui-toolkit/src/components/rounded-wrapper/rounded-wrapper.tsx
```diff
@@ -0,0 +1,23 @@
+import classnames from 'classnames';
+import type { ReactNode } from 'react';
+
+export interface RoundedWrapperProps {
+  children?: ReactNode;
+  paddingBottom?: boolean;
+}
+
+export const RoundedWrapper = ({
```
dexturr, line 9: question (non-blocking): this roundedness is not something i have seen in other components, but perhaps I have missed something. If we aren't going to be doing this outside of tables IMO a prop for key value table would be better. Happy to wait and see how it goes though

sam-keen: Hmm yeah ok. I'll keep this as is for now and catch up with Joelle when she's back, and make it a prop as a secondary ticket if necessary.

dexturr (approved):

PR #2402, opened 2022-12-14
feat(2036): add markets to the navbar

### LINKED ISSUE #2036: Navigation to / from the markets page
## Chore 

Work out what the navigation to and from the markets listing page should be.

## Sketch

Image

Logic 
* Common pattern on KUCOIN, Binance, Injective to have a tabbed "markets" view, the market switching modals should be improved separately but not as part of this ticket

macqbat:
# Related issues 🔗

Closes #2036 

# Description ℹ️

Add new entry to the _navbar_. Do small fix for better look on mobile devices

# Demo 📺

image

image

# Technical 👨‍🔧

Add  conditional  horizontal scrollbar

### DIFF
--- a/libs/ui-toolkit/src/components/nav/nav.tsx
+++ b/libs/ui-toolkit/src/components/nav/nav.tsx
@@ -40,13 +40,16 @@ export const Nav = ({
   titleContent,
   title,
 }: NavbarProps) => {
-  const themeWrapperClasses = classNames('w-full', {
-    dark: navbarTheme === 'dark',
-  });
+  const themeWrapperClasses = classNames(
+    'w-full overflow-y-hidden overflow-x-auto md:overflow-x-hidden',
+    {
+      dark: navbarTheme === 'dark',
+    }
+  );
 
   const isYellow = navbarTheme === 'yellow';
   const navbarClasses = classNames(
-    'flex items-stretch border-b px-4 border-default',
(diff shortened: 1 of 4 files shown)


MadalinaRaicu (approved):

libs/ui-toolkit/src/components/nav/nav.tsx
```diff
@@ -40,13 +40,16 @@ export const Nav = ({
   titleContent,
   title,
 }: NavbarProps) => {
-  const themeWrapperClasses = classNames('w-full', {
-    dark: navbarTheme === 'dark',
-  });
+  const themeWrapperClasses = classNames(
+    'w-full overflow-y-hidden overflow-x-auto md:overflow-x-hidden',
```
mattrussell36, line 44: suggestion(non blocking): I wouldn't bother with these or other styles to allow the overflow classes, we have design work in progress to handle the navigation layout on small screens.

mattrussell36 (approved):

libs/ui-toolkit/src/components/nav/nav.tsx
macqbat: yeah, so it's only few css and heals the view from the really ugly crash on mobile. So as temporary solution, I think it will play his role good :-)

PR #2426, opened 2022-12-16
chore: fix failing governance test

jtsang586:
# Related issues 🔗

Fixes test failing in the PR run. Test got broken after it got re-added into develop.

### DIFF
--- a/apps/token-e2e/src/support/governance.functions.js
+++ b/apps/token-e2e/src/support/governance.functions.js
@@ -89,6 +89,7 @@ Cypress.Commands.add('get_proposal_id_from_list', (proposalTitle) => {
   cy.contains(proposalTitle)
     .parent()
     .parent()
+    .parent()
     .within(() => {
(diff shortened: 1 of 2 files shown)


apps/token-e2e/src/support/governance.functions.js
```diff
@@ -87,6 +87,7 @@ Cypress.Commands.add(
 
 Cypress.Commands.add('get_proposal_id_from_list', (proposalTitle) => {
   cy.contains(proposalTitle)
+    .parent()
```
sam-keen, line 90: Question (non-blocking): are you able to use [parentsUntil](https://docs.cypress.io/api/commands/parentsuntil) instead? Seem like it'd be less brittle

jtsang586: Thanks Sam, I didn't know about that one! And I agree parentsUntil is much better here. 
Updated

sam-keen (approved):

PR #2435, opened 2022-12-20
chore(#2412): refactor theme context

### LINKED ISSUE #2412: Refactor theme provider to use zustand
## The Chore

It would be good to do a small refactor to use zustand. Benefits being
- align with other local state management
- no need to wrap apps with the theme provider
- can remove theme.js
- less passing of props

## Tasks

- [ ] Refactor using zustand
- [ ] Ensure default state is taken from local storage
- [ ] Remove theme.js

mattrussell36:
# Related issues 🔗

Closes #2412 

# Description ℹ️

Changes theme store to use zustand to allow removal of context and less prop drilling. Additionally sets theme immediately before react mounts so that theme.js in the static can be removed

### DIFF
--- a/libs/react-helpers/src/hooks/use-theme-switcher.ts
+++ b/libs/react-helpers/src/hooks/use-theme-switcher.ts
@@ -1,42 +1,68 @@
-import { useEffect, useState } from 'react';
+import { useEffect } from 'react';
+import create from 'zustand';
 import { LocalStorage } from '../lib/storage';
 
 const darkTheme = 'dark';
 const lightTheme = 'light';
-type themeVariant = typeof darkTheme | typeof lightTheme;
+type Theme = typeof darkTheme | typeof lightTheme;
 
 const darkThemeCssClass = darkTheme;
 
+const isValidTheme = (theme: string): theme is Theme => {
+  if (theme === 'light' || theme === 'dark') return true;
+  return false;
+};
+
+const setTheme = (theme: Theme) => {
+  if (typeof window === 'undefined') return;
+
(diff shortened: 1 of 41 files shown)


libs/react-helpers/src/hooks/use-theme-switcher.ts
```diff
@@ -1,42 +1,77 @@
...
-  const theme = LocalStorage.getItem('theme');
-  if (
-    theme === darkTheme ||
-    (!theme &&
-      typeof window !== 'undefined' &&
-      window.matchMedia('(prefers-color-scheme: dark)').matches)
-  ) {
-    return darkTheme;
+const validateTheme = (theme: string): theme is Theme => {
+  if (theme === 'light' || theme === 'dark') return true;
```
macqbat, line 13:
```suggestion
  if (theme === Themes.LIGHT || theme === Themes.DARK) return true;
```

macqbat (approved):

libs/react-helpers/src/hooks/use-theme-switcher.ts
macqbat, line 13:
```suggestion
  if ([Themes.LIGHT, Themes.DARK].includes(theme)) return true;
```

macqbat: Or like that...

mattrussell36: or one better `Object.values(Themes).includes(theme)`

PR #2436, opened 2022-12-20
fix(#2413): collateral table to have separate used available cells

### LINKED ISSUE #2413: Collateral table use separate cells for used/available to allow filtering/sorting
## The Chore

Lets separate the used and available cells so we can allow filtering and sorting on those values. This is also fixes a bug where the values can collide making it hard to see the true value.

We can remove the progress bar for now

### Tasks

- [ ] Separate cells
- [ ] Remove progress bar
- [ ] ~~Enable filtering/sorting~~

MadalinaRaicu:
# Related issues 🔗

Closes #2413 

# Description ℹ️

Collateral table to have separate used available cells to allow sorting.

# Demo 📺

Before
image

After
image

# Technical 👨‍🔧


libs/accounts/src/lib/breakdown-table.tsx
```diff
@@ -5,19 +5,47 @@ import {
...
+  data,
+  node,
+}: ValueFormatterParams): ValueProps['valueFormatted'] | undefined => {
+  if (!data || node?.rowPinned) {
+    return undefined;
+  }
+  const min = BigInt(data.used);
+  const mid = BigInt(data.available);
+  const max = BigInt(data.deposited);
+  const range = max > min ? max : min;
```
notbot00, line 33: comment (non-blocking): usually i find using `Math.max(min, max)` easier to use for cases like this, because i don't have to think the ternary branching. but this is absolutely fine as it is.

notbot00 (approved):

PR #2450, opened 2022-12-21
chore: console dapp routes for skeleton page (2361)

### LINKED ISSUE #2361: chore: change urls for skeleton (aka empty) market page
restructure urls a bit
* `/markets` - no market id, so you get the skeleton page
* `/markets/abcxyz` - market page renders as normal
* `/markets/all` - current /markets page

asiaznik:
# Related issues 🔗

Closes #2361

### DIFF
--- a/apps/trading/pages/client-router.tsx
+++ b/apps/trading/pages/client-router.tsx
@@ -1,8 +1,10 @@
 import { Suspense } from 'react';
+import type { RouteObject } from 'react-router-dom';
 import { useRoutes } from 'react-router-dom';
 import dynamic from 'next/dynamic';
 import { t } from '@vegaprotocol/react-helpers';
 import { Splash } from '@vegaprotocol/ui-toolkit';
+import trimEnd from 'lodash/trimEnd';
 
 const LazyHome = dynamic(() => import('../client-pages/home'), {
   ssr: false,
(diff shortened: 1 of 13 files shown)


apps/trading/pages/client-router.tsx
```diff
@@ -26,10 +26,23 @@ const LazyPortfolio = dynamic(() => import('../client-pages/portfolio'), {
...
   HOME = '/',
-  MARKETS = '/markets',
+  SKELETON_MARKET = '/markets',
+  MARKET = '/markets/:marketId',
+  MARKETS = '/markets/all',
   PORTFOLIO = '/portfolio',
+  LIQUIDITY = 'liquidity/:marketId',
 }
 
+export const Links = {
```
mattrussell36, line 36: suggestion (non blocking): could type this with something like `{ [R in Routes]: () => string }` so that we don't miss a link function for a route

apps/trading/pages/client-router.tsx
```diff
@@ -40,11 +53,15 @@ const routerConfig = [
     element: <LazyMarkets />,
   },
   {
-    path: 'markets/:marketId',
+    path: Routes.SKELETON_MARKET,
```
mattrussell36, line 56: suggestion (non blocking): I feel like it might be worth exploring a nested route here as /markets/abc is essentially a child of /markets

This way we might not need the skeleton prop, instead relying on the if there is a route match instead

mattrussell36 (approved):

MadalinaRaicu (approved):

apps/trading/pages/client-router.tsx
asiaznik: I've removed the prop as it was not really needed - also changed to the nested routes as suggested.

PR #2462, opened 2022-12-22
chore: add wallet client

### LINKED ISSUE #2457: Use the `Authorization` header when calling the wallet API
1. Set the `Authorization` header with the token, when calling the wallet API:

```
Autorization: VWT <TOKEN>
```

**It should still pass it as a parameter in the JSON-RPC payload.** That way, when the token update is shipped, the front end will work out of the box.

2. When calling the `client.connect_wallet`, today, the token is returned in the JSON-RPC payload. Later, it will be returned through the `Authorization` header. So we need to build a resolution mechanism, that check the `Authorization` header first, and if nothing is set, then it should look in the JSON-RPC payload.

3. If the API returns a 401 HTTP code, the header `WWW-Authenticate` header is set to `VWT` so you know which scheme is used for authentication.

4. There is a breaking change on the call to `GET /api/v2/methods` so we need to support 2 ways for reading this response, _temporarily_:
  * Before: `{ "registeredMethods": ["", "", ""]}` & `{"error":"...."}`
  * After: `{"result":{ "registeredMethods": ["", "", ""]}}` & {{"error":{ "message": "Single line explanation", "data": "a detailed error message"}}}
  
If the `registeredMethods` property is a the root of the response, we should read it the old way. If not, it's scoped under the `result` object.

The "clean up" issue should follow the release of the new token mechanism:
- https://github.com/vegaprotocol/frontend-monorepo/issues/2458

notbot00:
# Related issues 🔗

Closes #2457 

# Description ℹ️

Pull in the generated wallet client and replace the code in the wallet lib's jsonrpc connector to use the generated client instead of the hardcoded schemas.


MadalinaRaicu (approved):

libs/wallet/src/connectors/json-rpc-connector.ts
```diff
@@ -91,116 +28,104 @@ export class JsonRpcConnector implements VegaConnector {
...
+        'No client found. The connector needs to be initialized with a url.'
+      );
     }
-    const parseResult = ListKeysSchema.safeParse(result);
-    if (parseResult.success) {
-      return parseResult.data.result.keys;
-    } else {
+
+    try {
+      const { result } = await this.client.ListKeys({});
```
mattrussell36, line 90: question (non blocking): Why the empty object here? Add a comment?

libs/wallet/src/connectors/json-rpc-connector.ts
```diff
@@ -91,116 +28,104 @@ export class JsonRpcConnector implements VegaConnector {
...
 
   async disconnect() {
-    const cfg = getConfig();
-
-    if (cfg?.token) {
-      await this.request(Methods.DisconnectWallet, {
-        token: cfg.token,
-      });
+    if (!this.client) {
+      throw new Error(
```
mattrussell36, line 99: chore (non blocking): This could be DRY'd up a bit. Perhaps with a method `this.checkClient()` to check this.client is set? or storing the error as a static property?

mattrussell36 (approved):

libs/wallet/src/connectors/json-rpc-connector.ts
notbot00: it's a bug in the generator :( haven't figured out a way to be able to easily do this without bloating the template into some unreadable forever-entangled spaghetti. but since the pipeline is complaining on this, will see what i can do to clean this up, it's pretty annoying having to pass in empty objects as a consumer.

libs/wallet/src/connectors/json-rpc-connector.ts
notbot00: 1. i could implement a checkClient, but it wouldn't add any value in typescript. having a difficult time explaining it with words on a monday morning, but:

```
class JsonRpcConnector {
  checkClient() {
     if (!this.client) {
       throw new Error()
     }
   }
   
   someMethod() {
     this.checkClient()
     this.client.SomeMethod() // <= Error: Object possibly undefined
   }

}
```

this could be all scraped if the url was a required prop of the connector, but i'd need to rewrite a while bunch of things for that, and i was just not up for the challenge, but i think it would be a much nicer approach than having these checks for undefined all over the place. 

2. the error as a static property i think in general a not so good of a practice, because it totally swallows the stack trace, so when debugging, you wouldn't be able to identify which actual method threw the error. stack traces get injected upon initialisation, not a huge expert on the topic, but did the following little test to confirm:

```
// 1 - storing the pre-initialised error
const e = new Error('test stored')
function f1 () {
  throw e
}


// 2 - not storing it
function f2 {
  throw new Error('test not stored')
}
```

The result when calling them looks different:

Screenshot 2022-12-26 at 11 32 48

libs/wallet/src/connectors/json-rpc-connector.ts
notbot00: fixed this in https://github.com/vegaprotocol/vegawallet-ui/pull/55 so now i was able to remove the empty object params

libs/wallet/src/connectors/json-rpc-connector.ts
mattrussell36: Makes a lot of sense. Thanks for the detailed response

PR #2489, opened 2022-12-29
chore: redirect to last visited market instead first from the list

### LINKED ISSUE #2406: Persist the market I last viewed
## Story

As a user
I want Console to remember the market I was last looking at and load that market view when I close and re-open the dapp  
So that I don't have to browse the market list again to find the market I'm interested in 

## Acceptance Criteria

- [ ] I can close the browser tab/window that Console was open in and reopen in a new tab/window and be shown the market view I was last looking at (without the 'select a market to get started' modal opening) 

## Tasks

- [ ] Check local storage for a stored marekt ID in the startup logic before redirecting to market page

macqbat:
# Related issues 🔗

Closes #2406 

# Description ℹ️

Redirect from home page to the last visited market instead first from the list. Adjust welcome dialogs logic according to persisted markets

# Demo 📺

-

# Technical 👨‍🔧

-

### DIFF
--- a/apps/trading/client-pages/home/home.tsx
+++ b/apps/trading/client-pages/home/home.tsx
@@ -1,4 +1,8 @@
-import { marketsWithDataProvider } from '@vegaprotocol/market-list';
+import type { SingleMarketFieldsFragment } from '@vegaprotocol/market-list';
+import {
+  marketProvider,
+  marketsWithDataProvider,
+} from '@vegaprotocol/market-list';
 import {
   addDecimalsFormatNumber,
   titlefy,
@@ -6,7 +10,7 @@ import {
 } from '@vegaprotocol/react-helpers';
 import { AsyncRenderer } from '@vegaprotocol/ui-toolkit';
 import { Links, Routes } from '../../pages/client-router';
-import { useEffect } from 'react';
+import { useEffect, useMemo } from 'react';
 import { useNavigate } from 'react-router-dom';
 import { useGlobalStore, usePageTitleStore } from '../../stores';
 
@@ -17,18 +21,37 @@ export const Home = () => {
   const { data, error, loading } = useDataProvider({
     dataProvider: marketsWithDataProvider,
   });
-  const { update } = useGlobalStore((store) => ({
+  const { update, marketId } = useGlobalStore((store) => ({
     update: store.update,
+    marketId: store.marketId,
   }));
-
+  const variables = useMemo(
+    () => ({
+      marketId: marketId || '',
+    }),
+    [marketId]
+  );
+  const { data: singleMarketData, loading: singleMarketLoading } =
+    useDataProvider<SingleMarketFieldsFragment, never>({
+      dataProvider: marketProvider,
+      variables,
+      skip: !marketId,
+    });
+  const marketExists = Boolean(singleMarketData);
+  const marketNotExists = !marketExists && !singleMarketLoading;
   const { pageTitle, updateTitle } = usePageTitleStore((store) => ({
     pageTitle: store.pageTitle,
     updateTitle: store.updateTitle,
   }));
 
   useEffect(() => {
-    if (data) {
-      const marketId = data[0]?.id;
+    if (marketExists) {
+      navigate(Links[Routes.MARKET](marketId as string), {
+        replace: true,
+      });
+      update({ shouldDisplayWelcomeDialog: false });
+    } else if (data && marketNotExists) {
+      const marketDataId = data[0]?.id;
       const marketName = data[0]?.tradableInstrument.instrument.name;
       const marketPrice = data[0]?.data?.markPrice
         ? addDecimalsFormatNumber(
@@ -38,11 +61,11 @@ export const Home = () => {
         : null;
       const newPageTitle = titlefy([marketName, marketPrice]);
 
-      if (marketId) {
-        navigate(Links[Routes.MARKET](marketId), {
+      if (marketDataId) {
+        navigate(Links[Routes.MARKET](marketDataId), {
           replace: true,
         });
-        update({ marketId });
+        update({ marketId: marketDataId, shouldDisplayWelcomeDialog: true });
         if (pageTitle !== newPageTitle) {
           updateTitle(newPageTitle);
         }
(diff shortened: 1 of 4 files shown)


apps/trading/components/welcome-dialog/welcome-dialog.tsx
```diff
@@ -24,39 +25,58 @@ export const WelcomeDialog = () => {
...
+  const isRiskDialogNeeded =
+    riskAccepted !== 'true' && VEGA_ENV === Networks.MAINNET;
+  const isWelcomeDialogNeeded =
+    (pathname === '/' && !marketId) || shouldDisplayWelcomeDialog;
+  const onClose = useCallback(() => {
+    update({ shouldDisplayWelcomeDialog: isRiskDialogNeeded });
+    setDialog(null);
+  }, [setDialog, update, isRiskDialogNeeded]);
+
+  useEffect(() => {
```
mattrussell36, line 50: suggestion (non blocking): I realize this was already like this.. but hy are we setting components to state here? Can we just have some render logic? I think a small refactor might be good to simplify, increase readability and reduce renders. Something like:

```
// in component

const WelcomeDialog = () => {
  ...
  let dialog = null
  
  
  if (riskAccepted !== 'true' && VEGA_ENV === Networks.MAINNET) {
    dialog = {
      content: <WelcomeNoticeDialog />,
      title: t('WARNING'),
      size: 'medium'
    }
  } else if (pathname === '/' && data?.length === 0) {
    ...
  } else if (etc etc) {
    ...
  }
  
  return (
    <Dialog
      open={dialog !== null}
      title={dialog.title}
      size={dialog.size}
      onChange={onClose}
    >
      {dialog !== null && dialog.content}
    </Dialog>
  )  
}
```

apps/trading/client-pages/home/home.tsx
```diff
@@ -17,18 +21,37 @@ export const Home = () => {
...
+    marketId: store.marketId,
   }));
-
+  const variables = useMemo(
+    () => ({
+      marketId: marketId || '',
+    }),
+    [marketId]
+  );
+  const { data: singleMarketData, loading: singleMarketLoading } =
```
mattrussell36, line 34: suggestion (non blocking): For simplicity I think it would be better to not use the data provider here and just let the page load with an invalid market, and have the user see the 'Not found' message. It'll be more predictable and more obvious to tell the user that their expected market is gone rather than redirect them somewhere else

apps/trading/client-pages/home/home.tsx
```diff
@@ -17,18 +21,37 @@ export const Home = () => {
...
   const { pageTitle, updateTitle } = usePageTitleStore((store) => ({
     pageTitle: store.pageTitle,
     updateTitle: store.updateTitle,
   }));
 
   useEffect(() => {
-    if (data) {
-      const marketId = data[0]?.id;
+    if (marketExists) {
+      navigate(Links[Routes.MARKET](marketId as string), {
```
mattrussell36, line 49: chore (blocking): a null check would be better than a type cast (what happens if we call `navigate(null)`)?

apps/trading/client-pages/home/home.tsx
```diff
@@ -17,18 +21,37 @@ export const Home = () => {
...
+        replace: true,
+      });
+      update({ shouldDisplayWelcomeDialog: false });
```
mattrussell36, line 52: suggestion (non blocking): For me the home page logic should be only for deciding which market to redirect to. Is there any way we can change this so that we don't need this new piece of global state, and the WelcomeDialog component itself can contain the logic for deciding when to show and what dialog content to show?

mattrussell36 (changes requested):

apps/trading/components/welcome-dialog/welcome-dialog.tsx
macqbat: We rely here on routing parameter (_pathname_). If we gave up using _state_ here and started to simply return rendered component, changing pathname after a redirect would change the conditions and dialog could be hidden.
The goal was to avoid using store flag. Finally I had to use the store flag here due to growing complexity, so maybe I'll change the implementation, but still simply returning rendered content will be not easily feasible, because we will depend on store flag then  ¯\_(ツ)_/¯

apps/trading/client-pages/home/home.tsx
macqbat: marketId can't be null here, because marketExists flag is possible to be true only when results of call api with marketId return valid market. I could add there some guard assertion, but such assert type _as string_ seemed simpler to me

apps/trading/client-pages/home/home.tsx
macqbat: Hm, I didn't assume we want to redirect user to possibly not existing market. Yeah, I'm gonna rearrange this checking here

apps/trading/client-pages/home/home.tsx
macqbat: I'm gonna reconsider whole solution if it looks too messy and try to make it all simpler

macqbat: So, I pushed some changes. I added also some cleaning up of market page: moved update of last visiting market to update method of market data provider, rather than taking just param from url which can be easily misformatted.
Please @mattrussell36 take a look in a free while.

apps/trading/client-pages/home/home.tsx
```diff
@@ -17,40 +13,28 @@ export const Home = () => {
   const { data, error, loading } = useDataProvider({
     dataProvider: marketsWithDataProvider,
   });
-  const { update } = useGlobalStore((store) => ({
+  const { update, marketId } = useGlobalStore((store) => ({
```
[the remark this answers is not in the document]
macqbat: ok, although it's outside the scope of this task, it's improved now, check again please

apps/trading/client-pages/market/market.tsx
```diff
@@ -32,14 +32,13 @@ export interface SingleMarketData extends SingleMarketFieldsFragment {
...
 export const Market = () => {
-  const params = useParams();
+  const { marketId } = useParams();
   const navigate = useNavigate();
 
-  const marketId = params.marketId;
-
   const { w } = useWindowSize();
-  const { update } = useGlobalStore((store) => ({
+  const { update, marketId: lastMarketId } = useGlobalStore((store) => ({
```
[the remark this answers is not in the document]
macqbat: as above, it's improved now

bglownia (approved):

mattrussell36 (approved):

PR #2491, opened 2022-12-29
feat(explorer): use block height filter

### LINKED ISSUE #2490: Switch transaction per block fetch to use filter
## The Chore

 https://github.com/vegaprotocol/vega/pull/6958 added a proper filter for transactions by block, which is now deployed across all environments. Switch the block by ID page to use the new filter to get transactions for the ID in question, rather than the less-than-ideal cursor based solution

## Tasks

- [ ] Switch blocks by ID page to use filter

edd:
# Related issues 🔗

Closes #2490

# Description ℹ️

Switches the filter used when fetching transactions for a block from relying on a hack to using the correct filter.

### DIFF
--- a/apps/explorer/src/app/components/txs/txs-per-block.tsx
+++ b/apps/explorer/src/app/components/txs/txs-per-block.tsx
@@ -1,5 +1,4 @@
 import { Routes } from '../../routes/route-names';
-import { DATA_SOURCES } from '../../config';
 import { RenderFetched } from '../render-fetched';
 import { TruncatedLink } from '../truncate/truncated-link';
 import { TxOrderType } from './tx-order-type';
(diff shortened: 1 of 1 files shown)


apps/explorer/src/app/components/txs/txs-per-block.tsx
```diff
@@ -17,15 +17,11 @@ interface TxsPerBlockProps {
 const truncateLength = 5;
 
 export const TxsPerBlock = ({ blockHeight, txCount }: TxsPerBlockProps) => {
-  // TODO after https://github.com/vegaprotocol/vega/pull/6958/files is merged and deployed, use filter
-  // by block height instead
+  const filters = `filters[block.height]=${blockHeight}`;
+  const url = getTxsDataUrl({ limit: txCount.toString(), filters });
```
mattrussell36, line 21: suggestion (non blocking): hooks should be named with the useX convention

mattrussell36: Oh this is just a function within the file. Super minor but perhaps move to a more general helpers file?

mattrussell36 (approved):

apps/explorer/src/app/components/txs/txs-per-block.tsx
edd: I'll consider that next time I call it. So far it's only used twice.

PR #2493, opened 2022-12-30
chore: fix and improve types for market proposal

mattrussell36:
# Related issues 🔗

N/A

# Description ℹ️

- Fixed incorrect types for market proposals
- Improved types in some places with some unions


edd (approved):
👍 Looks good. 

In case any other reviewers are wondering, this change won't be backwards compatible before 0.65.1

PR #2498, opened 2023-01-03
feat(#2255): capsule tests for trading

### LINKED ISSUE #2255: Run console with capsule
## The Chore

Usage of vega wallet dummy requires capsule so that it doesn't make requests against a live environment. This leads to unstable testing environments and flakey tests

## Tasks

- [ ] Set up minimal capsule configuration (we will still be mocking all queries)
- [ ] Update github workflows to use capsule

mattrussell36:
# Related issues 🔗

Closes #2255 

# Description ℹ️

Adds setup scripts to create a market and place trades against it


libs/cypress/src/lib/capsule/ethereum-setup.ts
```diff
@@ -0,0 +1,191 @@
...
+}
+
+function timeout(time = 0, id: string) {
+  return new Promise((resolve, reject) => {
+    setTimeout(() => reject(new Error(`${id}: timeout triggered`)), time);
+  });
+}
+
+async function promiseWithTimeout(
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
MadalinaRaicu, line 80: we should leave Promise<any> here. we are using StakingBridge.wait() as a param

MadalinaRaicu (approved):

PR #2501, opened 2023-01-03
chore: fix percetnages

dexturr:
# Related issues 🔗

N/A

# Description ℹ️

API is erroring and giving an incorrect amount for total staked. Fortunately this is only used to calculate Share which is exactly stake score. So this PR presents the stake score as share as per the current testnet implmentation


asiaznik (approved):

edd (approved):

edd:
The code makes sense, but the PR description could be a little clearer

PR #2507, opened 2023-01-03
fix: error message consitency (2486)

### LINKED ISSUE #2486: Consistency in display error message for decimalPlaces - price
## The Chore

As a user 
I want to see error message of the same kind in the same way in different places in this case between deal ticket and edit order.
So that the user will have consistent information across application

Current behavior:
For deal ticket we have error message "Price accepts up to 5 decimal places"
For edit order we have "Please enter a valid value. The two nearest valid values are..."

Screenshot 2022-12-28 at 13 01 14

asiaznik:
# Related issues 🔗

Closes #2486 

# Demo 📺

image

# Technical 👨‍🔧

I had to move `validate-amount` because of the circular deps.


mattrussell36 (approved):
Nice fix and tidyup

PR #2513, opened 2023-01-04
feat(2408): trading data grid snags

### LINKED ISSUE #2408: Trading data grid snags
## The Chore

Improve visual consistency of data tables that use ag grid

## Tasks

- [x] Ensure row heights are consistent
- [x] Remove use of Button within cells and use ButtonLink
- [x] Replace custom market name cell in positions table with instrument.code and render single line text
- [ ] Remove cell focus globally
- [ ] Use pinned row for deposit/withdraw button (outside of the table)
- [x] TBD: Investigate ledger entry table and how we can compact the data so we have single rows

bglownia:
# Related issues 🔗

Closes #2408

# Description ℹ️
- fix line height in data grids
- remove custom rowHeight
- change all multiline cells to multiple columns
- merge pending and completed withdrawal tables 
- always display data tables, even if when data is loading
- render no data and loading messages from AsyncRenderer, instead of using ag-grid overlays
- don't mix row model types
- use separate columns for multiline Sender and Receiver columns in ledger table
- replace Button and button with ButtonLink
- fix useOrderListData avoid render condition
- move cancel all button from order list table pinned row to order list manager
- replace multiline market name cell in positions table with single line market name


libs/positions/src/lib/positions-table.tsx
```diff
@@ -413,12 +362,17 @@ export const PositionsTable = forwardRef<AgGridReact, Props>(
...
-              return {
-                component: ButtonCell,
-              };
-            }}
-            cellRendererParams={{ onClick: onClose }}
+            cellRenderer={({
+              data,
+              node,
+            }: VegaICellRendererParams<Position>) => (
+              <ButtonLink
```
mattrussell36, line 369: chore (non blocking): right align button (same as orders table buttons)

libs/accounts/src/lib/accounts-table.tsx
```diff
@@ -153,31 +152,30 @@ export const AccountTable = forwardRef<AgGridReact, AccountTableProps>(
...
             minWidth={250}
+            type="rightAligned"
             cellRenderer={({
               data,
             }: VegaICellRendererParams<AccountFields>) => {
               return data ? (
-                <div className="flex gap-2 justify-end">
-                  <Button
-                    size="xs"
+                <>
```
mattrussell36, line 160: chore (non blocking): lets move the break down link into the same cell so all actions are kept together

mattrussell36 (approved):

PR #2521, opened 2023-01-04
feat(2211): download proposal forms data as json

### LINKED ISSUE #2211: Allow downloading of proposals on proposal forms
## Story

As a User 
I want to be able to download the JSON of a proposal
So that I can post it on the forum and itterate on it

## Acceptance Criteria

- [ ] I can see a download button
- [ ] I can create a proposal without a wallet connected
- [ ] I can see the submit button is disabled if I am not connected to my Vega wallet

## Tasks

- [ ] Add button
- [ ] Ensure works without being connected to a vega wallet

- #2205

sam-keen:
# Related issues 🔗

Closes #2211

# Description ℹ️

Allows a user to download the current proposal form's data as json. Also allows proposal forms to be seen without needing to connect a wallet. Wallet connection is now only needed for form submission.

# Demo 📺

Screenshot 2023-01-05 at 11 52 27
Screenshot 2023-01-05 at 11 52 17


apps/token/src/lib/view-form-as-json-new-window.ts
```diff
@@ -0,0 +1,21 @@
...
+  // e.g. "2023-Jan-03-23-59-59"
+  const formattedDateTime = `${day}-${month}-${year}-${hours}-${minutes}-${seconds}`;
+
+  const blob = new Blob([jsonString], { type: 'application/json' });
+  const url = URL.createObjectURL(blob);
+  const a = document.createElement('a');
+  a.style.display = 'none';
+  a.href = url;
+  a.download = `${proposalTitle}-${formattedDateTime}.json`;
+  document.body.appendChild(a);
```
dexturr, line 18: question (blocking): We aren't cleaning up this link, is this intentional or should we add a 

`a.parentNode.removeChild(a);`

apps/token/src/routes/proposals/components/propose/proposal-form-download-json.tsx
```diff
@@ -0,0 +1,19 @@
+import { useTranslation } from 'react-i18next';
+import { Button } from '@vegaprotocol/ui-toolkit';
+
+interface ProposalFormDownloadJsonProps {
+  downloadJson: () => void;
+}
+
+export const ProposalFormDownloadJson = ({
```
dexturr, line 8: question (non-blocking): Why not just have this component take a blob of data and stringify and do the download logic itself rather than a function?

apps/token/src/routes/proposals/components/propose/proposal-form-submit.spec.tsx
```diff
@@ -0,0 +1,59 @@
...
+        pubKeys: [pubKey],
+      } as VegaWalletContextShape,
+      false
+    );
+    expect(screen.getByTestId('proposal-submit')).toHaveTextContent(
+      'Submit proposal'
+    );
+  });
+
+  it('should display submitting button text if wallet is connected and submitting', () => {
```
dexturr, line 46: chore: can we please add a unit test to check that the download function is called?

apps/token/src/routes/proposals/propose/update-market/propose-update-market.tsx
```diff
@@ -145,176 +147,182 @@ export const ProposeUpdateMarket = () => {
...
-        {() => (
-          <>
-            <ProposalMinRequirements
-              minProposalBalance={
-                params.governance_proposal_updateMarket_minProposerBalance
-              }
-              spamProtectionMin={params.spam_protection_proposal_min_tokens}
-              userAction={ProposalUserAction.CREATE}
-            />
+      data={{ ...params, ...marketsData }}
```
dexturr, line 169: follow up (non-blocking): can we just check quickly that none of the rest of the forms are likely to experience this issue?

apps/token/src/routes/proposals/components/propose/proposal-form-download-json.tsx
sam-keen: It seemed better to me for this to be as dumb as possible, and for the download logic to be a separate entity, which could be used independently. I don't feel strongly though, I can adjust if you think it's preferable to be in the component

apps/token/src/lib/view-form-as-json-new-window.ts
sam-keen: not intentional, good spot, thanks

apps/token/src/routes/proposals/propose/update-market/propose-update-market.tsx
sam-keen: no worries, all the rest simply take `params`

PR #2539, opened 2023-01-06
fix(#2530): market data not in sync

### LINKED ISSUE #2530: Trading mode in market selector and header do not match
Trading mode in market selector and header do not match.
Reproduce: Go on console testnet and switch between  BTCUSD and UNIDAI markets

Image

Image

mattrussell36:
TODO:

- tidy up market list test
- rename operation names
- rename market type alias

# Related issues 🔗

Closes #2530

# Description ℹ️

Fixes an issue where the data in the trading header and in the market dropdown are not in sync

# Technical 👨‍🔧

- Queries and subscribes once at the top of the market page component tree and passes data to the TradeHeader and the market dropdown as props. Note: this also means we can write smaller queries deeper in the tree and they will also be kept up to date.
- Ensures query for market dropdown is only called when opened (using cache and network) to ensure data is up to date
- Removes logic not required for HeaderStats
- Removes asset id as key field for product. This could be problematic for products that share an asset.
- Moves usePrevious to react-helpers


libs/market-list/src/lib/use-market.ts
```diff
@@ -0,0 +1,88 @@
...
+  useMarketHookQuery,
+} from './__generated__/MarketsHooks';
+import type {
+  MarketHookDataFragment,
+  MarketHookQuery,
+  MarketsDataHookSubscription,
+  MarketsDataHookSubscriptionVariables,
+} from './__generated__/MarketsHooks';
+
+export type MarketX = NonNullable<MarketHookQuery['market']>;
```
macqbat, line 16: can we such arbitrarily assume here _market_ in results of query will never be null or undefined?

libs/market-list/src/lib/use-markets.ts
```diff
@@ -0,0 +1,110 @@
...
+  }, [data]);
+
+  useEffect(() => {
+    const unsub = subscribeToMore<
+      MarketsDataHookSubscription,
+      MarketsDataHookSubscriptionVariables
+    >({
+      document: MarketsDataHookDocument,
+      variables: { marketIds: ids },
+      updateQuery: (prev, { subscriptionData }) => {
```
macqbat, line 37: maybe we will need some throttling here? Update can happen many times per second

libs/market-list/src/lib/use-markets.ts
```diff
@@ -0,0 +1,110 @@
...
+          };
+        });
+
+        return {
+          ...prev,
+          marketsConnection: {
+            ...prev.marketsConnection,
+            edges: updatedEdges,
+          },
+        } as MarketHookQuery;
```
macqbat, line 95: Shouldn't be here MarketsHookQuery instead single MarketHookQuery?

apps/trading/client-pages/market/market.tsx
```diff
@@ -26,21 +20,35 @@ const calculatePrice = (markPrice?: string, decimalPlaces?: number) => {
...
-  const lastMarketId = useGlobalStore((store) => store.marketId);
-
-  const pageTitle = usePageTitleStore((store) => store.pageTitle);
   const updateTitle = usePageTitleStore((store) => store.updateTitle);
+  const prevMarkPrice = usePrevious(data?.market?.data?.markPrice);
+
+  // Store last market id for easier so we can load
+  // previous last market when visiting index page
+  useEffect(() => {
+    update({ marketId });
```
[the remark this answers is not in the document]
macqbat: imho would fit here
```
useEffect(() => {
  if(data?.id){
    update({ marketId: data.id });
  }
}, [data?.id, update]);
```

apps/trading/client-pages/market/market.tsx
```diff
@@ -51,63 +59,17 @@ export const Market = () => {
...
-  });
-
   const tradeView = useMemo(() => {
+    if (!data?.market) {
+      return null;
+    }
+
     if (w > 960) {
-      return <TradeGrid market={data} onSelect={onSelect} />;
+      return <TradeGrid market={data.market} onSelect={onSelect} />;
```
macqbat, line 68: When we passing down to TradePanels constantly differentiating market object, we can cause a massive rerender. That's why previously we passed only constant _Market_ object with no changeable _MarketData_ part.

bglownia: I totally agree with @macqbat.
We on purpose separated market from market data.
Market don't change and it was passed down as a prop.
All components that were using market data were subscribing to updates through data provider and was reusing same data from provider. 
We avoid rerender of whole TradeGrid on marketData change. 
Additionally all components (such as MarketMarkPrice, MarketTradingMode or MarketVolume) that were using market data can be reused in any other place providing only marketId in props.  
It was all replaced by MarketHookData with subscribeToMore.

bglownia (changes requested):
MarketsHooks query should be removed. All changes connected to padding down market data from <Market> client page to <TradeGrid> and <TradePanels> should be reverted.

libs/market-list/src/lib/use-markets.ts
```diff
@@ -0,0 +1,110 @@
...
+    return data.marketsConnection.edges.map((e) => e.node.id);
+  }, [data]);
+
+  useEffect(() => {
+    const unsub = subscribeToMore<
+      MarketsDataHookSubscription,
+      MarketsDataHookSubscriptionVariables
+    >({
+      document: MarketsDataHookDocument,
+      variables: { marketIds: ids },
```
bglownia, line 36: In this place we are subscribing to market data (including mark price) for all existing markets.
I'm not sure if backend will be happy about it. For sure it will update multiple times per second and it will rerender <TradeGrid> in same frequency

mattrussell36: TradeGrid wont't be re-rendered here. This hook is only invoked when the market selector is opened and should only affect child components of SelectMarket.

I agree that it might not be great for the backend. But how else are we supposed to render a list of markets that receives updates via subscription? I noticed that the market list on /markets/all doesn't update at all, which is why I started experimenting with a different hooks/apollo based approach.

apps/trading/client-pages/market/market.tsx
mattrussell36: The market object here doesn't make it down to any of the windows, only really to TradeGridHeader. Each window only takes a marketId (I think they all do but need to double check) and should manage their own query, as they do now

libs/market-list/src/lib/use-markets.ts
mattrussell36: Additionally the marketsData subscription was recently flattened to improve perf for backend, so I think we should use it in parts of the app where we are rendering lists of markets

apps/trading/client-pages/market/market.tsx
bglownia: _Why do components get re-rendered?

[...]

Re-rendering of parent component: Whenever the components render function is called, all its subsequent child components will re-render, regardless of whether their props have changed or not._

If market data will change if will cause all sub components re render. We are connecting only very end components to 
live market data to avoid re rendering where it's not needed.
Even more. Components which are connected to market data subscription  (like MarketMarkPrice or  MarketVolume) are will  re render only when data significant to what they render will change and this update is throttled.

libs/market-list/src/lib/use-markets.ts
bglownia: To get live update of market data for every row in market selector only one line of the code needs to be changed.
https://github.com/vegaprotocol/frontend-monorepo/blob/develop/apps/trading/components/select-market/select-market-columns.tsx#L192
`const noUpdate = false`

PR #2541, opened 2023-01-06
feat(governance): pending balances: add event listeners

### LINKED ISSUE #2057: Pending balances: add event listeners
## The Chore

We need to add some state for pending association for the connected Vega keys.

## Tasks

- [x] Listen for events (https://docs.ethers.io/v5/concepts/events/#events--logs-and-filtering) from stake_deposited and stake_removed for both the [vesting](https://etherscan.io/address/0x23d1bFE8fA50a167816fBD79D7932577c06011f4#events) and [token](https://etherscan.io/address/0xcB84d72e61e383767C4DFEb2d8ff7f4FB89abc6e#events) contracts. These conform to the same IStake interface.
- [x] When the events are recived push them into a pending array
- [x] Once Vega registers the event remove the relevant entry from the array
- [x] Get the events from the last N (where N is the blockchain.ethereumConfig.requiredConfirmations)and set these as the inital state
- [x] Ensure the array is uniqued to prevent potential race conditions

sam-keen:
# Related issues 🔗

Closes #2057

# Description ℹ️

Adds pending balances functionality to provide some visibility of txs in transit. Displays in the wallet.

# Technical 👨‍🔧

Gets the number of confirmations as a network param, queries stake_deposited and stake_removed for both the [vesting](https://etherscan.io/address/0x23d1bFE8fA50a167816fBD79D7932577c06011f4#events) and [token](https://etherscan.io/address/0xcB84d72e61e383767C4DFEb2d8ff7f4FB89abc6e#events) contracts


apps/token/.env.testnet
```diff
@@ -1,7 +1,7 @@
 # App configuration variables
 NX_VEGA_ENV=TESTNET
 NX_VEGA_CONFIG_URL=https://static.vega.xyz/assets/testnet-network.json
-NX_VEGA_URL=https://api.n10.testnet.vega.xyz/graphql
+NX_VEGA_URL=https://api.n08.testnet.vega.xyz/graphql
```
MadalinaRaicu, line 4: (non-blocking): do you need to connect to a specific node by default?

MadalinaRaicu: also you may need to update name of the PR after last linter updates

sam-keen: I believe we need a default node. I think we switched to that one at the time when n10 was down, but it was a fairly arbitrary choice.

Thanks for the heads up on the linter update :)

apps/token/src/hooks/use-pending-balances-manager.ts
```diff
@@ -0,0 +1,35 @@
+import type { Event } from 'ethers';
+import uniqBy from 'lodash/uniqBy';
+
+import create from 'zustand';
+
+export type EthereumStore = {
```
dexturr, line 6:
```suggestion
export type PendingTxsStore  = {
```

apps/token/src/components/vega-wallet/vega-wallet.tsx
```diff
@@ -119,13 +120,38 @@ interface VegaWalletConnectedProps {
...
   const {
     appDispatch,
     appState: { decimals },
   } = useAppState();
   const { delegations, currentStakeAvailable, delegatedNodes, accounts } =
     usePollForDelegations();
+  const amountRemoved = BigNumber.sum.apply(null, [
+    new BigNumber(0),
+    ...pendingBalances
+      .filter(({ event }) => event === 'Stake_Removed')
```
dexturr, line 137: can we enum this?

apps/token/src/hooks/use-listen-for-pending-eth-events.ts
```diff
@@ -0,0 +1,103 @@
...
+        addPendingTxs([event]);
+        const tx = await event.getTransaction();
+        await tx.wait(numberOfConfirmations);
+        removePendingTx(event);
+      } catch (e) {
+        Sentry.captureException(`Error listening for pending eth events ${e}`);
+      }
+    };
+
+    contract?.on(filter, listener);
```
dexturr, line 38: conditional access here isn't needed

apps/token/src/hooks/use-listen-for-staking-events.ts
```diff
@@ -0,0 +1,55 @@
...
+  const { addPendingTxs, removePendingTx, resetPendingTxs } =
+    usePendingBalancesStore((state) => ({
+      addPendingTxs: state.addPendingTxs,
+      removePendingTx: state.removePendingTx,
+      resetPendingTxs: state.resetPendingTxs,
+    }));
+  const addFilter = useMemo(
+    () =>
+      vegaPublicKey && contract
+        ? contract.filters.Stake_Deposited(null, null, `0x${vegaPublicKey}`)
```
dexturr, line 20: is there a prepend ox function somewhere? I think there is

dexturr (approved):
Minor things

PR #2543, opened 2023-01-06
fix(#2532): show why market is in liquidity monitoring auction when liquidity supplied is > target

### LINKED ISSUE #2532: Update trading mode tooltip to cover why LP supplied is enough and market is still in liquidity monitoring auction
Change the Trading Mode tooltip to explain why even though there may be enough liquidity supplied, the market could still be in Monitoring auction - liquidity. We expect this to happen rarely, but it can happen in day 1 of trading launch and right now it does not look right in fairground.

Image

We could say something like these two:

- The market is in an auction until it has sufficient committed liquidity from LPs
- The market is in an auction because there are no priced limit orders, which are required to deploy LP pegged orders, so the order book is empty on one or both sides.

MadalinaRaicu:
# Related issues 🔗

Closes #2532

# Description ℹ️

- [x] update the liquidity supplied tooltip  - show why lp supplied is enough and the market is still in liquidity monitoring auction
- [x] rename `Estimated uncrossing price` to `Indicative price`
- [x] make sure in liquidity view header gets updated
- [x] show status and percentage in liquidity view

NOTE: long branch name because I generated it from GitHub issue.

# Demo 📺

image

# Technical 👨‍🔧
- [x] added condition to show the message


apps/trading/components/liquidity-supplied/liquidity-supplied.tsx
```diff
@@ -107,13 +108,26 @@ export const MarketLiquiditySupplied = ({
...
     <section>
       {compiledGrid && <DataGrid grid={compiledGrid} />}
       <br />
       <Link href={`/#/liquidity/${marketId}`} data-testid="view-liquidity-link">
         {t('View liquidity provision table')}
       </Link>
+      {showMessage && (
+        <p className="mt-4">
+          {t(
+            'If there is sufficient liquidity, the market is in an auction because there are no priced limit orders, which are required to deploy liquidity provision pegged orders. This means the order book is empty on one or both sides.'
```
candida-d, line 127: Will a sentence be displayed that also says the market is in an auction because there isn't enough supplied liquidity? Without that, it reads alike something's missing.

Alternatively, I think you could cut the first part of the sentence so it reads: `The market is in an auction because there are no priced limit orders, which are required to deploy liquidity commitment pegged orders. This means the order book is empty on one or both sides.`

MadalinaRaicu: Nice! I like that. But I should check with @barnabee too.

MadalinaRaicu: The sentence will be displayed only when supplied > target.

macqbat (approved):

PR #2548, opened 2023-01-08
feat: view apps as any pub key

### LINKED ISSUE #2598: View as vega user
## Story

As a user
I want to view console/token as a specific pubkey
So that i can easily inspect chain data

## Acceptance Criteria

- [ ] I can enter a pubkey 
- [ ] I can specify a pubkey via a url param
- [ ] I can disconnect my pubkey

dexturr:
# Related issues 🔗

Closes #2598 

# Description ℹ️

Allows users to input a pubkey into token and console and then view the site as if they were that user.

### DIFF
--- a/apps/token/src/app-loader.tsx
+++ b/apps/token/src/app-loader.tsx
@@ -27,7 +27,7 @@ const useVegaWalletEagerConnect = () => {
   if (query && !pubKey) {
     connect(Connectors['view']);
   }
-  return vegaConnecting
+  return vegaConnecting;
 };
 
 export const AppLoader = ({ children }: { children: React.ReactElement }) => {
--- a/apps/token/src/app-loader.tsx
+++ b/apps/token/src/app-loader.tsx
@@ -22,13 +22,11 @@ import { useSearchParams } from 'react-router-dom';
 const useUrlViewOnlyWalletConnect = () => {
   const { pubKey, connect } = useVegaWallet();
   const [searchParams] = useSearchParams();
-  const [query] = React.useState(
-    searchParams.get("address")
-  );
+  const [query] = React.useState(searchParams.get('address'));
   if (query && !pubKey) {
-    connect(Connectors['view'])
+    connect(Connectors['view']);
   }
-}
+};
 
 export const AppLoader = ({ children }: { children: React.ReactElement }) => {
   const { t } = useTranslation();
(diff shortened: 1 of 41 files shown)


apps/token/src/app-loader.tsx
```diff
@@ -17,17 +17,28 @@ import {
 import { useContracts } from './contexts/contracts/contracts-context';
 import { useRefreshAssociatedBalances } from './hooks/use-refresh-associated-balances';
 import { Connectors } from './lib/vega-connectors';
+import { useSearchParams } from 'react-router-dom';
+
+const useUrlViewOnlyWalletConnect = () => {
```
dexturr, line 22: move into helper

apps/token/src/components/token-input/token-input.tsx
```diff
@@ -100,6 +101,7 @@ export const TokenInput = ({
   approveText?: string;
   approveTxState?: TransactionState;
   approveTxDispatch?: React.Dispatch<TransactionAction>;
+  disabled?: boolean;
```
dexturr, line 104: remove all disabled

apps/token/src/lib/vega-connectors.ts
```diff
@@ -1,9 +1,17 @@
-import { RestConnector, JsonRpcConnector } from '@vegaprotocol/wallet';
+import {
+  RestConnector,
+  JsonRpcConnector,
+  ViewConnector,
+} from '@vegaprotocol/wallet';
+
+const urlParams = new URLSearchParams(window.location.search);
```
dexturr, line 7: do for console

apps/token/src/app-loader.tsx
```diff
@@ -17,6 +17,18 @@ import {
 import { useContracts } from './contexts/contracts/contracts-context';
 import { useRefreshAssociatedBalances } from './hooks/use-refresh-associated-balances';
 import { Connectors } from './lib/vega-connectors';
+import { useSearchParams } from 'react-router-dom';
+
+const useVegaWalletEagerConnect = () => {
+  const vegaConnecting = useEagerConnect(Connectors);
+  const { pubKey, connect } = useVegaWallet();
+  const [searchParams] = useSearchParams();
+  const [query] = React.useState(searchParams.get('address'));
```
edd, line 26: **thought (blocking)**: I think we need to make this check that address is a valid public key. I chose to be difficult and use `network`, which is valid, but I think at least requiring it is x length and valid makes sense?

edd (approved):
I used this on console, and it worked nicely. However the 'Submit Order' button was still allowed (I was viewing as `network`. Is that a follow up ticket?

dexturr:
> I used this on console, and it worked nicely. However the 'Submit Order' button was still allowed (I was viewing as `network`. Is that a follow up ticket?

@edd I discussed it with UX and their preference was that a message was shown on transaction attempt that says "You are connected in a view only state for public key: ${this.pubkey}. In order to send transactions you must connect to a real wallet."

So this is desired behaviour, moreover it makes it much easier to implement, if people want the buttons disabled we can do that as a follow up 👍

PR #2549, opened 2023-01-09
feat(2093): token sub nav

### LINKED ISSUE #2093: Reorganise token features into sub navigation
## The Chore

- [ ] There is a Token Home with existing landing page content
- [ ] Under VEGA token is a submenu / nav with: 
* Supply and vesting
* Withdraw assets 
* Redeem vested tokens (not in nav though? Or?)
- [ ] "Supply and vesting" links to the tranches page that's currently not in the nav 
- [ ] "Withdraw assets" links to the withdrawal page
- [ ] "Redeem vested tokens" links to the redeem page [if it's in the nav...]
- [ ] There is a useful and functioning version that works with no wallets connected

## Tasks

- [ ] and then what?
- [ ] Etc.

dexturr:
# Related issues 🔗

Closes #2093

# Description ℹ️

Adds token subnav.

# Demo 📺
Screenshot 2023-01-09 at 10 52 47
Screenshot 2023-01-09 at 10 52 56

### DIFF
--- a/apps/token-e2e/src/support/common.functions.js
+++ b/apps/token-e2e/src/support/common.functions.js
@@ -18,10 +18,12 @@ const navigation = {
   token: '[href="/token"]',
 };
 
+const topLevelRoutes = ['proposals', 'validators', 'rewards']
+
 Cypress.Commands.add('navigate_to', (page) => {
   const tokenDropDown = 'state-trigger';
 
-  if (page != 'proposals' && page != 'validators' && page != 'rewards') {
+  if (!topLevelRoutes.includes(page)) {
     cy.getByTestId(tokenDropDown, { timeout: 10000 }).click();
     cy.getByTestId('token-dropdown').within(() => {
       cy.get(navigation[page]).click();
--- a/apps/token-e2e/src/support/common.functions.js
+++ b/apps/token-e2e/src/support/common.functions.js
@@ -18,7 +18,7 @@ const navigation = {
   token: '[href="/token"]',
 };
 
-const topLevelRoutes = ['proposals', 'validators', 'rewards']
+const topLevelRoutes = ['proposals', 'validators', 'rewards'];
 
 Cypress.Commands.add('navigate_to', (page) => {
   const tokenDropDown = 'state-trigger';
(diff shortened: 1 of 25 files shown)


apps/token/src/components/nav/nav-dropdown.tsx
```diff
@@ -0,0 +1,206 @@
+import classNames from 'classnames';
```
sam-keen, line 1: Question (non-blocking): I know a lot of the navbar work is now shareable between apps, shouldn't this be in the ui-toolkit too?

apps/token/src/components/nav/nav-dropdown.tsx
```diff
@@ -0,0 +1,206 @@
...
+      'h-px my-1 mx-2 bg-neutral-700 dark:bg-black',
+      className
+    )}
+  />
+));
+
+export const NavDropDown = ({ navbarTheme }: { navbarTheme: NavbarTheme }) => {
+  const { t } = useTranslation();
+  const [isOpen, setOpen] = useState(false);
+  const subRoutes = [
```
sam-keen, line 153: Suggestion: I think we'd get some value from this config being separated out. We could then use it to not only build this menu, but also improve the cypress common function for page navigation

apps/token-e2e/src/support/common.functions.js
```diff
@@ -14,13 +14,23 @@ const navigation = {
...
   token: '[href="/token"]',
 };
 
 Cypress.Commands.add('navigate_to', (page) => {
-  return cy.get(navigation.section, { timeout: 10000 }).within(() => {
-    cy.get(navigation[page]).click({ force: true });
-  });
+  const tokenDropDown = 'state-trigger';
+
+  if (page != 'proposals' && page != 'validators' && page != 'rewards') {
```
sam-keen, line 24: Suggestion: I'm not a big fan of these routes being hard coded here. I've dropped a comment below about separating out the page structure from the subnav element, that config could then be used here to build this conditional

sam-keen (approved):

PR #2563, opened 2023-01-10
fix(#2553): show rejection reason & orderbook fixed dp

### LINKED ISSUE #2553: Tell me if my order was not fully filled due to self trading rule
## Story

As a trader
I want to see why my order was not filled, as I would expect, due to the self trading rule
So that I am not left guessing why a market order IOC or aggressive limit order got `partiallyFilled` when I wouldn't expect

Background: A market order (with IOC time in force) should always be filled (assuming market is in continuous and there is enough liquidity on the book). As should an aggressive limit order. I (as a trader) was  confused as to why my market order was not filled, as well as why my limit order was both unfilled and not in the book (it was GTC). Turns out it was because I had orders in the book that would have matched with the order I just placed. This is not permitted on Vega.

The API returns as `"rejectionReason": "ORDER_ERROR_SELF_TRADING"` on the order. - This should be used as a flag to highlight to the user what has happened.

I've written this ticket as a user story for my specific scenario.  But it has also been suggested that we could show the rejected reason whenever there is one. not just when status == partiallyfilled or rejected

## Acceptance Criteria

- [ ] on orders list: I can see that there is a reason that the order was not filled
- [ ] on orders list:I can see what the reason was
- [ ] on order feedback: I can see that there is a reason that the order was not filled
- [ ] on order feedback: I can see what the reason was

Here is [a PR ](https://github.com/vegaprotocol/specs/pull/1501)to add acs for this ticket to the mono repo for testing. Please ensure E2E tests are delivered as part of this or in a follow up ticket. 

## Tasks

- [ ] Explore and sketch
- [ ] Visual Design
- [ ] Team review
- [ ] Etc.

## Sketch

Here is how rejected reason is shown for orders with a status of rejected
Screenshot 2023-01-09 at 15 00 11

an example of a rejected reason

```
{
  party(id:"56d1e6739deac3c5c1ddc6fee876b3217e504a161b5b00fda96b40ed3e8f89b8"){
    ordersConnection{
      edges{
        node{
          rejectionReason
        }
      }
    }
    positionsConnection{
      edges{
        node{
          openVolume
          market{
            tradableInstrument{
              instrument{
                name
              }
            }
          }
        }
      }
    }
  }
  
}
```
gives
```
{
  "data": {
    "party": {
      "ordersConnection": {
        "edges": [
          {
            "node": {
              "rejectionReason": null
            }
          },
          {
            "node": {
              "rejectionReason": null
            }
          },
          {
            "node": {
              "rejectionReason": "ORDER_ERROR_SELF_TRADING"
            }
...
```

MadalinaRaicu:
# Related issues 🔗

Closes #2553 

# Description ℹ️

- [x] show rejection reason whenever it is available in both order feedback and list
- [x] show fixed order book dp

# Demo 📺

# Technical 👨‍🔧


apps/trading-e2e/src/integration/trading-orders.cy.ts
```diff
@@ -221,7 +221,10 @@ describe('subscribe orders', { tags: '@smoke' }, () => {
       id: orderId,
       status: Schema.OrderStatus.STATUS_PARKED,
     });
-    cy.getByTestId(`order-status-${orderId}`).should('have.text', 'Parked');
+    cy.getByTestId(`order-status-${orderId}`).should(
+      'have.text',
+      'Parked: Internal error'
```
MadalinaRaicu, line 226: @radoslaw-szpiech here I have a comment that I am n ot sure if I should keep anymore.
```
    // 7002-SORD-048
    // NOT COVERED:   must see an explanation of why parked orders happen
```

MadalinaRaicu: we are now showing the rejection reason but the rejection reason in this test is "Internal error". It could be sth else, do we need to change the mock? should we?

radoslaw-szpiech: I think it's cool. Please, remove `NOT COVERED` comment, leave `SORD` for AC coverage.

MadalinaRaicu: awesome. updated.

radoslaw-szpiech (approved):

PR #2564, opened 2023-01-10
fix(#2216): lp-dashboard grab market close from metadata

### LINKED ISSUE #2216: Time left and age e.g. "Settles in 7 days" "Market is 3 months old"  *
## The Chore

Add in the duration the market has left and the age of the market.

## Acceptance criteria 

* I can see for each market how old it is 
* I can see for each market when it settles 

## Tasks

- [ ] Ask from the frontend team if the market settlement date / timestamp is only in the governance API / has anybody already pulled this out via integration somewhere e.g. "expiry" in console
- [x] Implement market age "market is 6 days old"
- [ ] Etc.

ciaran-:
# Related issues 🔗

Closes #2216 

# Description ℹ️

Supersedes #2529 which attempted to display market settlement/closing time, but was incorrectly displaying instead the close time of _voting_ on the market.

This PR instead reads metadata and re-uses a helper to look for a market expiry, which if we find one is displayed on the dashboard for that market.

# Demo 📺

# Technical 👨‍🔧

@dexturr I've attached a screenshot of the nx dep graph as I know you mentioned we might want to refactor where the `getMarketExpiryDate` helper lives if it was too gnarly. What do you think?

Screenshot 2023-01-10 at 10 28 19

[Edit]
...and after the refactor, it' looks like this:
Screenshot 2023-01-24 at 14 07 36


ciaran-:
@dexturr confused by this unit test failure I'm getting - I initially thought I might have messed up refactoring `getExpiryDate` to avoid using the import we discussed, but the latest commit on this PR doesn't include moving & refactoring that function and still fails in the same way. This test isn't failing for me locally, which is the part confusing me most!

Any ideas?

libs/react-helpers/src/lib/market-expires.tsx
```diff
@@ -36,12 +36,14 @@ export const getMarketExpiryDateFormatted = (
...
 };
 
-export const getExpiryDate = (market: SingleMarketFieldsFragment): string => {
-  const metadataExpiryDate = getMarketExpiryDate(
-    market.tradableInstrument.instrument.metadata.tags
-  );
-  const marketTimestampCloseDate =
-    market.marketTimestamps.close && new Date(market.marketTimestamps.close);
+export const getExpiryDate = (
+  //   market: SingleMarketFieldsFragment,
```
MadalinaRaicu, line 40: why this field is commented?

ciaran-: Oh that's an oversight, sorry - I'll fix!

MadalinaRaicu: seems like all params of this method are optional. where can it get the data from?

ciaran-: It should have been removed when I expanded the arguments the function expects.

ciaran-: > seems like all params of this method are optional. where can it get the data from?

This is a good point - I wanted to avoid refactoring too much of the internals of this function as well as the signature, but I think you're right it doesn't make sense to have all these optional params. I'll rework this too.

ciaran-: OK I've updated this now @MadalinaRaicu if you want to have another look!

MadalinaRaicu (approved):

PR #2575, opened 2023-01-11
feat(2396): token delegations pagination

### LINKED ISSUE #2396: Fix delegations queries, to prevent mainnet incident
## The Chore

The current delegations queries request all delegeations. Due to the mainnet incident which occured, when the number of epochs becomes large then this causes many joins and long response times.

We have a paginated API for this, so by requesting only 50 of the most recent delegations this would guarantee at least a reasonable response time.

## Tasks

- [x] Use pagination within the delegations query

sam-keen:
# Related issues 🔗

Closes #2396

# Description ℹ️

To avoid any further mainnet incidents (where delegations queries swamped), set a pagination limit for Token delegations queries

# Technical 👨‍🔧

Limit controlled via an env var


libs/environment/src/utils/compile-environment.ts
```diff
@@ -82,6 +82,8 @@ const getBundledEnvironmentValue = (key: EnvKey) => {
       return (
         process.env['MAINTENANCE_PAGE'] || process.env['NX_MAINTENANCE_PAGE']
       );
+    case 'DELEGATIONS_PAGINATION':
+      return process.env['NX_DELEGATIONS_PAGINATION'];
```
dexturr, line 86: chore (blocking): the environemnet provider is for things shared across apps. I would just use this in the env.ts file rather than here

apps/token/src/routes/rewards/home/Rewards.graphql
```diff
@@ -20,7 +20,7 @@ fragment DelegationFields on Delegation {
   epoch
 }
 
-query Rewards($partyId: ID!) {
+query Rewards($partyId: ID!, $delegationsPagination: Pagination) {
```
dexturr, line 23: question: this says it's of pagination type but as near as I can tell this is provided a numebr?

sam-keen: answered in dm

dexturr (approved):

MadalinaRaicu (approved):

jeremyletang (approved):

PR #2607, opened 2023-01-12
feat: update market data components when they are in view

### LINKED ISSUE #2530: Trading mode in market selector and header do not match

bglownia:
# Related issues 🔗

Closes #2530

# Description ℹ️

Live updates enabled for all rows in market selector. To limit number of active subscriptions only components that are "in view" are subscribe for updates

# Technical 👨‍🔧

Intersection observer via hook with proper root is used to check if component is visible and should be subscribed for updated.
Static market data provider is added. It expose only low frequency updated market data.
Throttling of update in components via market provider update and setState is replaced by usage of throttled data provider hook


apps/trading/components/last-24h-volume/last-24h-volume.tsx
```diff
@@ -1,102 +1,61 @@
...
-      if (data) {
-        throttledSetCandles(data);
-      }
-      return true;
-    },
-    [throttledSetCandles]
-  );
-
-  const { error } = useDataProvider<Candle[], Candle>({
+  const { data } = useDataProvider<Candle[], Candle>({
```
macqbat, line 44: in such cases we're loosing memoization: `data` will be always a new object and will cause a re-render

macqbat: and `useThrottledDataProvider` shouldn't be use here?

apps/trading/components/market-mark-price/market-mark-price.tsx
```diff
@@ -1,84 +1,64 @@
...
-  const throttledSetMarketPrice = useRef(
-    throttle((price: string) => {
-      noUpdate || setMarketPrice(price);
-    }, constants.DEBOUNCE_UPDATE_TIME)
-  ).current;
-  const update = useCallback(
-    ({ data: marketData }: { data: MarketData | null }) => {
-      throttledSetMarketPrice(marketData?.markPrice || '');
-      return true;
+  const { data } = useThrottledDataProvider<
```
macqbat, line 35: yeah, this `data` I don't like too :-(

apps/trading/components/select-market/select-market.spec.tsx
```diff
@@ -151,6 +151,16 @@ const MARKET_B: PartialMarket = {
 };
 
 describe('SelectMarket', () => {
+  beforeEach(() => {
+    // IntersectionObserver isn't available in test environment
```
macqbat, line 155: instead this you can add to the jest setup file:

```
import { defaultFallbackInView } from 'react-intersection-observer';
defaultFallbackInView(true);
```

apps/trading/components/welcome-dialog/welcome-landing-dialog.spec.tsx
```diff
@@ -149,6 +149,16 @@ const MARKET_B: PartialMarket = {
 };
 
 describe('WelcomeLandingDialog', () => {
+  beforeEach(() => {
+    // IntersectionObserver isn't available in test environment
```
macqbat, line 153: like above...

libs/react-helpers/src/lib/grid/price-cell.tsx
```diff
@@ -1,37 +1,44 @@
...
-    if (
-      (!value && value !== 0) ||
-      (typeof value === 'number' && isNaN(Number(value)))
-    ) {
-      return <span data-testid="price">-</span>;
+export const PriceCell = memo(
+  forwardRef<HTMLSpanElement, IPriceCellProps>(
+    ({ value, valueFormatted, testId }: IPriceCellProps, ref) => {
+      if (
+        (!value && value !== 0) ||
```
macqbat, line 13: this whole messy condition can be probably replaced by  `isNumeric` from `react-helpers `

MadalinaRaicu (approved):

PR #2623, opened 2023-01-16
chore: make ag-grid cells copyable

MadalinaRaicu:
# Related issues 🔗

N/A

# Description ℹ️

Make ag-grid cells copyable.


mattrussell36 (changes requested):
Can we just make this part of the default props for tables? See my comment here

PR #2627, opened 2023-01-16
fix(#2569): market id copy update

### LINKED ISSUE #2569: Show market ID on markets table
## The Chore

Show the market Id on the markets table

MadalinaRaicu:
# Related issues 🔗

Closes #2569

# Description ℹ️

- [x] show market ID in markets table
- [x] make it copyable
- [x] update description

# Demo 📺

Screenshot 2023-01-16 at 17 32 11

# Technical 👨‍🔧


libs/market-list/src/lib/components/markets-container/market-list-table.tsx
```diff
@@ -30,6 +30,7 @@ export const MarketListTable = forwardRef<
   return (
     <AgGrid
       style={{ width: '100%', height: '100%' }}
+      enableCellTextSelection={true}
```
mattrussell36, line 33: suggestion (non blocking): Can we add this as a default prop to:

libs/ui-toolkit/src/components/ag-grid/ag-grid-lazy-themed.tsx
and
libs/ui-toolkit/src/components/ag-grid/ag-grid-dynamic-themed.tsx

I think you should be able to highlight and copy cells by default on all tables

mattrussell36 (approved):

libs/market-list/src/lib/components/markets-container/market-list-table.tsx
MadalinaRaicu: updated!

PR #2652, opened 2023-01-17
fix(#2345): positions table link to markets

### LINKED ISSUE #2345: Get to market page from the positions table
## Story

When looking at my positions
I want to click to the market of a position
so I can look at the state of the market (orderbook, chart etc)

## Acceptance Criteria

- [ ] I can click on a market from the positions view (inc portfolio) to get to the page of that market

MadalinaRaicu:
# Related issues 🔗

Closes #2345 

# Description ℹ️

Positions table - link to market.

# Demo 📺

image

# Technical 👨‍🔧


libs/positions/src/lib/positions-table.tsx
```diff
@@ -85,7 +85,20 @@ export const PositionsTable = forwardRef<AgGridReact, Props>(
...
-        <AgGridColumn headerName={t('Market')} field="marketName" />
+        <AgGridColumn
+          headerName={t('Market')}
+          field="marketName"
+          cellRenderer={({
+            value,
+            data,
+          }: VegaICellRendererParams<Position, 'marketName'>) =>
+            value ? (
+              <Link href={`/#/markets/${data?.marketId}`}>{value}</Link>
```
mattrussell36, line 96: suggestion (blocking): A few things to consider here

1. Hardcoding the '/#/' could be problematic. Other apps don't use hash routing (Of course the trading app is the only one using this component, but we should try to follow the principles of the monorepo). Normally you should use the Link component from react-router-dom and it will handle the hash for you.
2. @asiaznik  recently added some helpers for url construction it looks something like `to={Links[Routes.MARKET](market.id)}`

What I would say is given that this is in a lib. The positions table should take a new optional prop `onMarketClick` which you can use to pass a function `() => navigate(Links[routes.MARKET])(market.id)`.

Unfortunately this results in quite a bit of prop drilling, and might not be the best solution long term but for now it will be more robust

mattrussell36 (changes requested):

libs/positions/src/lib/positions-table.tsx
mattrussell36: Also to ensure that we get consistent link styles, you can still use the ui-toolkit `Link` component with an onClick handler, just omit the href prop and it will render a span with the same styles

MadalinaRaicu: ok, will do these changes to Orders and Fills tabs too

MadalinaRaicu: updated. please review!

mattrussell36 (approved):
If you have capacity, it might be nice to add a unit test for each table to click on the cell and ensure the callback is called with the right marketId

PR #2655, opened 2023-01-17
chore: ported token's web3 url connector to web3 lib

asiaznik:
# Related issues 🔗

Closes #[Issue number here]

# Description ℹ️

# Demo 📺

# Technical 👨‍🔧

### DIFF
--- a/libs/web3/src/lib/url-connector.ts
+++ b/libs/web3/src/lib/url-connector.ts
@@ -1,0 +1,170 @@
+import { ethers, Wallet } from 'ethers';
+import { Connector } from '@web3-react/types';
+import { Eip1193Bridge } from '@ethersproject/experimental';
+import type { ConnectionInfo } from '@ethersproject/web';
+import type { Actions } from '@web3-react/types';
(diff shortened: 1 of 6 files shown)


daro-maj (approved):

libs/web3/src/lib/url-connector.ts
```diff
@@ -0,0 +1,170 @@
+import { ethers, Wallet } from 'ethers';
+import { Connector } from '@web3-react/types';
+import { Eip1193Bridge } from '@ethersproject/experimental';
+import type { ConnectionInfo } from '@ethersproject/web';
+import type { Actions } from '@web3-react/types';
+import { initializeConnector } from '@web3-react/core';
+
+export class CustomizedBridge extends Eip1193Bridge {
```
dexturr, line 8: chore (blocking): can we move this into it's own file, doesn't feel like it belongs here, and can we make it so token is using it as well rather than duplicating :)

MadalinaRaicu (approved):

MadalinaRaicu (approved):

PR #2670, opened 2023-01-19
chore: remove console lite

### LINKED ISSUE #2640: Ignore / remove console lite from frontend-monorepo to remove all maintenance overhead
## The Chore

As we are unsure if / when work will start again on console-lite, and even if it does restart it could take a completely different form / focus etc. It has been decided that we will stop supporting Console-lite as it is becoming an overhead to keep the tests passing and otherwise maintaining the code as the other apps and shared libraries are rapidly developed. 

In discussion with the frontend leads, everyone is on board with this and open to whether we achieve this by leaving the code in the frontend mono-repo and just doing whatever needs to be done to just ignore it in test runs, builds etc. OR whether we actually move the code out into its own repository. Essentially whatever does the job to remove it as an overhead with least effort is fine with the teams. 

## Tasks

- [ ] TBC in refinement

I have just added the Trading label to this ticket as we agreed that the console team would do the work

dexturr:
# Related issues 🔗

Closes #2640

# Description ℹ️

Remove console lite code as maintaining a project that is not deployed or being actively worked on is slowing us down

It now lives here and is in a runnable state: https://github.com/vegaprotocol/console-lite


jtsang586 (approved):
Lgtm

notbot00 (approved):

edd (approved):

pscott31 (approved):

PR #2677, opened 2023-01-19
chore: simplify cypress workflows

### LINKED ISSUE #2592: Simplify or unify cypress workflows
We created separate workflows for each app, as they used different approaches (capsule/non-capsule, wallet, etc.). After #2255 they will all use capsule config, so there will be no difference between them. We may try to simplify them (e.g. using https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs ).

- add Node 16 support

radoslaw-szpiech:
# Related issues 🔗

Closes #2592 

# Description ℹ️

All cypress tests are using the same workflow now, as all of them use the same env config. Also simplified and introduced matrix strategy to run multiple projects e2e tests.


.github/workflows/cypress-pr.yml
```diff
@@ -0,0 +1,76 @@
...
+
+      # See affected apps
+      - name: See affected apps
+        run: |
+          affected=$(yarn nx print-affected --base=${{ env.NX_BASE }} --head=${{ env.NX_HEAD }} --select=projects)
+          projects=""
+          if [[ $affected == *"token"* ]]; then projects+='"token-e2e" '; fi
+          if [[ $affected == *"trading"* ]]; then projects+='"trading-e2e" '; fi
+          if [[ $affected == *"explorer"* ]]; then projects+='"explorer-e2e" '; fi
+          if [[ $affected == *"console-lite"* ]]; then projects+='"console-lite-e2e" '; fi
```
jtsang586, line 47: Console-lite got removed from the repo today so can be removed from all the workflows

radoslaw-szpiech: Awesome, I removed console-lite references in workflows.

daro-maj (approved):

PR #2690, opened 2023-01-20
feat(#2565): orderbook populate limit

### LINKED ISSUE #2565: Populate limit price when clicking on order book cell
## Story

**As a** trader using console
**I want** to click on one of the order book cells
**And** populate the limit price in the deal ticket
**So** that I can easily place orders

## Acceptance Criteria

- [ ] click on order book cells and populate limit price

**Feedback from community**

Image

MadalinaRaicu:
# Related issues 🔗

Closes #2565 

# Description ℹ️

Orderbook and trading populate limit price. 

# Demo 📺

Screenshot 2023-01-19 at 15 27 32

# Technical 👨‍🔧

- Adds a zustand store with persist middleware for storing all orders across all markets.
- Updates form state after a change to the store


libs/deal-ticket/src/components/deal-ticket/deal-ticket.tsx
```diff
@@ -84,8 +96,22 @@ export const DealTicket = ({ market, submit }: DealTicketProps) => {
     errors.summary?.type,
   ]);
 
+  useEffect(() => {
+    const priceUpdater = (event: CustomEvent) => {
```
mattrussell36, line 100: I'm not sure about this approach, I'd prefer to keep things in the react paradigm.

What I would suggest is to update the persisted order store to store a map of markets to orders, then we can just update the store as necessary and the deal ticket should receive the updates.

This also fits better with a suggestion I had for @asiaznik to update the local storage entries for the deal ticket to also be a map of markets to orders

mattrussell36:
```
const [persistedOrder, setPersistedOrder] = usePersistedOrder(market);
const {
  register,
  ...
  setValue
} = useForm({
  defaultValues: persistedOrder || getDefaultOrder(market),
});

...

// store local form changes
useEffect(() => setPersistedOrder(order), [order, setPersistedOrder]);

// update form from external changes
useEffect(() => {
  // set changed fields
}, [persistedOrder])
```

mattrussell36: This does make me feel like react-hook-form doesn't provide a lot of value here, instead it means we have to maintain order state in two places

MadalinaRaicu: So what you are suggesting was my initial approach and I discussed about this with @asiaznik as well. 

I ended up dispatching a custom event in order to avoid circular updating. 

The useEffect hooks from the example above would cause a circular update and the app will crash. 

The custom event seemed to be the simplest and cleanest solution.

mattrussell36: I see what you mean. Looks like the loop is caused by the persisted store and the useForm state relying on each other to stay in sync. I think it would be worth a refactor so that we have a single zustand store (with the persist middleware) for all orders across all markets. Then we should be able to just pass in the values to our already controlled form components. We may still need to update the form state after each update to ensure we get all the dirty field checking and validation that react-hook-form provides

asiaznik: 👋🏻 Folks, I have an idea on how to approach this with a little of the zustand magic.

asiaznik (approved):

mattrussell36 (approved):

PR #2696, opened 2023-01-23
feat: logging - working solution

### LINKED ISSUE #644: Logging

macqbat:
# Related issues 🔗

Closes #644 

# Description ℹ️

Add a logger module as a simple, easy to maintain and to extend, single js class. Use it to log apollo errors

# Demo 📺

- 

# Technical 👨‍🔧

-

### DIFF
--- a/libs/react-helpers/src/lib/local-logger.ts
+++ b/libs/react-helpers/src/lib/local-logger.ts
@@ -1,7 +1,18 @@
 import * as Sentry from '@sentry/browser';
-import type { SeverityLevel, Scope } from '@sentry/browser';
+import type { Scope } from '@sentry/browser';
 import type { Severity, Breadcrumb, Primitive } from '@sentry/types';
 
+const LogLevels = [
+  'fatal',
+  'error',
+  'warning',
+  'log',
+  'info',
+  'debug',
+  'critical',
+  'silent',
+];
+type LogLevelsType = typeof LogLevels[number];
 type ConsoleArg = string | number | boolean | bigint | symbol | object;
 type ConsoleMethod = {
   [K in keyof Console]: Console[K] extends (...args: ConsoleArg[]) => unknown
@@ -13,7 +24,7 @@ type ConsoleMethod = {
 interface LoggerConf {
   application?: string;
   tags?: string[];
-  logLevel?: SeverityLevel;
+  logLevel?: LogLevelsType;
 }
 
 const isPrimitive = (arg: ConsoleArg | undefined | null): arg is Primitive => {
@@ -23,7 +34,7 @@ const isPrimitive = (arg: ConsoleArg | undefined | null): arg is Primitive => {
 };
 
 export class LocalLogger {
-  static levelLogMap = {
+  static levelLogMap: Record<LogLevelsType, number> = {
     debug: 10,
     info: 20,
     log: 30,
@@ -33,7 +44,7 @@ export class LocalLogger {
     fatal: 70,
     silent: 80,
   };
-  private _logLevel: SeverityLevel = 'info';
+  private _logLevel: LogLevelsType = 'info';
   private get numberLogLevel() {
     return LocalLogger.levelLogMap[this._logLevel];
   }
@@ -68,7 +79,7 @@ export class LocalLogger {
     this._log('fatal', 'error', args);
   }
   private _log(
-    level: SeverityLevel,
+    level: LogLevelsType,
     logMethod: ConsoleMethod,
     args: ConsoleArg[]
   ) {
@@ -81,7 +92,7 @@ export class LocalLogger {
     this._transmit(level, args);
   }
   private _extractArgs(
-    level: SeverityLevel,
+    level: LogLevelsType,
     args: ConsoleArg[]
   ): [string, Error, Scope] {
     const arg = args.shift();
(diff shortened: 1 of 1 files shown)


libs/react-helpers/src/lib/local-logger.ts
```diff
@@ -0,0 +1,119 @@
+import * as Sentry from '@sentry/browser';
+import type { SeverityLevel, Scope } from '@sentry/browser';
+import type { Severity, Breadcrumb, Primitive } from '@sentry/types';
+
+type ConsoleArg = string | number | boolean | bigint | symbol | object;
```
mattrussell36, line 5: suggestion (non blocking): 
```
type ConsoleArg = Parameters<typeof console.log>
```

libs/react-helpers/src/hooks/use-logger.ts
```diff
@@ -0,0 +1,27 @@
...
+import type { LocalLogger } from '../lib/local-logger';
+import { localLoggerFactory } from '../lib/local-logger';
+
+interface Props {
+  dsn?: string;
+  application?: string;
+  tags?: string[];
+}
+
+export const useLogger = ({ dsn, ...props }: Props) => {
```
mattrussell36, line 13: question (non blocking): Whats the advantage of providing a hook?  Personally, I'd prefer to keep the component code cleaner and just instantiate the logger directly in the module file.

libs/react-helpers/src/lib/local-logger.ts
```diff
@@ -0,0 +1,119 @@
...
+const isPrimitive = (arg: ConsoleArg | undefined | null): arg is Primitive => {
+  return ['string', 'number', 'boolean', 'bigint', 'symbol'].includes(
+    typeof arg
+  );
+};
+
+export class LocalLogger {
+  tags: string[] = [];
+  application = 'trading';
+  constructor(conf: LoggerConf) {
```
mattrussell36, line 22: suggestion (non blocking): not sure if you looked at all at https://www.npmjs.com/package/loglevel

But it provides a nice way to set the log level with something like `logger.setLevel('info')`

I think it would be useful addition (with or without the library) to be able to call something like window.logging.setLevel('info') via the console so we can inspect deployed apps easily. By default this would be on 'silent' so no logs are shown.

libs/react-helpers/src/hooks/use-logger.ts
macqbat: hook gives additional possibility of initialization sentry. It can be useful in other apps than those build on nextjs, where initialization is config file

libs/react-helpers/src/lib/local-logger.ts
macqbat: So I implemented something like such functionality. Please check if this is it

libs/react-helpers/src/lib/local-logger.ts
macqbat: It would break a little bit further concept of typing params to the sentry :-(

libs/react-helpers/src/lib/local-logger.ts
```diff
@@ -0,0 +1,163 @@
...
+    return [msg, error || new Error(msg), scope];
+  }
+  private _transmit(level: LogLevelsType, args: ConsoleArg[]) {
+    const [msg, error, logEvent] = this._extractArgs(level, args);
+    switch (level) {
+      case 'debug':
+      case 'info':
+      case 'log':
+      case 'warning':
+        Sentry.captureMessage(msg, logEvent);
```
mattrussell36, line 133: chore (blocking): I don't think we should call captureMessage for each of these 4 levels. I think there is a use case where we might want to log without capturing the log in Sentry. Also I think it would make Sentry rather noisy with all these messages. Perhaps just call captureMessage for 'debug'?

mattrussell36 (approved):

libs/react-helpers/src/lib/local-logger.ts
```diff
@@ -0,0 +1,163 @@
+import * as Sentry from '@sentry/browser';
+import type { Scope } from '@sentry/browser';
+import type { Severity, Breadcrumb, Primitive } from '@sentry/types';
+
+const LogLevels = [
+  'fatal',
+  'error',
+  'warning',
+  'log',
```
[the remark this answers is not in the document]
macqbat: I thought we can just mimic what is in console object. In some cases aliases are useful.

libs/react-helpers/src/lib/local-logger.ts
macqbat: I think all we log on the client side can be useful and we should index it. Otherwise why log anything at all?

libs/react-helpers/src/lib/local-logger.ts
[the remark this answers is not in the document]
macqbat: Moreover sentry also distinguishes so many levels, and we can/should do it accordingly imho. The standards like distinguishes many levels quite clearly. `Critical` was also an AC in the task

PR #2701, opened 2023-01-23
feat(#1791): block height check

### LINKED ISSUE #1791: Update nodeHealth to use `x-block-height` header to check data node health

```
< x-block-height: 926035
```

## Acceptance Criteria

## Tasks

mattrussell36:
# Related issues 🔗

Closes #1791 

# Description ℹ️

Checks a random subset of nodes every 30s to see if the your current nodes block height is up to date. This is shown in the UI by an indicator in the footer along with the text Opertaional, Non operational, X blocks behind or Offline

# Demo 📺

drawing

drawing

drawing

drawing

# Technical
- Updated to createClient to take a few more options, for retry and connect to dev tools. This tidies up the network quite a bit when attempting connections and also doesn't mess with Apollo dev tools
- Fixed broken/evergreen tests for environment provider. Looks like this has been broken since [this commit](https://github.com/vegaprotocol/frontend-monorepo/commit/71ede25339e7a5cd2c8acd8198d99e72244e9a02#diff-771a20a1909c122b0792e947e22ec89f81061f1b41782d1bf7324536b464d14a)
- Adds useNavigatorOnline hook to detect if browser is connected or not

### DIFF
--- a/apps/trading/components/footer/footer.tsx
+++ b/apps/trading/components/footer/footer.tsx
@@ -1,5 +1,5 @@
 import { useEnvironment } from '@vegaprotocol/environment';
-import { t } from '@vegaprotocol/react-helpers';
+import { t, useNavigatorOnline } from '@vegaprotocol/react-helpers';
 import { ButtonLink, Indicator, Intent } from '@vegaprotocol/ui-toolkit';
 
 export const Footer = () => {
@@ -49,10 +49,15 @@ export const NodeHealth = ({
   blockDiff,
   openNodeSwitcher,
 }: NodeHealthProps) => {
+  const online = useNavigatorOnline();
+
   let intent = Intent.Success;
   let text = 'Operational';
 
-  if (blockDiff < 0) {
+  if (!online) {
+    text = t('Offline');
+    intent = Intent.Danger;
+  } else if (blockDiff < 0) {
     // Block height query failed and null was returned
     text = t('Non operational');
     intent = Intent.Danger;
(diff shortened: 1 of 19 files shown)


libs/environment/src/hooks/use-node-health.ts
```diff
@@ -0,0 +1,88 @@
...
+// Number of nodes to query against
+export const NODE_SUBSET_COUNT = 5;
+
+// Queries all nodes from the environment provider via an interval
+// to calculate and return the difference between the most advanced block
+// and the block height of the current node
+export const useNodeHealth = (clients: ClientCollection, vegaUrl?: string) => {
+  const [blockDiff, setBlockDiff] = useState(0);
+
+  useEffect(() => {
```
asiaznik, line 20: suggestion (non-blocking): I'm a little bit tempted to refactor this into `useSyncExternalStore` as the node health function is to literally sync with the block heights of other nodes. What do you think?

apps/trading/components/footer/footer.tsx
```diff
@@ -1,28 +1,70 @@
...
+const BLOCK_THRESHOLD = 3;
+
+export const NodeHealth = ({
+  blockDiff,
+  openNodeSwitcher,
+}: NodeHealthProps) => {
+  let intent = Intent.Success;
+  let text = 'Operational';
+
+  if (blockDiff < 0) {
```
asiaznik, line 55: question (non-blocking): How about a simple online/offline (browser, node) check here as well?

libs/environment/src/hooks/use-node-health.ts
mattrussell36: As discussed, the store is simply normal react state, so using useSyncExternalStore here doesn't provide much benefit

apps/trading/components/footer/footer.tsx
mattrussell36: Ill add a check and render 'Offline'

asiaznik (approved):

Issue #2587, opened 2023-01-11
 E2E tests for withdraw flow


mattrussell36:
## The Chore

Review current withdraw test coverage and add tests to cover withdraw UI specs using either mocks or capsule

## Tasks

- [x] Review current coverage of mocked tests and add any missing functionality
- [x] Add nightly capsule test run

## Additional details / background info

- UI specs https://github.com/vegaprotocol/specs/blob/master/user-interface/1002-WITH-withdraw.md

Issue #2585, opened 2023-01-11
E2E tests for deposit flow


mattrussell36:
## The Chore

Review current deposit test coverage and add tests

## Tasks

- [x] Check current coverage of mocked tests for deposits, and add any missing functionality
- [x] Add nightly capsule test run for deposits

## Additional details / background info

- Note: Deposit UI spec needs completing https://github.com/vegaprotocol/specs/blob/master/user-interface/1001-DEPO-desposit.md

Issue #2709, opened 2023-01-24
Deposit and withdrawal table doesn't update when success toast is rendered


daro-maj:
## The Chore

The deposit/withdrawals table does not update when the deposit/withdrawal is confirmed by toast message 

## Additional details / background info
![image](https://user-images.githubusercontent.com/119658839/214279376-b3b1b830-9129-4136-9faf-9ba31a0daab5.png)

Issue #2695, opened 2023-01-23
Edit order toast values are incorrect


daro-maj:
## Description

After editing the limit order, we have a new value on the toast confirmation instead of the old one in the line that shows the previous value

## Steps to Reproduce

1. Create limit order
2. Edit order - change e.g. price
3. Check value of old price (crossed out) on the toast confirmation ->  it is new one 

## Expected behavior

Old price value should be visible.

## Screenshots

![Screenshot 2023-01-23 at 13 08 32](https://user-images.githubusercontent.com/119658839/214036241-7967cb79-7bd5-4736-9e57-8a411bca0b9c.png)

PR #2720, opened 2023-01-24
chore: deposit and withdrawal capsule e2e tests

daro-maj:
# Related issues 🔗

#2587 
#2585 

# Description ℹ️

Added e2e tests flow for deposit and withdrawal in capsule.

### DIFF
--- a/apps/trading-e2e/src/integration/capsule.cy.ts
+++ b/apps/trading-e2e/src/integration/capsule.cy.ts
@@ -8,6 +8,7 @@ import {
 } from '@vegaprotocol/types';
 import { isBefore, isAfter, addSeconds, subSeconds } from 'date-fns';
 import { createOrder } from '../support/create-order';
+import { connectEthereumWallet } from '../support/ethereum-wallet';
 
 const orderSize = 'size';
 const orderType = 'type';
@@ -17,6 +18,16 @@ const orderPrice = 'price';
 const orderTimeInForce = 'timeInForce';
 const orderCreatedAt = 'createdAt';
 const orderUpdatedAt = 'updatedAt';
+const assetSelectField = 'select[name="asset"]';
+const amountField = 'input[name="amount"]';
+const txTimeout = Cypress.env('txTimeout');
+const btcName = 'BTC (local)';
+const btcSymbol = 'tBTC';
+const usdcSymbol = 'fUSDC';
+const toastContent = 'toast-content';
+const ordersTab = 'Orders';
+const depositsTab = 'Deposits';
+const toastCloseBtn = 'toast-close';
 
 // TODO: ensure this test runs only if capsule is running via workflow
 describe('capsule', { tags: '@slow' }, () => {
@@ -25,6 +36,7 @@ describe('capsule', { tags: '@slow' }, () => {
     cy.get('@markets').then((markets) => {
       cy.wrap(markets[0]).as('market');
     });
+    cy.updateCapsuleMultiSig();
   });
 
   beforeEach(() => {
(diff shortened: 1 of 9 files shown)


apps/trading-e2e/src/integration/capsule.cy.ts
```diff
@@ -116,26 +134,144 @@ describe('capsule', { tags: '@slow' }, () => {
...
+    connectEthereumWallet('Unknown');
+    cy.get(assetSelectField, txTimeout).select(btcName);
+    cy.getByTestId('deposit-approve-submit').click();
+    cy.getByTestId('dialog-title').should('contain.text', 'Approve complete');
+    cy.get('[data-testid="Return to deposit"]').click();
+    cy.get(amountField).clear().type('1');
+    cy.getByTestId('deposit-submit').click();
+    cy.getByTestId(toastContent, txTimeout).should(
+      'contain.text',
+      `Transaction completedYour transaction has been completedView on EtherscanDeposit 1.00 ${btcSymbol}`
```
mattrussell36, line 191: suggestion (non blocking): This is a little icky. Presumably the text is broken up by different elements `contain.text` treats it all as a joined string. I think this would be nicer if we targeted a single element containing the toast state which we can then assert is 'complete'

apps/trading-e2e/src/integration/capsule.cy.ts
```diff
@@ -116,26 +134,144 @@ describe('capsule', { tags: '@slow' }, () => {
...
+    );
+    cy.getByTestId(toastCloseBtn).click();
+    cy.getByTestId('Collateral').click();
+
+    cy.highlight('deposit verification');
+
+    cy.getByTestId('asset', txTimeout).should('contain.text', btcSymbol);
+    // need to reload page to see deposit history complete
+    cy.reload();
```
mattrussell36, line 200: question (non blocking): Is the reload really required? The deposit table should update via the subscription

apps/trading-e2e/src/integration/capsule.cy.ts
```diff
@@ -116,26 +134,144 @@ describe('capsule', { tags: '@slow' }, () => {
...
     cy.getByTestId('tab-orders')
       .get('.ag-center-cols-container')
       .children()
       .first()
-      .get(`[col-id='${orderStatus}']`)
+      .get(`[col-id='${orderStatus}']`, txTimeout)
       .should('contain.text', OrderStatusMapping.STATUS_CANCELLED);
   });
+
+  it('can deposit and withdrawal', function () {
```
mattrussell36, line 154: suggestion (non blocking): Can we split this into two separate tests one for deposit and one for withdrawal. We should probably also do the same for order submit, edit and cancel. I think even though each test relies on the previous test it should still result in better reporting if the test fails

mattrussell36 (approved):

apps/trading-e2e/src/integration/capsule.cy.ts
daro-maj: yes, I'm not a fan of this solution either, but currently the appearance of the deposit for the first time is very non-deterministic and takes time. I think after solving #2709  we will be able to get rid of this reload.

apps/trading-e2e/src/integration/capsule.cy.ts
daro-maj: it was done this way to save some time as we have dependent capsule tests anyway. I will check how the execution time will change after separating them and depending on the results we will decide what to do.

apps/trading-e2e/src/integration/capsule.cy.ts
daro-maj: Yes we can do it, but this is to check if everything really works e2e. This method found us a bug #2695

radoslaw-szpiech (approved):

PR #2733, opened 2023-01-25
feat: add seperate transaction toast for completed eth transaction

bglownia:
# Related issues 🔗

Closes #[Issue number here]

# Description ℹ️

Distinct Complete and Confirmed status of eth transaction in toast notification. 
Deposit transaction requires final confirmation by change in status of deposit from 'open' to 'finalized'. This change delays "Transaction confirmed" toast and displays "Processing deposit" toast until deposit will be finalized

# Demo 📺

image

### DIFF
--- a/libs/deposits/src/lib/use-deposits.ts
+++ b/libs/deposits/src/lib/use-deposits.ts
@@ -70,7 +70,6 @@ const updateQuery: UpdateQueryFn<
     subscriptionData.data.busEvents
   );
 
-  console.log({ subscriptionData, time: new Date() });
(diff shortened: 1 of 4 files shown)


libs/deposits/src/lib/use-deposits.ts
```diff
@@ -70,6 +70,7 @@ const updateQuery: UpdateQueryFn<
     subscriptionData.data.busEvents
   );
 
+  console.log({ subscriptionData, time: new Date() });
```
mattrussell36, line 73: chore (blocking): stray log

mattrussell36 (approved):

PR #2742, opened 2023-01-26
feat(token): 1823 tranches service

### LINKED ISSUE #1823: Tranches service with API for Token frontend
## Chore

Right now we need to scrape all the events ever to find out how much is in a tranche. 

This is impossible to do on the UI due to the sheer volume of the data, however understanding how much is in a tranche and how much has been withdrawn is within the staking ACs and has been on the site for a while.

The current solution involves generating a JSON file every few hours. This intermittently fails and can mean that data can be >8 hours out of date. Making users perhaps confused as to how they managed to double their money. It also makes the app hard to test because of this dependency on a JSON file which is probably out of step with the sidecar-ed ganache node running in capsule. 

The contract does not have APIs that support this currently. An alternative is we could deploy an indexer contract that gives us the API we need.

Some kind of API design, written in GQL syntax but after conversations we have decided a REST API would be easier to delvier. This is left as an instruction for how the response would ideally look.

```gql
fragment Withdrawals on Withdrawals {
  amount
  user
  tx
  trancheId
}

fragment Deposits on Deposit {
  amount
  user
  tx
  trancheId
}

fragment User on User {
  address
  tranches {
     ...Tranches
  }
  totalAdded
  totalWithdrawn
  totalRemaining
  withdrawals {
    ...Withdrawals
  }
  deposits {
    ...Deposits
  }
}

fragment Tranches on Tranches {
  id
  totalAdded
  totalRemaining
  totalRemoved
  startTime
  endTime
  deposits {
    ...Deposits
  }
  withdrawals {
    ...Withdrawals
  }
  users {
    ...User
  }
}

query user (address: string) {
  ...User
}
```

This ticket is a placeholder to design and build a service to query events, store in a small DB and provide an API for token frontend to access.

## Tasks

- [ ] Get service setup and running
- [ ] Replace the useGetUserTrancheBalances hook with the REST implementation
- [ ] Ensure that updates are happening as required
- [ ] Introduce some polling to ensure these balances update as required
- [ ] Ensure wallet is pulling data up to date

## Not needed now maybe later

```gql
query tranches {
  ...Tranches
}

query tranche (id: number) {
  ...Tranches
}
```

dexturr:
# Related issues 🔗

Closes #1823

# Description ℹ️

* Use tranches service for pulling tranche data
* Change tranche data view based on new data
* Allow entering of an arbitrary eth key and viewing tranches for that person


apps/token/src/routes/redemption/hooks.ts
```diff
@@ -0,0 +1,54 @@
...
+    });
+
+    const trancheBalances = await Promise.all(promises);
+    setUserTrancheBalances(trancheBalances);
+  }, [address, decimals, tranches, vesting]);
+  useEffect(() => {
+    loadUserTrancheBalances();
+  }, [loadUserTrancheBalances]);
+  return userTrancheBalances;
+};
```
edd, line 54: **praise (non-blocking)**: In this pretty tedious looking PR, the hooks in this stood out as a nice encapsulation of a lot of tedious work. Good refactor.

edd (approved):

MadalinaRaicu (approved):

PR #2759, opened 2023-01-26
chore: handle not found errors as correct response

### LINKED ISSUE #2757: Party data not updating when using a fresh wallet
## Description

Connecting with a fresh wallet results in a not found error which means that the subscription for further data is not started. This results in updates not being received for users first actions such as placing an order or having a position opened.

## Steps to Reproduce

1. Create a fresh wallet with 0 accounts
2. Go to the accounts table
3. Make a deposit
4. Notice that the accounts table never receives the asset

## Expected behavior

Assets are immediately shown in the table

## Screenshots

N/A

## Device details

All

## Additional context

This is likely also happening for other tables that query for party data (orders, positions, fills etc). Those tables should be fixed as well.

bglownia:
# Related issues 🔗

Closes #2757

# Description ℹ️

# Demo 📺

# Technical 👨‍🔧


libs/market-list/src/lib/market-data-provider.ts
```diff
@@ -18,14 +18,20 @@ import type {
...
-    Object.assign(draft, marketData);
-  });
+const update = (
+  data: MarketData | null,
+  delta: MarketDataUpdateFieldsFragment
+) => {
+  return (
+    data &&
+    produce(data, (draft) => {
+      const { marketId, __typename, ...marketData } = delta;
```
mattrussell36, line 28: Nice catch

mattrussell36 (approved):

PR #2763, opened 2023-01-27
fix: don't show `my markets` if there are no markets & reset deal ticket price validation before setting price 

MadalinaRaicu:
# Related issues 🔗

Closes N/A

# Description ℹ️

- [x] clear up price errors before setting value
- [x] don't show "My markets" if there are no markets

### DIFF
--- a/apps/trading/components/select-market/select-market.tsx
+++ b/apps/trading/components/select-market/select-market.tsx
@@ -170,7 +170,7 @@ export const SelectMarketPopover = ({
           </div>
         ) : (
           <table className="relative text-sm w-full whitespace-nowrap">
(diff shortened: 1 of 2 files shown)


bglownia (approved):

apps/trading/components/select-market/select-market.tsx
```diff
@@ -170,7 +170,7 @@ export const SelectMarketPopover = ({
           </div>
         ) : (
           <table className="relative text-sm w-full whitespace-nowrap">
-            {pubKey && (positions?.length ?? 0) > 0 ? (
```
bglownia, line 173: non blocking: no need to use nullish coalescing operator, null is also falsy
`pubKey && positions?.length && markets?.length`

PR #2765, opened 2023-01-27
feat(explorer): add long text component for hashes and tx viewer

### LINKED ISSUE #2638: Rename 'raw' and 'decoded' transaction views
## The Chore

The terms make sense if you're a developer... just about. Less so if you're anyone else. 

## Tasks
- [ ] Make a new component view that contains both the raw and 'decoded' views as different 'formats' a la etherscan

Screenshot 2023-01-27 at 11 29 03

edd:
# Related issues 🔗

Closes #2638

# Description ℹ️

A slightly bigger than it needs to be PR. Primarily, it adds a new component that renders the JSON or base64 encoded
transaction in a textbox. This component lives in the shared tx header, so it is available for all transaction types
immediately. This replaces the previous two details sections, raw transaction & decoded transaction.

The second addition is a 'hash' component that simply ensures that the text it renders has the appropriate CSS classes
to ensure that long strings like party IDs or market IDs will wrap. This won't magically fix all the scrollbars but will
be part of the solution.

# Demo 📺

Screenshot 2023-01-27 at 17 39 46
Screenshot 2023-01-27 at 17 39 59

https://user-images.githubusercontent.com/6678/215156637-e8585424-1863-4234-a9a9-2ebeaad15725.mov


MadalinaRaicu (approved):

edd:
There's a bug that got merged in this

PR #2766, opened 2023-01-27
fix: trades grid colors

MadalinaRaicu:
# Related issues 🔗

Closes N/A

# Description ℹ️

Fix trades grid colors. 

# Demo 📺

Screenshot 2023-01-27 at 17 46 41

# Technical 👨‍🔧

### DIFF
--- a/libs/trades/src/lib/trades-table.tsx
+++ b/libs/trades/src/lib/trades-table.tsx
@@ -13,31 +13,25 @@ import type { IDatasource, IGetRowsParams } from 'ag-grid-community';
 import type { CellClassParams, ValueFormatterParams } from 'ag-grid-community';
 import type { AgGridReactProps } from 'ag-grid-react';
 import type { Trade } from './trades-data-provider';
-import BigNumber from 'bignumber.js';
+import { Side } from '@vegaprotocol/types';
 
-export const UP_CLASS = 'text-vega-green dark:text-vega-green';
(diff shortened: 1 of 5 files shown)


libs/trades/src/lib/trades-table.tsx
```diff
@@ -13,31 +13,25 @@ import type { IDatasource, IGetRowsParams } from 'ag-grid-community';
...
-      const prevRowNode = api.getModel().getRow(rowIndex + 1);
-      const prevValue = prevRowNode?.data && prevRowNode.data[dataKey];
-      const valueNum = new BigNumber(value);
-
-      if (valueNum.isGreaterThan(prevValue)) {
-        colorClass = UP_CLASS;
-      } else if (valueNum.isLessThan(prevValue)) {
-        colorClass = DOWN_CLASS;
-      }
+  if (typeof rowIndex === 'number') {
```
mattrussell36, line 25: chore (blocking): There's no need to use rowIndex anymore. It was only needed as we needed to check the value of the previous row. Now we just use the row data it self.

mattrussell36 (changes requested):

libs/trades/src/lib/trades-table.tsx
MadalinaRaicu: updated.

jeremyletang (approved):

mattrussell36 (approved):

PR #2773, opened 2023-01-30
chore: capsule tests refactor and increase deposit ac

### LINKED ISSUE #2730: Additional mocked tests for deposits
## The Chore

There is still some missing coverage on some ACs for deposits which can be covered using mocked tests as part of CI

## Tasks

- [ ] Review ACs and determine which are not covered
- [ ] Add e2e tests to cover missing ACs

daro-maj:
# Related issues 🔗

Closes #2730 
# Description ℹ️

1. Capsule tests refactor:
   - removed reload() for deposit
   - splitted up tests 
2. Increase deposit ac:
   - 1001-DEPO-006
   - 1001-DEPO-005

### DIFF
--- a/apps/trading-e2e/src/integration/capsule.cy.ts
+++ b/apps/trading-e2e/src/integration/capsule.cy.ts
@@ -130,7 +130,7 @@ describe('capsule', { tags: '@slow' }, () => {
     cy.getByTestId('edit-order').find('[type="submit"]').click();
     cy.getByTestId(toastContent).should(
       'contain.text',
-      `ConfirmedYour transaction has been confirmed View in block explorerEdit order - activeTEST.24h+${size} @ ${newPrice}.00 ${usdcSymbol}+${size} @ ${newPrice}.00 ${usdcSymbol}`,
+      `ConfirmedYour transaction has been confirmed View in block explorerEdit order - activeTEST.24h+${size} @ ${price}.00 ${usdcSymbol}+${size} @ ${newPrice}.00 ${usdcSymbol}`,
       { matchCase: false }
     );
     cy.getByTestId(ordersTab).click();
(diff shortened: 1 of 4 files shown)


apps/trading-e2e/src/integration/capsule.cy.ts
```diff
@@ -160,44 +176,28 @@ describe('capsule', { tags: '@slow' }, () => {
...
-    cy.get('main[data-testid="/portfolio"]').should('exist');
-
-    cy.highlight('creating deposit');
+    // 1001-DEPO-010
 
     cy.getByTestId(depositsTab).click();
     cy.getByTestId('deposit-button').click();
     connectEthereumWallet('Unknown');
-    cy.get(assetSelectField, txTimeout).select(btcName);
+    cy.get(assetSelectField, txTimeout).select(btcName, { force: true });
```
jtsang586, line 184: Is the force true needed here?

MadalinaRaicu (approved):

apps/trading-e2e/src/integration/capsule.cy.ts
daro-maj: this is necessary due to the current implementation of app: select aria-hidden="true"

apps/trading-e2e/src/integration/capsule.cy.ts
```diff
@@ -113,35 +120,41 @@ describe('capsule', { tags: '@slow' }, () => {
...
-    //cancel order
+  });
+
+  it('can cancel order', function () {
     cy.getByTestId(ordersTab).click();
     cy.getByTestId('cancel').first().click();
     cy.getByTestId(toastContent).should(
       'contain.text',
-      `ConfirmedYour transaction has been confirmed View in block explorerCancel order - cancelledTEST.24h+0.0005 @ 200.00 ${usdcSymbol}`
+      `ConfirmedYour transaction has been confirmed View in block explorerCancel order - cancelledTEST.24h+${size} @ ${newPrice}.00 ${usdcSymbol}`,
```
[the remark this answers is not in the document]
daro-maj: The idea is to check if tests will be stable with checking the whole text and then we can be sure that we check everything. If the tests are not stable because the text will change, we will do it separately.

PR #2778, opened 2023-01-30
fix(trading): consolidate view as user mode

### LINKED ISSUE #2717: Error thrown when trying to send transactions when in view as user mode
## Chore

Disable or hide any buttons which execute transactions when user is in 'view mode'. This is because an error will be thrown if a tx is attempted

## Tasks

- [x] Disable/hide transaction executing buttons
- [ ] Exit early if send Tx function is invoked

## Screenshots

Image

Screenshot 2023-01-24 at 09 04 31

MadalinaRaicu:
# Related issues 🔗

Closes #2717

# Description ℹ️

Consolidate `view as` mode.

# Demo 📺

Screenshot 2023-01-31 at 15 50 45

Screenshot 2023-01-30 at 16 38 47

# Technical 👨‍🔧

### DIFF
--- a/libs/accounts/src/lib/accounts-manager.spec.tsx
+++ b/libs/accounts/src/lib/accounts-manager.spec.tsx
@@ -23,13 +23,23 @@ jest.mock('@vegaprotocol/react-helpers', () => ({
 describe('AccountManager', () => {
   it('change partyId should reload data provider', async () => {
     const { rerender } = render(
-      <AccountManager partyId="partyOne" onClickAsset={jest.fn} />
+      <AccountManager
+        partyId="partyOne"
+        onClickAsset={jest.fn}
+        isReadOnly={false}
+      />
     );
     expect(
       (helpers.useDataProvider as jest.Mock).mock.calls[0][0].variables.partyId
     ).toEqual('partyOne');
     await act(() => {
-      rerender(<AccountManager partyId="partyTwo" onClickAsset={jest.fn} />);
+      rerender(
+        <AccountManager
+          partyId="partyTwo"
+          onClickAsset={jest.fn}
+          isReadOnly={false}
+        />
+      );
     });
     expect(
       (helpers.useDataProvider as jest.Mock).mock.calls[1][0].variables.partyId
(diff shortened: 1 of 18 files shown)


libs/accounts/src/lib/accounts-table.spec.tsx
```diff
@@ -35,7 +35,11 @@ describe('AccountsTable', () => {
   it('should render correct columns', async () => {
     await act(async () => {
       render(
-        <AccountTable rowData={singleRowData} onClickAsset={() => null} />
+        <AccountTable
+          rowData={singleRowData}
+          onClickAsset={() => null}
+          isReadOnly={false}
+        />
```
edd, line 42: **chore (non-blocking)**: `isReadOnly` should probably get a test too. Non-blocking a it's such simple functionality

I've posted this twice - assume it's true for all the components that now take this prop

libs/accounts/src/lib/accounts-manager.spec.tsx
```diff
@@ -23,13 +23,23 @@ jest.mock('@vegaprotocol/react-helpers', () => ({
 describe('AccountManager', () => {
   it('change partyId should reload data provider', async () => {
     const { rerender } = render(
-      <AccountManager partyId="partyOne" onClickAsset={jest.fn} />
+      <AccountManager
+        partyId="partyOne"
+        onClickAsset={jest.fn}
+        isReadOnly={false}
```
edd, line 29: **chore (non-blocking)**: `isReadOnly` should probably get a test too. Non-blocking a it's such simple functionality

edd (approved):
✅ Left a comment that the new `isReadOnly` prop on most views feel like they should get a test, but marked it as non-blocking because it's such basic functionality.

Will re-review if disabling submit order gets added in to this PR, but as is - removing buttons in tables - this looks good

libs/accounts/src/lib/accounts-table.spec.tsx
MadalinaRaicu: yeah, I guess only false was purely tested. I am adding some extra cases now

PR #2779, opened 2023-01-30
chore(trading): new toasts designs

### LINKED ISSUE #2735: Improved toasts design with dark / light mode
## The Chore

Implement updated design for toasts considering dark and light mode, and colouring for status. 

## Tasks

- [ ] and then what?
- [ ] Etc.

## Design 
User Feedback

Figma link: https://www.figma.com/file/EaFndX7sV9CgZOdsE2Ilqp/dAPP-Components?node-id=510%3A8496

### LINKED ISSUE #2751: Toast timeout
## The Chore
Update toast behaviour:

- [ ] Toasts fade out after a certain period of time 
- [ ] I can see when the toast will fade out e.g. a progress bar 

Kraken Pro example
https://user-images.githubusercontent.com/46783148/214869144-af4d5ea3-87eb-48d8-aac5-e7d2af668acf.mov

asiaznik:
Closes #2735
Closes #2751

* fixes toast container re-renders (zustand subscriptions)
* fixes auto-closing and added pause auto-closing on hover
* added close all button
* changed to new colours and designs 
 
Designs: https://www.figma.com/file/EaFndX7sV9CgZOdsE2Ilqp/dAPP-Components?node-id=548%3A15951&t=RxpIUUf83DP4mmb3-0

https://user-images.githubusercontent.com/1980305/216690126-9fff2b62-b69e-46be-bf9b-6c652c7c2056.mov

### DIFF
--- a/libs/ui-toolkit/src/components/toast/toast.tsx
+++ b/libs/ui-toolkit/src/components/toast/toast.tsx
@@ -3,7 +3,7 @@ import styles from './toast.module.css';
 import type { IconName } from '@blueprintjs/icons';
 import { IconNames } from '@blueprintjs/icons';
 import classNames from 'classnames';
-import type { ForwardedRef, HTMLAttributes } from 'react';
+import type { HTMLAttributes, HtmlHTMLAttributes } from 'react';
 import { forwardRef, useEffect } from 'react';
 import { useCallback } from 'react';
 import { useLayoutEffect } from 'react';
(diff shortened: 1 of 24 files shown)


libs/ui-toolkit/src/components/toast/toast.tsx
```diff
@@ -118,26 +206,83 @@ export const Toast = ({
...
+            />
           )}
         </div>
         <div
-          className="flex-1 p-2 pr-6 text-sm overflow-auto dark:bg-black dark:text-white"
+          className={classNames(
+            'relative',
+            'overflow-auto flex-1 p-4 pr-[40px] [&>p]:mb-[2.5px]',
+            // toast heading
+            '[&>h3]:text-[14px] [&>h3]:leading-[13px] [&>h3]:uppercase [&>h3]:mb-[8px]'
```
mattrussell36, line 256: suggestion (non blocking): I don't think we should add these styles. If you were to create a different toast its not obvious that you should use an H3. Also the toast component shouldn't dictate what content and elements are required within it. I think we should one of the following

I would apply these styles to directly to the h3 element, maybe make a ToastHeading component to keep things DRY.

mattrussell36: Additionally, I'd like to try and avoid using specific px values where possible so we can keep consistency, yes it might not be pixel perfect compared to the designs but there is more value in the long run. The more we deviate from the TW config the more messy it will start to look over time

I just changed
`'[&>h3]:text-[14px] [&>h3]:leading-[13px] [&>h3]:uppercase [&>h3]:mb-[8px]'`
to
`'[&>h3]:text-sm [&>h3]:uppercase [&>h3]:mb-1'`

and the end result is almost identical

libs/ui-toolkit/src/components/toast/use-toasts.ts
```diff
@@ -1,11 +1,23 @@
...
 
-type ToastsStore = {
-  /**
-   * A list of active toasts
-   */
-  toasts: Toast[];
+export type Toasts = Record<string, Toast>;
+
+const isEqual = (a: Toast, b: Toast) => {
+  const h1 = hash(omit(a, 'onClose'));
```
mattrussell36, line 10: suggestion (non blocking): Maybe use lodash/isEqual instead of introducing a new dependency. This is already in use in the project and will do a deep comparison. I think it should work here

libs/ui-toolkit/src/components/toast/toast.tsx
```diff
@@ -33,20 +34,37 @@ const toastIconMapping: { [i in Intent]: IconName } = {
...
+export const TICKER = 100;
+export const CLOSE_AFTER = 5000;
 
-export const CLOSE_DELAY = 750;
+export const Panel = forwardRef<
+  HTMLDivElement,
+  {
+    ref: ForwardedRef<HTMLDivElement>;
+  } & HTMLAttributes<HTMLDivElement>
+>(({ children, ref, className, ...props }) => {
```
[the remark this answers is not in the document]
asiaznik: 🤦🏻

libs/ui-toolkit/src/components/toast/use-toasts.ts
asiaznik: Oops, I was just testing over few things here with a `stable-hash`, shouldn't have left it here. Removed.

MadalinaRaicu (approved):

PR #2800, opened 2023-01-31
feat(ui-toolkit): announcement banner to highlight mainnet sims

### LINKED ISSUE #2714: Highlight mainnet sims on frontend applications
## Background / context

The blog announcing the mainnet sim’s is going out today (24th Jan), and the first sim is due to start next Tuesday (31st Jan).

## The Chore

We would like to have something on our frontend applications to highlight the sims, either a new ‘banner’ at the top (we will be doing something similiar for vega.xyz shown in screengrabs below) or we could re-use a relevant existing common component i.e. the pop-up that has been built as the ‘warning message’ for console.

I would expect that we would do this for console, governance and block explorer (not wallet or docs) but this is TBC. 

## Tasks

- [x] Decide what component to use
- [ ] Design and content (as needed)
- [ ] Implement
- [ ] Test
- [ ] Launch (need to decide whether we would launch it before or on the 31st)

## Additional information

vega.xyz will be using the existing banner (top) and pop up (bottom right) components 

Rebecca & Nissa will be providing 2-3 word main heading, short secondary sentence and CTA for us to use

image

sam-keen:
# Related issues 🔗

Closes #2714

# Description ℹ️

Adds a shared announcement banner to the ui-toolkit for each app to announce mainnet sims. Accepts children so copy can be customised per app as required.

# Demo 📺

Screenshot 2023-01-31 at 15 11 56


edd (approved):

asiaznik (approved):

libs/ui-toolkit/src/components/announcement-banner/announcement-banner.tsx
```diff
@@ -0,0 +1,15 @@
+import classnames from 'classnames';
+import type { ReactNode } from 'react';
+
+export interface BannerProps {
+  children?: ReactNode;
+}
+
+export const AnnouncementBanner = ({ children }: BannerProps) => {
```
asiaznik, line 8: suggestion (non-blocking): How about `forwardRef` it? So it can be manipulated if needed (e.g. closable, hide-able)

sam-keen: I'm happy to add functionality if we decide it's needed, but I'd prefer not to build it in before it's necessary. As the vega.xyz version isn't going to use it for now, I'd rather not.

PR #2816, opened 2023-02-01
chore(trading): no data message in charts

### LINKED ISSUE #2411: Align font size of no data message in charts
## The Chore

The no data text for the depth chart and the candles chart is inconsistent with the rest of the app

Screenshot 2022-12-14 at 17 46 16

macqbat:
# Related issues 🔗

Closes #2411

# Description ℹ️

Add nicer "no data" message in chart components

# Demo 📺

image
Screenshot 2023-02-01 at 16 50 10

# Technical 👨‍🔧


libs/candles-chart/src/lib/candles-chart.tsx
```diff
@@ -168,6 +167,11 @@ export const CandlesChartContainer = ({
             chartType: chartType,
             overlays: overlays,
             studies: studies,
+            notEnoughDataText: (
+              <span className="text-xs text-center text-gray-800 dark:text-gray-200">
```
mattrussell36, line 171: chore(blocking): Could you switch to using text-neutral, we use those everywhere else

libs/market-depth/src/lib/depth-chart.tsx
```diff
@@ -207,6 +208,11 @@ export const DepthChartContainer = ({ marketId }: DepthChartManagerProps) => {
           theme={theme}
           volumeFormat={volumeFormat}
           priceFormat={priceFormat}
+          notEnoughDataText={
+            <span className="text-xs text-center text-gray-800 dark:text-gray-200">
```
mattrussell36, line 212: same here

macqbat: done

MadalinaRaicu (approved):

libs/candles-chart/src/lib/candles-chart.tsx
```diff
@@ -168,6 +167,11 @@ export const CandlesChartContainer = ({
             chartType: chartType,
             overlays: overlays,
             studies: studies,
+            notEnoughDataText: (
```
MadalinaRaicu, line 170: Could you please update the account history chart as well? That uses the PriceChart

macqbat: Unfortunately this type of charts has been omitted, and we don't have possibility now to set this param there as reactNode :-). I don't want to hold longer this solution, so I created a new issue for fixing it: https://github.com/vegaprotocol/frontend-monorepo/issues/2836

PR #2829, opened 2023-02-02
fix(trading): remove price flash cell from PNL

### LINKED ISSUE #2826: Remove flash cell from PNL cells
## Description

The PNL cells are already rendered green or red depending on profit/loss having the cell text change color when it goes up or down is confusing as you might have a profitable position that suddenly renders in red for a moment before going back to green.

## Steps to Reproduce

1. Place a trade to open a position
2. Observe the realised PNL cell
3. Keep observing... :-)
4. Eventually you will see it flash red if your PNL goes down

MadalinaRaicu:
# Related issues 🔗

Closes #2826 

# Description ℹ️

 Remove price flash cell from PNL 

# Demo 📺

# Technical 👨‍🔧

### DIFF
--- a/libs/positions/src/lib/positions-table.tsx
+++ b/libs/positions/src/lib/positions-table.tsx
@@ -19,6 +19,7 @@ import {
   signedNumberCssClassRules,
   DateRangeFilter,
   addDecimalsFormatNumber,
+  PriceCell,
 } from '@vegaprotocol/react-helpers';
 import { AgGridDynamic as AgGrid, Link } from '@vegaprotocol/ui-toolkit';
 import { AgGridColumn } from 'ag-grid-react';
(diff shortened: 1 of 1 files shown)


libs/positions/src/lib/positions-table.tsx
```diff
@@ -360,7 +360,7 @@ export const PositionsTable = forwardRef<AgGridReact, Props>(
               ? undefined
               : addDecimalsFormatNumber(data.unrealisedPNL, data.decimals)
           }
-          cellRenderer="PriceFlashCell"
+          cellRenderer="PriceCell"
```
mattrussell36, line 363: chore (blocking): You need to add PriceCell to the components prop of the AgGrid component

mattrussell36 (changes requested):

libs/positions/src/lib/positions-table.tsx
MadalinaRaicu: updated

mattrussell36 (approved):

Issue #176, opened 2022-03-31
Add Alpha Lyrae into typography section


dexturr:
Now we have alpha lyrae in the ui toolkit would be good to have:

* Demonstration of Alpha Lyrae font (using `alpha` class)
* Demonstration of contextual alternatives class (`calt`)

PR #2833, opened 2023-02-02
chore(trading): add alpha font to trading

### LINKED ISSUE #2723: Add Alpha Lyrae to Console
## The Chore

Switch to using Alpha Lyrae everywhere without the glitch effect only. `font-alpha liga-0-calt-0` should do the trick

## Tasks

- [x] Add `font-alpha liga-0-calt-0` to the body so that the font is applied without ligatures or contextual alternatives
- [x] Review and apply any fixes all views to ensure line heights and alignment is all okay
- [x] Add typography section to storybook to close https://github.com/vegaprotocol/frontend-monorepo/issues/176

macqbat:
# Related issues 🔗

Closes #2723 and #176

# Description ℹ️

Add  Alpha Lyrae to the body so that the font is applied without ligatures or contextual alternatives

# Demo 📺

image

# Technical 👨‍🔧


libs/ui-toolkit/src/primitives/typography.stories.tsx
```diff
@@ -0,0 +1,65 @@
...
+        {type}:
+      </div>
+      <div
+        className={classNames(
+          'flex-grow flex flex-col justify-end text-left items-start',
+          { 'font-alpha': isAlpha },
+          [alternatives, type]
+        )}
+      >
+        Lorem ipsum dolor sit amet, consectetur adipiscing elit
```
mattrussell36, line 27: suggestion (non blocking): Maybe add a section with ligatures as we'll start introducing that in some select places

mattrussell36 (approved):

libs/ui-toolkit/src/primitives/typography.stories.tsx
macqbat: done

PR #2842, opened 2023-02-03
fix(trading): update async renderer on error

MadalinaRaicu:
# Related issues 🔗

Closes N/A

# Description ℹ️
 Update `AsyncRenderer` on error

# Demo 📺

Screenshot 2023-02-03 at 15 00 35

# Technical 👨‍🔧

### DIFF
--- a/libs/ui-toolkit/src/components/async-renderer/async-renderer.tsx
+++ b/libs/ui-toolkit/src/components/async-renderer/async-renderer.tsx
@@ -1,6 +1,7 @@
 import { Splash } from '../splash';
 import type { ReactNode } from 'react';
 import { t } from '@vegaprotocol/react-helpers';
+import * as Sentry from '@sentry/react';
 
 interface AsyncRendererProps<T> {
   loading: boolean;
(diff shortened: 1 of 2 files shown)


macqbat (approved):

libs/ui-toolkit/src/components/async-renderer/async-renderer.tsx
```diff
@@ -26,13 +27,16 @@ export function AsyncRenderer<T = object>({
...
-          : t(`Something went wrong: ${error.message}`)}
-      </Splash>
-    );
+    Sentry.captureException(`Error rendering data: ${error.message}`);
+    if (!data) {
+      return (
+        <Splash>
+          {errorMessage
+            ? errorMessage
+            : t(`Something went wrong: ${error.message}`)}
```
macqbat, line 36: (non-blocking): I think displaying error.message is not necessary. Something went wrong would be enough here

MadalinaRaicu: I think we should still have the error message out there. People testing will see what is wrong more easily.

PR #2845, opened 2023-02-03
chore(trading): notification component redesign

### LINKED ISSUE #2665: Notification component
## The Chore

We have a notification component in ui-toolkit which needs to be b) added to figma/design system b) tweaked to align with the design system

Current:
Screenshot 2023-01-18 at 10 53 50

Screenshot 2023-01-18 at 11 00 09

## Tasks

- [x] Add to figma
- [x] Design tweaks
- [x] Dev

This is currently used in two places
- Asset dialog, to warn of a pending asset proposal
- Trade page header, to warn of a pending market proposal

macqbat:
# Related issues 🔗

Closes #2665 

# Description ℹ️

Upgrade view of notification component according to new design

# Demo 📺

image
image
image
image

image
image

# Technical 👨‍🔧


libs/ui-toolkit/src/components/notification/notification.tsx
```diff
@@ -40,7 +45,8 @@ export const Notification = ({
           'border-yellow-500': intent === Intent.Warning,
           'border-vega-pink': intent === Intent.Danger,
         },
-        'border rounded px-3 py-1 text-xs mb-1 mr-1'
+        'border rounded-[4px] text-xs mb-1 mr-1 p-4 flex items-start gap-2.5 bg-neutral-100 dark:bg-neutral-900',
```
mattrussell36, line 48: suggestion (non blocking): Can we just use `rounded` here. I'd like to avoid specifying exact pixel values in TW classes. The more we allow it the more we will end up with an inconsistent design later on..

macqbat: sure, done, thanks

MadalinaRaicu (approved):

PR #2851, opened 2023-02-03
fix(trading):  when order is rejected change tx toast intent

### LINKED ISSUE #2821: Rejected orders should use Intent.danger for the confirmation toast
## The Chore

Its possible that an order transaction can be successful but the order itself was rejected. Right now a successful transaction will always be in a successful green state which is confusing in the case where an order is rejected. We need to change the order submission toast to be in a red 'danger' state if the order was rejected

## Tasks

- [ ] Update toast logic so that rejected orders are in a red error state

https://vegaprotocol.slack.com/archives/C037STCKMGC/p1675287291581209

MadalinaRaicu:
# Related issues 🔗

Closes #2821 

# Description ℹ️

 When order rejected change tx toast intent

# Demo 📺

# Technical 👨‍🔧


libs/positions/src/lib/positions-table.tsx
```diff
@@ -85,7 +84,7 @@ export const PositionsTable = forwardRef<AgGridReact, Props>(
           filterParams: { buttons: ['reset'] },
           tooltipComponent: TooltipCellComponent,
         }}
-        components={{ AmountCell, PriceFlashCell, PriceCell, ProgressBarCell }}
```
MadalinaRaicu, line 88: price cell overrides class rules on PNL cells

macqbat (approved):

PR #2868, opened 2023-02-06
feat(explorer): change party to pubkey

### LINKED ISSUE #2866: Use public key instead of/alongside party id in explorer
The block explorer references 'party id', but users may expect to see the phrase 'public key'. Introduce the 'public key' terminology instead of (and perhaps alongside, in some places) in order to make that connection clearer.

edd:
# Related issues 🔗

Closes #2866

# Description ℹ️

In most places, 'public key' is a better label than 'party id'


apps/explorer/src/app/routes/parties/id/components/party-id-error.tsx
```diff
@@ -14,7 +14,7 @@ type PartyIdErrorProps = {
 const PartyIdError = ({ id, error }: PartyIdErrorProps) => {
   const end = isValidPartyId(id)
     ? t('No accounts or transactions found for: ')
-    : 'Invalid party id: ';
+    : 'Invalid pubkey: ';
```
sam-keen, line 17: Suggestion (non-blocking): for consistency's sake, this one should be 'public key' too

sam-keen (approved):

apps/explorer/src/app/routes/parties/id/components/party-id-error.tsx
candida-d: Agreed!

candida-d (approved):
But I second Sam's comment about pubkey vs public key

mattrussell36 (approved):

apps/explorer/src/app/routes/parties/id/components/party-id-error.tsx
edd, line 17:
```suggestion
    : 'Invalid public key: ';
```

PR #2872, opened 2023-02-06
fix(trading): use block height header to determine node health

### LINKED ISSUE #1791: Update nodeHealth to use `x-block-height` header to check data node health

mattrussell36:
# Related issues 🔗

Closes #1791 

# Description ℹ️

- Updates node health status to compare the the datanode's block height with the core block height
- Updates node switcher dialog to show each datanode's block height rather than the core block height

# Technical 👨‍🔧

- useNodeHealth now does not need to query all nodes, simply comparing the x-block-height header and statistics is simpler and more effective
- Added header store to update an object of urls and header results using a custom ApolloLink
- NodeSwitcher now renders blockHeights of headers rather than core result

### DIFF
--- a/apps/trading/client-pages/portfolio/account-history-container.tsx
+++ b/apps/trading/client-pages/portfolio/account-history-container.tsx
@@ -1,6 +1,6 @@
 import {
   addDecimal,
-  fromNanoSeconds,
+  fromISONanoSeconds,
   t,
   useThemeSwitcher,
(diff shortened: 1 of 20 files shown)


edd (approved):
👍 @mattrussell36 took me through this on Friday and it looks like a great iteration on the health check

PR #2882, opened 2023-02-08
feat(explorer): asset details

### LINKED ISSUE #2535: Change `assets` index, add details dialog
## The Chore

The assets index is similar to #2534, not all that useful in its current state. Unlike #2533, it's possible that a `/assets/` page *might* be useful. Let's remove the JSON views, render a basic list, and add details views.

## Tasks

- [ ] Render either the Asset Table from console, or a similar view
- [ ] Update `/assets` to be a very basic list of assets that link to the details view

asiaznik:
# Related issues 🔗

Closes #2535

# Description ℹ️

* removed default JSON views from `/assets`
* added a table that consists of all available assets with `View details` and `View JSON` buttons
* removed redundant queries

# Demo 📺

https://user-images.githubusercontent.com/1980305/217580514-4e7ec76f-59ac-4b22-b69c-32cc100ee703.mov


apps/explorer/src/app/components/links/asset-link/asset-link.tsx
```diff
@@ -1,36 +1,35 @@
...
 
-export type AssetLinkProps = Partial<ComponentProps<typeof Link>> & {
-  id: string;
+export type AssetLinkProps = Partial<ComponentProps<typeof ButtonLink>> & {
+  assetId: string;
 };
 
 /**
  * Given an asset ID, it will fetch the asset name and show that,
  * with a link to the assets list. If the name does not come back
```
edd, line 15:
```suggestion
 * with a link to the assets modal. If the name does not come back
```

apps/explorer/src/app/routes/assets/index.tsx
```diff
@@ -1,51 +1,108 @@
...
-          />
-        </section>
-      );
-    } else {
-      return <Loader />;
-    }
-  }
+  const { data, loading, error } = useAssetsDataProvider();
+  const ref = useRef<AgGridReact>(null);
+  const table = (
```
edd, line 28: **nitpick (non-blocking)**:  This Asset list inherits a bug that I've also seen on Governance that @sam-keen recently put in a fix for. If you size the window down to mobile-ish size, there's a large gap between the title (`Assets`) and the header. 

Mobile view is not a priority currently, but if there's a fix for this that's easy to slot it, let's slot it in before merging

libs/assets/src/lib/asset-details-table.tsx
```diff
@@ -105,7 +106,16 @@ export const rows: Rows = [
         return;
       }
 
-      return <ContractAddressLink address={asset.source.contractAddress} />;
+      return (
+        <>
+          <ContractAddressLink address={asset.source.contractAddress} />{' '}
```
edd, line 111: **question (non-blocking)**: I had missed this component. Is `links/eth-explorer-link` a complete duplicate? If so I'll ticket replacing that all over.

libs/tailwindcss-config/src/vega-custom-classes.js
```diff
@@ -15,14 +15,18 @@ const vegaCustomClasses = plugin(function ({ addUtilities }) {
     },
     '.syntax-highlighter-wrapper .hljs': {
       fontSize: '1rem',
-      fontFamily: "Consolas, Monaco, 'Andale Mono', 'Ubuntu Mono', monospace",
+      fontFamily: "'Roboto Mono', monospace",
```
edd, line 18: **nitpick (non-blocking)** : 🤨 :wink:

edd (approved):

edd:
**nitpick (non-blocking)**: Not a failing of this PR, but while using it I noticed two views that probably should have asset links that don't currently:

Screenshot 2023-02-08 at 16 29 08
Screenshot 2023-02-08 at 16 28 44

- The Order Submit Details card doesn't link to the asset modal
- The Order Submit table-at-the-top thing doesn't list the asset either. 
- Node Vote > Internal Asset Deposit does not list the asset id

apps/explorer/src/app/components/links/asset-link/asset-link.tsx
```diff
@@ -1,36 +1,35 @@
...
-    variables: { id },
-  });
-
-  let label: string = id;
-
-  if (data?.asset?.name) {
-    label = data.asset.name;
-  }
+export const AssetLink = ({ assetId, ...props }: AssetLinkProps) => {
+  const { data: asset } = useAssetDataProvider(assetId);
```
edd, line 19: This does make the query for rendering an asset link much heavier than the old query. It's probably OK to keep this heavyweight query, but I wonder if there's something more lightweight we can do in places where we just really want to render an asset name?

apps/explorer/src/app/routes/assets/index.tsx
asiaznik: It looks like it's `nav` related. It should be fixes once we change our sidebar to a proper menu :-)

libs/assets/src/lib/asset-details-table.tsx
asiaznik: Looks like it 👍🏻

libs/tailwindcss-config/src/vega-custom-classes.js
asiaznik: Tailwind's `font-mono` is basically that ⬆️, no reason to mix fonts.

apps/explorer/src/app/components/links/asset-link/asset-link.tsx
asiaznik: An asset entry is not that heavy and I don't think that mainating couple different queries related to asset(s) is worth the potential data savings. Also it shouldn't be a concern with apollo's cache in place.

pscott31 (approved):

PR #2884, opened 2023-02-08
chore: handle positions with market data gql errors

### LINKED ISSUE #2870: Handle null market data for terminated markets
## The Chore

Queries for markets that include marketData can return an error if that market has no marketData. This results in rendering of the AsyncRenderer component, blocking some UIs from displaying (or sometimes displaying on top of the UI).

Null marketData happens when a market is restored from a snapshot but is in a terminated state. This is because the marketData object is not stored in the snapshot and does not get recreated if the the market is already terminated.

For example, this can happen for positions as user's position _is_ restored from a snapshot and that position might have a reference to a now terminated market

## Tasks

- [ ] Add a specific condition for this error (similar for NotFound errors) so that the error does not get surfaced to the UI
- [ ] Ensure any UIs that rely on marketData are rendered appropriately, perhaps with a '-'

https://vegaprotocol.slack.com/archives/C037STCKMGC/p1675428151914079

macqbat:
# Related issues 🔗

Closes #2870 

# Description ℹ️

Display position data row in case we detect market data notFound error

# Demo 📺

Screenshot 2023-02-08 at 17 49 36

# Technical 👨‍🔧

### DIFF
--- a/libs/react-helpers/src/lib/apollo-client.ts
+++ b/libs/react-helpers/src/lib/apollo-client.ts
@@ -2,6 +2,7 @@ import type { ApolloError } from '@apollo/client';
 import type { GraphQLErrors } from '@apollo/client/errors';
 
 const NOT_FOUND = 'NotFound';
+const INTERNAL = 'Internal';
 
 const isApolloGraphQLError = (
   error: ApolloError | Error | undefined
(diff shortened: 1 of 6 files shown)


libs/react-helpers/src/lib/apollo-client.ts
```diff
@@ -2,28 +2,41 @@ import type { ApolloError } from '@apollo/client';
...
-  path?: string
+  path?: string[]
 ) => {
   return (
     isApolloGraphQLError(error) &&
     hasNotFoundGraphQLErrors(error.graphQLErrors, path)
   );
 };
+
+export const isOnlyMarketDataNotFoundErrors = (
```
bglownia, line 32:
```
const makeIsGraphQLError = (type: string ) => (
  error: Error | ApolloError | undefined,
  path?: string[]
) => {
  return (
    isApolloGraphQLError(error) &&
    error.graphQLErrors.some(
      (e) =>
        e.extensions &&
        e.extensions['type'] === type&&
        (!path || path.every((item, i) => item === e?.path?.[i]))
    );
  );
};

export const isNotFoundGraphQLError = makeIsGraphQLError(NOT_FOUND);
export const isOnlyMarketDataNotFoundErrors = makeIsGraphQLError(INTERNAL);
```

macqbat: `isNotFoundGraphQLError`  uses `some` method, `isOnlyMarketDataNotFoundErrors` needs `every`. Add third parameter? :-)

libs/react-helpers/src/lib/generic-data-provider.ts
```diff
@@ -248,6 +253,30 @@ function makeDataProviderInternal<
     callbacks.forEach((callback) => notify(callback, updateData));
   };
 
+  const call = (
```
bglownia, line 256: Please add unit tests to generic-data-provider.spec.ts with usage of errorPolicyGuard

bglownia (changes requested):

libs/react-helpers/src/lib/generic-data-provider.ts
macqbat: done

PR #2902, opened 2023-02-13
feat(explorer): add transfers tx view

### LINKED ISSUE #2118: Transfer transaction
Acceptance Criteria defined in the EPIC https://github.com/vegaprotocol/frontend-monorepo/issues/2082

Blocked by https://github.com/vegaprotocol/vega/issues/7462

edd:
# Related issues 🔗

Closes #2118

# Description ℹ️

Adds a view for transfers. One off transfers are quite simple, recurring transfers take a bit more work.

# Demo 📺
There are three types of transfer this covers (hence the size of the PR). The properties of these have been broken down in to three 'blocks' that get rendered depending on the type.

## A one-off transfer
Screenshot 2023-02-16 at 17 51 45

## A recurring transfer
Screenshot 2023-02-16 at 17 50 54

The middle block, the repetition one, is also the first extensive look at Epochs so it contains some new components to render Epoch info. It can look like:

### This transfer is *ongoing* - i.e. the start epoch has past and the end epoch is in the future
Screenshot 2023-02-16 at 14 49 11

#### Mousing over the epoch number for current/past epochs will show timestamps/blockheights
Screenshot 2023-02-16 at 14 49 08

## Epochs that have not occurred, or do not exist in the API, will get an estimated date based on the network parameter epoch length and distance from the current epoch
Screenshot 2023-02-16 at 14 44 37

## A rewards pool transfer
Screenshot 2023-02-16 at 17 51 12


jeremyletang (approved):

apps/explorer/src/app/components/epoch-overview/epoch-missing.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+/**
+ * Parses the interval string we get for the epoch length from the
+ * network parameter API. These are in the format '1D2H3m' for 1 day,
+ * 2 hours and 3 minutes.
+ *
+ * @param str Interval string
+ * @returns integer the number of seconds the interval represents
+ */
+export function getSecondsFromInterval(str: string) {
```
asiaznik, line 120: suggestion (non-blocking): This could live in the react-helpers lib.

apps/explorer/src/app/components/size-in-asset/size-in-asset.tsx
```diff
@@ -0,0 +1,43 @@
...
+  if (data) {
+    if (decimalSource === 'ASSET' && data.decimals) {
+      label = addDecimalsFormatNumber(size, data.decimals);
+    }
+  }
+
+  return (
+    <p>
+      <span>{label}</span>&nbsp;
+      <AssetLink assetId={assetId} showAssetSymbol={true} />
```
asiaznik, line 38: suggestion (non-blocking): `asDialog` prop could be added here so it opens the details dialog rather than navigating out.

apps/explorer/src/app/components/txs/details/transfer/blocks/transfer-participants.tsx
```diff
@@ -0,0 +1,118 @@
...
+          }
+        >
+          <span>
+            <Icon className="ml-3" name={'bank-account'} />
+          </span>
+        </Tooltip>
+        <br />
+
+        {/* This block of divs is used to render the inset arrow containing the transfer amount */}
+        <div className="bg-zinc-200 dark:bg-zinc-800 flex items-center justify-center my-4 relative">
```
asiaznik, line 68: chore (blocking): I think we should use the colours that are defined in our design system. Shades of grey are defined as `vega-light-XXX` and `vega-dark-XXX`.

Colours definition: `libs/tailwindcss-config/src/theme.js`

image
_`ui-toolkit:storybook`: `http://localhost:4400/?path=/story/primitives-colors--page`_

apps/explorer/src/app/components/txs/details/transfer/blocks/transfer-rewards.tsx
```diff
@@ -0,0 +1,99 @@
...
+        ) : null}
+        <li>
+          <strong>{t('Factor')}</strong>: {recurring.factor}
+        </li>
+      </ul>
+    </div>
+  );
+}
+
+const metricLabels: Record<Metric, string> = {
```
asiaznik, line 62: suggestion (non-blocking): would be nice to have this in `global-types-mapping`

apps/explorer/src/app/components/epoch-overview/epoc.spec.tsx
```diff
@@ -0,0 +1,56 @@
+import { render } from '@testing-library/react';
+import { IconForEpoch } from './epoch';
+
+const THE_PAST = 'Monday, 17 February 2022 11:44:09';
+const THE_FUTURE = 'Monday, 17 February 3023 11:44:09';
+
+describe('IconForEpoch', () => {
+  it('Handles malformed dates', () => {
+    const start = 'This is n0t a d4te';
```
asiaznik, line 9: 31337 H4X0R 😉

apps/explorer/src/app/components/txs/details/transfer/blocks/transfer-participants.tsx
```diff
@@ -0,0 +1,118 @@
...
+          <div className="bg-zinc-200 dark:bg-zinc-800 border w-full pt-5 pb-3 px-3 boder-zinc-200 dark:border-zinc-800 relative">
+            <div className="text-xs z-20 relative leading-none">
+              {transfer.asset ? (
+                <SizeInAsset assetId={transfer.asset} size={transfer.amount} />
+              ) : null}
+            </div>
+
+            {/* Empty divs for the top arrow and the bottom arrow of the transfer inset */}
+            <div className="z-10 absolute top-0 left-1/2 transform -translate-x-1/2 -translate-y-1/2 rotate-45 w-4 h-4 dark:border-zinc-800 border-zinc-200 bg-white dark:bg-black border-r border-b"></div>
```
asiaznik, line 77: question (non-blocking): How about changing these divs into SVG?
Simple triangle:
```
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 9">
	<path d="M0,0L8,9l8,-9Z"/>
</svg>
```

apps/explorer/src/app/components/txs/details/transfer/blocks/transfer-rewards.tsx
edd: This one was a *tiny* bit fiddly because the type from the Block Explorer API always includes `_UNSPECIFIED` whereas from the GraphQL query they don't. But if I said that was too much work, I'd be pretty lazy. So it's done.

apps/explorer/src/app/components/txs/details/transfer/blocks/transfer-participants.tsx
edd: Fixed across Explorer

apps/explorer/src/app/components/txs/details/transfer/blocks/transfer-participants.tsx
edd: Pushed- thanks for the SVG & the pointers in Slack

asiaznik (approved):

PR #2905, opened 2023-02-14
feat(trading): datanode block header for env and node switcher

### LINKED ISSUE #1791: Update nodeHealth to use `x-block-height` header to check data node health

mattrussell36:
# Related issues 🔗

Closes #1791 

# Description ℹ️

- Changes node health UI in trading footer to be determined by comparing header block height from the data node, and the 'core' block height in the statistics query. This has the benefit of not requiring multiple requests across different nodes to determine overall node health.
- NodeSwitcher component has also been updated to use datanode block height rather than the query result block height.
- Store selected node in localStorage and attempts to use that on startup

# Technical 👨‍🔧

- Adds a custom ApolloLink store a map of datanode urls to header values
- Environment hook and NodeSwitcher dialog have been refactored so that additional nodes state and client instance is only created in the dialog itself and garbage collected when the dialog is closed.
- Adds NodeGuard component to trading app to prevent rendering app connected to a non functioning node
- Adds AppFailure with DynamicLoader component fallback/loading states consistent


daunatv:
Did a CMD+SHIFT+R, still same, created a new incognito tab, still offline, accessed using Firefox and it was online.
https://user-images.githubusercontent.com/100769129/218757546-2139ebb7-ae43-4392-af44-60759f8bfa57.mov

apps/liquidity-provision-dashboard/src/app/app.tsx
```diff
@@ -1,18 +1,46 @@
+import type { InMemoryCacheConfig } from '@apollo/client';
+import { NetworkLoader, useInitializeEnv } from '@vegaprotocol/environment';
 import { useRoutes } from 'react-router-dom';
 
 import '../styles.scss';
 import { Navbar } from './components/navbar';
 
 import { routerConfig } from './routes/router-config';
 
+const cache: InMemoryCacheConfig = {
```
edd, line 10: **thought (non-blocking)**: @asiaznik added a shared config in https://github.com/vegaprotocol/frontend-monorepo/pull/2914, which might be useful. Or they might be totally separate. Worth a quick look anyway - perhaps there's some benefit to all apps ffrom these cache changes.

libs/environment/src/hooks/use-environment.spec.ts
```diff
@@ -0,0 +1,259 @@
...
+};
+
+describe('useEnvironment', () => {
+  const env = process.env;
+
+  beforeEach(() => {
+    jest.resetModules();
+    process.env = { ...env };
+
+    process.env['NX_VEGA_ENV'] = mockEnvVars.VEGA_ENV;
```
edd, line 64: **thought (non-blocking)**: Seeing this made me realise what a big task removing our dependence on env vars is going to be. But staying consistent here is definitely the right thing for now.

edd (approved):
👍 Two non blocking thoughts occured to me while looking at this. Good test coverage, looking forward to seeing it in

notbot00 (approved):

apps/liquidity-provision-dashboard/src/app/app.tsx
mattrussell36: This config looks different, to the new default one. I'll leave this as is for now. Don't wan't to introduce anything unexpected

libs/environment/src/components/node-guard/node-guard.tsx
```diff
@@ -0,0 +1,27 @@
...
+  children,
+  failure,
+  skeleton,
+}: {
+  children: ReactNode;
+  failure: ReactNode;
+  skeleton: ReactNode;
+}) => {
+  const { error, loading } = useStatisticsQuery();
+  const nonIdealWrapperClasses =
```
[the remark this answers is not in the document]
mattrussell36: Renamed to be a bit more generic

PR #2921, opened 2023-02-15
feat(trading): deposit to trade collateral tab pinned row

### LINKED ISSUE #2864: Deposit flow from collateral window
## Story

As a new user of vega
I want to immediately see how to make a deposit and start trading from the UI
So that I can get started without having to find the right place to go

## Acceptance Criteria

- [ ] There is a call to action to deposit collateral when you have not connected your vega wallet e.g."deposit collateral" when there is no wallet connected , on the collateral tab, with intro copy e.g. "Deposit funds & start trading."
- [ ] This is visible when you first come to console

## Tasks

- [x] Explore and sketch
- [ ] Visual Design
- [ ] Dev
- [ ] Test

### Collateral tab and its sibling tabs
[Miro source](https://miro.com/app/board/uXjVPUenJos=/?moveToWidget=3458764542524313091&cot=14)

image

- This task was split from https://github.com/vegaprotocol/frontend-monorepo/issues/2145. This ticket only concerns the entry points to the deposit form. The other ticket focuses on the experience _in_ the deposit form.

MadalinaRaicu:
# Related issues 🔗

Closes #2864 

# Description ℹ️

Add  `deposit to trade` pinned row in collaterals tab.

# Demo 📺

image

# Technical 👨‍🔧

- [x] pin only virtual row


libs/accounts/src/lib/accounts-table.tsx
```diff
@@ -23,19 +25,42 @@ export interface Datasource extends IDatasource {
...
+export type MarketAsset = Pick<Asset, 'symbol' | 'name' | 'id' | 'decimals'>;
+
 export interface AccountTableProps extends AgGridReactProps {
   rowData?: AccountFields[] | null;
   datasource?: Datasource;
   onClickAsset: (assetId: string) => void;
   onClickWithdraw?: (assetId: string) => void;
   onClickDeposit?: (assetId: string) => void;
   isReadOnly: boolean;
+  marketAsset?: MarketAsset;
```
mattrussell36, line 37: suggestion (non blocking): I think a better name for this prop would just be `pinnedAsset`. It more clearly describes what will happen if you use it and doesn't conceptually tie it to a market

libs/accounts/src/lib/accounts-table.tsx
```diff
@@ -137,38 +163,58 @@ export const AccountTable = forwardRef<AgGridReact, AccountTableProps>(
               type="rightAligned"
               cellRenderer={({
                 data,
+                api,
```
mattrussell36, line 166: chore (non blocking): doesnt look like this parameter gets used

libs/accounts/src/lib/accounts-table.tsx
```diff
@@ -137,38 +163,58 @@ export const AccountTable = forwardRef<AgGridReact, AccountTableProps>(
...
-                      data-testid="withdraw"
-                      onClick={() =>
-                        onClickWithdraw && onClickWithdraw(data.asset.id)
-                      }
-                    >
-                      {t('Withdraw')}
-                    </ButtonLink>
-                  </>
-                ) : null;
+                if (!data) return null;
```
mattrussell36, line 168: suggestion (non blocking): remove else keywords here to reduce indentation and improve readability

```
if (!data) return null
if (data.asset.id === marketAssetId ...) {
  // return pinned row ui
}
// return default ui
```

mattrussell36 (approved):

PR #2924, opened 2023-02-16
fix(trading): re-use input errors instead of notification component

### LINKED ISSUE #2919: Revert input field errors to not use notification component

MadalinaRaicu:
# Related issues 🔗

Closes #2919 

# Description ℹ️

Revert using notification component instead of input error.

# Demo 📺

image

# Technical 👨‍🔧


mattrussell36 (changes requested):
Lets remove the NotificationError component as its not used

mattrussell36 (approved):

PR #2925, opened 2023-02-16
feat(explorer): new proposals index page

### LINKED ISSUE #2912: Create Proposals and Proposal details pages instead of raw JSON listings

asiaznik:
# Related issues 🔗

Closes #2912 

# Demo 📺

https://user-images.githubusercontent.com/1980305/219442584-b735941b-8d7f-43cc-858b-8f83f4ee53d3.mov

## Mobile view tweaks

image


apps/explorer/src/app/components/proposals/proposals-table.tsx
```diff
@@ -0,0 +1,221 @@
...
+          headerName={t('Title')}
+          field="rationale.title"
+        />
+        <AgGridColumn
+          colId="type"
+          maxWidth={180}
+          headerName={t('Type')}
+          field="terms.change.__typename"
+        />
+        <AgGridColumn
```
edd, line 109: **question (non-blocking)**:  Is it possible to prefill the filter selection on this? I think it's currently treating it like a string, where really it's an enumb of (something like) `[Enacted, Open]`. On the network I'm looking at, all proposals are `Enacted` and that gives me no clue what the alternatives would be if I wanted to filter to just the Active ones.

Screenshot 2023-02-17 at 09 58 16

libs/governance/src/lib/proposals-data-provider/proposals.mock.ts
```diff
@@ -50,7 +65,33 @@ export const marketUpdateProposal: ProposalListFieldsFragment = {
...
+            dataSourceSpecForTradingTermination: {
+              __typename: 'DataSourceDefinition',
+              sourceType: {
+                __typename: 'DataSourceDefinitionInternal',
+                sourceType: {
+                  __typename: 'DataSourceSpecConfigurationTime',
+                  conditions: [],
+                },
+              },
+            },
```
edd, line 94: **thought (non-blocking)**: I feel like we're missing a trick with mock generation from the schema. I haven't looked in to it because it works this way, and I suspect part of the problem is everything is a string - but there might be a way of adding in `example` annotations in the schema.

Definitely a thought. Absolutely definitely not a blocking thought.

edd (approved):
Looks good. Only non-blocking comments added - can be addressed here or moved out to a separate ticket

apps/explorer/src/app/components/proposals/proposals-table.tsx
asiaznik: There's an out of the shelf filter like that in Enterprise edition of Ag Grid. For our current Community edition we'd need to make our own.

AnExsomnis (approved):

PR #2931, opened 2023-02-17
fix(governance): navigation token drop down behaviour

### LINKED ISSUE #2656: New navigation token drop down behaviour
## Description

When using the token drop down menu in the main site navigation, I can accidentally jump to the token landing page without ever seeing that there is a menu of icons, based on where I click. 

## Steps to Reproduce

1. Go to 'https://token.fairground.wtf/token'
2. Click underneath "Token" in the primary navigation
3. Notice you have jumped to the token page without seeing the token drop down. This behaviour does not occur if you click on the word Token, only on the space underneath. Suspect the bug is related to the tab pattern - it's like you click where the tab indicator would be and that's what's caused this problem.

## Expected behaviour

I'd expect a click to trigger the menu opening for the token sub menu, and that be **the only way you can reach the pages** contained within the token drop down.

### LINKED ISSUE #2875: [Token]: Token drop down menu should not be opaque on tranches page
The Token dropdown menu z-index needs to be higher than content in the 'main' component

Screenshot 2023-02-08 at 10 42 30

sam-keen:
# Related issues 🔗

Closes #2656
Closes #2875

# Description ℹ️

Ensures Token dropdown menu opens consistently regardless of where in the link is clicked. Also added the correct cursor (good spot @neildawson) and as it was a one liner, I also fixed the z-index issue with the dropdown itself (#2875)

### DIFF
--- a/apps/token/src/components/nav/nav-dropdown.tsx
+++ b/apps/token/src/components/nav/nav-dropdown.tsx
@@ -18,7 +18,7 @@ export const NavDropDown = ({ navbarTheme }: { navbarTheme: NavbarTheme }) => {
       <AppNavLink
         name={
           <NavDropdownMenuTrigger
-            className="w-auto flex items-center"
+            className="w-auto flex items-center -m-3 p-3 cursor-pointer"
             data-testid="state-trigger"
             onClick={() => setOpen(!isOpen)}
           >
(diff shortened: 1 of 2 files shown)


notbot00 (approved):

edd (approved):
:thumbs That'll do it

PR #2934, opened 2023-02-17
feat(wallet): tighten up the flow for "view as vega user"

### LINKED ISSUE #2725: Tighten up the flow for "view as vega user"
## Chore
Catching a couple of things to make the "View as vega user" more easily understandable. 

## Tasks
On wallet connect popup
- [ ] Separate "View as vega user" from the wallet connect options with "OR"

Screenshot 2023-01-25 at 11 21 31

On View as Vega user state / page
- [ ] Add explanation of what you are doing by viewing as a vega user "Browse from the perspective of another Vega user in read-only mode"
- [ ] Add title "View as Vega user"
- [ ] Update CTA to "Browse network"
- [ ] Add navigation to go back

Screenshot 2023-01-25 at 11 24 17

sam-keen:
# Related issues 🔗

Closes #2725

# Description ℹ️

Makes changes to the 'view as vega user' wallet connection option. Also adds a back button to the hosted wallet and vega user connection options.

# Demo 📺

Screenshot 2023-02-17 at 16 20 55

Screenshot 2023-02-17 at 16 20 45

Screenshot 2023-02-17 at 16 20 38

### DIFF
--- a/libs/wallet/src/connect-dialog/view-connector-form.tsx
+++ b/libs/wallet/src/connect-dialog/view-connector-form.tsx
@@ -1,5 +1,11 @@
 import { t } from '@vegaprotocol/react-helpers';
-import { Button, FormGroup, Input, InputError } from '@vegaprotocol/ui-toolkit';
+import {
+  Button,
+  FormGroup,
+  Icon,
+  Input,
+  InputError,
+} from '@vegaprotocol/ui-toolkit';
 import { useForm } from 'react-hook-form';
 import type { ViewConnector } from '../connectors';
 import { useVegaWallet } from '../use-vega-wallet';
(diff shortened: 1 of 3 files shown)


libs/wallet/src/connect-dialog/view-connector-form.tsx
```diff
@@ -41,24 +49,46 @@ export function ViewConnectorForm({
...
-        {t('Connect')}
-      </Button>
-    </form>
+    <>
+      <button
+        onClick={reset}
+        className="absolute p-2 top-0 left-0 md:top-2 md:left-2"
+        data-testid="back-button"
+      >
+        <Icon name={'chevron-left'} ariaLabel="back" size={4} />
```
notbot00, line 58: suggestion (non-blocking): maybe a text with "Back" could be useful and more obvious for the user then just the small icon here, but this is just my biased take

notbot00 (approved):

mattrussell36 (approved):

dexturr (approved):

libs/wallet/src/connect-dialog/view-connector-form.tsx
sam-keen: Thanks for the suggestion. I think this is unobtrusive, and if the user is looking for a back button, they'll notice it. It follows the same convention as the close button which is a similarly positioned, similarly sized X. I think the two in conjunction, along with their placement, makes for a nice piece of simple UI.

PR #2944, opened 2023-02-20
chore(trading): filtering and sorting for ledger entries

### LINKED ISSUE #2818: Filtering and sorting for ledger entries does not work
## Description

Ledger entries provide a lot of data. The filtering doesn't seem to work, and thus the table is pretty much unusable for an LP or any other user that will have created lots of data.

Additionally the ledger entry table does not have a date range applied by default.. It should to avoid a huge data set and/or possibly very long wait for data.

Raised on behalf of an LP 

## Steps to Reproduce

1. Place a lot of orders
2. Click on Portfolio > Ledger entries
3. Choose the filter option
4. See that the results don't change

## Screen recording

https://user-images.githubusercontent.com/62548908/216098795-c9e40321-e88c-4edd-9974-274ea3405b66.mov

macqbat:
# Related issues 🔗

Closes #2818 

# Description ℹ️

Add time range limits to the ledgerEntries query, for making possible client side sorting and filtering. Default query is set on last 7 days. User can modified it to shorter or earlier period to max last 30 days. Buttons in date pickers will be dynamically enabled and disabled regarding to users selects.

# Demo 📺

image
image

# Technical 👨‍🔧

We implemented single query instead infinite model - one bigger request (for max 5000 entries) provides data for review it by the user. The request has default range last 7 days. 
The earliest data a user can query is 30 days ago.
The max range of time a user can query is 7 days. 
Min and max date in date pickers will be limited acording to rules above.
Sorting by column works on existing set of data on client side as well as filtering by accounts type.
Filtering by  date range works on server side (due to querying server with given range parameters).


MadalinaRaicu (approved):

mattrussell36 (changes requested):
It appears that the table flashes when using filters, I think this is because its causing additional requests on filter? Will send you a screen recording

libs/ledger/src/lib/ledger-table.tsx
```diff
@@ -35,24 +37,35 @@ export const TransferTooltipCellComponent = ({
   );
 };
 
+const defaultRangeFilter = { start: formatRFC3339(subDays(Date.now(), 7)) };
+const dateRangeFilterParams = {
+  maxSubDays: 30,
+  maxDaysRange: 7,
```
mattrussell36, line 43: question (non blocking): Why 7 here? From trying out the table it seems a little restrictive. Was there a noticeable increase in time for the query to resolve if you increased it?

libs/react-helpers/src/lib/grid/date-range-filter.tsx
```diff
@@ -1,113 +1,248 @@
...
-        new Date(event.target.value).toISOString().replace('Z', '000000Z'),
-    });
-  };
-
-  useEffect(() => {
-    props?.filterChangedCallback();
-  }, [value, props]);
+    const Notification = () => {
+      const not = error ? (
+        <div className="border-vega-pink bg-vega-pink-300 dark:bg-vega-pink-650 border rounded p-2 flex items-start gap-2.5 m-4 w-full">
```
mattrussell36, line 177: chore(blocking): Use InputError underneath the date inputs

libs/ledger/src/lib/ledger-table.tsx
mattrussell36: Additionally, was there a reason for restricting it to 30 days in the past?

If we have a sensible range, I think we should allow any previous date

mattrussell36 (changes requested):

libs/react-helpers/src/lib/grid/date-range-filter.tsx
macqbat: can't easily do this because there are circular dependencies between `react-helpers` and `ui-toolkit`

libs/ledger/src/lib/ledger-table.tsx
macqbat: I removed restriction of max 30 days in past.

libs/react-helpers/src/lib/grid/date-range-filter.tsx
```diff
@@ -1,113 +1,248 @@
...
-      [event.target.name]:
-        event.target.value &&
-        new Date(event.target.value).toISOString().replace('Z', '000000Z'),
-    });
-  };
-
-  useEffect(() => {
-    props?.filterChangedCallback();
-  }, [value, props]);
+    const Notification = () => {
```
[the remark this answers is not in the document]
macqbat: yeah, that stayed from some previous version - now it's normal fragment from `useMemo`

libs/react-helpers/src/lib/grid/date-range-filter.tsx
```diff
@@ -1,113 +1,248 @@
...
+      if (validate(name, date, update)) {
+        setValue({
+          ...value,
+          ...update,
+        });
+      }
     };
-  });
+    useEffect(() => {
+      props?.filterChangedCallback();
```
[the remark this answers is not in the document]
macqbat: we can't. Component started to work differently when I did it - refresh callback happened still with old value

libs/react-helpers/src/lib/grid/date-range-filter.tsx
```diff
@@ -1,113 +1,248 @@
...
-
-        return { value };
-      },
-
-      setModel(model?: { value: Schema.DateRange } | null) {
-        setValue(model?.value || defaultFilterValue);
-      },
+      }
+      if (validate(name, date, update)) {
+        setValue({
```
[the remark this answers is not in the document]
macqbat: done, thanks

bglownia (approved):

libs/react-helpers/src/lib/grid/date-range-filter.tsx
macqbat: Moreover we need rather to put errors above, because if native date picker is expanded, it cover everything what is under - that's why I left some space over inputs

mattrussell36: Yikes, we'll need to fix that. Okay makes sense to go above, but can we just make it red text so it looks the same as InputError

libs/ledger/src/lib/ledger-table.tsx
mattrussell36: What about the maxDaysRange? I think 7 is is good for the initial query but I don't think we should restrict users from selecting a wider range unless there are performance reasons not too.

libs/react-helpers/src/lib/grid/date-range-filter.spec.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+describe('DateRangeFilter', () => {
+  it('should be properly rendered', async () => {
+    const defaultRangeFilter = {
+      start: '2023-02-14T13:53:01+01:00',
+      end: '2023-02-21T13:53:01+01:00',
+    };
+    const displayStartValue = '2023-02-14T12:53:01.000';
+    const displayEndValue = '2023-02-21T12:53:01.000';
+    render(
```
mattrussell36, line 17: @macqbat I've pushed some changes to this test to simplify the code. Please note:

- removed `act`. testing-library/react automatically wraps with act when you use `render`. Normally you shouldn't need to wrap components with act when rendering. We often do when working with AgGrid as there are multiple events that update state after AgGrid Mounts
- Removed storing the container element in a variable. Just use  `screen` where possible
- Combined input assertions. getByLabelText will return the `<input>` for a corresponding label. It will error if the input isn't found so its safe to immediately check the value on it too.

macqbat: nice, thanks

macqbat:
> 

It has been fixed already in some previous works

macqbat:
> It appears that the table flashes when using filters, I think this is because its causing additional requests on filter? Will send you a screen recording

It has been fixed already in some previous works

libs/ledger/src/lib/ledger-table.tsx
macqbat: I'm afraid we will let the users for requesting a tremendous amount of data, hundreds thousands of rows personalised, hard to cache entries and then we will face many timeout errors and high load of servers. I can remove easily period of days limitation - actually it's just remove of one filter parameter, but I wouldn't do it :-|

mattrussell36: Yeah trying it again and even with 7 days the query takes a long time resolve, not sure if this is causing any danger to the datanode, but lets not risk it for now. We can experiment/discuss with core in future

mattrussell36 (approved):

PR #2960, opened 2023-02-22
feat(governance): tooltips for the validator details page

### LINKED ISSUE #2653: Add tooltips for the validator details page
## The Chore

Add tooltips for the validator details page so users can understand what they're looking at.

## Tasks

- [x] Populate definitions @JonRay15 
- [ ] Do it @sam-keen 

For each one I list the section and the field, if I havent listed it assume it doesnt need a tooltip.

Profile
About this validator > External URL provided by the validator linking to informafion about themselves.
Status > Consensus, Standby, or Pending (Candidate) depending on how much stake the validator has attracted

Stake 
Staked by operator > The stake provided as self-stake by the node operator, must be at least the minimum stake as defined by network parameter
Staked by delegates > The stake delegated to the node by other users
Pending stake > copy what is on the main page already 
Stake share > copy what is on the main page already 

Penalties 
Overstaked penalty > A penalty applied for having more stake than the optimal stake for the network. Designed to avoid concentration of voting power with a small number of validators.
Performance score > A measure of how often a validator proposed blocks in the last epoch relative to how many they should be expected to propose based on their voting power.
Performance penalties > The penalty applied for having a performance score of less than 1

Voting Power 
Unnormalised voting power > The voting power of the validator based on their final validator score after all penalties have been applied
Normalised voting power > The voting power of the validator, adjusted to ensure all validator scores sum to 1, used for distribution of rewards.

Suggest we also include some words at the top similar to the main page directing people to the docs for more detailed information on any of the things shown here.

sam-keen:
# Related issues 🔗

Closes #2653

# Description ℹ️

Provides descriptions for the various validator table values, displayed in tooltips


apps/token/src/routes/staking/shared.ts
```diff
@@ -95,3 +95,8 @@ export const getTotalPenalties = (
     2
   );
 };
+
+export const getStakePercentage = (total: BigNumber, stakedOnNode: BigNumber) =>
+  total.isEqualTo(0) || stakedOnNode.isEqualTo(0)
```
dexturr, line 100: this seems wrong, if stakedOnNode is 0 then stake percentage is 0 not - no?

dexturr (approved):

PR #3031, opened 2023-02-28
test(trading): fixed capsule tests

### LINKED ISSUE #3028: Fix two capsule tests:
## The Chore

Fix two capsule tests:
    1) can place and receive an order
    2) approved amount is less than deposit

daro-maj:
# Related issues 🔗

Closes #3028 

# Description ℹ️

Fix two capsule tests:
1) can place and receive an order
2) approved amount is less than deposit

### DIFF
--- a/apps/trading-e2e/src/integration/capsule.cy.ts
+++ b/apps/trading-e2e/src/integration/capsule.cy.ts
@@ -40,6 +40,7 @@ const newPrice = '200';
 const completeWithdrawalBtn = 'complete-withdrawal';
 const submitTransferBtn = '[type="submit"]';
 const transferForm = 'transfer-form';
+const depositSubmit = 'deposit-submit';
 
 // Because the tests are run on a live network to optimize time, the tests are interdependent and must be run in the given order.
 describe('capsule - without MultiSign', { tags: '@slow' }, () => {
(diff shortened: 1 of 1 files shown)


apps/trading-e2e/src/integration/capsule.cy.ts
```diff
@@ -410,12 +410,18 @@ describe('capsule', { tags: '@slow' }, () => {
...
     cy.get(assetSelectField, txTimeout).select(btcName, { force: true });
+    cy.contains('Deposits of tBTC not approved').should('not.exist');
+    cy.contains('Use maximum').should('be.visible');
     cy.get(amountField).clear().type('20000000');
-    cy.getByTestId('deposit-approve-submit').should('be.visible');
+    cy.getByTestId(depositSubmit).should('be.visible');
+    cy.getByTestId(depositSubmit).click();
+    cy.getByTestId('input-error-text').should(
+      'contain.text',
+      'Amount is above approved amount.Update approve amount'
```
MadalinaRaicu, line 423: I don't understand this text.

MadalinaRaicu: Is this the input error we have? maybe we should update with 
```suggestion
      'Amount is above the approved amount. Update the amount'  ? 
```

MadalinaRaicu (approved):

PR #3042, opened 2023-03-01
fix(token): router incorrectly configured, causing query spamming

dexturr:
# Description ℹ️

Mainnet on token is spamming requests due to a poorly configured router

The router was nesting two routes incorrectly. Due to the type policy the first would fire off and then the second, which would cause changes to the first query causing it to refetch, repeat infinitely. These did not need to be passing data between each other, and should not have been used together at all as there was a different container for that route, which this PR is now using.

### DIFF
--- a/apps/token/src/routes/router-config.tsx
+++ b/apps/token/src/routes/router-config.tsx
@@ -52,10 +52,18 @@ const LazyRedemptionTranche = React.lazy(
       /* webpackChunkName: "route-redemption-tranche", webpackPrefetch: true */ './redemption/tranche'
     )
 );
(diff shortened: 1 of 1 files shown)


edd (approved):
Could you explain *why* this fixes it? I don't understand from reading the changelog. I'm willing to take your word that it does though.

dexturr:
@edd Description updated

edd:
> @edd Description updated

Thanks! Subtle.

asiaznik (approved):

sam-keen (approved):

dexturr:
These unit tests were broken before this PR, so I am merging.

PR #3043, opened 2023-03-01
fix(trading): remove margin warning semicolon and word break on notification message

MadalinaRaicu:


asiaznik (approved):

libs/ui-toolkit/src/components/notification/notification.tsx
```diff
@@ -82,7 +82,7 @@ export const Notification = ({
             {title}
           </div>
         )}
-        <div className="text-sm">{message}</div>
+        <div className="text-sm [word-break:break-word]">{message}</div>
```
MadalinaRaicu, line 85: tailwind break-word comes with overflow-wrap and it won't work. https://tailwindcss.com/docs/word-break

PR #3070, opened 2023-03-02
fix: unsubscribe from watch order input

### LINKED ISSUE #3041: Deal ticket decimals input bug
Bug:

> when entering a price in the deal ticket box, if you are to correct it, and erase the things after the decimal e.g: you have 1631.60 and want to make it 1631.70. you start by removing the 0 and then the 7 and you d'expect the . to still be there, but the console remove it for you. It's attrocious. I've had to re-add the . every single time because of that. Is there rational about it? I would think that you should not try to be too smart about that, and let the user remove the . if they want to.
Same field, but if you are to enter an invalid character (say a , instead of a .) that clears out of the input

https://user-images.githubusercontent.com/16125548/222100539-e9593622-bb51-4e17-9013-2b9e504e6c7a.mov

PR that introduced it together with the orderbook feature: 
https://github.com/vegaprotocol/frontend-monorepo/commit/7ea7edc1e2246ac897e5b0254a49ee3662e3e8b4#diff-5d8b312316d8e4fb74566cfa37e079b2c9c6200535548558048c7ec0f552ac44

- usage of react-hook-form watch with the callback pattern seems incorrect
- Theres a lot more renders occurring on value change and this might in turn be writing to local storage every time?
-  `checkForErrors` was it intentional to not return early and therefor not submit an order if errors were found?

MadalinaRaicu:
# Related issues 🔗

Closes #3041 

# Description ℹ️

# Demo 📺

# Technical 👨‍🔧

### DIFF
--- a/libs/deal-ticket/src/components/deal-ticket/deal-ticket.tsx
+++ b/libs/deal-ticket/src/components/deal-ticket/deal-ticket.tsx
@@ -37,6 +37,7 @@ import {
   usePersistedOrderStoreSubscription,
 } from '@vegaprotocol/orders';
 import { OrderType } from '@vegaprotocol/types';
+import isNumber from 'lodash/isNumber';
 
 export type TransactionStatus = 'default' | 'pending';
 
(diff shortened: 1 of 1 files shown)


libs/deal-ticket/src/components/deal-ticket/deal-ticket.tsx
```diff
@@ -82,14 +83,17 @@ export const DealTicket = ({
...
-      setPersistedOrder(orderData as DealTicketFormFields);
-    }
-  });
+  useEffect(() => {
+    const { unsubscribe } = watch((orderData) => {
+      const persistable = !(
+        orderData.type === OrderType.TYPE_LIMIT && orderData.price === ''
+      );
+      if (persistable) {
+        setPersistedOrder(orderData as DealTicketFormFields);
```
MadalinaRaicu, line 92: on firefox the cursor does not move, but on chromium after first decimal decimal point is erased and cursor is moved at the start of the input. 

not sure where else to persist the order

mattrussell36: @MadalinaRaicu I was tinkering with this today and have a nice solution, lets chat tomorrow

MadalinaRaicu:
contuinuing on [fix/order-store-watch](https://github.com/vegaprotocol/frontend-monorepo/tree/fix/order-store-watch)

PR #3073, opened 2023-03-03
fix(trading): add a timeout to show subscriptions as failed

mattrussell36:
# Related issues 🔗

N/A

# Description ℹ️

Add a 3 second timeout to fail a node in the node switcher if data from a TimeUpdate bus event is not received within that period. Now nodes will correctly be marked as 'failed' and not stuck in a pending state

# Demo 📺

Screenshot 2023-03-02 at 22 30 44


edd (approved):
Tried it out, works nicely 👍

PR #3133, opened 2023-03-09
fix(trading,explorer,governance): consistent zustand usage

mattrussell36:
# Related issues 🔗

N/A

# Description ℹ️

Makes all zustand stores follow the typescript recommendations in the docs. This makes mocking the stores and ensuring the stores are cleaned up after tests consistent and should help eliminate hard to track down testing errors


libs/orders/src/lib/order-hooks/use-order-store.spec.ts
```diff
@@ -9,6 +9,8 @@ import { OrderType } from '@vegaprotocol/types';
 
 jest.mock('zustand');
 
+jest.mock('zustand');
```
bglownia, line 12: `jest.mock('zustand')` called twice

PR #3143, opened 2023-03-09
feat(orders): use default filters in orders list

### LINKED ISSUE #3072: Filter orders by default on trading page
## Story

As a user
I want the order list to be filtered when on the trading page
So that I only see relevant orders

## Acceptance Criteria

- [x] I only see active and parked orders by default when viewing orders on the trading page
- [x] I can filter the list
- [x] I can understand that I am looking at a filtered view of orders (not all orders) (ideally using existing AG Grid elements)
- [x] Orders are shown for all markets
- [x] Infinite scroll is removed
- [ ] Sorting and filtering is applied client side only
- [x] The initial request for orders should fetch the latest ~~1000~~ 5000 orders
- [x] Subsequent data is fetched by using the date filter and using the selected dates as the DateRange for the query
- [x] Orders arriving via subscription are immediately applied to the grid
- [ ] Apply a default sort order on type column to group LP orders to the bottom

## Tasks

- [x] Check with Lewis on liquidity pegged orders and whether they should be shown (and if so, at the bottom?)
- [ ] Dev
- [ ] Test

https://vegaprotocol.slack.com/archives/C037STCKMGC/p1677749496425519

bglownia:
# Related issues 🔗

Closes #3072

### DIFF
--- a/libs/orders/src/lib/components/order-list-manager/order-list-manager.tsx
+++ b/libs/orders/src/lib/components/order-list-manager/order-list-manager.tsx
@@ -16,6 +16,7 @@ import { useOrderListData } from './use-order-list-data';
 import { useHasActiveOrder } from '../../order-hooks/use-has-active-order';
 import type { Filter, Sort } from './use-order-list-data';
 import { useBottomPlaceholder } from '@vegaprotocol/react-helpers';
+import { OrderStatus } from '@vegaprotocol/types';
 import {
   normalizeOrderAmendment,
   useVegaTransactionStore,
@@ -53,6 +54,12 @@ const CancelAllOrdersButton = ({
   ) : null;
 };
 
+const initialFilter: Filter = {
+  status: {
+    value: [OrderStatus.STATUS_ACTIVE, OrderStatus.STATUS_PARKED],
+  },
+};
+
 export const OrderListManager = ({
   partyId,
   marketId,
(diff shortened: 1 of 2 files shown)


libs/orders/src/lib/components/order-list-manager/order-list-manager.tsx
```diff
@@ -53,6 +54,15 @@ const CancelAllOrdersButton = ({
   ) : null;
 };
 
+const initialFilter: Filter = {
+  status: {
+    value: [OrderStatus.STATUS_ACTIVE, OrderStatus.STATUS_PARTIALLY_FILLED],
+  },
+  type: {
+    value: [OrderType.TYPE_LIMIT],
```
mattrussell36, line 62: question (blocking): Are you intentionally filtering by order type too here? Pretty sure we want all regular (lp orders aside) types showing

mattrussell36 (changes requested):
I think it would make sense to break out the filtering by lp/pegged orders into a separate task.

mattrussell36 (approved):

libs/orders/src/lib/components/order-list-manager/order-list-manager.tsx
bglownia: It's already there. `marketId` is passed to as variable to data provider next to `filter `

PR #3145, opened 2023-03-09
chore(environment): fetch networks configuration from repo

### LINKED ISSUE #3101: Fetch network configs from networks repos
## The Chore

Currently we fetch a list of nodes from json files deployed to static.vega.xyz. These need to be fetched directly from the networks repo

## Tasks

- [ ] Work with dev ops to add a CI check on the validity of the .toml files
- [x] Update start up logic to fetch network configs directly from https://github.com/vegaprotocol/networks and https://github.com/vegaprotocol/networks-internal
- [x] Parse .toml result
- [x] Ensure that we properly handle the experience if github is down
- [ ] Delete old config files
- [x] Update env vars to point to the relevant .toml files

macqbat:
# Related issues 🔗

Closes #3101 

# Description ℹ️

Parse toml files from github network repo for nodes configuration

# Demo 📺

-

# Technical 👨‍🔧

-


libs/environment/src/hooks/use-environment.ts
```diff
@@ -159,7 +160,14 @@ export const useInitializeEnv = () => {
...
-  const cfg = await res.json();
+  let cfg: Record<string, unknown>;
+  if (url.match(/github(.+)\.toml$/)) {
+    const content = await res.text();
+    const parsed = tomlParse(content);
+    cfg = { hosts: parsed.API.GraphQL.Hosts };
+  } else {
+    cfg = await res.json();
+  }
   const result = configSchema.parse(cfg);
```
mattrussell36, line 171: chore (blocking): can we update the zod schema to match what expected correct output of parsing the toml file will be. This way it will correctly throw if the underling toml file is invalid. With this change as is, parsing with zod is pointless

mattrussell36 (changes requested):

libs/environment/src/utils/validate-configuration.ts
```diff
@@ -3,3 +3,7 @@ import z from 'zod';
 export const configSchema = z.object({
```
mattrussell36, line 3: chore (blocking): Can we just remove the configSchema now? Its kind of pointless as we are validating the structure of the object we literally just created on line 176

mattrussell36 (changes requested):
We need to sync up with the other explorer/governance as this will break their apps

libs/environment/src/hooks/use-environment.ts
```diff
@@ -159,7 +163,20 @@ export const useInitializeEnv = () => {
 const fetchConfig = async (url?: string) => {
   if (!url) return [];
   const res = await fetch(url);
-  const cfg = await res.json();
+  let cfg: Record<string, unknown>;
+  if (url.match(/github(.+)\.toml$/)) {
+    const content = await res.text();
+    const parsed = tomlParse(content);
```
dexturr, line 169: what if this is not valid toml or res.text fails, probably need some kind of catch

libs/environment/src/hooks/use-environment.ts
```diff
@@ -159,7 +163,20 @@ export const useInitializeEnv = () => {
 const fetchConfig = async (url?: string) => {
   if (!url) return [];
   const res = await fetch(url);
```
dexturr, line 165: if github is down or similar we need a way of safely failing in a reasonable way. We don't want to be susceptible to crashes from the downtime of an application we have no control over.

libs/environment/src/hooks/use-environment.ts
mattrussell36: This is wrapped with try catch in the main body of the initialize function. If it fails initialization will continue, and the app will attempt to load via VEGA_URL or what is in localstorage

dexturr: Noice

libs/environment/src/utils/validate-configuration.ts
macqbat: we still have old alternative working. I thought leaving old validation for this branch would be still useful. Maybe removal of it should be make along with this old alternative, just after deploying and changing envs on netlify for save consistent transition on new configuration source?

mattrussell36: I see what you are saying. Personally I think we should just make a clean switch to _only_ using the toml files rather than take on this debt. Every app has VEGA_CONFIG_URLs configured by .env files so lets lets just update those for each app in this PR. Thoughts @dexturr ?

dexturr: As long as we don't remove the static files as they are currently, I'm all for simplicity. The live token site is using those, but that is off the main branch which is super diverged from this one and will be a clean release once the next mainnet release goes out. cc @notbot00

mattrussell36: Okay yeah lets not touch them. @macqbat can you update this PR so that we only ever fetch from the toml file in github then we can remove the url matching condition keeping things simpler. We'll also need to update all the .env files to point to the github file

macqbat: done

mattrussell36 (approved):

PR #3169, opened 2023-03-13
chore(trading): display latency in block height progressing as warn or error

### LINKED ISSUE #3074: Handle entire network being down
## The Chore

If the entire network goes down the app will still report things as 'Operational' due to no block heights progressing and therefore the x-block-height header and query result block height being in sync

Existing data:

- Block heights (header and API)
- Timestamp (header and API)
- Block height number

Might want some kind of Amber state that explains that there is n time since last block (so that could indicate a problem but not necessarily).

## Tasks

- [ ] Establish what is an acceptable delay, and what might be an 'amber' delay
- [ ] Check that the block height is increasing at an expected rate or use TimeUpdate to ensure evens are being emitted

macqbat:
# Related issues 🔗

Closes #3074 

# Description ℹ️

Save previous time in store in case when block height is not changing (increase nor decrease). Latencies are temporary set on (no matter what's on screenshots):
- "Amber" latency on 10 sec
-  Error latency on 20 sec 

These settings as well as used wording are to improve :-)
Some re-look from design/UX perspective would also may be needed

# Demo 📺

image
image

# Technical 👨‍🔧

-


apps/trading/components/footer/footer.tsx
```diff
@@ -11,7 +11,10 @@ export const Footer = () => {
   const setNodeSwitcher = useGlobalStore(
     (store) => (open: boolean) => store.update({ nodeSwitcherDialog: open })
   );
-  const { blockDiff, datanodeBlockHeight } = useNodeHealth();
+  const { blockDiff, datanodeBlockHeight, datanodeVegaTime } = useNodeHealth();
+  const blockUpdateMsLatency = datanodeVegaTime
+    ? Date.now() - datanodeVegaTime.valueOf()
```
mattrussell36, line 16: suggestion (non blocking): using getTime is a little more 'standard' javascript

apps/trading/components/footer/footer.tsx
```diff
@@ -70,13 +79,19 @@ export const NodeUrl = ({ url }: NodeUrlProps) => {
 
 interface HealthIndicatorProps {
   blockDiff: number | null;
+  blockUpdateMsLatency: number;
 }
 
 // How many blocks behind the most advanced block that is
 // deemed acceptable for "Good" status
 const BLOCK_THRESHOLD = 3;
```
mattrussell36, line 87: suggestion (non blocking): I think we should move these consts to the use-node-health file so that all 'health' logic is kept together

apps/trading/components/footer/footer.tsx
macqbat: done

apps/trading/components/footer/footer.tsx
macqbat: So I let myself for a small refactoring of use-node-health for close whole logic in one hook

MadalinaRaicu (approved):

PR #3172, opened 2023-03-13
fix(ui-toolkit): anchor button should be same size as button by default

dexturr:
# Related issues 🔗

N/A

# Description ℹ️

Anrchor button did not have text centered or the same spacing by default as a standard button. These should be consistent.

### DIFF
--- a/libs/ui-toolkit/src/components/button/button.tsx
+++ b/libs/ui-toolkit/src/components/button/button.tsx
@@ -18,6 +18,9 @@ const md = 'px-10 py-2 text-base';
 const lg = 'px-14 py-4';
 const fillClasses = 'block w-full';
 const defaultClasses = [
+  'px-10 py-2',
+  'text-base',
+  'text-center',
   'border-neutral-500',
   'bg-transparent',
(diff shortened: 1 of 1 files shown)


libs/ui-toolkit/src/components/button/button.tsx
```diff
@@ -18,6 +18,9 @@ const md = 'px-10 py-2 text-base';
 const lg = 'px-14 py-4';
 const fillClasses = 'block w-full';
 const defaultClasses = [
+  'px-10 py-2',
```
mattrussell36, line 21: chore (blocking): defaultClasses is for a button variant (IE border,bg,text colors only) and so shouldn't contain padding/sizing stuff. Additionally this will apply multiple padding classes from which I'm not sure which styles will take precedence.

Suggest moving text-base and text center to the `base` variable. Also it looks like AnchorButton uses size of `lg` by default whereas the regular button doesn't, this is why the sizes are different. Lets just update AnchorButton to use `md` button by default (and remove padding classes) though you'll need to find all usage of AnchorButton and make sure the correct size is used

mattrussell36 (changes requested):

PR #3177, opened 2023-03-13
fix(trading): deal ticket validation on zero account balance and margin warnings

MadalinaRaicu:
# Related issues 🔗

Closes N/A

# Description ℹ️

- [x] deal ticket validation on zero account balance and margin warnings


libs/deal-ticket/src/components/deal-ticket/deal-ticket.tsx
```diff
@@ -141,7 +141,6 @@ export const DealTicket = ({
     pubKey,
     setError,
     clearErrors,
-    errors.summary,
```
MadalinaRaicu, line 144: this was causing crazy re-renders

PR #3187, opened 2023-03-14
fix(trading): auction triggers liquidity API change in testnet

### LINKED ISSUE #3150: Update liquidity auction triggers - API change
https://github.com/vegaprotocol/vega/pull/7803

Auction triggers have been updated.

MadalinaRaicu:
# Related issues 🔗

Closes #3150 

# Description ℹ️

Update auction triggers for liquidity. `testnet`


apps/liquidity-provision-dashboard/src/app/components/status/status.tsx
```diff
@@ -52,14 +53,16 @@ export const Status = ({
...
-const getTooltipDescription = (status: string) => {
+const getTooltipDescription = (status: string, trigger?: string) => {
   let tooltipDescription = '';
   switch (status) {
     case Schema.MarketTradingModeMapping.TRADING_MODE_CONTINUOUS:
       tooltipDescription =
         'This is the standard trading mode where trades are executed whenever orders are received';
       break;
-    case `${Schema.MarketTradingModeMapping.TRADING_MODE_MONITORING_AUCTION} - ${Schema.AuctionTriggerMapping.AUCTION_TRIGGER_LIQUIDITY}`:
+    case `${
```
mattrussell36, line 63: suggestion (non blocking): Not to be fixed in this PR here, but I don't love having this concatenated string as a dynamic case in the switch statement. 

I think this could be improved by a small refactor to make `getStatus` and `getTooltipDescription` accept only the raw enum values (NOT the mappings) and only returning raw strings for rendering, and also not depend on each other.

Additionally I would 'mark' each returned string in getTooltipDescription by using the `t` function individually for each result.

Non urgent but ping @ciaran-

mattrussell36 (approved):

mattrussell36 (changes requested):
@MadalinaRaicu build is failing because libs/deal-ticket/src/utils/validate-type.ts needs to handle the changes to auction trigger types

MadalinaRaicu:
This fix will be merged in stagnet1 and release/testnet only.

apps/liquidity-provision-dashboard/src/app/components/status/status.tsx
MadalinaRaicu: @ciaran-  I updated these methods as described above

ciaran-: That's better, yeah - thanks @MadalinaRaicu ! 

How do we feel about this once the LP dashboard moves to the website - do we still think we'd want a specific-to-the-dashboard function to do this mapping or do we think we could re-use the console component for that purpose?

MadalinaRaicu:
> @MadalinaRaicu build is failing because libs/deal-ticket/src/utils/validate-type.ts needs to handle the changes to auction trigger types

issue fixed by @mattrussell36 's commit [02829d8](https://github.com/vegaprotocol/frontend-monorepo/pull/3187/commits/02829d88e40f634079e01a37ab0ca5bb140780a0)

PR #3189, opened 2023-03-14
fix(trading): hide transfer button in view as mode

### LINKED ISSUE #3079: Remove link to "transfer" when viewing as another user
## The Chore

in this case I am viewing as another VEGA key
The link to transfer should not be here, in the same way as there is no link to cancel.

Screenshot 2023-03-03 at 12 45 40

MadalinaRaicu:
# Related issues 🔗

Closes #3079 

# Description ℹ️

 Hide the transfer button in `view as` mode.

# Demo 📺

image

# Technical 👨‍🔧


asiaznik (approved):

edd:
Tested it out after the fact, but it works fine

edd:
Tested it out after the fact, but it works fine

Screenshot 2023-03-14 at 15 19 28
Screenshot 2023-03-14 at 15 19 38

PR #3202, opened 2023-03-15
chore(ui-toolkit): rework intent colours

ciaran-:
Replace the pink/yellow with blue for 'Primary' intent. Pink is our danger colour.

# Related issues 🔗

Closes #[Issue number here]

# Description ℹ️

Changes the Intent.Primary to blue from neutral,  addresses suggestions from this thread: https://vegaprotocol.slack.com/archives/C8WS7M50E/p1678290959049379

### DIFF
--- a/libs/ui-toolkit/src/utils/intent.ts
+++ b/libs/ui-toolkit/src/utils/intent.ts
@@ -11,7 +11,8 @@ export const getIntentBorder = (intent = Intent.None) => {
     border: true,
     'border-danger': intent === Intent.Danger,
     'border-warning': intent === Intent.Warning,
-    'border-neutral-500': intent === Intent.None || intent === Intent.Primary,
+    'border-neutral-500': intent === Intent.None,
+    'border-vega-blue-300': intent === Intent.Primary,
     'border-success': intent === Intent.Success,
   };
 };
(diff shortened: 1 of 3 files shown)


libs/ui-toolkit/src/utils/intent.ts
```diff
@@ -20,9 +21,9 @@ export const getIntentBackground = (intent?: Intent) => {
   return {
     'bg-neutral-200 dark:bg-neutral-800': intent === undefined,
     'bg-black dark:bg-white': intent === Intent.None,
-    'bg-vega-pink dark:bg-vega-yellow': intent === Intent.Primary,
-    'bg-danger': intent === Intent.Danger,
-    'bg-warning': intent === Intent.Warning,
+    'bg-vega-blue-300 dark:bg-vega-blue-700': intent === Intent.Primary,
+    'bg-vega-pink': intent === Intent.Danger,
```
ciaran-, line 25: Anyone with any ideas why this change made a difference? I couldn't get a colour back from `bg-danger` even though I can see in the theme file it should be defined as the pink. Changing this line to `bg-vega-pink` made passing `Intent.Danger` to this function apply a colour properly. Pretty confused!

mattrussell36: I just tried making the same changes as you and couldn't recreate the problem, this has happened to me and it was because I had added the classname in a file which wasn't included in the tailwind.config.js. Could that be the case here?

ciaran-: I suppose that's possible, but I'm then confused as to why e.g. `bg-vega-pink` works in the same place! 🤔

mattrussell36 (approved):

libs/ui-toolkit/src/utils/intent.ts
ciaran-: I dropped this change out of the PR

PR #3235, opened 2023-03-20
chore(trading): 2764 link to mainnet incident reports

### LINKED ISSUE #2764: Link to mainnet incident reports from apps
## The Story

As a user 
I want to be able to find all incident reports that Vega have published
So that I can understand current issues, their status and get a feel for how Vega responds to them

## Tasks

- [ ] Add link in app (footer?) to 'Mainnet status & incidents' https://blog.vega.xyz/tagged/vega-incident-reports

(https://vegaprotocol.slack.com/archives/C02GZ0Q0NDC/p1673003925560669)

macqbat:
# Related issues 🔗

Closes #2764 

# Description ℹ️

Add link to the footer. Set "manually" width of the container with block height to avoid annoying jumps of its width

# Demo 📺

image
image

# Technical 👨‍🔧

### DIFF
--- a/apps/explorer/src/app/app.tsx
+++ b/apps/explorer/src/app/app.tsx
@@ -71,8 +71,8 @@ function App() {
           )}
         >
           <div>
-            <Header />
             <MainnetSimAd />
+            <Header />
           </div>
           <div>
             <Main />
--- a/apps/governance/src/components/page-templates/template-sidebar.tsx
+++ b/apps/governance/src/components/page-templates/template-sidebar.tsx
@@ -21,9 +21,9 @@ export function TemplateSidebar({ children, sidebar }: TemplateSidebarProps) {
     <>
       <AnnouncementBanner>
         <div className="font-alpha calt uppercase text-center text-lg text-white">
-          <span className="pr-4">Mainnet sim 3 is live!</span>
-          <ExternalLink href="https://fairground.wtf/">
-            Come help stress test the network
+          <span className="pr-4">Wait no longer, SIM3 is here!</span>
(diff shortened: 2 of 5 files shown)


apps/trading/components/footer/footer.tsx
```diff
@@ -27,19 +28,36 @@ export const NodeHealth = () => {
...
-      </FooterButtonPart>
-      <FooterButtonPart>
-        <NodeUrl url={VEGA_URL} />
-      </FooterButtonPart>
-      <FooterButtonPart>
-        <span title={t('Block height')}>{datanodeBlockHeight}</span>
-      </FooterButtonPart>
-    </FooterButton>
-  ) : null;
+    <>
```
mattrussell36, line 37: suggestion (non blocking): rather than rendering the link twice, in both true _and_ false cases of the ternary check, I think lift the fragment so it can be rendered once in a single place

```
return (
  <>
    {VEGA_URL && (
      <FooterButton>...</FooterButton>
    )}
    <ExternalLink className="ml-1" href={MAINNET_INCIDENT_LINK}>
      {t('Mainnet status & incidents')}
    </ExternalLink>
  </>
)
```

apps/trading/components/footer/footer.tsx
```diff
@@ -17,7 +17,8 @@ export const Footer = () => {
     </footer>
   );
 };
-
+const MAINNET_INCIDENT_LINK =
```
mattrussell36, line 20: chore (blocking): we'll probably want to move this to env var

macqbat: done as well

mattrussell36 (approved):

PR #3252, opened 2023-03-22
feat(announcements): announcements

### LINKED ISSUE #3244: Add generic announcement banner
Create a generic component to connect to the [announcements repo](https://github.com/vegaprotocol/announcements) and pulls in the correct announcement from the static data there, so that the apps can display a banner.

notbot00:
# Related issues 🔗

Closes #3244 

# Description ℹ️

Add generic announcement banner which fetches the data from the [announcements](https://github.com/vegaprotocol/announcements) repo and displays the relevant message + link for a dApp.

# Demo 📺

Screenshot 2023-03-22 at 11 05 15

### DIFF
--- a/.github/workflows/pr-trigger.yml
+++ b/.github/workflows/pr-trigger.yml
@@ -33,10 +33,10 @@ jobs:
           node-version: ${{ steps.node-version.outputs.npmVersion }}
 
       # Check SHAs
-      # - name: Derive appropriate SHAs for base and head for `nx affected` commands
-      #   uses: nrwl/nx-set-shas@v3
-      #   with:
-      #     main-branch-name: develop
+      - name: Derive appropriate SHAs for base and head for `nx affected` commands
+        uses: nrwl/nx-set-shas@v3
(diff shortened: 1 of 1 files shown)


libs/announcements/src/lib/hooks/use-announcement.ts
```diff
@@ -0,0 +1,67 @@
...
+        setState({
+          loading: false,
+          data: null,
+          error: `${err}`,
+        });
+      });
+  }, [name, url, setState]);
+
+  useEffect(() => {
+    fetchData();
```
mattrussell36, line 60: suggestion (non blocking): should probably add cleanup logic here to prevent setting state on unmounted components

mattrussell36 (approved):

libs/announcements/src/lib/hooks/use-announcement.ts
notbot00: good point, put a fix in there for this. thanks Matt!

PR #3261, opened 2023-03-23
feat(trading): persist state of chart intervals studies and overlays

### LINKED ISSUE #2761: Keep state of chart intervals, studies and overlays
## Story

As a trader 
I want to update the interval preferences on my candlestick chart
And console remember by preferences 
So that I can navigate to other chart components and complete my price analysis without extra steps.

### Acceptance Criteria

- [ ] I can change interval, chart type, studies and overlays and the selection is persisted in local storage
- [ ] The selection is NOT unique per market. IE if I select interval 1m in market A switching to market B will also have 1m interval selected

### Feedback
I have noticed something quite annoying on Console:

- Go to the price chart and change the interval from 15m to 1m
- Change to the depth chart
- Change back to the price chart
- Observe that the interval is 15m again, rather than 1m (I think it should be 1m still)

bglownia:
# Related issues 🔗

Closes #2761 

# Description ℹ️

Persist state of chart intervals studies and overlays

### DIFF
--- a/libs/candles-chart/src/lib/candles-chart.tsx
+++ b/libs/candles-chart/src/lib/candles-chart.tsx
@@ -39,6 +39,26 @@ const chartTypeIcon = new Map<ChartType, IconName>([
   [ChartType.OHLC, IconNames.WATERFALL_CHART],
 ]);
 
+const getValidItem = <T,>(
+  value: T | null | undefined,
+  set: T[],
+  defaultValue: T
+) =>
+  value !== null && value !== undefined && set.includes(value)
+    ? value
+    : defaultValue;
+
+const getValidSubset = <T,>(
+  value: T[] | null | undefined,
+  set: T[],
+  defaultValue: T[]
+) =>
+  value !== null && value !== undefined
+    ? value.filter((item) => set.includes(item))
+    : defaultValue;
+
+const SET_SEPARATOR = '\t';
+
 export type CandlesChartContainerProps = {
   marketId: string;
 };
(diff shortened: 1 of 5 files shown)


daro-maj (approved):

libs/candles-chart/src/lib/candles-chart.tsx
```diff
@@ -47,10 +70,48 @@ export const CandlesChartContainer = ({
   const { pubKey } = useVegaWallet();
   const { theme } = useThemeSwitcher();
 
-  const [interval, setInterval] = useState<Interval>(Interval.I15M);
-  const [chartType, setChartType] = useState<ChartType>(ChartType.CANDLE);
-  const [overlays, setOverlays] = useState<Overlay[]>([]);
-  const [studies, setStudies] = useState<Study[]>([Study.VOLUME]);
+  const [storedInterval, setInterval] = useLocalStorageSnapshot(
+    'console-candels-chart-interval'
```
mattrussell36, line 74: chore(blocking): typo `candels`

libs/candles-chart/src/lib/candles-chart.tsx
```diff
@@ -47,10 +70,48 @@ export const CandlesChartContainer = ({
   const { pubKey } = useVegaWallet();
   const { theme } = useThemeSwitcher();
 
-  const [interval, setInterval] = useState<Interval>(Interval.I15M);
-  const [chartType, setChartType] = useState<ChartType>(ChartType.CANDLE);
-  const [overlays, setOverlays] = useState<Overlay[]>([]);
-  const [studies, setStudies] = useState<Study[]>([Study.VOLUME]);
+  const [storedInterval, setInterval] = useLocalStorageSnapshot(
```
mattrussell36, line 73: suggestion (non blocking): I think its probably worth reaching for  a zustand store with persist middleware for this. It would tidy up the state management, use a single key (which would mean a single read from the store) and allow allow easier default state if nothing is stored

mattrussell36 (changes requested):

mattrussell36 (approved):

mattrussell36:
@bglownia Looks good. Still some failing tests though. It would also be good to add some unit tests for the store and possible one for the component to ensure the right values get passed down

PR #3267, opened 2023-03-23
feat(explorer,trading,governance): use nx to know which apps needs building

mkjmdski:
closes: https://github.com/vegaprotocol/devops-infra/issues/1761

### DIFF
--- a/.github/workflows/pr-trigger.yml
+++ b/.github/workflows/pr-trigger.yml
@@ -1,4 +1,4 @@
-name: Cypress tests - PR
+name: PR Validations
 
 on:
   push:
--- a/.github/workflows/pr-trigger.yml
+++ b/.github/workflows/pr-trigger.yml
@@ -12,7 +12,7 @@ on:
       - synchronize
       - ready_for_review
 jobs:
-  define-affected-apps:
+  pr:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout frontend mono repo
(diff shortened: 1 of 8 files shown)


radoslaw-szpiech (approved):

.github/workflows/pr-trigger.yml
```diff
@@ -20,14 +20,22 @@ jobs:
...
-      - name: Remove package.json & yarn.lock to avoid installing everything
-        run: rm package.json yarn.lock
-
-      - name: Install nx
-        run: yarn add nx
+      - name: Setup nx
+        run: |
+          rm package.json yarn.lock
+          yarn add nx
+          git restore package.json yarn.lock
```
edd, line 38: **thought (non-blocking)**: An alternative to this would be to use `npx`, which is like `npm` but executes it. Something like:

```bash
npx  --silent --yes nx affected...
```

Should be equivalent to this. I'm not sure if the package.json version of `nx` is significant or not (i.e. if we should make sure we are installing the same major version of `nx`), but `npx` would handled that with `npx nx@17` or a similar semver string

edd (approved):
I am not so good at reviewing workflows but I *think* this makes sense. I've suggested an alternative approach for the `nx` / package.json switching

.github/workflows/pr-trigger.yml
mkjmdski: nice one! added

mkjmdski: At the end I had issues with making this work, but I have used your suggestion to use pinned version of `nx` to make sure we use the same one

PR #3273, opened 2023-03-24
chore(trading): add "holding CMD + click" external link

### LINKED ISSUE #3188: CMD+Click to open a market in another tab
## The Chore

Make it so I can CMD+Click (or whatever the windows equivalent is) on a market name and have that market open in another browser tab, just like a normal link would. 

(Right click and open in new tab does appear to work). 

Multi monitors and watching many markets at once is such a common use case that we should support the way the browser normally works.

## Tasks

- [ ] Makes in the the positions, orders etc panels
- [x] Markets in the market switcher
- [x] Markets in the markets list

macqbat:
# Related issues 🔗

Closes #3188 

# Description ℹ️

Make conditionally external link: CMD/Control + click should open market in new tab/window 

# Demo 📺

image

# Technical 👨‍🔧


MadalinaRaicu (approved):

mattrussell36 (changes requested):
Sorry @macqbat I don't think we should merge this. IMO We shouldn't make libs depend on a specific router implementation (IE using react-router-dom directly in libs). This makes re-use harder and complicates testing. I don't see any problem with just passing props.

Can we just make the simple change for now (and ensure its tested) and check for the meta key in the callback?

If we really want to avoid passing props, I'd recommend passing the `navigate` function down from the app using a context, that way the app can provide a specific navigate function

macqbat:
> Sorry @macqbat I don't think we should merge this. IMO We shouldn't make libs depend on a specific router implementation (IE using react-router-dom directly in libs). This makes re-use harder and complicates testing. I don't see any problem with just passing props.
> 
> Can we just make the simple change for now (and ensure its tested) and check for the meta key in the callback?
> 
> If we really want to avoid passing props, I'd recommend passing the `navigate` function down from the app using a context, that way the app can provide a specific navigate function

OK, I've reverted the commit where I pushed this simplification, since I placed it totally in one commit it was quite easy. However  In my opinion it was quite useful solution, and If in the future it would turn out as problematic, we could easily reformat the hook only to be more "router solution independent". 
Now again we pass from top to bottom through almost the entire component tree of the trading and wallet pages two different, almost identical handlers :-(

libs/datagrid/src/lib/cells/market-name-cell.tsx
```diff
@@ -0,0 +1,40 @@
...
+      if (onMarketClick) {
+        onMarketClick(id, ev.metaKey);
+        return;
+      }
+    },
+    [id, onMarketClick]
+  );
+  if (!data) return null;
+  return (
+    <Link to={marketLink} onClick={handleOnClick}>
```
mattrussell36, line 36: chore (blocking): We should remove the dependency on react-router-dom here. I'd suggest just using a span but with appropriate attributes
```suggestion
    <span onClick={handleOnClick} role="link" tabIndex="0">
```

libs/datagrid/src/lib/cells/market-name-cell.tsx
```diff
@@ -0,0 +1,40 @@
...
+}
+
+export const MarketNameCell = ({
+  value,
+  data,
+  idPath,
+  onMarketClick,
+}: MarketNameCellProps) => {
+  const id = data ? get(data, idPath ?? 'id', 'all') : '';
+  const marketLink = `/${MARKET_PATH}/${id}`;
```
mattrussell36, line 22: chore (blocking): remove router implementation details. Lets just use the onMarketClick callback and not render a link with a url

mattrussell36 (changes requested):

mattrussell36:
Thanks for changing :-)

I don't understand the aversion to passing props (in most cases its 3/4 levels). It keeps things much simpler. 

I'd rather not have to wrap tests with MemoryRouter (or other contexts) for them to work and keep components more pure.

If you are concerned with the identical click handler logic perhaps make a hook `useMarketClickHandler` to share that logic

Issue #2983, opened 2023-02-23
Review loading/error states


edd:
## Description

These aren't currently handled uniformly:
- Going to a party that does not exist will show a title of 'Party' and nothing else
- Some things use the Loader component, others don't

# Pages to check
- [x] Transactions list
    - [x] Loading transactions list
    - [x] Transactions list error- __Bad - all rows replaced by errors__
    - [ ] Transaction by ID
        - [x] Loading transaction by id - __Should use spinner__
        - [x] Transaction by ID error (e.g. transaction doesn't exist)
- [x] Blocks list
    - [x] Loading block list
    - [x] Error block list - __infinite spinner if you scroll down and more fail to load in__
    - [x] Block by ID
        - [x] Loading block by ID - __Should use spinner__
        - [x] Block by ID error (e.g. block doesn't exist)
- [x] Oracles -  __Should use AsyncRenderer__
    - [x] Loading oracles
    - [x] Error oracles
    - [x] Oracle by ID
        - [x] Loading Oracle data __Should use spinner__
        - [x] Oracle by ID error (oracle does not exist)
- [x] Parties
    - [x] #3178
        - [x] Party by ID loading state - __Should use spinner__
        - [x] Party by ID error state - __Should have any sort of an error__
    - [x] Party transactions loading & error
- [x] Assets
    - [x] Loading assets
    - [x] Error loading assets
- [ ] Genesis parameters - __Should use async renderer__
    - [x] Loading genesis parameters
    - [ ] Error loading genesis
- [x] Network parameters - __Should use async renderer__
    - [x] Loading network parameters -
    - [x] Error loading network parameters - __Something went wrong: NetworkError when attempting to fetch resource.__
- [ ] Validators
    - [x] Loading validators
    - [ ] Error validators -  __Improve error__

PR #3274, opened 2023-03-24
fix(explorer): improve party page loading

### LINKED ISSUE #3178: Party by ID error handling
Party by ID page is slightly different from the rest of #2983 - it requests multiple different bits of data, but renders 'loading' or 'error' depending on only if the Party has accounts. Some users may submit transactions without any accounts - they won't be accepted, probably, but they might be vald enough that the block explorer should be showing them.

# Tasks
- [x] Switch party by ID to use `AsyncRenderer`
- [ ] Make each subsection load even if the other doesn't:
   - [ ] Account balances should load even if there are no TXs
   - [ ] TXs should load and display even if there are no accounts
   - [ ] Staking ditto, independent of the above

edd:
- feat(explorer): make oracles and txs and blocks and genesis use asyncrender
- fix(explorer): loading states were incorrect
- feat(explorer): wip party page
- fix(explorer): fix incorrect merge result
- fix(explorer): expand party query
- feat(explorer): make separate accounts route
- fix(explorer): unrelated fix to improve rendering of tx list

# Related issues 🔗

Closes #3178. Partially addresses #2983

# Description ℹ️
There are two parts to this chunky ticket:

1. Making most views (see #2983 for the full list) to use AsyncRenderer, which improves both the loading an error handling
2. Re-layout the Party page so that the shape of the page doesn't change drastically depending on how many accounts/txs the user has.
    1. Add a new route for accounts-by-asset for an address 

## Old vs New
| **old** | **new** |
|-------------|---------|
|  Screenshot 2023-05-05 at 12 29 53 |  Screenshot 2023-05-05 at 12 31 47  |

It adds an extra click to see a parties full account balance, but this (very roughly) follows the same pattern as Etherscan's view for an address. What Etherscan does better than I could manage in this PR is let you check the 'top level' balance for a user's various tokens/assets within the page, without navigating away. I chose to make it an extra click for now - if we can come up with a nice filter view, I'd be happy to add it.


apps/explorer/src/app/routes/router-config.tsx
```diff
@@ -75,14 +75,43 @@ const partiesRoutes: Route[] = flags.parties
...
-            handle: {
-              breadcrumb: (params: Params<string>) => (
-                <Link to={linkTo(Routes.PARTIES, params.party)}>
-                  {truncateMiddle(params.party as string)}
-                </Link>
-              ),
-            },
+            element: <Party />,
+
+            children: [
```
edd, line 80: This feels really clunky, so I think I have done this incorrectly - if anyone has suggestions for how to improve it, I'm happy to hear them!

MadalinaRaicu (approved):

PR #3311, opened 2023-03-29
feat(trading): add post only and reduce only orders

### LINKED ISSUE #3238: Add post only and reduce only order options to deal ticket
## Story

As a trader
I want to be able to place post only and reduce only orders (aka execution flags or order conditions)
So that can minimize risk and maximise profit

## Acceptance Criteria

- [ ] I can place a post only order
- [ ] I can place a reduce only order

more detailed ACs here: https://github.com/vegaprotocol/specs/pull/1640

## Tasks

- [x] Explore and sketch
- [ ] Visual Design
- [ ] Team review
- [ ] Dev
- [ ] Test

### Changes to deal ticket

- Add two check boxes to the deal ticket: one for post only one for reduce only
- Add logic so that only one of these can be selected at the same time. (Radio buttons were considered instead but Checkboxes are more appropriate)
- Possibly a later ticket: add tool tips on hover of each with an explanation of what they mean

Suggested draft content

> Selecting Post only will ensure the order is not filled immediately but is placed on the book as a passive order. When the order is processed it is either be stopped (if it would be filled immediately) Or placed in the book as a passive order until and price taker matches with it

> Selecting Reduce only will ensure that this order will not increase the size of an open position. When the order is matched it will only trade enough volume to bring your open volume towards 0 but never change the direction of your position. If applied to a limit order that is not instantly filled the order will be stopped.

- Link to documentation from these tool tips: https://github.com/vegaprotocol/documentation/pull/544

- When the TIF is FOC or IOC the Post only check box should be unchecked and disabled 
-The tool tip when hovering over the disabled check box should tell you that this can not be used for the selected TIF

> "Post only" can not be used on "Fill or Kill" or "Immediate-or-canceled" orders

- The selected state of the Execution flags should be remembered as part of the general deal ticket state (size value etc). This includes the change to state that happen when the box is unchecked automatically.

### Changes to Orders table

Image

- Add any selected execution flag to the TIF column. Simply append the flag to the end of the time in force text. e.g. "Good til' close. Reduce only" or "Good til` time" if no execution flag is applied.
- ensure that the "[reason](https://docs.vega.xyz/mainnet/api/graphql/enums/order-rejection-reason)" is being displayed in the status if there is one. I know this is the case for rejected orders but not sure if it has been implemented for Stopped orders too. background:  "...it is instead immediately Stopped with a reason informing the trader that the order was stopped to avoid a trade occurring" [source](https://github.com/vegaprotocol/specs/blob/master/protocol/0014-ORDT-order_types.md#execution-flags).

### Changes to toast

none

### 

Specs PR: https://github.com/vegaprotocol/specs/pull/1625
Related core issue: https://github.com/vegaprotocol/vega/issues/7851
Related core pull request: https://github.com/vegaprotocol/vega/pull/7852

MadalinaRaicu:
# Related issues 🔗

Closes #3238 

# Description ℹ️

Add `Post only` and `Reduce only` options when submitting order.

# Demo 📺
Screenshot 2023-03-29 at 15 43 14

# Technical 👨‍🔧

### DIFF
--- a/apps/trading-e2e/src/integration/trading-deal-ticket.cy.ts
+++ b/apps/trading-e2e/src/integration/trading-deal-ticket.cy.ts
@@ -84,6 +84,8 @@ describe('must submit order', { tags: '@smoke' }, () => {
       type: Schema.OrderType.TYPE_MARKET,
       side: Schema.Side.SIDE_BUY,
       timeInForce: Schema.OrderTimeInForce.TIME_IN_FORCE_FOK,
+      postOnly: false,
+      reduceOnly: false,
       size: '100',
     };
     createOrder(order);
@@ -98,6 +100,8 @@ describe('must submit order', { tags: '@smoke' }, () => {
       side: Schema.Side.SIDE_SELL,
       timeInForce: Schema.OrderTimeInForce.TIME_IN_FORCE_IOC,
       size: '100',
+      postOnly: false,
+      reduceOnly: false,
     };
     createOrder(order);
     testOrderSubmission(order);
@@ -112,6 +116,8 @@ describe('must submit order', { tags: '@smoke' }, () => {
       side: Schema.Side.SIDE_BUY,
       timeInForce: Schema.OrderTimeInForce.TIME_IN_FORCE_GTC,
       size: '100',
+      postOnly: false,
+      reduceOnly: false,
       price: '200',
     };
     createOrder(order);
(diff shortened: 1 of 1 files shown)


apps/trading/lib/hooks/use-vega-transaction-toasts.tsx
```diff
@@ -474,15 +474,17 @@ const VegaTxCompleteToastsContent = ({ tx }: VegaTxToastContentProps) => {
...
     return (
       <>
         <ToastHeading>{t('Order rejected')}</ToastHeading>
-        <p>
-          {t(
-            'Your order has been rejected because: %s',
-            getRejectionReason(tx.order) || ''
-          )}
-        </p>
+        {rejectionReason ? (
```
MadalinaRaicu, line 481: Fallback. But we need this ticket merged https://github.com/vegaprotocol/vega/pull/8046 @mattrussell36  and @candida-d

MadalinaRaicu: And after that generate again types on our end.

macqbat (approved):

PR #3318, opened 2023-03-30
feat(deal-ticket): show total margin available for an order, and where it will come from

### LINKED ISSUE #3215: Show total margin available for an order, and where it will come from
## Story

As a user placing an order
I want to know how much collateral I can use, and which account(s) it will come from
So I understand how my assets will be used

## Acceptance Criteria

- [ ] I can see the maximum amount of margin I could commit to this order
- [ ] I can find an explanation of how that maximum is calculated
- [ ] I can see where that margin will come from (general account, market margin account, or both)

## Tasks

- [ ] Write tooltips for "deduction from collateral" and "deduction from margin" fields
- [ ] Double check if you can use margin down to the maintenance level, or only down to search level. If it's the latter, update the tooltip for the "Total margin available" field to say so.
- [ ] Implement "Total margin available" field in deal ticket
- [ ] Implement "Deduction from collateral" field in deal ticket
- [ ] ~~Implement "Deduction from margin" field in deal ticket~~ (Barney, Bart, and Neil agreed to remove this on 30th March 2023, see comments for Slack link)
- [ ] Dev and design QA

## Sketch

These fields should be shown underneath the "Margin required" field currently in the deal ticket.

It would be useful if the total margin available, deduction from collateral, and deduction from margin fields were indented a little bit, to show they are all related. However this is not necessary. The ticket https://github.com/vegaprotocol/frontend-monorepo/issues/2834 will add an expanding/collapsing area to collect this extra detail.

image

Tooltip for "Total margin available" field:
> Total margin available = general account balance (166 tDAI) + margin balance (32 tDAI) - maintenance level (10.00 tDAI).

### Calculating available margin
Redrawn from a sketch by Barney. The grey area might actually start at maintenance level, not search level, we need to check this.

Image

- This task was split out from https://github.com/vegaprotocol/frontend-monorepo/issues/2834

bglownia:
# Related issues 🔗

Closes #3215

# Description

image

Values calculated as follows:

Margin required = estimatedTotalInitialMargin - currentInitialMargin

Total margin available = generalAccountBalance + marginAccountBalance - currentMaintenanceMarginLevel

Deduction from collateral = estimatedTotalInitialMargin - marginAccountBalance
Deduction value is displayed as 0 if it is  lower than 0

### DIFF
--- a/libs/deal-ticket/src/hooks/use-fee-deal-ticket-details.tsx
+++ b/libs/deal-ticket/src/hooks/use-fee-deal-ticket-details.tsx
@@ -15,6 +15,8 @@ import {
   NOTIONAL_SIZE_TOOLTIP_TEXT,
   MARGIN_ACCOUNT_TOOLTIP_TEXT,
   MARGIN_DIFF_TOOLTIP_TEXT,
+  DEDUCTION_FROM_COLLATERAL_TOOLTIP_TEXT,
+  TOTAL_MARGIN_AVAILABLE,
 } from '../constants';
 import { useOrderCloseOut } from './use-order-closeout';
 import { useMarketAccountBalance } from '@vegaprotocol/accounts';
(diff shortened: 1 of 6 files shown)


libs/deal-ticket/src/hooks/use-fee-deal-ticket-details.tsx
```diff
@@ -85,24 +87,33 @@ export const useFeeDealTicketDetails = (
...
-  balance,
+  marginAccountBalance,
+  generalAccountBalance,
   assetSymbol,
   estimateOrder,
   market,
   notionalSize,
-  totalMargin,
+  estimatedTotalInitialMargin,
+  // estimatedInitialMargin,
```
macqbat, line 111: If it should stay like that (commented) it needs a short explanation why

macqbat (approved):

PR #3324, opened 2023-03-30
chore(web3): wallet connect v2

### LINKED ISSUE #2904: WalletConnect v1.0 migration
## The Chore

WalletConnect v1.0 is being deprecated

## Tasks

- [ ] Figure out what we need to do to migrate to v2.0 of WalletConnect
- [ ] Dev
- [ ] Test

https://medium.com/walletconnect/walletconnect-v1-0-sunset-notice-and-migration-schedule-8af9d3720d2e

asiaznik:
# Related issues 🔗

Closes #2904


libs/web3/src/constansts.ts
mattrussell36, line 1: chore(blocking): typo in filename

mattrussell36 (commented):
I couldn't get this working with TrustWallet and MetaMask mobile. Also tried with a different projectId. Any ideas?

asiaznik:
> I couldn't get this working with TrustWallet and MetaMask mobile. Also tried with a different projectId. Any ideas?

I couldn't make it working with MetaMask mobile, but I'm not sure they're already on v2 - Tested it with the test wallet https://react-web3wallet.vercel.app/walletconnect and works fine

mattrussell36:
> > I couldn't get this working with TrustWallet and MetaMask mobile. Also tried with a different projectId. Any ideas?
> 
> I couldn't make it working with MetaMask mobile, but I'm not sure they're already on v2 - Tested it with the test wallet https://react-web3wallet.vercel.app/walletconnect and it seems to work fine

Perhaps we need to not replace v1 with v2  but provide both options? Not sure the status of Metamask and when they might change to v2, we'll need to look into it

asiaznik:
> > > I couldn't get this working with TrustWallet and MetaMask mobile. Also tried with a different projectId. Any ideas?
> > 
> > 
> > I couldn't make it working with MetaMask mobile, but I'm not sure they're already on v2 - Tested it with the test wallet https://react-web3wallet.vercel.app/walletconnect and it seems to work fine
> 
> Perhaps we need to not replace v1 with v2 but provide both options? Not sure the status of Metamask and when they might change to v2, we'll need to look into it

I've added v1 as Legacy, also while I was at it I've added the Coinbase connector and icons to make it look better.

How much attached are you to the web3-react? I've found that wagmi does the same and is pretty well documented (https://wagmi.sh/).

jxom:
Hey! wagmi author here!

Would be great if y’all considered integrating wagmi for WalletConnect v2. There are a few nuances with WalletConnect v2 UX which the wagmi connector handles (mainly around ensuring that chains are authorized - which I don’t believe web3-react handles). Happy to answer any questions.

apps/governance/src/lib/web3-connectors.ts
```diff
@@ -20,23 +20,27 @@ export const createDefaultProvider = (providerUrl: string, chainId: number) => {
...
-          },
-        }),
-      [chainId]
-    );
+  const [coinbase, coinbaseHooks] = initializeCoinbaseConnector(providerUrl);
+  const [walletconnect, walletconnectHooks] = initializeWalletConnector(
+    chainId,
+    providerUrl
+  );
+  const [legacy, legacyHooks] = initializeWalletConnectLegacyConnector(
```
mattrussell36, line 34: suggestion (non blocking): I think a more explicity name here would be helpful perhaps `walletconnectLegacy` or `walletconnectV1`

libs/web3/src/constants.ts
```diff
@@ -0,0 +1,5 @@
+/**
+ * WalletConnect project id
+ * https://cloud.walletconnect.com/sign-in
+ */
+export const PROJECT_ID = 'fe8091dc35738863e509fc4947525c72';
```
mattrussell36, line 5: chore (blocking): env var?

libs/web3/src/lib/use-eager-connect.ts
```diff
@@ -18,10 +18,18 @@ export const useEagerConnect = () => {
...
-      stored[0].connectEagerly();
-    }
+    const tryConnectEagerly = async () => {
+      // found a valid connection option
+      if (stored && stored[0].connectEagerly) {
+        try {
+          await stored[0].connectEagerly();
+        } catch (err) {
+          // NOOP - no active session
+          console.log('ERR_WEB3_EAGER_CONNECT', (err as Error).message);
```
mattrussell36, line 28: suggestion (non blocking): use logger from utils?

libs/web3/src/lib/web3-provider.tsx
```diff
@@ -1,14 +1,23 @@
...
 import type { Connector } from '@web3-react/types';
+import { useMemo } from 'react';
 
 interface Web3ProviderProps {
   children: JSX.Element | JSX.Element[];
   connectors: [Connector, Web3ReactHooks][];
 }
 
 export const Web3Provider = ({ children, connectors }: Web3ProviderProps) => {
+  const key = useMemo(
```
mattrussell36, line 12: suggestion (non blocking): can you add a comment as to why we need this key? Its unclear why this key is required

mattrussell36 (changes requested):

mattrussell36:
@asiaznik coinbase wallet is working ok, but you kind of get stuck on the select wallet type dialog when using the coinbase app (as opposed to coinbase wallet), I think this is because you can't connect to test networks, but the experience isn't great. Happy for it to be some follow up work

mattrussell36 (approved):

PR #3328, opened 2023-03-30
fix(smart-contracts): add mainnet gas estimation

### LINKED ISSUE #3325: Incorrect estimation of gas fees on a small number of transactions
At 04:01 on the 29th March 2023 a user reported being unable to stake all of their Vega tokens (subsequent staking of a smaller portion of their tokens was successful). The transaction was approved on the Ethereum chain however was not successful due to an incorrect gas fee estimation, resulting in a loss of gas fees.

* The non-library code for the Governance dApp has been checked for any adjustment to gas fee estimation. We can confirm that estimations are not adjusted.

* We are currently investigating three potential sources for the problem:

1. A problem with the version of the Ethers library ~~we are currently using~~
~~2. An issue with Metamask (see https://github.com/MetaMask/metamask-extension/issues/18297)~~
~~3. It seems that the failures this user had coincide with a new release of geth. The gas estimate was off by just 12 gas (0.02%) so this may be a latent cause.~~

**Update 31st March**: we discovered the issue to be a problem with the Ethers library. This is a known issue: https://github.com/ethers-io/ethers.js/issues/3627. To rectify this, we have introduced a gas fee overestimation 20% higher than that provided directly by Ethers. This fix has now been deployed.

mattrussell36:
# Related issues 🔗

Closes #3325 

# Description ℹ️

Increase gas limit by 20% on the original gas estimation

### DIFF
--- a/apps/token/README.md
+++ b/apps/token/README.md
@@ -2,7 +2,7 @@
 
 **_Control panel for your VEGA tokens_**
 
-<img width="1438" alt="Screenshot 2021-12-11 at 06 32 51" src="https://user-images.githubusercontent.com/13255539/145666935-563fc1ff-35bc-4cd9-ae6d-cf711cc23454.png">
+<img width="1437" alt="Screenshot 2021-12-11 at 06 32 51" src="https://user-images.githubusercontent.com/13255539/145666935-563fc1ff-35bc-4cd9-ae6d-cf711cc23454.png">
 
 ## Features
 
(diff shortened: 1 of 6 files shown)


C0deMunk33 (approved):
This 20% buffer is the appropriate solution to combat a faulty estimation coming from the node or metamask. This is the best way to mitigate that failure. If 20% isn't enough most the dapps out there will have gone off the rails. This is also how uniswap does it.

sam-keen (approved):

dexturr (approved):

PR #3347, opened 2023-04-03
fix(trading): order stopped intent toasts

### LINKED ISSUE #3238: Add post only and reduce only order options to deal ticket

MadalinaRaicu:
# Related issues 🔗

Closes #3238 

# Description ℹ️

- [x] Fix order intent and rejection error
- [x] Add order list label in TIF

# Demo 📺

image
image

# Technical 👨‍🔧


jeremyletang (approved):

libs/deal-ticket/src/components/deal-ticket/deal-ticket.tsx
```diff
@@ -179,6 +185,8 @@ export const DealTicket = ({
   // if an order doesn't exist one will be created by the store immediately
   if (!order || !normalizedOrder) return null;
 
+  console.log('errors', errors);
```
bglownia, line 188: console.log

libs/accounts/src/lib/transfer-form.tsx
```diff
@@ -232,25 +232,43 @@ export const TransferFee = ({
...
+        </div>
       </div>
-      <div
-        data-testid="transfer-fee"
-        className="text-neutral-500 dark:text-neutral-300"
-      >
-        {value}
+      <div className="flex justify-between gap-4 items-center flex-wrap">
+        <Tooltip
+          description={t(
```
bglownia: Do not translate variables
`t("The total amount plus the transfer fee, which is set by the network parameter %s", feeFactor)`

libs/deal-ticket/src/components/deal-ticket/deal-ticket.tsx
MadalinaRaicu: thanks will remove now testing some undefined rejection reasons

libs/orders/src/lib/components/order-list/order-list.tsx
```diff
@@ -218,7 +218,14 @@ export const OrderListTable = memo(
                 return `${Schema.OrderTimeInForceMapping[value]}: ${expiry}`;
               }
 
-              return value ? Schema.OrderTimeInForceMapping[value] : '';
+              const tifLabel = value
+                ? Schema.OrderTimeInForceMapping[value]
+                : '';
+              const label = `${tifLabel}${data?.postOnly ? '. Post Only' : ''}${
+                data?.reduceOnly ? '. Reduce only' : ''
```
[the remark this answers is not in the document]
MadalinaRaicu: true, updating it now

libs/accounts/src/lib/transfer-form.tsx
MadalinaRaicu: removed this change

libs/orders/src/lib/components/order-list/order-list.tsx
```diff
@@ -218,7 +219,14 @@ export const OrderListTable = memo(
                 return `${Schema.OrderTimeInForceMapping[value]}: ${expiry}`;
               }
 
-              return value ? Schema.OrderTimeInForceMapping[value] : '';
+              const tifLabel = value
+                ? Schema.OrderTimeInForceMapping[value]
```
mattrussell36, line 223: chore (blocking): This looks incorrect. This is the status column we should be using the status mapping

mattrussell36: oh nvm my mistake

libs/deal-ticket/src/components/deal-ticket/deal-ticket.tsx
```diff
@@ -158,6 +158,12 @@ export const DealTicket = ({
     return disabled;
   }, [order]);
 
+  useEffect(() => {
```
mattrussell36, line 161: chore(blocking): we remove this effect and instead move this logic to the TIF callback function.

MadalinaRaicu: I did this initially but it's not enough. 
Whenever it gets disabled we need to make sure postOnly gets false

MadalinaRaicu: updated again reduceOnly

mattrussell36 (approved):

PR #3348, opened 2023-04-03
chore(deposits): persistence of deposit form state

### LINKED ISSUE #3211: Preserve state of deposit form

macqbat:
# Related issues 🔗

Closes #3211 

# Description ℹ️

Save state of deposit form in local storage

# Demo 📺

-

# Technical 👨‍🔧

-


libs/deposits/src/lib/use-persistent-deposit.ts
```diff
@@ -0,0 +1,32 @@
...
+export const usePersistentDeposit = (
+  assetId?: string
+): [PersistedDeposit, (entry: PersistedDeposit) => void] => {
+  const [value, saveValue] = useLocalStorageSnapshot(STORAGE_KEY);
+  const parsedValues = useMemo(
+    () => JSON.parse(value || '{}') as PersistedDepositData,
+    [value]
+  );
+  const savedData = useMemo(() => {
+    const index = assetId || Object.keys({ ...parsedValues }).pop() || '';
```
mattrussell36, line 20: suggestion (non blocking): This part seems a little odd to me. Can you add a comment or alternatively it may be better to let the hook return undefined if there isn't a matching deposit for the asset found.

mattrussell36 (approved):

libs/deposits/src/lib/use-persistent-deposit.ts
```diff
@@ -0,0 +1,32 @@
+import { useLocalStorageSnapshot } from '@vegaprotocol/react-helpers';
+import { useCallback, useMemo } from 'react';
+
+const STORAGE_KEY = 'DEPOSIT_FORM_STORAGE';
```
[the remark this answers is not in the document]
macqbat: done

libs/deposits/src/lib/use-persistent-deposit.ts
macqbat: now it looks little different, but basically it just returns last added value in case when no key (assetId) is given

PR #3363, opened 2023-04-04
feat(governance): protocol upgrade proposals list and details

### LINKED ISSUE #3257: Highlight network upgrade proposals
## Story

As a governance dapp user
I want to know when a validator has proposed a network upgrade  
So that I have visibility of what's happening with the network, can assess how it might effect me and decide whether I need to take any action

## Acceptance Criteria

- [ ] I can understand that network update proposals are made by validators (and not Vega) 
- [ ] Network upgrade proposals are highlighted on the governance dapp landing page
- [ ] Network upgrade proposals are highlighted on the governance dapp landing page 

## Tasks

- [ ] Explore and sketch
- [ ] Visual Design
- [ ] Team review
- [ ] Etc.

sam-keen:
# Related issues 🔗

Closes #3257

# Description ℹ️

Adds support for protocol upgrade proposals, as list items on the homepage and proposals list pages, and a basic proposal details page. 

# Demo 📺

Screenshot 2023-04-04 at 16 11 21
Screenshot 2023-04-04 at 15 59 32
Screenshot 2023-04-04 at 14 43 27
Screenshot 2023-04-11 at 17 10 55


apps/governance/src/i18n/translations/dev.json
```diff
@@ -159,6 +159,7 @@
   "Tokens are held in different <trancheLink>Tranches</trancheLink>. Each tranche has its own schedule for how the tokens are unlocked.": "Tokens are held in different <trancheLink>Tranches</trancheLink>. Each tranche has its own schedule for how the tokens are unlocked.",
   "proposals": "Proposals",
   "proposal": "Proposal",
+  "PROPOSAL": "PROPOSAL",
```
notbot00, line 162: question (non-blocking): seems weird at first to see the same piece of text with different casing in the translations dictionary. if this is purely for styling, i'd probably expect this to be done via a tailwind `uppercase` class where needed, or using `toUpperCase` if that's not possible. not sure

apps/governance/src/i18n/translations/dev.json
```diff
@@ -168,6 +169,7 @@
   "enactedOn": "Enacted on",
   "status": "Status",
   "state": "State",
+  "STATE": "STATE",
```
notbot00, line 172: ^ same as above for the `"PROPOSAL"`

apps/governance/src/i18n/translations/dev.json
```diff
@@ -767,13 +770,28 @@
...
   "homeVegaTokenButtonText": "Manage tokens",
-  "downloadProposalJson": "Download proposal as JSON"
+  "downloadProposalJson": "Download proposal as JSON",
+  "networkUpgrade": "Network Upgrade",
+  "PROTOCOL_UPGRADE_PROPOSAL_STATUS_APPROVED": "Approved",
+  "PROTOCOL_UPGRADE_PROPOSAL_STATUS_PENDING": "Pending",
+  "PROTOCOL_UPGRADE_PROPOSAL_STATUS_REJECTED": "Rejected",
+  "PROTOCOL_UPGRADE_PROPOSAL_STATUS_UNSPECIFIED": "Unspecified",
+  "vegaRelease{release}": "Vega Release {{release}}",
+  "UPGRADE BLOCK HEIGHT": "UPGRADE BLOCK HEIGHT",
```
notbot00, line 788: question (non-blocking)L i guess the same for all of these, too, all the hardcoded uppercase-only strings above and below

apps/governance/src/routes/home/index.tsx
```diff
@@ -136,6 +139,35 @@ const HomeNodes = ({
   );
 };
 
+interface SimpleValidatorRendererProps {
+  id: string;
+  avatarUrl: string | null | undefined;
+  name: string;
+}
+
+export const SimpleValidatorRenderer = ({
```
notbot00, line 148: nit (non-blocking): the naming feels weird here, overly verbose without actually telling much. including `simple` in the name usually makes sense when there are multiple variations, which is not the case here, if i'm interpreting this correctly. renderer is basically every react component, because that's just what they do :) I might consider something like `ValidatorItem`, `ValidatorLink`, etc, not sure if it makes any sense.
also not very important, feel free to dismiss please

apps/governance/src/routes/home/index.tsx
```diff
@@ -163,11 +205,38 @@ const GovernanceHome = ({ name }: RouteChildProps) => {
...
-          ).slice(0, 3)
+          )
         : [],
     [proposalsData]
   );
 
+  const protocolUpgradeProposals = useMemo(
+    () =>
+      protocolUpgradesData
+        ? getNotRejectedProtocolUpgradeProposals<ProtocolUpgradeProposalFieldsFragment>(
```
notbot00, line 216: suggestion (non-blocking): i'd consider either using a different function here, where you don't need to double-filter it, or adding filter criteria (as a callback maybe?) to the function, so it can support your use-case here, where you want non-rejected proposals within a block height range.

notbot00 (approved):

apps/governance/src/i18n/translations/dev.json
sam-keen: Yes, that's a fair point. Thanks

apps/governance/src/routes/home/index.tsx
sam-keen: Thanks - there is a different validator renderer for the validator's table, but it's a poor naming choice. Why is naming so hard!? 😭

guoguojin (approved):
GraphQL query looks fine to me.

PR #3409, opened 2023-04-10
fix(trading): fix formatting on clicking on orderbook cell after decimal resolution update

### LINKED ISSUE #3408: Clicking on orderbook cell after decimal resolution update does not get formatted correctly
## Bug

Image

MadalinaRaicu:
# Related issues 🔗

Closes #3408 

# Description ℹ️

Fix formatting when user clicks on orderbook cell after selecting decimal resolution.

# Demo 📺

image

Issue being fixed:

image

# Technical 👨‍🔧

- [x] The formatting should happen at the level of the orderbook row. 
- [x] In the orderbook manager we should just update the price value in the store.

### DIFF
--- a/libs/market-depth/src/lib/orderbook.spec.tsx
+++ b/libs/market-depth/src/lib/orderbook.spec.tsx
@@ -20,7 +20,7 @@ describe('Orderbook', () => {
   const decimalPlaces = 3;
   it('should scroll to mid price on init', async () => {
     window.innerHeight = 11 * rowHeight;
-    const result = render(
+    render(
       <Orderbook
         decimalPlaces={decimalPlaces}
         positionDecimalPlaces={0}
@@ -30,7 +30,7 @@ describe('Orderbook', () => {
       />
     );
     await waitFor(() => screen.getByTestId(`bid-vol-${params.midPrice}`));
-    expect(result.getByTestId('scroll').scrollTop).toBe(91 * rowHeight);
+    expect(screen.getByTestId('scroll').scrollTop).toBe(91 * rowHeight);
   });
 
   it('should keep mid price row in the middle', async () => {
@@ -45,7 +45,7 @@ describe('Orderbook', () => {
       />
     );
     await waitFor(() => screen.getByTestId(`bid-vol-${params.midPrice}`));
-    expect(result.getByTestId('scroll').scrollTop).toBe(91 * rowHeight);
+    expect(screen.getByTestId('scroll').scrollTop).toBe(91 * rowHeight);
     result.rerender(
       <Orderbook
         decimalPlaces={decimalPlaces}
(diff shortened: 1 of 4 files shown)


libs/market-depth/src/lib/orderbook.spec.tsx
```diff
@@ -183,4 +183,53 @@ describe('Orderbook', () => {
     );
     expect(result.getByTestId('scroll').scrollTop).toBe(6 * rowHeight);
   });
+
+  it('should format correctly the numbers on resolution change', async () => {
+    const onClickSpy = jest.fn();
+    window.innerHeight = 11 * rowHeight;
+    const result = render(
```
mattrussell36, line 190: suggestion (non blocking): prefer not using the return value of `render` and use the direct import `screen` instead

libs/market-depth/src/lib/orderbook.spec.tsx
```diff
@@ -183,4 +183,53 @@ describe('Orderbook', () => {
...
+      <Orderbook
+        decimalPlaces={decimalPlaces}
+        positionDecimalPlaces={0}
+        onClick={onClickSpy}
+        fillGaps
+        {...generateMockData(params)}
+        onResolutionChange={onResolutionChange}
+      />
+    );
+    await waitFor(() => screen.getByTestId(`bid-vol-${params.midPrice}`));
```
mattrussell36, line 200: suggestion (non blocking): this would be better as 
```
expect(await screen.findByTestId(`bid-vol-${params.midPrice}`)).toBeInTheDocument()
```

libs/market-depth/src/lib/orderbook.spec.tsx
```diff
@@ -183,4 +183,53 @@ describe('Orderbook', () => {
...
+    const scrollElement = result.getByTestId('scroll');
+
+    fireEvent.click(result.getByTestId('price-122934'));
+    expect(onClickSpy).toBeCalledWith('122.934');
+
+    expect(scrollElement.scrollTop).toBe(91 * rowHeight);
```
mattrussell36, line 206: question (non blocking): Do we need all the test code relating to scroll position? They seem like they are unrelated to this test

mattrussell36 (approved):

libs/market-depth/src/lib/orderbook.spec.tsx
```diff
@@ -183,4 +183,53 @@ describe('Orderbook', () => {
...
+    expect(result.getByTestId('scroll').scrollTop).toBe(1);
+    const resolutionSelect = result.getByTestId(
+      'resolution'
+    ) as HTMLSelectElement;
+
+    fireEvent.change(resolutionSelect, { target: { value: '10' } });
+    expect(onResolutionChange.mock.calls.length).toBe(1);
+    expect(onResolutionChange.mock.calls[0][0]).toBe(10);
+    fireEvent.click(result.getByTestId('price-122994'));
+    expect(onClickSpy).toBeCalledWith('122.994');
```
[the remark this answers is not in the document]
MadalinaRaicu: updated

libs/market-depth/src/lib/orderbook.spec.tsx
MadalinaRaicu: removed

libs/market-depth/src/lib/orderbook.spec.tsx
MadalinaRaicu: updated

libs/market-depth/src/lib/orderbook.spec.tsx
MadalinaRaicu: updated

PR #3411, opened 2023-04-10
chore(trading): prevent node poll if query fails

mattrussell36:
# Related issues 🔗

N/A

# Description ℹ️

Prevents polling in the node switcher if a stats query fails for a given node. This should help with the number of Sentry alerts

### DIFF
--- a/libs/environment/src/components/node-switcher/row-data.spec.tsx
+++ b/libs/environment/src/components/node-switcher/row-data.spec.tsx
@@ -1,6 +1,6 @@
 import type { MockedResponse } from '@apollo/react-testing';
 import { MockedProvider } from '@apollo/react-testing';
-import { render, screen, waitFor } from '@testing-library/react';
+import { act, render, screen, waitFor } from '@testing-library/react';
 import { RadioGroup } from '@vegaprotocol/ui-toolkit';
 import type {
   BlockTimeSubscription,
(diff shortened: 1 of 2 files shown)


libs/environment/src/components/node-switcher/row-data.spec.tsx
```diff
@@ -9,6 +9,7 @@ import type {
 import { BlockTimeDocument } from '../../utils/__generated__/Node';
 import { StatisticsDocument } from '../../utils/__generated__/Node';
 import type { RowDataProps } from './row-data';
+import { POLL_INTERVAL } from './row-data';
```
MadalinaRaicu, line 12:
```suggestion
```

libs/environment/src/components/node-switcher/row-data.spec.tsx
```diff
@@ -25,7 +26,7 @@ const statsQueryMock: MockedResponse<StatisticsQuery> = {
   result: {
     data: {
       statistics: {
-        blockHeight: '1234',
+        blockHeight: '1234', // the actual value used in the component is the from the header store
```
MadalinaRaicu, line 29:
```suggestion
        blockHeight: '1234', // the actual value used in the component is the value from the header store
```

PR #3424, opened 2023-04-12
fix(orders): update order toast intent and title 

MadalinaRaicu:
# Related issues 🔗

Closes N/A

# Description ℹ️

Update order toast intent and title.

# Demo 📺

Bug observed:
image

# Technical 👨‍🔧

### DIFF
--- a/libs/orders/src/lib/utils.ts
+++ b/libs/orders/src/lib/utils.ts
@@ -82,10 +82,10 @@ export const getOrderToastIntent = (
       return Intent.Warning;
     case Schema.OrderStatus.STATUS_REJECTED:
     case Schema.OrderStatus.STATUS_STOPPED:
-    case Schema.OrderStatus.STATUS_CANCELLED:
       return Intent.Danger;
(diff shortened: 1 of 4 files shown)


libs/orders/src/lib/utils.ts
```diff
@@ -82,10 +82,10 @@ export const getOrderToastIntent = (
       return Intent.Warning;
     case Schema.OrderStatus.STATUS_REJECTED:
     case Schema.OrderStatus.STATUS_STOPPED:
-    case Schema.OrderStatus.STATUS_CANCELLED:
       return Intent.Danger;
     case Schema.OrderStatus.STATUS_FILLED:
     case Schema.OrderStatus.STATUS_ACTIVE:
+    case Schema.OrderStatus.STATUS_CANCELLED:
```
MadalinaRaicu, line 88: When the order is cancelled it is triggered by a user action. 
The intent should be Success.

bglownia (approved):

PR #3432, opened 2023-04-13
feat(trading): add liquidity provision link to orders table

### LINKED ISSUE #3372: Liquidity provision link in orders datagrid
## Story

As a liquidity provider
I want to be able to click and see the full liquidity provision
So that I know what orders belong to what provision

## Acceptance Criteria

- [x] I can click through to the relevant liquidity provision from the order list

## Tasks

- [x] Dev
- [ ] Test

I've broken this issue out from #1716

MadalinaRaicu:
# Related issues 🔗

Closes #3372 

# Description ℹ️

Add liquidity provision link to orders table for that market. 

# Demo 📺
Click on `Liquidity provision` cell:  
Screenshot 2023-04-13 at 14 42 40

... and navigate to
Screenshot 2023-04-13 at 14 42 46

# Technical 👨‍🔧

### DIFF
--- a/.github/workflows/ci-cd-trigger.yml
+++ b/.github/workflows/ci-cd-trigger.yml
@@ -50,7 +50,7 @@ jobs:
     secrets: inherit
 
   lint-test-build:
-    timeout-minutes: 10
+    timeout-minutes: 20
     needs: node-modules
     runs-on: ubuntu-22.04
     name: '(CI) lint + unit test + build'
--- a/.github/workflows/publish-dist.yml
+++ b/.github/workflows/publish-dist.yml
@@ -15,7 +15,7 @@ jobs:
         app: ${{ fromJSON(inputs.projects) }}
     name: ${{ matrix.app }}
     runs-on: ubuntu-22.04
-    timeout-minutes: 10
+    timeout-minutes: 20
     steps:
       - name: Check out code
         uses: actions/checkout@v3


libs/datagrid/src/lib/cells/order-type-cell.tsx
```diff
@@ -0,0 +1,62 @@
...
+import * as Schema from '@vegaprotocol/types';
+
+interface OrderTypeCellProps {
+  value?: Schema.OrderType;
+  data?: {
+    id?: string;
+    marketId?: string;
+    market?: { id: string };
+    peggedOrder?: any;
+    liquidityProvision?: any;
```
mattrussell36, line 16: chore (blocking): Why not remove the `any`s with

```
peggedOrder: Order['peggedOrder']
liquidityProvision: Order['liquidityProvision']
```

libs/datagrid/src/lib/cells/order-type-cell.tsx
```diff
@@ -0,0 +1,62 @@
...
+  onClick?: (marketId: string, metaKey?: boolean) => void;
+}
+
+export const OrderTypeCell = ({
+  value,
+  data: order,
+  idPath,
+  onClick,
+}: OrderTypeCellProps) => {
+  const id = order ? get(order, idPath ?? 'id', 'all') : '';
```
mattrussell36, line 28: suggestion (non blocking): I think this would be cleaner and simpler if the onClick function just passed the order object. Then the app can handle it however it pleases and you don't need to make sure the id path matches up

mattrussell36 (changes requested):

libs/datagrid/src/lib/cells/order-type-cell.tsx
mattrussell36: I'd even say we could go further and pass the actual browser event too

libs/datagrid/src/lib/cells/order-type-cell.tsx
mattrussell36: Or actually

```
data?: Order
```

libs/datagrid/src/lib/cells/order-type-cell.tsx
```diff
@@ -0,0 +1,62 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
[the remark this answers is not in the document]
MadalinaRaicu: updated

libs/datagrid/src/lib/cells/order-type-cell.tsx
MadalinaRaicu: updated

libs/datagrid/src/lib/cells/order-type-cell.tsx
MadalinaRaicu: yes, cause this one is not reused. and it is very order type specific. 
for now only for liquidity provision orders we're rendering a link

MadalinaRaicu: I updated it, it is very similar with market-name-cell but not as re-used so we don't need that id path.

mattrussell36 (approved):

PR #3521, opened 2023-04-23
fix: use correct vote value string for casting vote

### LINKED ISSUE #3520: Unable to vote on governance proposals
## Description

An issue has been identified with users being unable to vote upon the [governance proposal](https://token.vega.xyz/governance/d2157929132456dbc66eecbd478307156066243cd8769306e71f31882c22344f) that proposes to change the network parameter `reward.staking.delegation.delegatorShare` to `0.8`.

## Steps to Reproduce

1. Navigate to the open [governance proposal ](https://token.vega.xyz/governance/d2157929132456dbc66eecbd478307156066243cd8769306e71f31882c22344f)
2. Connect a Vega wallet that has Vega tokens thus allowing voting
3. Attempt to vote on the proposal 
4. Note that it is not working as would be expected

(alternatively submit a `tx send` command using the [example payload from the docs site ](https://docs.vega.xyz/mainnet/tools/vega-wallet/cli-wallet/latest/guides/build-send-commands#example-commands)

## Expected behavior

Votings to be registered 

## Screenshots

image

mattrussell36:
# Related issues 🔗

Closes #3520 

# Description ℹ️

Fixes casting a vote being broken in the token app. 

# Technical 👨‍🔧

The wrong string was being passed in the transaction payload, not picked up on because the types were also incorrect

### DIFF
--- a/apps/token/src/routes/governance/components/vote-details/use-user-vote.tsx
+++ b/apps/token/src/routes/governance/components/vote-details/use-user-vote.tsx
@@ -95,12 +95,17 @@ export function useUserVote(
 
     setVoteState(VoteState.Requested);
 
+    const voteMapping = {
+      [VoteValue.Yes]: 'VALUE_YES',
+      [VoteValue.No]: 'VALUE_NO',
+    } as const;
+
     try {
       const variables = {
         pubKey: keypair.pub,
         propagate: true,
         voteSubmission: {
-          value: value,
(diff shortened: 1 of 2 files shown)


jeremyletang (approved):

edd (approved):
Fix looks fine. Is there anything that we could automate to make the types get updated more reliably?

PR #3536, opened 2023-04-25
feat(explorer): add epoch to block explorer homepage

### LINKED ISSUE #3362: Adding Epochs to the Explorer front page
hey team! Would be great if we could also add the Epoch number on the Explorer front page :) 

Screenshot 2023-04-04 at 15 23 29

edd:
# Related issues 🔗

Closes #3362

# Description ℹ️

Adds a new block to the homepage stats on block explorer. It's the epoch number! For the current epoch!

Note that the GraphQL statistics endpoint does not contain the epoch ID for the current epoch, while the REST equivalent does. That's why this query uses the `Epoch` top level query.

# Demo 📺
Screenshot 2023-04-25 at 13 53 36


apps/explorer-e2e/src/integration/home.cy.js
```diff
@@ -9,64 +9,48 @@ context('Home Page', function () {
...
-      cy.get(statsValue)
-        .eq(12)
-        .invoke('text')
-        .should('match', /v\d+\.\d+\.\d+/i);
-      cy.get(statsValue)
-        .eq(13)
-        .invoke('text')
-        .should('match', /\d+\.\d+\.\d+/i);
       cy.get(statsValue).eq(14).should('not.be.empty');
       cy.get(statsValue).eq(15).should('not.be.empty');
```
dexturr, line 48: Should we add a test here for stats at 16? Now there are 17 times?

dexturr (approved):

apps/explorer-e2e/src/integration/home.cy.js
edd: I've removed the value tests so there isn't inconsistent testing of some fields.

sam-keen (approved):

PR #3541, opened 2023-04-25
feat(trading): divide order tab into open, closed, rejected and all

### LINKED ISSUE #3452: When monitoring my orders I want to jump to see what is active, get feedback on what has "failed" or changed, as well as All orders
## Story

~~As a~~ when monitoring the positions I have in Vega,
I want to know what orders have have active, see feedback on orders that have changed status or failed as well as see all orders (and apply bespoke filters
So that I don't have to re-apply filters each time

image

Proposal part 1: A tab for each order status (or order status group)
- `Active` + `Parked`. aka **Open**
- `Canceled` + `Expired` + `filled` + `Partially filled` + `stopped` - aka **Closed**
- `Rejected` - aka **Rejected**
- All

When the viewport is wide enough for two tables:

Screenshot 2023-04-18 at 09 59 48

On the narrow viewport I (Michael) suggest showing all tabs (existing behaviour)

Barney suggests a single "Orders" tab, perhaps with the same default as implemented.

Screenshot 2023-04-18 at 10 01 34

Proposal part 2: Show a badge / count on each tab

- Open shows a badge with a count of all the "open" orders (parked and active)
- Badge could flash or light up each time this changes
- Canceled shows a badge (not a count) to show that this has changed since the last time the user looked at it (on this device). 
  - However it may not be necessary to update this badge if the change happended due to a user action in console e.g... they hit the cancel order button (although it probably should happen if they place a market order that gets part filled) Lets play with this on a deploy preview and decide.
- Rejected has badge that is reset each time the user views the tab

[Slack discussion](https://vegaprotocol.slack.com/archives/C037STCKMGC/p1680862379891279)

## Acceptance Criteria

- [ ] See above and [existing ACs](https://github.com/vegaprotocol/specs/blob/master/user-interface/7003-MORD-manage_orders.md#manage-orders).

## Tasks

- [x] Explore and sketch
- [ ] Visual Design
- [ ] Team review
- [ ] Etc.

## Sketch

See above

### LINKED ISSUE #3528: Orders table timing out due to filters
## Description

The orders query can easily time out due to filters. We need to introduce the live flag or handle orders differently

## Additional context

https://vegaprotocol.slack.com/archives/C02KVKMAE82/p1682003156032039

bglownia:
# Related issues 🔗

Closes #3452
Closes #3528


libs/datagrid/src/lib/filters/set-filter.tsx
```diff
@@ -9,85 +9,90 @@ import {
...
+                disabled={readonly}
+                className="mr-1"
+                checked={value.includes(key)}
+                onChange={onChange}
+              />
+              <span>{props.colDef.filterParams.set[key]}</span>
+            </label>
+          ))}
+        </fieldset>
+        {!readonly && (
```
mattrussell36, line 83: question (non blocking): why conditionally render the reset button? It doesn't seem necessary to me? If you are in view as mode, you'd probably still want to be able to filter

libs/orders/src/lib/components/order-list-manager/order-list-manager.tsx
```diff
@@ -1,25 +1,41 @@
...
 } from '@vegaprotocol/wallet';
-import isEqual from 'lodash/isEqual';
 import type { OrderTxUpdateFieldsFragment } from '@vegaprotocol/wallet';
 import { OrderEditDialog } from '../order-list/order-edit-dialog';
-import type { Order, OrderEdge } from '../order-data-provider';
+import type { Order } from '../order-data-provider';
+import { OrderStatus } from '@vegaprotocol/types';
+
+export enum Filter {
+  'Open' = 1,
```
mattrussell36, line 23: question (non blocking): This looks a little weird to me, might be worth adding a comment. Why override the default indexing that the enum creates?

mattrussell36 (approved):

libs/datagrid/src/lib/filters/set-filter.tsx
bglownia: If you are on active, closed, rejected tabs orders status filter value is preselected and filter is readonly. You can see what is selected but input's are disabled. input[type=checkbox] get's different styles so it's obvious for the user that he cannot change value.
If you cannot change value why we should render reset button? We can make it disabled, but it will not change styles. Of course we can change styles but it's just easier and less confusing just to hide it

MadalinaRaicu (approved):

PR #3568, opened 2023-04-28
feat(governance): eth chain should not inhibit usage

### LINKED ISSUE #2996: Eth chain should not inhibit usage of site
As a user
I want to be able to see the token site when I have change network in my ethereum wallet
So that I can continue to view the token site 

##Acceptance criteria

- [ ] After connecting my ethereum wallet, when I change my network I can see see and navigate the token site
- [ ] When I attempt to initiate any transaction that involves my ethereum wallet I am prompted to change to the right network

 ## Tasks 

- [ ] Establish all the transactions that involve the ethereum wallet (and therefore need to be prompted to change network from)  
- [ ] Implement 
- [ ] Test
- [ ] Release

Image

sam-keen:
# Related issues 🔗

Closes #2996

# Description ℹ️

If a user connects their eth wallet to an unsupported network, they are disconnected automatically, and a notice is displayed in the eth wallet area of the sidebar, but they are otherwise not inhibited in their use of the app. If they wish to initiate a transaction, they can reconnect their wallet, which should automatically connect them to the correct network.

# Demo 📺

Screenshot 2023-04-28 at 17 31 50

### DIFF
--- a/apps/governance/src/components/web3-connector/web3-connector.tsx
+++ b/apps/governance/src/components/web3-connector/web3-connector.tsx
@@ -88,10 +88,8 @@ export const Web3Content = ({ children, appChainId }: Web3ContentProps) => {
         if (previousChainId !== undefined && !appState.disconnectNotice) {
           showDisconnectNotice(true);
         }
-      } else {
-        if (appState.disconnectNotice) {
-          showDisconnectNotice(false);
-        }
+      } else if (appState.disconnectNotice) {
+        showDisconnectNotice(false);
       }
     }
   }, [
--- a/apps/governance/src/components/web3-connector/web3-connector.tsx
+++ b/apps/governance/src/components/web3-connector/web3-connector.tsx
@@ -1,6 +1,5 @@
 import { Button, Splash } from '@vegaprotocol/ui-toolkit';
 import {
-  getChainName,
   useWeb3ConnectStore,
   useWeb3Disconnect,
   Web3ConnectDialog,
@@ -26,7 +25,6 @@ export function Web3Connector({
   connectors,
   chainId,
 }: Web3ConnectorProps) {
-  const [showDisconnectBanner, setShowDisconnectBanner] = useState(false);
   const { appState, appDispatch } = useAppState();
   const setDialogOpen = useCallback(
     (isOpen: boolean) => {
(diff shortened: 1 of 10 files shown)


apps/governance/src/components/web3-connector/web3-connector.tsx
```diff
@@ -52,23 +49,61 @@ export function Web3Connector({
...
+        // If the user was previously connected, show the disconnect explanation notice.
+        if (previousChainId !== undefined && !appState.disconnectNotice) {
+          showDisconnectNotice(true);
+        }
+      } else {
+        if (appState.disconnectNotice) {
+          showDisconnectNotice(false);
+        }
+      }
+    }
```
dexturr, line 96: nitpick (non-blocking): would probably makes this an else if rather than an else and an if

apps/governance/src/components/web3-connector/web3-connector.tsx
```diff
@@ -52,23 +49,61 @@ export function Web3Connector({
...
+  useEffect(() => {
+    if (chainId !== undefined) {
+      // We use this to detect when the user switches networks.
+      setPreviousChainId(chainId);
+
+      if (chainId !== appChainId) {
+        disconnect();
+
+        // If the user was previously connected, show the disconnect explanation notice.
+        if (previousChainId !== undefined && !appState.disconnectNotice) {
```
dexturr, line 88: question (blocking): `previousChainId !== undefined` is always true because of if on line 80.

dexturr: Talked through on slack

dexturr (approved):

apps/governance/src/components/web3-connector/web3-connector.tsx
sam-keen: Discussed in DMs, but essentially:

- It's possible for previousChainId to be undefined whilst the current chain id is not undefined
- If I remove that 'previousChainId !== undefined' check, and if a user still has a metamask connection and they're on the wrong network, we show the user that they've been disconnected all the time. This is true, but it's needlessly noisy. I'm aiming to only show them the disconnected notice after they deliberately change network. Whether this is the right thing to do is up for debate

PR #3588, opened 2023-05-03
fix(datagrid): prevent unmount of aggrid on theme change

### LINKED ISSUE #3208: Changing theme loses datagrid filters and sort state
## Description

If you apply filters to a data grid and then change theme, the filter state is reset
As part of this ticket we also want to change to rely only on React.lazy and remove next.js dynamic

## Steps to Reproduce

1. Go to a table and apply a filter
2. Change theme
3. Note that the filter state has been lost

## Expected behavior

- Filter state should be preserved
- Sorting state should be preserved

## Additional context

- This is because light and dark them for datagrids is handled using separate components in order to set dynamic custom css properties. We should be able to dynamically switch _only_ the aggrid css and custom properites when the theme changes, allowing the filter selection to be preserved.
- We should also probably use custom tailwind classes in the tailwind config to keep all styles in the same place

mattrussell36:
# Related issues 🔗

Closes #3208 

# Description ℹ️

Fixes a problem where changing theme resulted in table state being lost (filters, column resize etc), due to having different components rendered for light and dark themes. Now only the classname changes and different css variables are used.

# Technical 👨‍🔧

- Also switches all AgGrid to use only React.lazy imported grids


MadalinaRaicu (approved):

apps/governance/src/routes/staking/home/validator-tables/standby-pending-validators-table.tsx
```diff
@@ -248,11 +247,13 @@ export const StandbyPendingValidatorsTable = ({
     );
 
     return (
-      <div data-testid="standby-pending-validators-table">
+      <div
+        data-testid="standby-pending-validators-table"
+        className="validators-table"
```
macqbat, line 252: non-blocking: why we need this className? Can't see anything related to it in this PR. Maybe some styles from `customThemeParams`, are being removed here, should be implemented using it?

libs/accounts/src/lib/accounts-manager.spec.tsx
```diff
@@ -66,7 +66,7 @@ describe('AccountManager', () => {
 
     it('update method should return proper result', async () => {
       let rerenderer: (ui: React.ReactElement) => void;
-      await act(() => {
+      await act(async () => {
```
macqbat, line 69: non-blocking: don't think `async` here is necessary: return value of act is always a thenable, so `await` is useful, but `async` on callback if there is no any `await` inside a callback is redundant

apps/governance/src/routes/staking/home/validator-tables/standby-pending-validators-table.tsx
mattrussell36: Well spotted this was for some custom styles for some aggrids on the governance app. This has been merged so I will make a follow up Pr

PR #3595, opened 2023-05-03
feat(trading): persist table column sizes

### LINKED ISSUE #3393: Persist data grid column sizes (per pane)
## Story

As a user
When I have adjusted the table layout to my preferences
I want column sizes to be persisted (per pane) when I refresh my browser
So that I don't have to resize to my desired table layout every time

## Acceptance Criteria

- [x] I can resize columns
- [x] When I switch to view another market, I can see the table column sizes I specified are preserved
- [x] When I refresh the page I am looking at, I can see the table column sizes I specified are preserved
- [x] When I switch between light and dark themes, I can still see the table column sizes I specified are preserved
- [x] All markets use the same column sizes
- [x] This same behaviour is true for every table in console where I can adjust column sizes
- [ ] There is a way to reset to the default column sizes (and potentially other things that are configurable by the user)

## Tasks

- [ ] Determine whether this should be applied per pane or per component (class or instance)
- [ ] Explore and sketch how / where the user would reset column sizes (and how this effects the other things we are persisting)
- [ ] Set up versioning in local storage to handle user configurable settings that might no longer work i.e. a different number of columns
- [ ] Implementation
- [ ] Write tests (unit and e2e) 
- [ ] Code review
- [ ] Verify acceptance criteria met
- [ ] Exploratory testing to find bugs and edge cases
- [ ] Write automated regression tests for the A/Cs and add to spec doc (expanding on A/Cs where possible)

macqbat:
# Related issues 🔗

Closes #3393

# Description ℹ️

Make column sizes persistent by save state of their dimensions in localStorage

# Demo 📺

# Technical 👨‍🔧


libs/datagrid/src/lib/ag-grid/use-column-sizes.ts
```diff
@@ -0,0 +1,138 @@
...
+    const width = getWidthOfAll();
+    if (width && sizes['width'] && width !== sizes['width']) {
+      setCalculatedSizes(recalculateSizes(sizes));
+    }
+  }, [getWidthOfAll, recalculateSizes, sizes]);
+  useResizeObserver(container?.current as Element, onResize);
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const handleOnChange = useCallback(
+    debounce((columns: Column[]) => {
```
mattrussell36, line 88: chore (blocking): you can remove debounced logic as AgGrid provides a property event.finished which will be true once dragging is finished

libs/datagrid/src/lib/ag-grid/use-column-sizes.ts
```diff
@@ -0,0 +1,138 @@
...
+          aggr[column.getColId()] = column.getActualWidth();
+          return aggr;
+        }, {} as Record<string, number>);
+        sizesObj['width'] = getWidthOfAll();
+        valueSetter(id, sizesObj);
+      }
+    }, COLUMNS_SET_DEBOUNCE_TIME),
+    [valueSetter, id]
+  );
+  const reshapeAgGridChildren = useCallback(
```
mattrussell36, line 100: chore (blocking): I think we can make this simpler by doing the resizing using the api in the `onGridReady` callback, then we don't need to modify children and coldefs.

Additionally we can call `sizeColumnsToFit` if not stored sizes are found so that the columns fill available space. Right now there is unused space on some tables

Something like this should work 🤔 

```
onGridReady={(event) => {
  if (!sizes) {
    event.columnApi.sizeColumnsToFit();
  } else {
    // convert sizes to an array of column size in this shape: { key: "the-col-id", size: number }
    const sizes = Object.entries(sizes).map(
      ([key, size]) => ({
        key,
          newWidth: size,
        })
      );
      event.columnApi.setColumnWidths(sizes);
    }
}}
```

libs/datagrid/src/lib/ag-grid/ag-grid-lazy-themed.tsx
```diff
@@ -1,30 +1,56 @@
...
+  children,
   ...props
 }: (AgGridReactProps | AgReactUiProps) & {
   style?: React.CSSProperties;
   gridRef?: React.ForwardedRef<AgGridReact>;
+  id?: string;
+  children?: ReactNode[];
 }) => {
+  const containerRef = useRef<HTMLDivElement | null>(null);
+  const [setValues, reshapeChildren, reshapeProps] = useColumnSizes({
```
mattrussell36, line 23: I don't think we should be enforcing every grid to store columns. It should be opt in via prop, or the hook should be used in each parent component of the table

mattrussell36 (changes requested):

libs/datagrid/src/lib/ag-grid/ag-grid-lazy-themed.tsx
macqbat: It is opt in, it's enableable via `id` property which is also the key in the store

libs/datagrid/src/lib/ag-grid/use-column-sizes.ts
macqbat: payload of the call when the `event.finished` is fired is much poorer than in other calls (with no `event.finished` = true). When event.finished is true, there is only one column dimension passed. Otherwise we have more siblings column, we can save simultaneously. I thought it can be more appropriate in our case.

libs/datagrid/src/lib/ag-grid/use-column-sizes.ts
macqbat: I'm afraid, relying on `onGridReady` can bring us blink effect, since it's fired usually after some time

macqbat: ok, after some checks, it doesn't look bad, I mean `setColumnWidths` instead iterating children and props. Please take a look.

libs/datagrid/src/lib/ag-grid/ag-grid-lazy-themed.tsx
asiaznik: ~~`id`~~ -> `storeKey`

macqbat: done

libs/datagrid/src/lib/ag-grid/use-column-sizes.ts
mattrussell36: `event.columnApi.getColumnState()` does the trick

libs/datagrid/src/lib/ag-grid/ag-grid-lazy-themed.tsx
```diff
@@ -1,30 +1,78 @@
...
+          setValues(event.columns);
+        }
+      },
+      [setValues]
+    ),
   };
+  const { onGridReady } = props;
+  const onGridReadyInternal = useCallback(
+    (event: GridReadyEvent) => {
+      onGridReady?.(event);
```
mattrussell36, line 45: @macqbat I don't like all this logic being in this component. What if you want to just render a simple grid that doesn't store column widths or one that just uses a flex layout? For example in the governance app. They now have local storage entries which are not required.

I would suggest moving more of the logic to the hook and let the consumer dictate when to use the store.

If we can get the hook's API to look something more like the below it would be more explicit and declarative and doesn't require you to know the inner workings of our AgGrid component (IE what happens if you use storeKey). As well as not polluting other apps localstorage.

This also has the added benefit of tidying hook logic because we can guarantee that the required arguments are provided.

```
const MyGrid = forwardRef((props, ref) => {
  const { onGridReady, onColumnResized } = useColumnSizes({
    storeKey: 'MyGrid',
    gridRef: ref
  })

  return (
    <AgGird
       ref={ref}
       onGridReady={onGridReady}
       onColumnResized={onColumnResized}
    />
  )
})
```

If you really dont want to have to useColumnSizes in multiple places I would suggest an HOC around the grid component

```
const PersistedDataGrid = ({ storeKey: { storeKey: string }) => {
    const callbacks = useColumnSizes({
      storeKey,
      ...
    });
    return <AgGrid {...callbacks} />
}
```

bglownia (approved):

libs/datagrid/src/lib/ag-grid/ag-grid-lazy-themed.tsx
macqbat: I'm afraid the use of HOC here would't reduce amount of places where changes should be implemented, since our usual hierarchy of ag-grid components (container => manager => table) is quite deep, and potential HOC should somehow passing it all by as well as take into consideration lazy import of ag-grid children.
But I made some tidying up as you've suggested, as closing up more logic into hook, and in case of not adding storeKey all other components can actually omit results of this hook. I think it looks quite good now.. :-)

mattrussell36 (approved):

PR #3612, opened 2023-05-04
feat(ci): establish build based on docker to produce ipfs-hashes

mkjmdski:
closes: https://github.com/vegaprotocol/devops-infra/issues/1762

### DIFF
--- a/.github/workflows/publish-dist.yml
+++ b/.github/workflows/publish-dist.yml
@@ -48,7 +48,7 @@ jobs:
       - name: Setup node
         uses: actions/setup-node@v3
         with:
-          node-version-file: ".nvmrc"
+          node-version-file: '.nvmrc'
           # https://stackoverflow.com/questions/61010294/how-to-cache-yarn-packages-in-github-actions
           cache: yarn
 
@@ -159,8 +159,8 @@ jobs:
           AWS_S3_BUCKET: ${{ env.BUCKET_NAME }}
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-          AWS_REGION: "eu-west-1"
-          SOURCE_DIR: "dist-result"
+          AWS_REGION: 'eu-west-1'
+          SOURCE_DIR: 'dist-result'
 
       - name: Add preview label
         uses: actions-ecosystem/action-add-labels@v1


README.md
```diff
@@ -103,25 +103,95 @@ In CI linting, formatting and also run. These checks can be seen in the [CI work
...
+## Build Instructions
+
+### dist build
+
+Packaging prepared dist into [`nginx`](https://hub.docker.com/_/nginx)([server configuration](./nginx/nginx.conf)) docker image involves building the application on docker host machine from source.
+
+As a prerequisite you need to perform build of `dist` directory and move its content for specific application to `dist-result` directory.
+
+This is an example script written in bash with possible implementation (using `--env=mainnet` flag to control `.env` file):
+
```
edd, line 127: **question (non-blocking)**: Could this be turned in to a make script rather?

README.md
```diff
@@ -103,25 +103,95 @@ In CI linting, formatting and also run. These checks can be seen in the [CI work
...
 
 ```bash
-docker build --dockerfile dockerfiles/Dockerfile.cra . --build-arg APP=[YOUR APP] --tag=[TAG]
+docker build --dockerfile dockerfiles/dist.Dockerfile . --tag=[TAG]
 ```
 
-In order to run a container:
+### node in docker build
+
+Using mutlistage dockerfile dist is compiled using [node](https://hub.docker.com/_/node) image and later packed to nginx as in [dist build](#dist-build) example.
```
edd, line 150:
```suggestion
Using multistage dockerfile dist is compiled using [node](https://hub.docker.com/_/node) image and later packed to nginx as in [dist build](#dist-build) example.
```

README.md
```diff
@@ -103,25 +103,95 @@ In CI linting, formatting and also run. These checks can be seen in the [CI work
...
+docker run --rm [TAG] cat /ipfs-hash
+```
+
+To recalculate hash again you need to run
+
+```bash
+docker run --rm [TAG] ipfs add -rw /usr/share/nginx/html
+```
+
+or to skip progress bar:
```
edd, line 172: **thought (non-blocking)**: Let's drop this - it makes this section too long

README.md
```diff
@@ -103,25 +103,95 @@ In CI linting, formatting and also run. These checks can be seen in the [CI work
...
+
+At the moment this feature is important only for `trading` (console) releases.
+
+Each docker build finishes with hash calculation for dist directory. Resulting hash is added to file named as `/ipfs-hash`. Once docker image is produced you can run following commad to display ipfs-hash:
+
+```bash
+docker run --rm [TAG] cat /ipfs-hash
+```
+
+To recalculate hash again you need to run
```
edd, line 166: **thought (non-blocking)**: Let's drop this

README.md
```diff
@@ -103,25 +103,95 @@ In CI linting, formatting and also run. These checks can be seen in the [CI work
...
+```
+
+### Computing ipfs-hash of the build
+
+At the moment this feature is important only for `trading` (console) releases.
+
+Each docker build finishes with hash calculation for dist directory. Resulting hash is added to file named as `/ipfs-hash`. Once docker image is produced you can run following commad to display ipfs-hash:
+
+```bash
+docker run --rm [TAG] cat /ipfs-hash
```
edd, line 163: **thought (blocking)**: This should be moved to a script or makefile

README.md
```diff
@@ -103,25 +103,95 @@ In CI linting, formatting and also run. These checks can be seen in the [CI work
...
-The [Dockerfile](./dockerfiles) for running the frontends is pretty basic, merely building the application with the APP arg that is passed in and serving the application from [nginx](./nginx/nginx.conf). The only complexity that exists is that there is a script which allows the passing of run time environment variables to the containers. See configuration below for how to do this.
+There are two [Dockerfiles](./dockerfiles) for running the frontends: **dist** build and **node in docker** build. For specificc build instructions follow [build instructions](#build-instructions)
+
+In order to run a container on port 3000:
+
+```bash
+docker run -p 3000:80 [TAG]
+```
+
+## Build Instructions
```
edd, line 118: **thought (non-blocking)**: We should rename this section. This isn't required for most people, but it is a reasonable way of setting up a hosted version for the web, or pinning a build on IPFS. 

How about `Hosting a Console`

README.md
```diff
@@ -103,25 +103,95 @@ In CI linting, formatting and also run. These checks can be seen in the [CI work
...
+
+### dist build
+
+Packaging prepared dist into [`nginx`](https://hub.docker.com/_/nginx)([server configuration](./nginx/nginx.conf)) docker image involves building the application on docker host machine from source.
```
edd, line 122:
```suggestion
Packaging prepared dist into [`nginx`](https://hub.docker.com/_/nginx)([server configuration](./nginx/nginx.conf)) docker image involves building the application on docker host machine from source.
```

I don't think this section is clear enough.  Perhaps something more along the lines of:

'The `docker` subfolder has some docker configurations for easily setting up your own hosted version of console either for the web, or ready for pinning on IPFS'

README.md
```diff
@@ -103,25 +103,95 @@ In CI linting, formatting and also run. These checks can be seen in the [CI work
...
+
+As a prerequisite you need to perform build of `dist` directory and move its content for specific application to `dist-result` directory.
```
edd, line 124: **thought (blocking)**: If we made this part of a `makefile`, we can drop this section of the docs - and probably drop the `dist-result` folder as well

README.md
```diff
@@ -103,25 +103,95 @@ In CI linting, formatting and also run. These checks can be seen in the [CI work
...
+
+```bash
+docker run --rm [TAG] ipfs add -rwQ /usr/share/nginx/html
+```
+
+**updating hash:** recompiling dist directory (even if there are no changed to source code) results in different hash computed by ipfs command.
+
+### Verifying ipfs-hash of existing current application version
+
+it is advised to compare computed hash with the one from github [release notes](https://github.com/vegaprotocol/frontend-monorepo/releases) to ensure you are running exact same version.
```
edd, line 182: **nitpick (blocking)**: I was going to correct the missing capital letter at the start of this sentence, but I think we could probably make this clearer.

> An IPFS CID will be attached to every release. If you are intending to pin an application on IPFS, you can check that your build matches by running the following steps:

README.md
```diff
@@ -103,25 +103,95 @@ In CI linting, formatting and also run. These checks can be seen in the [CI work
...
+
+1. run `docker pull vegaprotocol/<APP NAME>:<TAG>`
+2. follow instructions from [previous step](#computing-ipfs-hash-of-the-build) and use image name from step 1 as `[TAG]` to recalculate html directory
+3. you should see exactly same hash produced by ipfs command as one placed with the release notes
+4. if you want to extract dist from docker image to your local filesystem you can run following command:
```
edd, line 187: **thought (non-blocking)**: Again I think this could be a `makefile` target called something like `eject`, so that end to end it's more like

`make release-tag; make release-tag-ipfs; make eject`

edd (changes requested):
More of this could be automated which would let us simplify the documentation section.

PR #3634, opened 2023-05-08
feat(trading): hook up new EstimatePosition and EstimateFees api methods

### LINKED ISSUE #3216: Hook up new `estimatePosition` and `estimateFees` queries
## The Chore

Once the `EstimateOrder` API method has been refactored (see https://github.com/vegaprotocol/vega/issues/7847) we need to ensure console is using this new logic in the estimations on the deal ticket.

## Tasks

- [ ] Hook up new EstimateOrder API method
- [ ] Restore the "estimated liquidation price" field
- [ ] Compare the accuracy of the new estimations with the old method
- [ ] Check against #3531 

- Dependent on https://github.com/vegaprotocol/vega/issues/7847

bglownia:
# Related issues 🔗

Closes #3216


libs/accounts/src/lib/use-account-balance.tsx
```diff
@@ -33,9 +33,9 @@ export const useAccountBalance = (assetId?: string) => {
 
   return useMemo(
     () => ({
-      accountBalance,
-      accountDecimals,
+      accountBalance: pubKey ? accountBalance : '',
```
macqbat, line 36: why we need this? Because state isn't reset after partyId change? If so, maybe such kind of reset should be implemented instead?

libs/accounts/src/lib/use-market-account-balance.tsx
```diff
@@ -32,9 +32,9 @@ export const useMarketAccountBalance = (marketId: string) => {
 
   return useMemo(
     () => ({
-      accountBalance,
-      accountDecimals,
+      accountBalance: pubKey ? accountBalance : '',
```
macqbat, line 35: here as well

guoguojin (approved):

libs/environment/src/hooks/use-node-health.ts
```diff
@@ -38,7 +38,7 @@ export const useNodeHealth = () => {
       return;
     }
 
-    if (!('Cypress' in window)) {
+    if (!('Cypress' in window) && window.location.hostname !== 'localhost') {
```
macqbat, line 41: is this intentional?

libs/proposals/src/lib/protocol-upgrade-proposals/use-time-to-upgrade.ts
```diff
@@ -8,20 +8,29 @@ const durations = [] as number[];
...
     skip: durations.length === polls,
   });
 
+  useEffect(() => {
+    if (error) {
+      stopPolling();
+      return;
+    }
+
+    if (!('Cypress' in window) && window.location.hostname !== 'localhost') {
```
macqbat, line 23: here as well.

macqbat (approved):

PR #3675, opened 2023-05-09
feat(trading): show pegged order offset and reference in the order table

### LINKED ISSUE #3551: See reference and offset (as well as price) for pegged orders
## Story

~~As a~~ When I have a pegged order
I want to see the reference and offset (as well as the price)
So that I can confirm that the order is when I expect it to be

## Acceptance Criteria

- [ ] I can see the Reference (bid, ask or mid)
- [ ] I can see the offset (e.g. + 0.001 for a Sell, or -1231.023 for a Buy)
- [ ] I can also see the current price (if unfilled), the last price it had (if it was filled canceled etc) or the rejected/stopped reason if there is one (note this is already implemented but I mention it because it is still required

## Tasks

- [x] Explore and sketch
- [ ] Visual Design
- [ ] Team review
- [ ] Etc.

## Sketch

image

One could argue that the word limit is not needed as that is all it could be, but in the future it could be possible to peg stop orders

MadalinaRaicu:
# Related issues 🔗

Closes #3551 

# Description ℹ️

Show pegged order details in the order list. 

# Demo 📺
Screenshot 2023-05-09 at 18 43 05

# Technical 👨‍🔧


libs/datagrid/src/lib/cells/order-type-cell.tsx
```diff
@@ -23,7 +24,13 @@ export const OrderTypeCell = ({
...
     if (!value) return '-';
     if (order?.peggedOrder) {
-      return t('Pegged');
+      const offset = addDecimalsFormatNumber(
+        order.peggedOrder?.offset,
+        order.market.decimalPlaces
+      );
+      const reference =
+        Schema.PeggedReferenceMapping[order.peggedOrder?.reference];
+      return t('%s - %s Peg limit', [reference, offset]);
```
edd, line 33: **question (non-blocking)**:  Is this going to cover all the possible pegs? Just worrying if it'll mis-describe some

MadalinaRaicu: agreed, just updated! it needs to be different based on buy/sell

jeremyletang (approved):

guoguojin (approved):
GQL looks fine to me

libs/datagrid/src/lib/cells/order-type-cell.tsx
```diff
@@ -23,7 +24,16 @@ export const OrderTypeCell = ({
     }
     if (!value) return '-';
     if (order?.peggedOrder) {
-      return t('Pegged');
+      // reference (bid, ask or mid)
```
MadalinaRaicu, line 27:
```suggestion
```

Issue #3672, opened 2023-05-09
Node checker shouldnt stop you selecting a node that is failing the checks


JonRay15:
## The Chore

The node checker function on the apps is a bit ropey and stops you selecting nodes that are actually fine.  We should disable that part so that even if it thinks the node is bad it still lets you select it anyway.

## Tasks

- [x] Do it 
- [ ] Test it 
- [ ] Tell Jon Ray

pennyandrews:
Duplicate ticket

PR #3678, opened 2023-05-09
feat(explorer,governance): always allow selecting a node, add node guard

mattrussell36:
# Related issues 🔗

#3672 

# Description ℹ️

Allows selecting to any node, regardless of start up checks. Added NodeGuard component to allow switching nodes in case your selected node wasn't suitable

### DIFF
--- a/libs/environment/src/components/node-guard/NodeGuard.graphql
+++ b/libs/environment/src/components/node-guard/NodeGuard.graphql
@@ -1,7 +1,5 @@
 query NodeGuard {
-  statistics {
-    blockHeight
-  }
+  lastBlockHeight
   networkParametersConnection {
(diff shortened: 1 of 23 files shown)


dexturr (approved):

MadalinaRaicu (approved):

libs/environment/src/components/node-guard/NodeGuard.graphql
```diff
@@ -0,0 +1,13 @@
+query NodeGuard {
+  statistics {
+    blockHeight
+  }
```
jeremyletang, line 4: you could use lastBlockHeight here, being not proxied to the core node, it should be faster

PR #3699, opened 2023-05-10
fix(explorer): fix bug where no votes displayed a yes icon

### LINKED ISSUE #3689: Block explorer showing thumbs up for a no vote
image

edd:
# Related issues 🔗

Closes #3689

# Description ℹ️

The icon displayed on a proposal transaction always displayed a thumbs up emoji, even if the vote was no. That's obviously
bad and misleading, but using an emoji was also a little lazy so this PR fixes the bug but also improves the display

- Add custom lozenge-thing to display an icon and text representing a vote
- Update vote TX view to use new component
- Update TX list view to use new component for proposals
- Update TX list view to de-emphasise hearbeat TXs
- Update TX success code to use an icon rather than an emoji

# Demo 📺


apps/explorer/src/app/components/vote-icon/vote-icon.spec.tsx
```diff
@@ -0,0 +1,37 @@
...
+  });
+
+  it('should use the noText for no votes if specified', () => {
+    const no = render(<VoteIcon vote={false} noText="Test" />);
+    expect(no.getByTestId('label')).toHaveTextContent('Test');
+  });
+
+  it('should display the delete icon for no votes', () => {
+    const no = render(<VoteIcon vote={false} />);
+    expect(no.getByRole('img')).toHaveAttribute('aria-label', 'delete icon');
```
sam-keen, line 35: Suggestion (non-blocking): 'delete icon' doesn't sound like a useful aria label for a 'no votes' component. 'no votes icon' would be fine

sam-keen (approved):

apps/explorer/src/app/components/vote-icon/vote-icon.spec.tsx
edd: Fair point. I'll do it later (this PR is getting old) - I checked and I can easily override the label. This one is just what it did by default. Lazy me.

PR #3708, opened 2023-05-10
fix(governance): tranches data

### LINKED ISSUE #3690: Issues with Tranches data
There seems to be something not quite right about the tranches data on mainnet https://governance.vega.xyz/token/tranches

[Lewis noticed](https://vegaprotocol.slack.com/archives/C038EB6V2RE/p1683709176604629) that we are showing negative numbers for locked tokens in tranches, where it should be zero

Image

AND

[Daniel noticed](https://vegaprotocol.slack.com/archives/CD4AKTFQT/p1683715619513609) that funds are showing as being unlocked in 2021, but red color shows 100% is locked

image (2)

sam-keen:
# Related issues 🔗

Closes #3690

# Description ℹ️

Tranche progress calculation required a couple of tweaks:

- Items in the past could have a 'lockedDecimal' greater than 1, resulting in incorrect totals being calculated (negative totals)
- Tranches that finish immediately were being treated as if they were always locked. Added some logic to check if their cliff_start value was in the past, giving them an unlocked value if so.

I also tweaked the layout of the tranche item to make the values more readable. See the 'unlocking starts' and 'fully unlocked' values below:

before:

Screenshot 2023-05-10 at 16 21 10

after:

Screenshot 2023-05-10 at 16 22 01

### DIFF
--- a/apps/governance/src/lib/tranches/tranches-store.tsx
+++ b/apps/governance/src/lib/tranches/tranches-store.tsx
@@ -50,7 +50,16 @@ export const useTranches = create<TranchesStore>()((set) => ({
         ?.map((t) => {
           const tranche_progress =
             t.duration !== 0 ? (now - t.cliff_start) / t.duration : 0;
-          const lockedDecimal = tranche_progress < 0 ? 1 : 1 - tranche_progress;
+          const lockedDecimal =
+            t.duration !== 0
+              ? tranche_progress < 0
+                ? 1
+                : 1 - tranche_progress
+              : now < t.cliff_start
+              ? 1
+              : 0;
+          const clampedLockedDecimal = Math.max(0, Math.min(1, lockedDecimal));
+
           return {
             tranche_id: t.tranche_id,
             tranche_start: secondsToDate(t.cliff_start),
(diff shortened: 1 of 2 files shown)


MadalinaRaicu (approved):

apps/governance/src/lib/tranches/tranches-store.tsx
```diff
@@ -50,7 +50,16 @@ export const useTranches = create<TranchesStore>()((set) => ({
...
             t.duration !== 0 ? (now - t.cliff_start) / t.duration : 0;
-          const lockedDecimal = tranche_progress < 0 ? 1 : 1 - tranche_progress;
+          const lockedDecimal =
+            t.duration !== 0
+              ? tranche_progress < 0
+                ? 1
+                : 1 - tranche_progress
+              : now < t.cliff_start
+              ? 1
+              : 0;
```
dexturr, line 60: Sorry to be annoying but can we make this an if statement or a function or something, this is hard to read.

dexturr (approved):

dexturr (approved):

PR #3716, opened 2023-05-10
fix(): change proposal sort order

### LINKED ISSUE #3681: Change proposal sort order
## The Chore

On the proposal screen and on the homepage the proposals are ordered so that most recent is top 

It should be ordered by ;
1. Closing Timestamp 
2. Submission timestamp

So that the ones that will close soonest are shown at the top since they are the most pressing. 

## Tasks

- [ ] Do it 
- [ ] Test it

sam-keen:
# Related issues 🔗

Closes #3681

# Description ℹ️

Orders by most recent by closing date, then by proposal submission date

### DIFF
--- a/apps/governance/src/routes/proposals/components/proposals-list/proposals-list.spec.tsx
+++ b/apps/governance/src/routes/proposals/components/proposals-list/proposals-list.spec.tsx
@@ -24,7 +24,6 @@ const openProposalClosesNextMonth = generateProposal({
   },
   terms: {
     closingDatetime: nextMonth.toString(),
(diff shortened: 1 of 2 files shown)


macqbat (approved):

edd (approved):

edd (approved):
👍 A test to check that the ordering is maintained would be good, but I think you can get away with it because the sorts are so simple it's more like testing AG Grid.

I'd still rather see one, but I'm not going to block for it or make a ticket that will never get prioritised.

PR #3726, opened 2023-05-11
chore(trading): suppress sentry logs when eth fails

### LINKED ISSUE #2581: BalanceOf call on Ethereum sometimes fails
Error: call revert exception [ See: https://links.ethers.org/v5-errors-CALL_EXCEPTION ] (method="balance...

Sentry Issue: [TRADING-6T](https://sentry.io/organizations/vega-o3/issues/3801249197/?referrer=github_integration)

```
Error: call revert exception [ See: https://links.ethers.org/v5-errors-CALL_EXCEPTION ] (method="balanceOf(address)", data="0x", errorArgs=null, errorName=null, errorSignature=null, reason=null, code=CALL_EXCEPTION, version=abi/5.7.0)
  at Error: call revert exception [ See:  (/v5-errors-CALL_EXCEPTION%20]%20(method=%22balanceOf(address)%22,%20data=%220x%22,%20errorArgs=null,%20errorName=null,%20errorSignature=null,%20reason=null,%20code=CALL_EXCEPTION,%20version=abi/5.7.0))
  at Generator.next (<anonymous>)
...
(2 additional frame(s) were not displayed)
```

macqbat:
# Related issues 🔗

Closes #2581 

# Description ℹ️

Limit sentry logs in cases when it seems to be not relevant, because error exists on ethereum side.

# Demo 📺

-

# Technical 👨‍🔧

-


libs/withdraws/src/lib/use-complete-withdraw.ts
```diff
@@ -55,7 +55,12 @@ export const useCompleteWithdraw = () => {
           approval.signatures
         );
       } catch (err) {
-        captureException(err);
+        const logger = localLoggerFactory({ application: 'deposits' });
```
mattrussell36, line 58: I don't think this helps solve the problem, it only moves the sentry error to a sentry message.

Instead lets try and capture more information on how we are calling the smart contracts. Then we can check that we are calling it with the right arguments (which is a likely culprit)

Lets do the following:

A) call logger.info with the supplied arguments
B) just call logger.error in the catch. We should be doing this more overall in the application so we aren't pulling Sentry directly into modules.

```
const submit = useCallback(() => {
  ...
  try {
    ...
    logger.info('calling withdraw_asset', ...args)
   } catch (err) {
     logger.error(err)
   }
 }, [])
```

mattrussell36 (changes requested):

libs/withdraws/src/lib/use-complete-withdraw.ts
macqbat: ok, I've pushed such a changes. Please take a look

PR #3740, opened 2023-05-11
fix(governance): fix time delay notice for withdrawal of 0.00 assets

### LINKED ISSUE #3553: Time delay for withdrawal of 0.00 assets
## Description
If user fills in the withdrawal form and the asset has a time delay for number of assets below 0.00, the following message is shown in the screen shot.

At a glance it looks like there's a delay withdrawing 0 tokens which doesn't look right. 

## Expected behaviour
If number of tokens for delay is the lowest token value then text should be:
"Withdrawals of any xxx will be delayed for xxx"

else if number of tokens is above the lowest value but below 0.00, then the value of 0.00 should be mouse hoverable (tooltip) to show the "real" number of tokens for delay

Screenshot 2023-04-26 at 15 01 09

sam-keen:
# Related issues 🔗

Closes #3553

# Description ℹ️

The withdrawal form message was configured to show a different message if the threshold for a withdrawal delay was 0, but it wasn't working. Subsequently, you'd see the scenario detailed in #3553.

### DIFF
--- a/libs/withdraws/src/lib/withdraw-form.tsx
+++ b/libs/withdraws/src/lib/withdraw-form.tsx
@@ -70,13 +70,14 @@ const WithdrawDelayNotification = ({
   return (
     <Notification
       intent={Intent.Warning}
+      key={symbol}
       testId={
-        threshold.isFinite()
-          ? 'amount-withdrawal-delay-notification'
-          : 'withdrawals-delay-notification'
+        threshold.isEqualTo(0)
+          ? 'withdrawals-delay-notification'
+          : 'amount-withdrawal-delay-notification'
       }
       message={[
-        !threshold.isFinite()
+        threshold.isEqualTo(0)
           ? t('All %s withdrawals are subject to a %s delay.', replacements)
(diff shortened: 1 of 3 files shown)


libs/withdraws/src/lib/withdraw-form.tsx
```diff
@@ -70,13 +70,14 @@ const WithdrawDelayNotification = ({
...
       testId={
-        threshold.isFinite()
-          ? 'amount-withdrawal-delay-notification'
-          : 'withdrawals-delay-notification'
+        threshold.isEqualTo(0)
+          ? 'withdrawals-delay-notification'
+          : 'amount-withdrawal-delay-notification'
       }
       message={[
-        !threshold.isFinite()
```
bglownia, line 79: threshold is set to Infinity if it equals 0

https://github.com/vegaprotocol/frontend-monorepo/blob/develop/libs/web3/src/lib/use-get-withdraw-threshold.tsx#L28

So in theory it should work fine

PR #3755, opened 2023-05-12
fix(governance): avoid unknown pubkey causing validator page error

### LINKED ISSUE #3742: Can't view validator details page when Vega wallet is connected with a key not known to the network
## Steps to Reproduce

1. Go to [governanve](https://governance.vega.xyz/validators)
2. Connect a Vega wallet and select/switch to a key with no stake and no transactions on Vega
3. Click on a validator's name to open the details page
4. Sad times ("Something went wrong")

## Expected behavior

Validator details

## Screenshots

Screenshot 2023-05-11 at 17 40 05

## Device details

- OS: macOS Ventura
- Browser: Firefox

sam-keen:
# Related issues 🔗

Closes #3742

# Description ℹ️

If a pubkey hasn't interacted with the chain before, a gql query to `party` will cause it to error, which happens on validator pages, which then display the error.

I've added a check for the particular error the API returns in that case, and instead of rendering the error, I call a refetch with an empty pubkey, which returns null for `party` but gives us the rest of the required data.

### DIFF
--- a/apps/governance/src/routes/staking/node/nodes-container.tsx
+++ b/apps/governance/src/routes/staking/node/nodes-container.tsx
@@ -1,3 +1,4 @@
+import { useEffect, useState } from 'react';
 import { ENV } from '../../../config';
 import { Callout, Intent, Splash } from '@vegaprotocol/ui-toolkit';
 import { useVegaWallet } from '@vegaprotocol/wallet';
@@ -28,6 +29,7 @@ export const NodeContainer = ({
   const { t } = useTranslation();
   const { pubKey } = useVegaWallet();
   const { delegationsPagination } = ENV;
+
   const { data, loading, error, refetch } = useStakingQuery({
     variables: {
       partyId: pubKey || '',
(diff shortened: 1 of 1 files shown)


apps/governance/src/routes/staking/node/nodes-container.tsx
```diff
@@ -38,6 +40,20 @@ export const NodeContainer = ({
         : undefined,
     },
   });
+
+  const [isRefetching, setIsRefetching] = useState(false);
+
+  useEffect(() => {
```
mattrussell36, line 46: This isn't the right approach. If the query errors due to party not found... I'd consider that not an error for the user and just ignore it.

```
if (error && !error.message.includes('failed to get party fo ID')) {
  // render error ui
}
```

You may want an additional hook or some or a utility function if we commonly want to ignore this error. We might also want to not report this error to Sentry.

mattrussell36 (changes requested):

apps/governance/src/routes/staking/node/nodes-container.tsx
sam-keen: @mattrussell36 I think it might be the right approach. If the API errors, we never get a data object. We need to retry

sam-keen: For posterity: the issue was Apollo's default policy for not returning data if there's an error. Setting `errorPolicy: 'all'` in the staking query allowed data to be passed to the children and the error to be ignored

mattrussell36 (approved):

PR #3776, opened 2023-05-15
feat(trading): sidebar market list

### LINKED ISSUE #3457: Replace Market switcher drop down with a markets side bar and remove the first use markets modal
Dev depends on https://github.com/vegaprotocol/frontend-monorepo/issues/3587

## The Chore

Following on from [this spike about the target market list experience](https://github.com/vegaprotocol/frontend-monorepo/issues/3327)

- Create a new draw/side panel that lists markets (in a compact form, e.g. cards)
- Remove the existing mega drop down and have this one replace it
- Remove the select a market to get started modal and have the draw be open on first load
- The draw contains a link to "all markets" (`/Markets`)
- The draw will contain some quick filters
- The draw will contain a way to switch between different lists (e.g. top movers, new markets, top markets, these lists to be defined later), in the first iteration it could default to the same as the existing mega drop down)
- The landing dialog market list is removed
- Market draw is open by default when I land on Console  
- Console remembers the last state of the draw i.e. if the user closes it, when he next opens console the draw is closed

### Detail

- The draw can be pinned opened or closed so that the user can monitor markets A, B and C whilst looking at D
- The market you are viewing should appear selected in the list

- For each market:
  - Show `Code` of the market as the primary header
  - Show the `Name` of the market
  - Show a single price for the market (`last traded` or `mark price`)
  - show the % price change of the market (using the existing method, should be 24 hours 0pen ->Close )
  - Show a sparkline using the one from the existing dropdown
  - if the market is in auction: treatment to show things are not "normal"

## Tasks

- [x] Sketch

Image

Image

from [here](https://www.figma.com/file/JZQ1d6etF7yzOb94N4K0nS/Markets-list?node-id=0%3A1&t=NPd1ilaUapfW3dUt-1)

There is also a draft for the [Acs here](https://github.com/vegaprotocol/specs/blob/user-interface/MARK-starting-content/user-interface/6001-MARK-find_markets.md)

- [ ] Visual design
- [ ] stakeholder review
- [ ] build it
- [ ] tests and Acs 

David is doing something similar with the new landing page for the website so might be worth aligning with him

mattrussell36:
# Related issues 🔗

Closes #3457 

# Description ℹ️

Replaces the market popover with a togglable sidebar containing a virtualized list of markets which can be filtered and sorted

# Demo 📺

Screenshot 2023-05-15 at 16 46 50


MadalinaRaicu (approved):

daro-maj (changes requested):
Tests are failing because in tab All we can't use filter because they grey-out. Looks like issue in this pr because on develop works fine.

daro-maj (approved):

PR #3793, opened 2023-05-16
feat(governance): proposal terms in table, full proposal json, description …

### LINKED ISSUE #3779: SPIKE: how much proposal detail can we render nicely and quickly
Get the JSON into tables as nicely as possible without custom conditional rendering

sam-keen:
…collapsible, layout tweak

# Related issues 🔗

Closes #3779

# Description ℹ️

Request for proposals to have their 'terms' in tables rather than just in JSON format. 
Puts the full proposal into JSON format
Puts the description into a collapsible block

Note that this is a last minute request, so the component that deals with terms isn't particularly refined. It recursively iterates over data (the depth of nesting changes dependent on proposal type) and renders it into a fairly readable format, that will likely be tweaked in the coming weeks. 

Screenshot 2023-05-16 at 20 47 30

Screenshot 2023-05-16 at 15 20 34

Screenshot 2023-05-16 at 15 12 55


apps/governance/src/routes/proposals/components/proposal-terms/proposal-terms.tsx
```diff
@@ -0,0 +1,132 @@
...
+    <KeyValueTableRow>
+      {key}
+      {value}
+    </KeyValueTableRow>
+  </KeyValueTable>
+);
+
+const renderArray = (key: string, array: unknown[]) => {
+  if (array.every((item) => typeof item === 'string')) {
+    return renderKeyValue(key, array.join(', '));
```
mattrussell36, line 47: suggestion (non blocking): Isn't this just a component? Shouldn't it be upper cased and called with angle brackets?

apps/governance/src/routes/proposals/components/proposal-terms/proposal-terms.tsx
```diff
@@ -0,0 +1,132 @@
...
+            <ProposalTermsRenderer data={item as Record<string, unknown>} />
+          </div>
+        ))}
+      </div>
+    );
+  }
+};
+
+// Working with 'unknown' type as a proposal's terms can be in many shapes
+const renderTerm = (key: string, value: unknown) => {
```
mattrussell36, line 63: Same here. Not sure what the reasoning is here, but for me functions that end up returning JSX\HTML should be denoted as such by being called with `<MyComponent />` rather than `myComponent()`

mattrussell36 (approved):

apps/governance/src/routes/proposals/components/proposal-terms/proposal-terms.tsx
sam-keen: Yes it should - this function/component went through a couple of changes to it's functionality and that got overlooked, as it did with the others. Thanks

PR #3813, opened 2023-05-17
fix(governance): clear wallet state if user changes to a non existent party

### LINKED ISSUE #3085: Switching Vega keys does not update the page or wallet side bar until page reload.
## Description

When I use the switch key feature I expect the side bar and all other content to update to my newly selected key, however, it does not.

I can reproduce this only when switching from a key that has a large number of assets, associated tokens to etc to one that doesn't. see the video, in the first section it doesn't switch then I reload and it does. then I can switch from my second key to the first without issue

https://user-images.githubusercontent.com/13255539/222774676-c1cce271-8acc-4616-9e3b-97b3de6f010e.mov

If I refresh the app, then it does register the new key I had selected.

mattrussell36:
# Related issues 🔗

Closes #3085 

# Description ℹ️

- If query errors check, if its a party not found error, if it is clear wallet state
- de-dupe magic strings for error checking

### DIFF
--- a/apps/governance/src/components/vega-wallet/hooks.ts
+++ b/apps/governance/src/components/vega-wallet/hooks.ts
@@ -26,6 +26,7 @@ import type {
   WalletDelegationFieldsFragment,
 } from './__generated__/Delegations';
 import { DelegationsDocument } from './__generated__/Delegations';
+import { isPartyNotFoundError } from '../../lib/party';
 
 export const usePollForDelegations = () => {
   const { token: vegaToken } = useContracts();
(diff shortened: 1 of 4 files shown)


apps/governance/src/components/vega-wallet/hooks.ts
```diff
@@ -75,6 +76,9 @@ export const usePollForDelegations = () => {
           })
           .then((res) => {
             if (!mounted) return;
+
+            console.log(res.error);
```
sam-keen: Chore (blocking): remove console.log

sam-keen (approved):

PR #3817, opened 2023-05-18
feat(ci): add pipeline for rollback console deployments

mkjmdski:
closes: https://github.com/vegaprotocol/devops-infra/issues/1880

### DIFF
--- a/.github/workflows/rollback-console.yml
+++ b/.github/workflows/rollback-console.yml
@@ -11,15 +11,8 @@ on:
 jobs:
   rollback:
     runs-on: ubuntu-22.04
-    timeout-minutes: 25
+    timeout-minutes: 10
     steps:
-      - name: Set up QEMU
-        id: quemu
-        uses: docker/setup-qemu-action@v2
-
-      - name: Available platforms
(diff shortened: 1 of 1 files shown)


.github/workflows/rollback-console.yml
```diff
@@ -0,0 +1,84 @@
...
+    inputs:
+      version:
+        description: 'Version that should be set on rollback'
+        required: true
+        type: string
+
+jobs:
+  rollback:
+    runs-on: ubuntu-22.04
+    timeout-minutes: 25
```
edd, line 14: **Question (non-blocking)**: why so long? Shouldn't 5 minutes be sufficient?

.github/workflows/rollback-console.yml
```diff
@@ -0,0 +1,84 @@
...
+    steps:
+      - name: Set up QEMU
```
daniel1302, line 16: Do we need Quemu for that? We do not build multi-platform here. If we have multiplatform docker image We should retag all of them. Docker does not support multiplatform itself does it? I think it is required to use dockerx for multiplatform. But for docker we also do not need quemu

PR #3821, opened 2023-05-18
feat(trading): do not show oracle banner when data is loading

bglownia:
# Description ℹ️

Change source of data source specs from market info query to markets query in useMarketOracle hook.
If data needed in oracle banner is loading return null instead of no oracle notification

### DIFF
--- a/libs/markets/src/lib/components/oracle-banner/oracle-banner.tsx
+++ b/libs/markets/src/lib/components/oracle-banner/oracle-banner.tsx
@@ -6,30 +6,20 @@ import {
   NotificationBanner,
   ButtonLink,
 } from '@vegaprotocol/ui-toolkit';
-import { OracleDialog } from '../market-info';
-
-export const oracleStatuses = {
-  UNKNOWN: t(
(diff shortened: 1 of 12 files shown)


libs/markets/src/lib/components/oracle-banner/oracle-banner.tsx
```diff
@@ -6,30 +6,20 @@ import {
...
-  );
+  const { data: settlementOracle, loading: settlementOracleLoading } =
+    useMarketOracle(marketId);
+  const {
+    data: tradingTerminationOracle,
+    loading: tradingTerminationOracleLoading,
+  } = useMarketOracle(marketId, 'dataSourceSpecForTradingTermination');
+  if (settlementOracleLoading || tradingTerminationOracleLoading) {
+    return null;
+  }
```
MadalinaRaicu, line 22: Please let's remove `<div>{t('There is no oracle for this market yet.')} </div>` condition altogether

MadalinaRaicu (approved):

PR #3826, opened 2023-05-18
fix(orders): cancel all and pegged orders amendable

### LINKED ISSUE #3684: Make 'cancel all' cancel  everything for all markets
## The Chore

The cancel all button should cancel all orders, regardless of the current market

## Tasks

- [ ] Verify if we need a confirmation dialog to warn about closing all
- [ ] Remove marketId from the cancel all orders transaction so that all orders are called
- [ ] Remove cancel all button from closed, rejected and all order tabs. We only need cancel all on the active orders list.

MadalinaRaicu:
# Related issues 🔗

Closes #3684 
# Description ℹ️

Cancel all and pegged orders amendable

# Demo 📺

Screenshot 2023-05-18 at 16 05 28

# Technical 👨‍🔧

### DIFF
--- a/libs/orders/src/lib/components/order-list/order-list.tsx
+++ b/libs/orders/src/lib/components/order-list/order-list.tsx
@@ -329,7 +329,7 @@ export const isOrderActive = (status: Schema.OrderStatus) => {
 };
 
 export const isOrderAmendable = (order: Order | undefined) => {
-  if (!order || order.peggedOrder || order.liquidityProvision) {
+  if (!order || order.liquidityProvision) {
     return false;
   }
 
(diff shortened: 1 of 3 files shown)


libs/orders/src/lib/components/order-list/order-list.tsx
```diff
@@ -329,7 +329,7 @@ export const isOrderActive = (status: Schema.OrderStatus) => {
 };
 
 export const isOrderAmendable = (order: Order | undefined) => {
-  if (!order || order.peggedOrder || order.liquidityProvision) {
+  if (!order || order.liquidityProvision) {
```
MadalinaRaicu, line 332: https://vegaprotocol.slack.com/archives/CFXA4EGN7/p1684421782550409

PR #3827, opened 2023-05-18
feat(governance): use rest to expose proposal data

dexturr:
# Related issues 🔗

TBC

# Description ℹ️

Show JSON based on REST api as Graphql API is missing some data


edd (approved):
Yuck but nicely constrained yuck

PR #3864, opened 2023-05-21
chore(governance): remove process tranches script

### LINKED ISSUE #3002: Remove mainnet tranches workflow
This is something that needs to be done after next mainnet release but can be ignored until then

## The Chore

Remove the mainnet tranches JSON file, workflow files and script for generating tranches.

mattrussell36:
# Related issues 🔗

Closes #3002 

# Description ℹ️

Removes the process tranches scripts as its no longer required since upgrading mainnet to 0.71.4


dexturr (changes requested):
You can also delete all the tranches json files in here too:

apps/static/src/assets

PR #3870, opened 2023-05-22
fix(trading): price and volume in the last 24h can be incorrectly reported

### LINKED ISSUE #3476: 24hr price and volume can be incorrectly reported
## Description

If there was a network upgrade or any reason that there might be a gap in candles we may report a 24hr price or volume change incorrectly. 

This is because candles are fetched using a 'since' date and if a network upgrade has occurred after the since date then less than 24 candles might be returned. The code takes the first and last candles from the returned data which means we end up displaying the change over less than a 24hr period (the actual period being determined by whatever number of candles are returned)

We need to actually check the timestamp of each candle (and possibly the total candle length) to determine what to show.

## Acceptance criteria

- [ ] We don't incorrectly report data for a 24hr period, either updating the labeling accordingly or by rendering a placeholder (usually we use a '-') 

## Proposed solution

- Increase `since` date for candles query to be for 5 days (rather than the current 1 day)
- Add logic that finds the candle that is 24h ago
- If there is no candle with the correct timestamp, use the next oldest candle.
- If there is no candle older than the correct timestamp, do not show a 24 hour change value. In its place, write "Unknown" and add a tooltip to explain.
- The tooltip should also include the price change DO know about, and the length of time over which that change occurred. 

## Sketch

image

MadalinaRaicu:
# Related issues 🔗

Closes #3476 

# Description ℹ️

Price and volume in the last 24h can be incorrectly reported.

- [x] show `Unknown`
- [x] and a tooltip with the price change in the past five days

# Demo 📺

image

Screenshot 2023-05-22 at 14 19 38

Screenshot 2023-05-22 at 14 19 42

# Technical 👨‍🔧


libs/react-helpers/src/hooks/use-yesterday.ts
```diff
@@ -10,13 +11,23 @@ export const now = (roundBy = 1) => {
 /**
  * Returns the yesterday's timestamp rounded by given number (in milliseconds; 5 minutes by default)
  */
-export const useYesterday = (roundBy = DEFAULT_ROUND_BY_MS) => {
-  const yesterday = useRef<number>(now(roundBy) - TWENTY_FOUR_HOURS_MS);
+export const useYesterday = (
```
asiaznik, line 14: chore (blocking): As this one is now more generic than initially designed I'd rename it to `useAgo` and also make it return a function `(roundBy) => number`. Having that I'd remake `useYesterday` and `useFiveDaysAgo` as:
```
export const useYesterday = useAgo(TWENTY_FOUR_HOURS_MS);
export const useFiveDaysAgo = useAgo(FIVE_DAYS_MS);
```

asiaznik (approved):

libs/markets/src/lib/components/last-24h-price-change/last-24h-price-change.tsx
```diff
@@ -24,30 +27,75 @@ export const Last24hPriceChange = ({
...
+    candles
       ?.map((candle) => candle?.close)
       .filter((c): c is CandleClose => c !== null) || initialValue;
 
+  if (
+    fiveDaysCandles &&
+    fiveDaysCandles.length > 0 &&
+    (!oneDayCandles || oneDayCandles?.length === 0)
+  ) {
+    console.log('data', data);
```
MadalinaRaicu, line 57: to remove this and align

PR #3886, opened 2023-05-22
chore(trading): dont show upgrade banner unless APP_VERSION is set

mattrussell36:
# Related issues 🔗

N/A

# Description ℹ️

- Only renders the upgrade banner if NX_APP_VERSION is set
- Fixes color of copy icon in banner


MadalinaRaicu (approved):

edd (approved):
**thought (non-blocking)**: I am too lazy to check that a ENV that isn't set === null - I'd have been extra careful and checked for `string` && `string.length`

But I suspect you checked that

PR #3944, opened 2023-05-25
fix(trading): performance issues with many orders

### LINKED ISSUE #3788: Performance issues when you have lots of orders
## Description

The app eventually crashes if you have a lot of orders.

## Steps to Reproduce

1. Go to console
2. Connect with a pub key with lots of orders (eg a market maker)
3. Wait a few seconds
4. Notice that the app has slowed or crashed

## Expected behavior

Should be able to handle lots of orders

## Additional context

https://vegaprotocol.slack.com/archives/C037STCKMGC/p1684248771348679

mattrussell36:
# Related issues 🔗

Closes #3788

# Description ℹ️

Improves performance of the app when connected as a party with lots of orders.

# Technical 👨‍🔧

- Removes top level subscription for all orders, only subscribing to active orders
- Prevents caching of OrderUpdate types
- Prevents caching of transaction update subscription data
- Prevents caching of estimateFees and estimatePosition queries

### DIFF
--- a/apps/trading/components/app-loader/app-loader.tsx
+++ b/apps/trading/components/app-loader/app-loader.tsx
@@ -59,7 +59,7 @@ const cacheConfig: InMemoryCacheConfig = {
       keyFields: false,
     },
     TradableInstrument: {
-      keyFields: ['instrument'],
+      keyFields: false,
     },
     Product: {
       keyFields: ['settlementAsset', ['id']],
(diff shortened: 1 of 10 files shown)


mattrussell36:
@bglownia I don't think we should be adding these bigger changes onto this PR. Can you branch off to save your work and revert this one so we can get the perf improvements merged (even if we still have an issue with the subscription)

bglownia (approved):

PR #3969, opened 2023-05-25
chore(trading): update vega.trading to console.vega.xyz

MadalinaRaicu:


edd (approved):
Thanks for catching these references I missed in #3962

PR #3985, opened 2023-05-30
fix: market info refactor and update test

### LINKED ISSUE #3791: Refactor market info accordion
## The Chore

The current market info accordion will render all sections and requires all market data to be present for it to work. 

Ideally the accordion would be able to be composed together using different components to construct which parts of the data you would like to show. This would allow it to be used more easily in different apps

## Tasks

- [ ] Refactor market info accordion
- [ ] Check with governance team to ensure its suitable for the governance app
- [ ] Test

https://vegaprotocol.slack.com/archives/C037STCKMGC/p1684257295525009

MadalinaRaicu:
# Related issues 🔗

Closes #3791 

# Description ℹ️

Market details refactor. 

# Demo 📺

# Technical 👨‍🔧

### DIFF
--- a/apps/explorer/src/app/components/markets/market-details.tsx
+++ b/apps/explorer/src/app/components/markets/market-details.tsx
@@ -1,5 +1,6 @@
 import { t } from '@vegaprotocol/i18n';
 import type { MarketInfoWithData } from '@vegaprotocol/markets';
+import { PriceMonitoringBoundsInfoPanel } from '@vegaprotocol/markets';
 import {
   LiquidityInfoPanel,
   LiquidityMonitoringParametersInfoPanel,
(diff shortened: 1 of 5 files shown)


apps/explorer/src/app/components/markets/market-details.tsx
```diff
@@ -39,133 +39,79 @@ export const MarketDetails = ({ market }: { market: MarketInfoWithData }) => {
...
-        title: t(`Price monitoring trigger ${i + 1}`),
-        content: <MarketInfoTable noBorder={false} data={trigger} />,
-      })
-    ),
-    ...(market.data?.priceMonitoringBounds || []).map((trigger, i) => ({
-      title: t(`Price monitoring bound ${i + 1}`),
-      content: (
+  return (
+    <div>
+      <KeyDetailsInfoPanel noBorder={false} showTitle={true} market={market} />
```
mattrussell36, line 49: chore (blocking): boolean props such as noBorder and showTitle are a bit of a code smell to me. Rather than passing a prop just to hide show something why not pass the title it self. Its much more declarative and obvious from the component consumer what the behaviour is then.

mattrussell36 (changes requested):

apps/explorer/src/app/components/markets/market-details.tsx
MadalinaRaicu: agree. addressed your comments. could you take another look?

libs/markets/src/lib/components/market-info/market-info-accordion.tsx
```diff
@@ -85,6 +85,25 @@ export const MarketInfoAccordion = ({
     market.accountsConnection?.edges
   );
 
+  const settlementData = market.tradableInstrument.instrument.product
```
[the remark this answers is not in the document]
MadalinaRaicu: based on examples from testnet, there is a possibility they're going to have only one oracle for both settlement and trading termination

mattrussell36 (approved):

Issue #3981, opened 2023-05-27
24h volume inconsistent across different places in console


davidsiska-vega:
## Description

The 24h volume seems inconsistent across different places in console.

In the top right of the screenshot (part of the market view) it seems to update continuously as trades happen (a subscription I guess). 

In the market picker on the left it updates on console reload (so it is consistent initially but then they diverge). 

## Screenshots

<img width="836" alt="image" src="https://github.com/vegaprotocol/frontend-monorepo/assets/62546419/8146ab45-31aa-4a48-b9b7-3d1c64f629cb">

PR #3989, opened 2023-05-30
fix(trading): market selector volume update

### LINKED ISSUE #3476: 24hr price and volume can be incorrectly reported

### LINKED ISSUE #3987: Market volume in market selector does not match volume in trade header and market info
## Problem 
 Console says volume on BTC/USDT is 19.62 on the market info and on the header section 
... but 3.64 on the market selector tile on the left
Also what units are any of this measured in?  What is the actual correct volume?

Image

MadalinaRaicu:
# Related issues 🔗

Closes #3476 #3981 

# Description ℹ️

Market volume was not correctly calculated without periodStart. 

# Demo 📺

Screenshot 2023-05-30 at 11 41 29

# Technical 👨‍🔧


libs/markets/src/lib/markets-provider.ts
```diff
@@ -154,13 +154,13 @@ export const marketListProvider = makeDerivedDataProvider<
 );
 
 export const useMarketList = () => {
-  const yesterday = useYesterday();
+  const fiveDaysAgo = useFiveDaysAgo();
```
MadalinaRaicu, line 157: when a network upgrade happens, we might not get the last 24h candles accurately. 
it'd be good to fetch last 5 days candles and then filter which candles are from 24h period.

apps/trading/client-pages/market/market-selector-item.tsx
```diff
@@ -106,25 +103,34 @@ const MarketData = ({ market }: { market: MarketMaybeWithDataAndCandles }) => {
           </p>
         )}
       </div>
-      <DataRow value={volume} label={t('24h vol')} />
+      <DataRow
+        value={
+          <Last24hVolume
```
mattrussell36, line 108: suggestion (blocking): Lets not re-use this component as the design is different. Instead refactor so that the biz logic can be shared by a hook that the header-stat and market-selector-item can use to retrieve the formatted data but render different UI

mattrussell36 (changes requested):

libs/markets/src/lib/components/last-24h-volume/last-24h-volume.tsx
```diff
@@ -72,13 +54,14 @@ export const Last24hVolume = ({
           </div>
         }
       >
-        <span ref={ref}>{t('Unknown')} </span>
+        <span ref={ref}>{'-'} </span>
```
MadalinaRaicu, line 57:
```suggestion
        <span ref={ref}>-</span>
```

apps/trading/client-pages/market/market-selector-item.spec.tsx
```diff
@@ -17,14 +17,29 @@ import {
 } from '@vegaprotocol/types';
 import { addDecimalsFormatNumber } from '@vegaprotocol/utils';
 
+jest.mock('@vegaprotocol/markets', () => ({
```
mattrussell36, line 20: There is a MockedProvider set up for this test. We can gain more value from this test if you create a mock for the query used in marketCandlesProvider and pass it to the mocks array

mattrussell36 (changes requested):
This is looking really nice. But I've left some comments for things that should be changed

libs/markets/src/lib/hooks/use-candles.ts
```diff
@@ -0,0 +1,43 @@
+import { useThrottledDataProvider } from '@vegaprotocol/data-provider';
```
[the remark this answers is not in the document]
MadalinaRaicu: added

libs/markets/src/lib/components/last-24h-price-change/last-24h-price-change.tsx
```diff
@@ -27,28 +22,10 @@ export const Last24hPriceChange = ({
   inViewRoot,
 }: Props) => {
   const [ref, inView] = useInView({ root: inViewRoot?.current });
```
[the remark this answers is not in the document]
MadalinaRaicu: updated

libs/markets/src/lib/hooks/use-candles.ts
```diff
@@ -0,0 +1,43 @@
...
+import { useFiveDaysAgo, useYesterday } from '@vegaprotocol/react-helpers';
+import type { MarketCandlesFieldsFragment } from '../__generated__';
+import { marketCandlesProvider } from '../market-candles-provider';
+import { Interval } from '@vegaprotocol/types';
+
+export const useCandles = ({
+  marketId,
+  inView,
```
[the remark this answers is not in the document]
MadalinaRaicu: updated

apps/trading/client-pages/market/market-selector-item.spec.tsx
MadalinaRaicu: I see, I mocked useCandles hook here, and added tests to useCandles. 

I had some trouble trying to mock this, it expects data undefined on mocked response, not sure what I'm missing. 
image

mattrussell36: @MadalinaRaicu for the query to successfully resolve within the test you need to ensure that variables (and return data) defined in the mock _exactly_ match what is provided to the query in the code. 

Here, it looks like you might have modified what variables get passed to the data provider, but internally the data provider will still be calling the _actual query_ with `marketId`, `interval` and `since`, so you'll need to update the `MockedResponse` to use the correct variables.

Have a read here: https://www.apollographql.com/docs/react/development-testing/testing/#defining-mocked-responses

MadalinaRaicu: I have tried to mock both the query and the subscription from marketCandlesDataProvider 

image


but since this test is from trading 
I can't manage to not mock useCandles hook which is part of `@vegaprotocol/markets` lib 
so the 24h volume would be 0.00

mattrussell36: Okay two things 

1. The variables and return type must exactly match. The interval variable here is 1D when actually the query uses 1H, so that would've resulted in the mock query not 'matching' the query used in the test.
2. Any query which uses a date as a variable can be tricky as the date timestamp must match between the mock and what actually runs in the test, so often its necessary to mock the date constructor so that there isn't a delay between when the test runner creates the date and when the code under test creates the date. 

Have a look at the latest commit. I've added a mock for the date creation to ensure whenever you call Date.now you get the same date. Secondly I've updated the variables so that hourly intervals are used.

We now have the added benefit that we are
A) also inherently testing the useCandles hook
B) also testing the date provider

mattrussell36 (approved):

PR #3990, opened 2023-05-30
chore(react-helpers,utils,logger): create logger lib and move sentry/logger utils there

ciaran-:
# Description ℹ️

Creates a 'logger' lib and move logger/sentry utils there from utils package. 
Removes `useLogger` from react-helpers

We want to avoid Sentry as a dependency for published libs - See [this thread](https://vegaprotocol.slack.com/archives/C8WS7M50E/p1684837078128609) for why we're doing this.


libs/react-helpers/.eslintrc.json
```diff
@@ -7,15 +7,27 @@
...
+              "ag-grid-community",
+              "graphql",
+              "graphql-tag",
+              "graphql-ws",
+              "ethers",
+              "@ethersproject"
+            ]
+          }
+        ],
+        "no-restricted-imports": [
```
mattrussell36, line 25: Can the sentry pattern not be combined with the dupe rule above?

mattrussell36 (approved):

libs/react-helpers/src/hooks/use-mutation-observer.ts
```diff
@@ -34,7 +33,8 @@ export function useMutationObserver(
     try {
       observer.observe(target, options.config);
     } catch (err) {
-      captureException(err);
+      console.error("Can't observe mutation target", err);
```
mattrussell36, line 36: I think lets just remove the entire try catch, we'll let Sentry pick up and log the error

libs/react-helpers/src/hooks/use-resize-observer.ts
```diff
@@ -38,7 +37,7 @@ export function useResizeObserver(
     try {
       observer.observe(target, options.config);
     } catch (err) {
-      localLoggerFactory({ application: 'react-helpers' }).debug(err as Error);
+      //ignore error for now - we don't want to use the logger here
```
mattrussell36, line 40: Lets remove the entire try catch and let Sentry pick up the error

PR #3993, opened 2023-05-30
fix(withdraws): minimal withdrawal amount validation

### LINKED ISSUE #3915: Use quantum for minimum withdrawal amount
## The Chore

The withdrawal form will let you attempt to withdraw the minimum viable denomination for an asset. However, the chain will reject such a withdrawal if it is less than the assets quantum value. We need to update the form to prevent submission of withdrawals less than the quantum value.

## Tasks

- [ ] Update form validation to account for quantum
- [ ] Update tests accordingly
- [ ] Test

asiaznik:
# Related issues 🔗

Closes #3915 

# Description ℹ️

Asset's quantum is minimal amount eligible to withdraw, but the de-facto value is quantum * `spam.protection.minimumWithdrawalQuantumMultiple`. This PR adds the factor to the calculation of the minimal amount.

### DIFF
--- a/libs/withdraws/src/lib/use-withdraw-asset.tsx
+++ b/libs/withdraws/src/lib/use-withdraw-asset.tsx
@@ -9,7 +9,7 @@ import {
   useGetWithdrawThreshold,
 } from '@vegaprotocol/web3';
 import { useWithdrawStore } from './withdraw-store';
-import { useNetworkParams } from '@vegaprotocol/network-parameters';
+import { useNetworkParam } from '@vegaprotocol/network-parameters';
 
 export const useWithdrawAsset = (
   assets: Asset[],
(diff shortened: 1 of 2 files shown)


mattrussell36 (commented):
@asiaznik Looks good. Can you update the specs repo to capture that the minimum withdrawal should use `spam_protection_minimumWithdrawalQuantumMultiple` and ensure we add test coverage for this

asiaznik:
> @asiaznik Looks good. Can you update the specs repo to capture that the minimum withdrawal should use `spam_protection_minimumWithdrawalQuantumMultiple` and ensure we add test coverage for this

https://github.com/vegaprotocol/specs/pull/1725

daro-maj (approved):

PR #4015, opened 2023-06-01
feat(trading): do not merge orders with markets

### LINKED ISSUE #4004: Maintain market lookup in memory

bglownia:
# Related issues 🔗

Closes #4004


libs/markets/src/lib/components/markets-data/markets-data.tsx
```diff
@@ -0,0 +1,21 @@
+import { useMarketNamesMap } from '../../hooks/use-market-names-map';
+import { useMarketsMap } from '../../hooks/use-markets-map';
+import { useDataProvider } from '@vegaprotocol/data-provider';
+import { marketsProvider } from '../../markets-provider';
+
+export const MarketsData = () => {
+  const updateMarketNames = useMarketNamesMap((state) => state.update);
```
mattrussell36, line 7: Lets just have a single store for all **static** market data. I don't see any advantage to having two stores one for names and one for everything else. It just creates a little more overhead when starting up the app

mattrussell36: Additionally I think we might want to wrap our app (or the page) with these start up queries and only render once they are complete. This will help simplify things by allowing us to know for certain that once we are deeper in the component tree, the market lookup _will exist_

libs/orders/src/lib/components/order-list/order-list.spec.tsx
```diff
@@ -23,6 +23,17 @@ jest.mock('@vegaprotocol/react-helpers', () => ({
   }),
 }));
 
+jest.mock('@vegaprotocol/markets', () => ({
```
mattrussell36, line 26: I feel like this is going to become painful if we need to mock this for every test that will consume the markets map. Perhaps we can use a __mocks__ folder to mock this everywhere at once. Or maybe using context will be a cleaner solution

mattrussell36 (commented):
This is going to make data management much easier. Still some things I think we need to address though

libs/markets/src/lib/components/markets-data/markets-data.tsx
```diff
@@ -0,0 +1,21 @@
+import { useMarketNamesMap } from '../../hooks/use-market-names-map';
+import { useMarketsMap } from '../../hooks/use-markets-map';
+import { useDataProvider } from '@vegaprotocol/data-provider';
+import { marketsProvider } from '../../markets-provider';
+
+export const MarketsData = () => {
```
mattrussell36, line 6: I'd also add that I think we should just keep MarketsData and the useMarketsMap hook in the trading app. And not add it to libs yet

PR #4016, opened 2023-06-01
feat(trading): 3945 orderbook enhancements

### LINKED ISSUE #3945: Orderbook changes
## The Chore

Changes to the order book to improve experience

## Tasks

- [x] Never show price levels with 0 orders (don't allow switching between modes)
- [x] Remove scrolling
- [x] Ensure the mid is centered
- [x] Show a single column for bid/ask volume
- [x] Improve aggregation to allow for more and less precision. Use quantum and asset decimals to determine min/max precision
- [x] Remove column headers
- [x] Prices should be vega-green for bids and vega-pink for asks
- [x] Ensure columns are price, volume, cumulative volme (from left to right)
- [x] Remove grid lines
- [x] Reduce padding within cells
- [ ] Don't show cumulative volume unless wide enough
- [x] Show as many rows as that can fit within each side

macqbat:
# Related issues 🔗

Closes #3945 

# Description ℹ️

Improve view and logic of aggregation

# Demo 📺

image

# Technical 👨‍🔧

-


libs/market-depth/src/lib/orderbook.tsx
```diff
@@ -477,157 +61,119 @@ export const Orderbook = ({
...
   ) : null;
 
-  const c = theme === 'dark' ? colors.neutral[600] : colors.neutral[300];
-  const gradientStyles = isContinuousMode
-    ? `linear-gradient(${c},${c}) 33.4% 0/1px 100% no-repeat, linear-gradient(${c},${c}) 66.7% 0/1px 100% no-repeat`
-    : `linear-gradient(${c},${c}) 25% 0/1px 100% no-repeat, linear-gradient(${c},${c}) 50% 0/1px 100% no-repeat, linear-gradient(${c},${c}) 75% 0/1px 100% no-repeat`;
+  const tableBodyDown = bidRows?.length ? (
+    <div
+      className="text-right w-full flex flex-col justify-start"
+      style={{ height: 'calc(50% - 20px)' }}
```
mattrussell36, line 104: Can we just use AutoSizer to give us the height? We'll end up with less hard coded values for layout. We can then also remove the useResizeObserver callback logic.

libs/market-depth/src/lib/orderbook-row.tsx
```diff
@@ -89,42 +131,46 @@ export const OrderbookContinuousRow = React.memo(
...
-          testId={`cumulative-vol-${price}`}
-          positionDecimalPlaces={positionDecimalPlaces}
-          bid={cumulativeBid}
-          ask={cumulativeAsk}
-          relativeAsk={cumulativeRelativeAsk}
-          relativeBid={cumulativeRelativeBid}
-          indicativeVolume={indicativeVolume}
-        />
-      </>
+        <div className="grid gap-1 text-right auto-rows-[17px] grid-cols-3 w-full grid-rows-1">
```
mattrussell36, line 144: Lots of these classes are not required, we don't need to account for rows because this grid only has one row

libs/market-depth/src/lib/orderbook-manager.tsx
```diff
@@ -30,9 +31,10 @@ export const OrderbookManager = ({ marketId }: OrderbookManagerProps) => {
   const variables = { marketId };
   const resolutionRef = useRef(resolution);
   const [orderbookData, setOrderbookData] = useState<OrderbookData>({
-    rows: null,
+    asks: null,
+    bids: null,
   });
-  const dataRef = useRef<OrderbookData>({ rows: null });
+  const dataRef = useRef<OrderbookData>({ asks: null, bids: null });
```
mattrussell36, line 37: I realise this is what we have currently, but it seems odd to need to store the data both in useState and as a ref. Trying to understand why exactly we need to do this, but something seems wrong here

libs/market-depth/src/lib/orderbook-data.ts
```diff
@@ -286,6 +301,41 @@ export const updateCompactedRows = (
...
+) => {
+  const data = cloneDeep(oldData as OrderbookRowData[]);
+  uniqBy(reverse(newData || []), 'price')?.forEach((delta) => {
+    partiallyUpdateCompactedRows(dataType, data, delta, resolution);
+  });
+
+  updateCumulativeVolumeByType(data, dataType);
+  let index = 0;
+  // remove levels that do not have any volume
+  while (index < data.length) {
```
mattrussell36, line 327: Why do we need this? We shouldn't need to fill in empty price levels. That was a previous requirement

mattrussell36 (changes requested):
@macqbat 

Left you some comments. I really find the data management side of this extremely over-engineered and overly complex. This is no doubt because of the original requirements, which are no longer applicable, I think we can make this much more simple if we more aggressively strip it back and rewrite it.

libs/market-depth/src/lib/orderbook.tsx
macqbat: we can't use it since we have rich memoization in hooks. We would need to completely refactor the solution for it.

libs/market-depth/src/lib/orderbook-manager.tsx
macqbat: refs we need for better memoization in hooks

libs/market-depth/src/lib/orderbook-row.tsx
macqbat: removed redundant classes

libs/market-depth/src/lib/orderbook-data.ts
macqbat: empty bids or asks can be a result of updating existing data with delta from subscription

ben-vegaprotocol:
Couple of thoughts:

- Should we reduce the spacing between the rows, simmilar to the example Barney gave in the ticket.
- Sould we have a currency for the spread?
- When you hover over a price it might be nice if the color change was across the whole row? Maybe increase the clickable area as well to make it a bit easier to use? (Simmilar to [dYdX](https://trade.stage.dydx.exchange/trade/ETH-USD))

macqbat:
> Couple of thoughts:
> 
> * Should we reduce the spacing between the rows, simmilar to the example Barney gave in the ticket.
> * Sould we have a currency for the spread?
> * When you hover over a price it might be nice if the color change was across the whole row? Maybe increase the clickable area as well to make it a bit easier to use? (Simmilar to [dYdX](https://trade.stage.dydx.exchange/trade/ETH-USD))

Such a stuff we should possibly tweak according to the delivered design

libs/market-depth/src/lib/orderbook-manager.tsx
```diff
@@ -1,113 +1,90 @@
...
-      deltaRef.current.sell = [];
-      setOrderbookData(dataRef.current);
-    }, 250)
-  );
 
-  useEffect(() => {
-    deltaRef.current.buy = [];
-    deltaRef.current.sell = [];
-  }, [marketId]);
+  const initialDataSort = useCallback((data?: MarketDepthQuery['market']) => {
```
mattrussell36, line 33: There doesn't appear to be any sorting going on in this function. I'd rename to something more accurate

libs/market-depth/src/lib/orderbook-manager.tsx
```diff
@@ -129,8 +106,7 @@ export const OrderbookManager = ({ marketId }: OrderbookManagerProps) => {
 
   const marketDataUpdate = useCallback(
     ({ data }: { data: MarketData | null }) => {
-      marketDataRef.current = data;
-      updateOrderbookData.current();
+      setMarkPrice(data?.markPrice || '');
```
mattrussell36, line 109: Why not just let the internal `update` function handle the update. Then we can avoid storing an extra value in state and handling the update

I see no issue with just using markPrice like this
```
const {
    data: marketData,
    error: marketDataError,
    loading: marketDataLoading,
  } = useDataProvider({
    dataProvider: marketDataProvider,,
    variables,
  });
  
  // just use the value
  marketData.markPrice 
```

libs/market-depth/src/lib/orderbook.tsx
```diff
@@ -1,677 +1,184 @@
...
+                      resolution={resolution}
+                      decimalPlaces={decimalPlaces}
+                      positionDecimalPlaces={positionDecimalPlaces}
+                      onClick={onClick}
+                    />
+                    <div
+                      className={`flex flex-shrink h-[${MID_PRICE_HEIGHT}px] items-center justify-center text-lg`}
+                    >
+                      <PriceCell
+                        value={Number(markPrice)}
```
mattrussell36, line 132: We actually have `midPrice` in market data now so we should use that

mattrussell36 (changes requested):

libs/market-depth/src/lib/orderbook.tsx
```diff
@@ -1,677 +1,184 @@
...
 
-const HorizontalLine = ({ top, testId }: { top: string; testId: string }) => (
-  <div
-    className="absolute border-b border-default inset-x-0 hidden"
-    style={{ top }}
-    data-testid={testId}
-  />
-);
+// 17px of row height plus 3px gap
+export const rowHeight = 20;
```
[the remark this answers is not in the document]
macqbat: I wouldn't do this. The view is therefore little bit overloaded and blurry:
image

libs/market-depth/src/lib/orderbook-manager.tsx
macqbat: renamed for `initData` :-)

libs/market-depth/src/lib/orderbook-manager.tsx
macqbat: done

libs/market-depth/src/lib/orderbook-row.tsx
```diff
@@ -1,130 +1,125 @@
...
-          testId={`cumulative-vol-${price}`}
-          positionDecimalPlaces={positionDecimalPlaces}
-          bid={cumulativeBid}
-          ask={cumulativeAsk}
-          relativeAsk={cumulativeRelativeAsk}
-          relativeBid={cumulativeRelativeBid}
-          indicativeVolume={indicativeVolume}
-        />
-      </>
+      <div className="relative w-full">
```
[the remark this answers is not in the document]
macqbat: done

libs/market-depth/src/lib/orderbook-row.tsx
```diff
@@ -1,130 +1,125 @@
...
+        <CumulationBar cumulativeValue={cumulativeRelativeValue} type={type} />
+        <div className="grid gap-1 text-right auto-rows-[17px] grid-cols-3">
```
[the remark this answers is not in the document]
macqbat: done

libs/market-depth/src/lib/orderbook.tsx
```diff
@@ -1,677 +1,184 @@
...
+                    <OrderbookTable
+                      rows={askRows}
+                      type={VolumeType.ask}
+                      resolution={resolution}
+                      decimalPlaces={decimalPlaces}
+                      positionDecimalPlaces={positionDecimalPlaces}
+                      onClick={onClick}
+                    />
+                    <div
+                      className={`flex flex-shrink h-[${MID_PRICE_HEIGHT}px] items-center justify-center text-lg`}
```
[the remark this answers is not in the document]
macqbat: sure, removed

libs/market-depth/src/lib/orderbook-manager.tsx
```diff
@@ -1,113 +1,89 @@
...
-      setOrderbookData(dataRef.current);
-    }, 250)
-  );
 
-  useEffect(() => {
-    deltaRef.current.buy = [];
-    deltaRef.current.sell = [];
-  }, [marketId]);
+  const initData = useCallback((data?: MarketDepthQuery['market']) => {
+    const bids = (data?.depth.buy || []).filter((item) => item.volume !== '0');
```
bglownia, line 33: If there is a need to filter out zero volume price levels it would be better to do it in data provider

bglownia (changes requested):

bglownia (approved):

PR #4022, opened 2023-06-02
feat(explorer): transaction list filtering'

edd:
- chore(explorer): regenerate types
- chore(explorer): form filter
- feat: initial dropdown for filtering transactions by type
- chore(explorer): rearrange tx list filtering


apps/explorer/src/app/components/txs/tx-filter.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+  return (
+    <span>
+      {t('Filters')}: <code>{Array.from(filters)[0]}</code>
+    </span>
+  );
+}
+
+// Effectively a copy & paste of the Vega DropdownMenuItem styles
+const itemClass = classNames(
```
sam-keen, line 96: Suggestion (non-blocking): as above, it'd be nice maintainability win to import exported styles from the dropdownMenuItem element

sam-keen (approved):

apps/explorer/src/app/hooks/use-txs-data.ts
```diff
@@ -65,6 +65,14 @@ export const useTxsData = ({ limit, filters }: IUseTxsData) => {
     }
   }, [setTxsState, data]);
 
+  useEffect(() => {
+    setTxsState((prev) => ({
```
mattrussell36, line 69: setting state from an effect is an anti pattern. Can this state setter be lifted?

apps/explorer/src/app/components/blocks/blocks-refetch.tsx
```diff
@@ -1,36 +1,33 @@
-import { useState, useEffect } from 'react';
-import { useTendermintWebsocket } from '../../hooks/use-tendermint-websocket';
 import { t } from '@vegaprotocol/i18n';
-import { ButtonLink } from '@vegaprotocol/ui-toolkit';
+import { Icon } from '@vegaprotocol/ui-toolkit';
```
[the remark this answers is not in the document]
edd: I can see why I could, but I'm not going now as it involves finding a refresh icon.

apps/explorer/src/app/hooks/use-txs-data.ts
edd: You are right and in all normal scenarios I should rewrite this. However the next ticket is junking this loader and rewriting it entirely (#4070) so instead I will make sure I don't port the mistake there.

apps/explorer/src/app/components/txs/tx-filter.tsx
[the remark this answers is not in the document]
edd: Yes. I've done this

edd: As above. Fixed.

apps/explorer/src/app/components/blocks/blocks-refetch.tsx
```diff
@@ -1,36 +1,33 @@
...
+
+// These are cribbed from dropdown-menu.tsx, so that the classes match
+// the filter button this sits next to on the TXs list page
+const defaultClasses = [
```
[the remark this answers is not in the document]
edd: I fixed this by going in another direction - removing the default styling from the DropdownTrigger by providing a className, so now it's a more standard button.

PR #4030, opened 2023-06-02
chore(trading): add quantum formatting to deal ticket

### LINKED ISSUE #3952: Add quantum formatting to deal ticket
## Story

As a user
I want to not see my estimated margin calculated to 0.000001 USDT 
So that my display looks nicer and doesn't end up text wrapping 

## Acceptance Criteria

- [x] I can see anything in the deal ticket currently using settlement asset decimals formatted using "Quantum Formatting" 
- [x] "Quantum Formatting" is defined as `Number DP = MAX(0, LOG10(100/Quantum))`
- [x] In the case where "Number DP" resolves as decimal round by normal logic to decide the number of decimal places
- [ ] There is a way to see the full number if I want to know the fully accurate number (e.g. with a tooltip)
- [ ] There is a way to find a definition of quantum, how it is calculated and why we use it in Console (i.e. [link to docs](https://github.com/vegaprotocol/documentation/issues/640))

## Tasks

- [ ] UX (if needed)
- [ ] Design (if needed)
- [ ] Implementation
- [ ] Code review
- [ ] QA review

NB. Ideally the 100 in the above would be configurable in the component, because we will probably use this in LP Dashboard too and in that we probably dont care about 1c .... we would only care about 1 USD min so having that configurable gives us flexibility.

MadalinaRaicu:
# Related issues 🔗

Closes #3952 

# Description ℹ️

- [x] Add quantum formatting to the deal ticket.
- [x] Added tooltips to show full numbers
- [x] Added tests

# Demo 📺
image

# Technical 👨‍🔧


libs/deal-ticket/src/hooks/use-fee-deal-ticket-details.tsx
```diff
@@ -53,19 +53,22 @@ export interface FeeDetails {
...
 const formatValue = (
   value: string | number | null | undefined,
-  formatDecimals: number
+  formatDecimals: number,
+  quantum?: string
 ): string => {
+  const numberDp = Math.max(0, Math.log10(100 / Number(quantum) || 0));
   return isNumeric(value)
-    ? addDecimalsFormatNumber(value, formatDecimals)
+    ? addDecimalsFormatNumber(value, formatDecimals, numberDp)
```
bglownia, line 61: Most probably we will be using this function also in other places so it will be useful to have this in `libs/utils/sec/lib/format/number.ts` as something like this

```
export const addDecimalsFormatNumberQuantum = (
  rawValue: string | number,
  decimalPlaces: number,
  quantum: number | string
) =>
  addDecimalsFormatNumber(
    rawValue,
    decimalPlaces,
    Math.max(0, Math.log10(100 / Number(quantum) || 0))
  );
```

MadalinaRaicu: added

libs/deal-ticket/src/hooks/use-fee-deal-ticket-details.tsx
```diff
@@ -1,5 +1,9 @@
 import { FeesBreakdown } from '@vegaprotocol/markets';
```
edd, line 1: **chore (non-blocking)**: I'd like to see more tests for this hook. Initially I thought `formatValue` and formatRange`  were untested, but @MadalinaRaicu pointed me to the tests for the new `addDecimalsFormatNumberQuantum` function that does a great job of encapsulating how that function works. 

I still want to see more hook tests, but the `formatValue` function has enough branches that some tests would be good to show a few paths through the code. Decimal formatting has been error prone in the past, so the more tests at all levels, the better off we'll be.

libs/utils/src/lib/format/number.ts
```diff
@@ -90,6 +90,17 @@ export const formatNumberFixed = (
...
 
+export const addDecimalsFormatNumberQuantum = (
+  rawValue: string | number,
+  decimalPlaces: number,
+  quantum: number | string
+) =>
+  addDecimalsFormatNumber(
+    rawValue,
+    decimalPlaces,
+    Math.max(0, Math.log10(100 / Number(quantum) || 0))
```
edd, line 101: **chore (non-blocking)**: Sounding like a broken record, but I'd also test this function for this line only. It takes some effort to pull apart what it's doing.

libs/deal-ticket/src/hooks/use-fee-deal-ticket-details.tsx
MadalinaRaicu: I'm gonna add some while this PR is still in draft

libs/utils/src/lib/format/number.spec.ts
```diff
@@ -23,6 +24,21 @@ describe('number utils', () => {
     }
   );
 
+  it.each([
+    { v: new BigNumber(123000), d: 5, o: '1.23', q: 0.1 },
+    { v: new BigNumber(123000), d: 3, o: '123.00', q: 0.1 },
+    { v: new BigNumber(123000), d: 1, o: '12,300.00', q: 0.1 },
+    { v: new BigNumber(123001), d: 2, o: '1,230.01', q: 0.1 },
+    { v: new BigNumber(123001000), d: 2, o: '1,230,010.00', q: 0.1 },
```
edd, line 32: **comment (non-blocking)**: I copied in David for a review of these but they *look* good to me.

edd (approved):

libs/utils/src/lib/format/number.ts
MadalinaRaicu: added more tests

apps/trading-e2e/src/integration/trading-deal-ticket-order.cy.ts
```diff
@@ -100,9 +100,8 @@ describe('deal ticker order validation', { tags: '@smoke' }, () => {
         .within(() => {
           cy.get('[data-state="closed"]').should(
             'have.text',
-            'Total margin available'
+            'Total margin available100,000.01 tDAI'
```
jeremyletang, line 103: I realise I'm supposed to review the graphql, but I can't help than think it's weird theire's no spaces between available and 100,....

jeremyletang (approved):

PR #4040, opened 2023-06-06
fix(governance): ensure cached finalized vote is only used when pubkey is c…

### LINKED ISSUE #3862: governance dapp: The "You voted" doesn’t update when I switch between wallets.
## Description

I have 1 vegawallet with 2 keys - each with its associated tokens.
I switch between them using the Manage button, but the problem is that the You voted section under Proposal does not update.

### Example 1

Starting point: I haven’t voted yet from any wallet
1. I open the proposal, and vote from my first key - the `You voted` correctly shows value
2. Then I click Manage and switch to the other key
3. The `You voted`, doesn't show correct value that I haven't voted yet
4. Instead it shows state of my first wallet, that I have already voted
5. Refreshing dApp helps

### Example 2

Starting point: I voted from both of my keys: one `For` and one `Against
1. I open the proposal with first key - the `You voted` shows correct value: `For`
2. Then I click Manage ans switch to the other key
3. The `You voted` shows incorrect value `For`, because I voted `Against` from my 2nd key
4. Refreshing dApp helps

## Expected behavior

The `You voted` should update when I switch between keys.

## Device details

Latest version of FF on MacOS.

sam-keen:
…orrect

# Related issues 🔗

Closes #3862

# Description ℹ️

After a user has voted, their vote is cached and used to display vote state thereafter. However if a user has voted with multiple pubkeys, and switched their pubkey (without reloading the page causing a new query), the cached vote could be incorrect. To address this I've now added the pubkey to the cached object and now check whether to display the cached value based on the value matching the current pubkey.


edd (approved):

edd:
Failing governance e2e tests on this one

PR #4053, opened 2023-06-07
fix(trading): order book mid

### LINKED ISSUE #4051: Orderbook view is incorrect
## Description

midPrice is not in the middle

## Steps to Reproduce

1. Open market page
2. See error: mid price is not in the center of orderbook panel

## Device details

**Smartphone (please complete the following information):**

macqbat:
# Related issues 🔗

Closes #4051 

# Description ℹ️

- Adds asset symbol to the mid
- Fixes the layout of the grid in the order book which wasn't taking into account the configured sizes
- Fixes row sizes which resulted in one less row per side being shown, even though there was space for it
- Adds a small 1px gap between rows

# Technical 👨‍🔧

- Changed to use some inline styles instead of tailwind as tailwind was not compiling the dynamically created rowHeight, midHeight sizes

### DIFF
--- a/libs/market-depth/src/lib/orderbook.tsx
+++ b/libs/market-depth/src/lib/orderbook.tsx
@@ -52,7 +52,10 @@ const OrderbookTable = ({
         )
       }
     >
-      <div className={`grid auto-rows-[${rowHeight}px]`}>
+      <div
+        className="grid"
+        style={{ gridAutoRows: rowHeight }} // use style as tailwind won't compile the dynamically set height
+      >
         {rows.map((data) => (
           <OrderbookRow
             key={data.price}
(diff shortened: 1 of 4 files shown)


libs/market-depth/src/lib/orderbook.tsx
```diff
@@ -107,7 +107,7 @@ export const Orderbook = ({
             const bidRows = groupedBids?.slice(0, limit) ?? [];
             return (
               <div
-                className={`overflow-hidden grid grid-rows-[1fr_${midHeight}px_1fr]`}
+                className={`overflow-hidden grid grid-rows-[1fr_30px_1fr]`}
```
mattrussell36: Ah I guess tailwind didn't compile the class. Rather than this it would better to use an inline style

mattrussell36 (approved):

PR #4056, opened 2023-06-08
fix(ui-toolkit): display of loader in large variant

mattrussell36:
# Related issues 🔗

N/A

# Description ℹ️

Fixes the gaps in the loader when displayed as large

# Demo 📺

Screenshot 2023-06-07 at 20 42 05


macqbat (approved):
I let myself for push something for save consistency in css for console app preloader. I added also preloader.scss as missing source to static app

PR #4059, opened 2023-06-08
fix(markets): back with updating by grid api

### LINKED ISSUE #4058: market list loses visibility of context menu
## Description

Redundant rerenders cause unwanted closing of contextual row menu

## Steps to Reproduce

1. Go to market list
2. Click on any row menu button for open it
3. Menu is open, wait a couple of seconds
4. See error: menu is closed by itself :-(

## Expected behavior

Menu should keep open state

## Screenshots

-

## Device details

**Smartphone (please complete the following information):**

macqbat:
# Related issues 🔗

Closes #4058 

# Description ℹ️

Let's make update of grid data direct by api to prevent re-renders which cause closing open context menu

# Demo 📺

-

# Technical 👨‍🔧

-


libs/markets/src/lib/components/markets-container/markets-container.tsx
```diff
@@ -30,11 +40,14 @@ export const MarketsContainer = ({ onSelect }: MarketsContainerProps) => {
     };
   }, [reload]);
 
+  const handleOnGridReady = useCallback(() => {
```
mattrussell36, line 43: suggestion (blocking): just tried out using useMemo to create col defs. This fixes the issue and is more simple

mattrussell36 (changes requested):

libs/markets/src/lib/components/markets-container/markets-container.tsx
mattrussell36: you can merge this change

macqbat: merged and little bit improved

mattrussell36 (approved):

PR #4083, opened 2023-06-13
chore(wallet): use types generated from vega proto files POC

### LINKED ISSUE #3709: Use generated transaction types
## The Chore

Right now we manually maintain types for transaction payloads in [vega-connector.ts](https://github.com/vegaprotocol/frontend-monorepo/blob/develop/libs/wallet/src/connectors/vega-connector.ts)

We should be using generated types based off the protos.

## Tasks

- [ ] Import and use generated types
- [ ] Remove manually maintained types

macqbat:
# Related issues 🔗

Closes #3709

# Description ℹ️

Replace manually updated types with generated from vega core protos

# Demo 📺

-

# Technical 👨‍🔧

-


libs/wallet/src/connectors/vega-connector.ts
```diff
@@ -1,68 +1,93 @@
...
+import type { WithdrawSubmission as OriginalWithdrawSubmission } from '@vegaprotocol/protos/dist/vega/commands/v1/WithdrawSubmission';
+import type { ProposalSubmission as OriginalProposalSubmission } from '@vegaprotocol/protos/dist/vega/commands/v1/ProposalSubmission';
+import type { BatchMarketInstructions as OriginalBatchMarketInstructions } from '@vegaprotocol/protos/dist/vega/commands/v1/BatchMarketInstructions';
+import type { Transfer } from '@vegaprotocol/protos/dist/vega/commands/v1/Transfer';
+import type { ProposalTerms } from '@vegaprotocol/protos/dist/vega/ProposalTerms';
+import type { WithdrawExt } from '@vegaprotocol/protos/dist/vega/WithdrawExt';
+
+// generated type doesn't match in many places with current implementation. It has to be overwrite here
+// and re-exported for saving consistency.
+export type ProposalSubmission = Omit<
```
mattrussell36, line 18: I don't want to introduce these overriding types. The types exported here should be exactly as the wallet requires them. If there are differences we should make sure any transactions sent to the wallet conform to the interface

libs/wallet/src/utils.ts
```diff
@@ -1,16 +1,44 @@
...
+  OrderSubmission,
+  OrderAmendment,
 } from './connectors';
 import type { Exact } from 'type-fest';
+import { vega as vegaProtos } from '@vegaprotocol/protos';
+import type { Side } from '@vegaprotocol/protos/dist/vega/Side';
+import type { TimeInForce } from '@vegaprotocol/protos/dist/vega/Order/TimeInForce';
+import type { Type } from '@vegaprotocol/protos/dist/vega/Order/Type';
+
+export interface DealTicketOrderSubmission {
```
mattrussell36, line 20: These types shouldn't live here. They would be better in the deal ticket lib. The deal ticket should be required to conform to the interfaces provided by the wallet

libs/wallet/src/connectors/vega-connector.ts
```diff
@@ -1,68 +1,93 @@
...
+import type { OrderSubmission as OriginalOrderPosition } from '@vegaprotocol/protos/dist/vega/commands/v1/OrderSubmission';
+import type { OrderCancellation as OriginalOrderCancellation } from '@vegaprotocol/protos/dist/vega/commands/v1/OrderCancellation';
+import type { OrderAmendment as OriginalOrderAmendment } from '@vegaprotocol/protos/dist/vega/commands/v1/OrderAmendment';
+import type { VoteSubmission } from '@vegaprotocol/protos/dist/vega/commands/v1/VoteSubmission';
+import type { WithdrawSubmission as OriginalWithdrawSubmission } from '@vegaprotocol/protos/dist/vega/commands/v1/WithdrawSubmission';
+import type { ProposalSubmission as OriginalProposalSubmission } from '@vegaprotocol/protos/dist/vega/commands/v1/ProposalSubmission';
+import type { BatchMarketInstructions as OriginalBatchMarketInstructions } from '@vegaprotocol/protos/dist/vega/commands/v1/BatchMarketInstructions';
+import type { Transfer } from '@vegaprotocol/protos/dist/vega/commands/v1/Transfer';
+import type { ProposalTerms } from '@vegaprotocol/protos/dist/vega/ProposalTerms';
+import type { WithdrawExt } from '@vegaprotocol/protos/dist/vega/WithdrawExt';
```
mattrussell36, line 14: Can we export all these types from here like this this way so we don't introduce @vegaprotocol/protos as a dep for other libs

```
import type { vega as vegaProtos } from '@vegaprotocol/protos';

export type OrderSubmission =
  vegaProtos.commands.v1.OrderSubmission.OrderSubmission;
export type OrderCancellation =
  vegaProtos.commands.v1.OrderCancellation.OrderCancellation;
export type LiquidityProvisionSubmission =
  vegaProtos.commands.v1.LiquidityProvisionSubmission.LiquidityProvisionSubmission;
export type DelegateSubmission =
  vegaProtos.commands.v1.DelegateSubmission.DelegateSubmission;
export type UndelegateSubmission =
  vegaProtos.commands.v1.UndelegateSubmission.UndelegateSubmission;
export type OrderAmendment =
  vegaProtos.commands.v1.OrderAmendment.OrderAmendment;
export type VoteSubmission =
  vegaProtos.commands.v1.VoteSubmission.VoteSubmission;
export type WithdrawSubmission =
  vegaProtos.commands.v1.WithdrawSubmission.WithdrawSubmission;
export type ProposalSubmission =
  vegaProtos.commands.v1.ProposalSubmission.ProposalSubmission;
export type BatchMarketInstructionSubmission =
  vegaProtos.commands.v1.BatchMarketInstructions.BatchMarketInstructions;
export type TransferSubmission = vegaProtos.commands.v1.Transfer.Transfer;
```

libs/accounts/src/lib/transfer-container.tsx
```diff
@@ -6,7 +6,7 @@ import {
   useNetworkParam,
 } from '@vegaprotocol/network-parameters';
 import { useDataProvider } from '@vegaprotocol/data-provider';
-import type { Transfer } from '@vegaprotocol/wallet';
+import type { Transfer } from '@vegaprotocol/protos/dist/vega/commands/v1/Transfer';
```
mattrussell36, line 9: I'd prefer these types exported from the wallet. I've left a comment on vega-connectors.ts, if we make that change can we then import this type like (same goes for other files in the PR):

`import type { TransferSubmission } from '@vegaprotocol/wallet'`

libs/wallet/src/connectors/vega-connector.ts
mattrussell36: I see now why we need these :). Can we simplify it though?

for example I think is more clear
```
export type OrderSubmission = SetOptional<
  protos.commands.v1.OrderSubmission.OrderSubmission,
  'reference' | 'peggedOrder' | 'price' | 'expiresAt'
>;
```

macqbat: done

libs/wallet/src/connectors/vega-connector.ts
macqbat: done

libs/wallet/src/connectors/vega-connector.ts
```diff
@@ -1,98 +1,29 @@
...
-        receiverAddress: string;
-      };
-    };
-  };
-}
+import type { SetOptional } from 'type-fest';
+import type { vega as vegaProtos } from '@vegaprotocol/protos';
+import type { RecurringTransfer } from '@vegaprotocol/protos/dist/vega/commands/v1/RecurringTransfer';
+
+export type OriginalOrderPosition =
```
bglownia, line 6: did you mean `OriginalOrderSubmission`?

libs/withdraws/src/setup-tests.ts
```diff
@@ -1,4 +1,5 @@
 import '@testing-library/jest-dom';
+import '../../../setup-test';
```
mattrussell36, line 2: I think we should probably set these set up in the jest.config.ts, rather than importing here

libs/withdraws/src/lib/create-withdrawal-dialog.tsx
```diff
@@ -34,7 +35,7 @@ export const CreateWithdrawalDialog = () => {
                 },
               },
             },
-          });
+          } as Transaction);
```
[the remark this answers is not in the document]
macqbat: removed

libs/withdraws/src/setup-tests.ts
macqbat: I couldn't find any other easy way of share jest setup configuration :-(
Moving it to jest.config caused crash because we cannot import from over local rootDir

libs/wallet/src/connectors/vega-connector.ts
```diff
@@ -302,51 +233,84 @@ export interface ProposalSubmission {
...
 }
 
-export interface ProposalSubmissionBody {
-  proposalSubmission: ProposalSubmission;
+// generated type doesn't match in many places with current implementation. It has to be overwrite here
+// and re-exported for saving consistency.
+export type WithdrawSubmission = Omit<OriginalWithdrawSubmission, 'ext'> &
+  WithdrawExt;
+
+export type OrderSubmission = SetOptional<
```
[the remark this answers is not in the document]
macqbat: yeah, this is done

mattrussell36 (changes requested):
It doesn't seem sensible to me to introduce these changes as they are now. This is because

- We are quite substantially modifying the generated types to get this to work. Devaluing the whole idea behind this piece of work
- We are introducing lots of extra config around cypress and jest
- We also have to modify BigInt's prototype (which is bad practice) so that the BigInt types can be serialized for the network request. This means that we are converting OrderObj strings into BigInts which then need to get converted back to strings for the json rpc payload. This all just introduces greater surface area for bugs

I'm not sure what the best path forward is right now, but we IMO shouldn't merge this, and should try to figure out a better solution

mattrussell36:
@macqbat Closing this as we are likely going to get better types exported from a new package which will mean we don't need so many work-arounds and overriding of types

PR #4114, opened 2023-06-15
fix(trading): quantum formatting changes

### LINKED ISSUE #2341: Use quantum formatting for liquidity provisions table
## Story

**As a** user
**I want** to see values in the liquidity provisions table that are more meaningful
**So that** I can better understand the real value of the numbers in the LP table

## Acceptance Criteria

- [ ] Numeric values in the LP table are formatted using asset.quantum
- [ ] I can interact with a column header to change the formatting to show the full raw number for that column
- [ ] There is a way to see the full number if I want to know the fully accurate number (e.g. with a tooltip)
- [ ] There is a way to find a definition of quantum, how it is calculated and why we use it in Console (i.e. [link to docs](https://github.com/vegaprotocol/documentation/issues/640))

## Tasks

- [ ] UX (if needed)
- [ ] Design (if needed)
- [ ] Implementation
- [ ] Code review
- [ ] QA review

Created from: https://github.com/vegaprotocol/frontend-monorepo/issues/2132

MadalinaRaicu:
# Related issues 🔗

Closes #2341 

# Description ℹ️

- [x] Use quantum formatting in liquidity provision tables.
- [x] Fixes quantum calculation by formatting settlement asset quantum with asset decimals
- [x] Fixes navigation to liquidity view
- [x] Fixes `Notional` and `Liquidation` estimates to be formatted with market decimals and shown in quote, not in settlement asset symbol. This is how it has been before but it was changed recently

# Demo 📺
image

# Technical 👨‍🔧


libs/liquidity/src/lib/liquidity-table.tsx
```diff
@@ -35,21 +37,152 @@ export interface LiquidityTableProps
...
+          minWidth: 160,
+          valueFormatter: assetDecimalsQuantumFormatter,
+          tooltipValueGetter: assetDecimalsFormatter,
+        },
+        {
+          headerName: t('Obligation'),
+          field: 'commitmentAmount',
+          type: 'rightAligned',
+          headerTooltip: t(
+            `The liquidity provider's obligation to the market, calculated as the liquidity commitment amount multiplied by the value of the stake_to_ccy_siskas network parameter to convert into units of liquidity volume. The obligation can be met by a combination of LP orders and limit orders on the order book.`
```
candida-d, line 139: I believe it's no longer called stake_to_ccy_siskas.

apps/trading-e2e/src/integration/market-liquidity.cy.ts
```diff
@@ -84,7 +84,7 @@ describe('liquidity table - trading', { tags: '@smoke' }, () => {
     cy.get(rowSelector)
       .first()
       .find(colAverageEntryValuation)
-      .should('have.text', '685,852.93692');
+      .should('have.text', '685,852.94');
```
mattrussell36, line 87: chore (blocking): Can we add a new AC for this https://github.com/vegaprotocol/specs/blob/master/user-interface/5002-LIQP-liquidity.md

mattrussell36 (changes requested):

apps/trading-e2e/src/integration/market-liquidity.cy.ts
MadalinaRaicu: https://github.com/vegaprotocol/specs/pull/1769

libs/liquidity/src/lib/liquidity-table.tsx
MadalinaRaicu: Yes, it is stakeToCcyVolume in the code. I'm updating the tooltip

MadalinaRaicu:
```suggestion
            `The liquidity provider's obligation to the market, calculated as the liquidity commitment amount multiplied by the value of the stake_to_ccy_volume network parameter to convert into units of liquidity volume. The obligation can be met by a combination of LP orders and limit orders on the order book.`
```

MadalinaRaicu:
```suggestion
            `The liquidity provider's obligation to the market, calculated as the liquidity commitment amount multiplied by the value of the stake_to_ccy_volume network parameter to convert into units of liquidity volume. The obligation can be met by a combination of LP orders and limit orders on the order book.`
```

apps/trading/lib/hooks/use-market-click-handler.ts
```diff
@@ -20,20 +20,8 @@ export const useMarketClickHandler = (replace = false) => {
...
-      const link = Links[Routes.LIQUIDITY](selectedId);
-      if (metaKey) {
-        window.open(`/#${link}`, '_blank');
-      } else if (selectedId !== marketId || !isLiquidityPage) {
-        navigate(link, { replace });
-      }
-    },
-    [navigate, marketId, replace, isLiquidityPage]
-  );
+export const useMarketLiquidityClickHandler = () => {
```
mattrussell36, line 23: Why this change here?

mattrussell36: I think we probably only want _blank if metaKey is true

mattrussell36 (approved):

apps/trading/lib/hooks/use-market-click-handler.ts
MadalinaRaicu: updated to keep the metakey. 
the method was changed because it was not working correctly. navigation on LP view was broken
even in mainnet

libs/deal-ticket/src/components/deal-ticket/deal-ticket-fee-details.tsx
```diff
@@ -267,19 +267,7 @@ export const DealTicketFeeDetails = ({
...
-        ? liquidationEstimateBestCase
-        : liquidationEstimateWorstCase
-      ).toString(),
-      (liquidationEstimateBestCase > liquidationEstimateWorstCase
-        ? liquidationEstimateBestCase
-        : liquidationEstimateWorstCase
-      ).toString(),
-      assetDecimals,
-      quantum
+      marketDecimals
```
bglownia, line 270: liquidation price should use assetDecimals, when using market decimals we are getting wrong values
image

MadalinaRaicu: this might be a problem. is it a bug?
as far as I know from @barnabee `liquidation price` and `notional` are dependent on quote, and not settlement asset. so we should use market dp instead. 

why liquidation price uses asset dp instead?

bglownia: It uses asset decimals and it has fractional part also.
image

@witgaw @barnabee should we change it in API to use market decimal places?

JonRay15:
The value is coming in `asset dp` and this will not change imminently. 

We need to convert it using `asset dp` and then display it to the level of precision given in `market dp`.

PR #4119, opened 2023-06-15
refactor(trading): pane stores for grid state

### LINKED ISSUE #3396: Persist table filters
## Story

As a user
When I have set filters on data tables to my preferences
I want those filters to be persisted when I refresh my browser
So that I don't have to reset filters to my preferences every time

## Acceptance Criteria

- [ ] I can apply filters to data tables
- [ ] When I switch to view another market, the filters I applied are preserved
- [ ] When I refresh the page I am looking at, I can see the filters I applied are preserved
- [ ] When I switch between light and dark themes, I can see the filters I applied are preserved
- [ ] All markets use the same filters
- [ ] This same behaviour is true for every table in console where I can apply filters

## Tasks

- [ ] Implementation
- [ ] Write tests (unit and e2e) 
- [ ] Code review
- [ ] Verify acceptance criteria met
- [ ] Exploratory testing to find bugs and edge cases
- [ ] Write automated regression tests for the A/Cs and add to spec doc (expanding on A/Cs where possible)

### LINKED ISSUE #3397: Persist table sorting
## Story

As a user
When I have sort data in tables to my preferences
I want that sorting to be persisted when I refresh my browser
So that I don't have to sort to my preferences every time

## Acceptance Criteria

- [ ] I can sort data in tables
- [ ] When I switch to view another market, the sorting I applied is preserved
- [ ] When I refresh the page I am looking at, I can see the sorting I applied is preserved
- [ ] When I switch between light and dark themes, I can see the sorting I applied is preserved
- [ ] All markets use the same sorting
- [ ] This same behaviour is true for every table in console where I can apply sorting

## Tasks

- [ ] Implementation
- [ ] Write tests (unit and e2e) 
- [ ] Code review
- [ ] Verify acceptance criteria met
- [ ] Exploratory testing to find bugs and edge cases
- [ ] Write automated regression tests for the A/Cs and add to spec doc (expanding on A/Cs where possible)

mattrussell36:
# Related issues 🔗

Closes #3396 
Closes #3397 

# Description ℹ️

Adds a way for panes to store state and connects panes that display data grids to store the current filters and sort orders. Going forward we can use these stores to manage other pane based settings.

# Demo 📺

N/A

# Technical 👨‍🔧

- Adds useDataGridEvents hook to connect resize, filter and sort events to a component
- Adds zustand stores for each pane. These are initialized in the app only, not in libs so as not to leak local storage into other apps.
- Remove useBottomPlaceholder hook. For now buttons might overlay the bottom rows, the incoming pane context menu will resolve this.
- Remove useColumnSizes


apps/trading/components/accounts-container/accounts-container.tsx
```diff
@@ -75,3 +85,31 @@ export const AccountsContainer = ({
...
+  filterModel?: { [key: string]: any };
+  columnState?: ColumnState[];
+};
+
+const useAccountStore = create<{
+  gridStore: Store;
+  update: (gridStore: Store) => void;
+}>()(
+  persist(
+    subscribeWithSelector((set) => ({
```
bglownia, line 100: `subscribeWithSelector` will not give any performance gain if selector returns new array each time.
`Object.is([store.gridStore, store.update], [store.gridStore, store.update])` is always false 

To make it work effectively we can split usage in two hook calls.

```
const gridStore = useAccountStore((store) => store.gridStore);
const update =   useAccountStore((store) => store.update);
```

apps/trading/components/accounts-container/accounts-container.tsx
```diff
@@ -75,3 +85,31 @@ export const AccountsContainer = ({
     </div>
   );
 };
+
+type Store = {
```
bglownia, line 89: Store shape for each table is the same. Do we need to defined it for each table separately?

```
const makeTableStateStore = (name: string) =>
  create<{
    gridStore: Store;
    update: (gridStore: Store) => void;
  }>()(
    persist(
      subscribeWithSelector((set) => ({
        gridStore: {},
        update: (newStore) => {
          set((curr) => ({
            gridStore: {
              ...curr.gridStore,
              ...newStore,
            },
          }));
        },
      })),
      { name }
    )
  );

const useLiquidityStore = makeTableStateStore('vega_ledger_store');
```

bglownia, line 89: Store shape is the same as `State` from libs\datagrid\src\lib\use-datagrid-events it will be better if we can import and reuse it

apps/trading/components/orders-container/orders-container.tsx
```diff
@@ -0,0 +1,125 @@
...
+  );
+};
+
+type Store = {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  filterModel?: { [key: string]: any };
+  columnState?: ColumnState[];
+};
+
+const useOrderListStore = create<{
```
bglownia, line 67: We can make it more generic


```
import { immer } from 'zustand/middleware/immer';

const makeMultiinstanceTableStore = (name: string) =>
  create<{
    gridState: Record<string, Store>;
    update: (gridStore: Store, key: string) => void;
  }>()(
    persist(
      subscribeWithSelector(
        immer((set) => ({
          gridState: {},
          update: (newState: Store, key) => {
            set((state) => {
              state.gridState[key] = newState;
            });
          },
        }))
      ),
      {
        name,
      }
    )
  );

const useOrderListStore = makeMultiinstanceTableStore('vega_order_list_store');
```
and then use it in component
```
const StoreKey: Record<Filter, string> = {
  [Filter.Closed]: 'closed',
  [Filter.Open]: 'open',
  [Filter.Rejected]: 'rejected',
};

export const OrdersContainer = ({ marketId, filter }: OrderContainerProps) => {
  [...]
  const storeKey = filter ? StoreKey[filter] : 'all';
  const gridStore = useOrderListStore((store) => store.gridState[storeKey]);
  const update = useOrderListStore((store) => store.update);
  const gridStoreCallbacks = useDataGridEvents(gridStore, (colState) => {
    update(colState, storeKey);
  })
  [...]
}

```

bglownia:
I'm getting empty rows after applying sort by size in closed orders table and then go to rejected and go back to closed
image

apps/trading/components/accounts-container/accounts-container.tsx
mattrussell36: Actually not sure why I'm using subscribeWithSelector. Its not required. Will remove

mattrussell36: Updated to use different selectors

apps/trading/components/orders-container/orders-container.tsx
mattrussell36: I think that each pane should use a single store for all its state, right now that only includes grid state but in future it might also have state for other things (formatting, small balances, chart options, etc etc).

We can use slices (as per zustand documentation) to organise the state and share logic.

mattrussell36:
@bglownia pushed some changes.

- Removed subscribeWithSelector
- Fixed filters. I had actually broken these. The onGridReady function for the order list that set initial filter state was being overridden by the one from useDataGridEvents. I've updated orders-container to use a hook which handles the special cases for the order list
- Added slices to prevent grid store functionality repetition

apps/trading/components/accounts-container/accounts-container.tsx
```diff
@@ -75,3 +84,14 @@ export const AccountsContainer = ({
     </div>
   );
 };
+
+const useAccountStore = create<DataGridSlice>()(
+  persist(
```
bglownia, line 89: it can be just `persist(createDataGridSlice, { name: 'vega_fills_store' })`

bglownia (approved):

apps/trading/components/accounts-container/accounts-container.tsx
mattrussell36: Done

PR #4122, opened 2023-06-16
feat(deal-ticket): improve no collateral validation

### LINKED ISSUE #4017: Reduce order does not work when "most" margin has been used
## Description

So when I put an order that is large, eats up most my margin, and then want to short right after, I am not able to without depositing funds, clicking reduce only does not work either. The only way to get out of the position is to close the whole position in market at once. So my option is to add funds or close it all, I cannot reduce just some of the size. Looks to be the same on mainnet

## Steps to Reproduce

1. Go to a market e.g. https://console.fairground.wtf/#/markets/4cb06ca74c44c2c4aea5018ca9106bd771ee0f8904fa4127bcab637bd17801a1
2. Click on a market order that eats up most your margin
3. Try to close by putting an order with the same size in the opposite position, try to reducing it to the smallest position size.

## Expected behavior

Should be able to close part of my position.

## Screenshots

https://github.com/vegaprotocol/frontend-monorepo/assets/100769129/508d8241-4df1-4e44-82bd-c96e6bccc713

https://github.com/vegaprotocol/frontend-monorepo/assets/100769129/5820fd72-d400-4936-a309-cf5ec071feab

## Device details

- OS: MacOS
- Browser Chrome

bglownia:
# Related issues 🔗

Closes #4017

# Description ℹ️

Check both general and margin account for no collateral validation

### DIFF
--- a/libs/accounts/src/lib/accounts.mock.ts
+++ b/libs/accounts/src/lib/accounts.mock.ts
@@ -151,19 +151,53 @@ export const amendGeneralAccountBalance = (
   marketId: string,
   balance: string
 ) => {
-  if (accounts.party?.accountsConnection?.edges) {
-    const marginAccount = accounts.party.accountsConnection.edges.find(
-      (edge) => edge?.node.market?.id === marketId
-    );
-    if (marginAccount) {
-      const generalAccount = accounts.party.accountsConnection.edges.find(
-        (edge) =>
-          edge?.node.asset.id === marginAccount.node.asset.id &&
-          !edge?.node.market
-      );
-      if (generalAccount) {
-        generalAccount.node.balance = balance;
-      }
-    }
+  if (!accounts.party?.accountsConnection?.edges) {
+    return accounts;
   }
+  const marginAccount = accounts.party?.accountsConnection?.edges?.find(
+    (edge) => edge?.node.market?.id === marketId
+  );
+  if (marginAccount) {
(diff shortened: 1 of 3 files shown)


libs/accounts/src/lib/accounts.mock.ts
```diff
@@ -167,3 +167,18 @@ export const amendGeneralAccountBalance = (
     }
   }
 };
+
+export const amendMarginAccountBalance = (
```
mattrussell36, line 171: Not a huge fan of mutating arguments passed to functions. I'd prefer to just pass an override object to accountsQuery and add create functions. Alternatively I'd keep these amend functions just scoped to the test itself. 

```
const accounts = aaccountsQuery({
  party: {
    accountsConnection: {
      edges: [
        createAccountFragment({
          balance: '100',
          market: {
            id: '1'
          }
        }),
        createAccountFragment({
          ...
        })
      ]
    }
  }
})
```

libs/deal-ticket/src/components/deal-ticket/deal-ticket.tsx
```diff
@@ -195,7 +195,11 @@ export const DealTicket = ({
     }
 
     const hasNoBalance =
+<<<<<<< HEAD
```
mattrussell36, line 198: Cant merge this ;-)

mattrussell36 (changes requested):

mattrussell36 (approved):

PR #4126, opened 2023-06-16
feat(ci): rewrite publish-dist bash to nodejs

mkjmdski:
closes: https://github.com/vegaprotocol/devops-infra/issues/1916


tools/scripts/package-lock.json
mattrussell36, line 1: Might be worth sticking with yarn as that what is the project as a whole is using

tools/scripts/publish-dist-set-vars.test.mjs
```diff
@@ -0,0 +1,170 @@
+import {
+  validateAppName,
+  AppsThatDeployToMainnetFromDevelop,
+  S3BucketNameForApp,
+  getBucketName,
+  getEnvironmentFromDevelop,
+} from './publish-dist-set-vars.mjs';
+import { test } from 'tape';
```
mattrussell36, line 8: Was there a particular reason for using tape? I think it would be beneficial to stick with jest as thats what we've been using across the project already

mattrussell36 (changes requested):

tools/scripts/package-lock.json
edd: Done 👍

tools/scripts/publish-dist-set-vars.test.mjs
edd: I didn't like doing this (Jest takes seconds to start up, Tape is nice and quick and doesn't rely on globals and all sorts of things) but it has the side benefit of making it easy to add in to the main test suite. 

Done 👍

edd:
Marked as Draft - @mkjmdski doesn't like my approach and is goign to demonstrate another one

PR #4140, opened 2023-06-19
feat(governance): show multisig penalty in tooltip

### LINKED ISSUE #3802: Show Multsig penalty in the staking penalties tooltip
## The Chore

This isnt actually an issue since we dont show thie currently but if we do part 2 then IF a multisig issue occured the overall penalty would jump to 100% and it wouldnt be obvious why since we dont currently show this.

We should conditionally show this in the penalty popup.  If `rewardScore.multisigScore` for any validator is zero then add to the popup "Multisig Penalty: 100%", otherwise dont show it.  I would get rid of the Total Penalties in the popup since we already know that, its in the table.  Then you can just append that message when relevant.

image

## Tasks

- [ ] and then what?
- [ ] Etc.

sam-keen:
# Related issues 🔗

Closes #3802

# Description ℹ️

Removes the 'total penalties' field inside the validator penalties tooltip, and if there's a multisig error, shows a 'Multisig penalty' of 100%. 

Screenshot 2023-06-19 at 17 56 07


edd (approved):
👍 I am not convinced that this is the right solution on how to display this information, but the code is good and the information isn't *wrong*, so I am ticking it.

PR #4150, opened 2023-06-20
fix(governance): proposal deadline validation missing

### LINKED ISSUE #4006: Governance - proposal forms deadline validation missing
## Description

There used to be validation in the proposal forms that prevented user trying to create a proposal with a deadline of `0.1` or value higher than `8760` but that seems to be gone now. 

Issue found in all the proposal forms

## Expected behaviour
Form should prevent user from downloading json with a deadline below or higher than the min/max values

Screenshot 2023-05-31 at 15 43 02

sam-keen:
# Related issues 🔗

Closes #4006

# Description ℹ️

Previously, the submit button would highlight issues with deadline validation. However, the messaging left room for improvement. Along the way during the refactor, some validation was lost in removing the submit button and keeping only the 'download json' option. This PR adds validation for when voting or enactment deadlines are set outside the min and max values determined by network parameters, and gives better warning messaging that the proposal will fail with the current values. 

Screenshot 2023-06-20 at 16 00 29
Screenshot 2023-06-20 at 16 00 42


edd (approved):
Nice and clean test coverage 👍

PR #4163, opened 2023-06-22
fix(trading): market info number formatting

### LINKED ISSUE #4166: Market info formatting decimal numbers less than 0 (triggering ratio)
## Issue

`formatNumber`  is not a good solution for numeric values with decimals, it gets rounded.

MadalinaRaicu:
# Related issues 🔗

Closes #4166 

# Description ℹ️

Market info number formatting. 
formatNumber method was used before with 0 dp 
``` return getNumberFormat(formatDecimals).format(Number(rawValue));```  - it rounded values less < 1 (ex. 0.7) to 1 

# Demo 📺

image

format number decimals 0 rounds the number 

image

# Technical 👨‍🔧

### DIFF
--- a/apps/trading-e2e/src/integration/market-info.cy.ts
+++ b/apps/trading-e2e/src/integration/market-info.cy.ts
@@ -182,7 +182,7 @@ describe('market info is displayed', { tags: '@smoke' }, () => {
       .contains('Liquidity monitoring parameters')
       .click();
 
-    validateMarketDataRow(0, 'Triggering Ratio', '0');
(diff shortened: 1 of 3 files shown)


asiaznik (approved):

bglownia (commented):
We should amend e2e tests to check proper formatting.
triggeringRatio is mocked as 0 so this change do not affect rendered value
libs\markets\src\lib\components\market-info\market-info.mock.ts:97
apps\trading-e2e\src\integration\market-info.cy.ts:185

PR #4186, opened 2023-06-26
fix(explorer): liquidity amend order showed incorrect fee

edd:
# Description ℹ️

The fee figure shown on amend LP TXs was not formatted as a percentage. Now it is.

### DIFF
--- a/apps/explorer/src/app/components/txs/details/tx-liquidity-amend.tsx
+++ b/apps/explorer/src/app/components/txs/details/tx-liquidity-amend.tsx
@@ -34,6 +34,8 @@ export const TxDetailsLiquidityAmendment = ({
     txData.command.liquidityProvisionAmendment;
   const marketId: string = amendment.marketId || '-';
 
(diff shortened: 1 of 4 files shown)


apps/explorer/src/app/components/txs/details/tx-liquidity-amend.tsx
```diff
@@ -34,6 +34,8 @@ export const TxDetailsLiquidityAmendment = ({
     txData.command.liquidityProvisionAmendment;
   const marketId: string = amendment.marketId || '-';
 
+  const fee = amendment.fee ? parseFloat(amendment.fee) * 100 : '-';
```
sam-keen, line 37: Suggestion (blocking): multiplying any rounding issues introduced by parseFloat by 100 could result in a fair amount of inaccuracy here. Using Bignumber would be a better approach.

sam-keen (changes requested):

edd:
I just realised I should be pulling this in to someone elses branch so I'm going to close this and open a very similar one

PR #4187, opened 2023-06-26
chore(trading): ag-grid upgrade

bglownia:
Upgrade ag-grid by two major versions from 27 to 29


libs/liquidity/src/lib/liquidity-table.tsx
```diff
@@ -188,7 +188,9 @@ export const LiquidityTable = forwardRef<AgGridReact, LiquidityTableProps>(
       <AgGrid
         style={{ width: '100%', height: '100%' }}
         overlayNoRowsTemplate={t('No liquidity provisions')}
-        getRowId={({ data }) => data.id}
+        getRowId={({ data }: { data: LiquidityProvisionData }) =>
+          `${data.party.id}-${data.status}`
```
mattrussell36, line 192: _I think_ this might be a mistake? Are you deliberately changing this back? Madalina deployed today to fix the LP table which had an issue caused by this ID

libs/accounts/src/lib/accounts-table.spec.tsx
```diff
@@ -44,62 +44,79 @@ describe('AccountsTable', () => {
...
+
     const cells = await screen.findAllByRole('gridcell');
     const expectedValues = ['tBTC', '1,256.00', '1,256.00', '2,512.00', ''];
     cells.forEach((cell, i) => {
       expect(cell).toHaveTextContent(expectedValues[i]);
     });
-    const rows = await screen.findAllByRole('row');
-    expect(rows.length).toBe(6);
+    const rows = container.querySelector('.ag-center-cols-container');
+    expect(rows?.childElementCount).toBe(1);
```
mattrussell36, line 61: We are losing some coverage here as we aren't checking the correct number of rows are rendered. Not really bothered though as thats testing the implementation of aggrid

mattrussell36 (changes requested):

mattrussell36:
@bglownia This looks all good baring the one comment on the LP table.

libs/accounts/src/lib/accounts-table.spec.tsx
bglownia: There is only one row rendered. Previously number of rows was checked by counting elements with `role="row"`  was 6 because there are multiple elements which have this row in whole table. 
Now we are checking exact number of rows, which should be 1

PR #4190, opened 2023-06-26
feat(governance): add toggle for closed proposals list

### LINKED ISSUE #4145: Add toggle for network upgrades vs governance proposals to proposals list
## Chore 

Completed network upgrades always show at the top of the closed proposals list because they do not have an enacted time and "hide" the governance proposals. We cant easily fix this unless Core make changes ... which they wont. 

This chore is to add a toggle in the CLOSED PROPOSALS section. 

## Acceptance criteria 
- [ ] There is a way to toggle between closed network governance proposals and closed network upgrade proposals
- [ ] When I toggle to network upgrades, I only see closed network upgrade proposals 
- [ ] When I toggle to network governance I only see closed network governance proposals 
- [ ] Open proposals is unaffected by this toggle 
- [ ] The closed network proposals are sorted with the most recent first 
- [ ] When I use the filter proposals search, only matching proposals will show on the page
- [ ] When I use the filter proposals search, and there are only results in one category of closed proposals, not the other, the view toggles to that view (right now it's only ever going to have results in network governance)
- [ ] I can see how many search results there are e.g. "(1)"
- [ ] When no proposals match my search, there is a message that tells me no proposals were found

## Tasks 

- [x] Sketch 
- [ ] Discuss 
- [ ] Implement
- [ ] Test
- [ ] 

## Design 
https://www.figma.com/file/X9l0t8TZUF1tU8u0ILV3ws/Governance?type=design&node-id=2838-18127&mode=design&t=O0exK4U6stOxMWet-0

## Sketch 
Governance (  Token Utility) - Toggle network upgrade _ governance proposal

## Background

* Add the toggle pattern from the validators page
* Values on it would be `Network Governance / Network Upgrades`.
* OPEN PROPOSALS should be left alone, because its unlikely that many will be open at once, and we dont want to hide anything.  the filter should only apply to CLOSED PROPOSALS. 
* The network proposals should be sorted by blockheight in descending order, so most recent first.

sam-keen:
# Related issues 🔗

Closes #4145

# Description ℹ️

- Separates closed governance proposals from closed network upgrade proposals, using a toggle to view.
- Also improved proposal list filter UI, and ensures filters work correctly with network upgrades (first pass here is that network upgrades are hidden when the filter is engaged, as they lack the required fields for filtering)
- As part of testing this work, I realised it would be more useful for rejected proposals to have a link at the top to take the user back to the rejected proposals list, rather than the non-rejected one, so added that

# Demo 📺

Screenshot 2023-06-26 at 17 28 42

Screenshot 2023-06-26 at 17 28 50

Screenshot 2023-06-27 at 15 20 22

### DIFF
--- a/apps/governance/src/routes/proposals/components/proposals-list/proposals-list.tsx
+++ b/apps/governance/src/routes/proposals/components/proposals-list/proposals-list.tsx
@@ -216,7 +216,7 @@ export const ProposalsList = ({
                   >
                     <div className="w-[440px]">
                       <Toggle
-                        name="validators-view-toggle"
+                        name="closed-proposals-toggle"
                         toggles={[
                           {
                             label: t(
--- a/apps/governance/src/routes/proposals/components/proposals-list/proposals-list.tsx
+++ b/apps/governance/src/routes/proposals/components/proposals-list/proposals-list.tsx
@@ -1,6 +1,6 @@
 import orderBy from 'lodash/orderBy';
 import { isFuture } from 'date-fns';
-import { useState, useMemo } from 'react';
+import { useState, useMemo, useEffect } from 'react';
 import { useTranslation } from 'react-i18next';
 import { Heading, SubHeading } from '../../../../components/heading';
 import { ProposalsListItem } from '../proposals-list-item';
@@ -75,6 +75,14 @@ export const ProposalsList = ({
     useState<ClosedProposalsViewOptions>(
       ClosedProposalsViewOptions.NetworkGovernance
     );
+
+  useEffect(() => {
+    if (filterString.length > 0) {
+      // If the filter is engaged, ensure the user is viewing governance proposals,
+      // as network upgrades do not have IDs to filter by and will be excluded.
+      setClosedProposalsView(ClosedProposalsViewOptions.NetworkGovernance);
+    }
+  }, [filterString]);
 
   const sortedProposals: SortedProposalsProps = useMemo(() => {
     const initialSorting = proposals.reduce(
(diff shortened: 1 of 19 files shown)


apps/governance/src/routes/proposals/components/proposals-list/proposals-list.tsx
```diff
@@ -54,13 +59,30 @@ export const orderByUpgradeBlockHeight = (
...
   lastBlockHeight,
 }: ProposalsListProps) => {
   const { t } = useTranslation();
   const [filterString, setFilterString] = useState('');
+  const [closedProposalsView, setClosedProposalsView] =
+    useState<ClosedProposalsViewOptions>(
+      ClosedProposalsViewOptions.NetworkGovernance
+    );
+
+  useEffect(() => {
```
mattrussell36, line 79: no reason for this affect. Call `setClosedProposalsView` directly where needed

```
<ProposalsListFilter
      filterString={filterString}
      setFilterString={(value) => {
        setFilterString(value)
        if (value.length > 0) {
          setClosedProposalsView(ClosedProposalsViewOptions.NetworkGovernance)
        }
      }}
    />
```

See: https://react.dev/learn/you-might-not-need-an-effect

mattrussell36 (changes requested):

apps/governance/src/routes/proposals/components/proposals-list/proposals-list.tsx
sam-keen: Ah damnit, fell into that trap! Thanks

PR #4196, opened 2023-06-27
chore(trading): migrate nx to latest version

bglownia:
Migrate nx from version 14 to 16

`nx migrate latest` was executed this updated not only nx but also:
next from version 12 to 13
storybook from version 6 to 7
jest from version 27 to 29
webpack from version 5 to 6

Custom nx executors needs to be removed, `yarn nx run {command} --env {env}` do not work any more
Please use yarn `env-cmd -f .\apps\{app-name}\.env.{env} yarn nx run {command}`


edd (commented):
This looks like it was really tedious. Thank you for doing it!

mattrussell36 (approved):
Overall less config and some nice tidying of of stuff in here :-)

PR #4199, opened 2023-06-27
fix(ui-toolkit): healthbar tooltips text colour

### LINKED ISSUE #4118: Health bar numbers issues
We fixed some decimals issues on the main columns but some of these persist _within_ the healthbar. 

The heathbar seems to have two hover-overs.

### Target Stake line 

This one seems fine. It matches target stake, but we should remember to check this picks up quantum formatting when we do that. 

### Over a block of liquidity 

These are hard to interpret currently because there is only one commitment on each one.

1. The LP Fee is incorrectly formatted .... needs to be x 100. 
2. I'm not sure what the second part is showing.  
    a. For BTC & LINK markets it seems to show the supplied stake, but it looks like it needs to be formatted with `asset decimals` and isnt. 
    b. For ETH its doing something else. The decimals are wrong but its also showing 32,320 instead of 50,000...and I have no idea whatsoever where it is getting that number! 

Image

ciaran-:
# Related issues 🔗

Closes #4118 

Specify the text colour for tooltips rendered by the Healthbar.

### DIFF
--- a/libs/ui-toolkit/package.json
+++ b/libs/ui-toolkit/package.json
@@ -1,4 +1,4 @@
 {
   "name": "@vegaprotocol/ui-toolkit",
-  "version": "0.12.5"
+  "version": "0.12.6"
 }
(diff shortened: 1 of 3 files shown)


ciaran-:
No longer needed - fixed at use instead

MadalinaRaicu (approved):

Issue #4215, opened 2023-06-29
Improve block explorer pagination


edd:
## The Chore

After #4070 is complete, the inifinite loading will be gone! Pagination will rule again. Except it relies on a client side hack, because the API didn't support pagination in a way that allowed for going backwards nicely. This is coming in https://github.com/vegaprotocol/vega/pull/8627 - which is merged, but not redeployed. When it is, we can remove the hack and make some improvements

## Tasks

- [ ] Remove `previousCursor` from the state
- [ ] Update the URL with the cursor each time the page moves
- [ ] Take the cursor from the URL if it is set
- [ ] Persist the filters in the URL along with the cursor
- [ ] Take the filter (currently just one) from the URL (https://github.com/vegaprotocol/vega/pull/8543)

PR #4217, opened 2023-06-29
feat(explorer): remove infinite tx loader

### LINKED ISSUE #4070: Transaction list: remove infinite loading
## The Chore

Inifinite loading works in some places, but it is not a good system for looking through history for transactions. We should remove it and switch to a manually paged list of transactions

## Tasks
- [ ] This should change all TX lists - the one on the party page and the one on the transactions page
- [x] Remove infinite loading wrapper
- [ ] Remove react-window wrapper around table
- [ ] Add more standard pagination controls. As a starting point, let's 
   - [ ] This can be iterated on, but must have:
   - [ ] Previous page button, which should be disabled on page 1
   - [ ] Current page number
   - [ ] Next page button
- [x] Page size should be 25
    - [ ] A follow up ticket can make this configurable with a dropdown

## Notes
- This doesn't need to change the layout. That's a separate ticket.
- It can change the layout if it's better done all at once.

edd:
- feat(explorer): remove infinite loader on tx index page
- fix(utils): 0 length end should be valid
- feat(explorer): paging
- feat: progress on tx list
- feat: tx list reorganisation
- feat: tweak vote type indicator to match smaller size
- feat: functional back navigation
- feat: layout tweaks for the transaction list
- chore: move nav out to component and add it to party txs
- fix(explorer): address broken tests

# Related issues 🔗

Closes #4070

# Description ℹ️

Removes infinite scroll for transactions, in favour of normal pagination. #4215 will improve on this functionality.

This also updated PartyLink to render a validator's name instead of their key in the transaction list.

# Demo 📺
screencast


libs/utils/src/lib/format/strings.spec.ts
```diff
@@ -28,7 +29,7 @@ describe('shorten', () => {
     { i: '12345678901234567890', l: 10, o: `123456789${ELLIPSIS}` },
     { i: '12345678901234567890', l: 20, o: `1234567890123456789${ELLIPSIS}` },
     { i: '12345678901234567890', l: 30, o: `12345678901234567890` },
-  ])('should shorten given string by specific limit', ({ i, l, o }) => {
+  ])(`should shorten given string by specific limit`, ({ i, l, o }) => {
```
dexturr, line 32: chore (non-blocking): multiple tests with same name is probably not dresired, can we add the parameters in the name or similar to make this easier to debug in future?

dexturr (approved):

ettec (approved):

PR #4228, opened 2023-06-30
feat(explorer): add pegged order details

### LINKED ISSUE #3489: Show reference, offset and current price for pegged orders
Block explorer currently uses what looks like a Limit order view for pegged orders. this results in a few issues...

## Story

~~As a~~ When checking a transaction with a pegged order, 
I want to see the reference and offset that was used (as well as the current price)
So that I can verify Vega has processed my order as I intended (i.e. that the ref and offset I entered match what was received, and I've got the number of digits after the decimal right).

Here is what it currently looks like: [e.g.](https://explorer.fairground.wtf/txs/98795E58C492F943ECD3E4AAC6C41D211D76D8EECAD1CC06FC86593B74AC0FA9)

Screenshot 2023-04-21 at 12 04 14

Note that a pegged order doesn't have a price, it has an offset from a reference price. The price changes as the reference moves.

I would expect to see these offset and reference as primary, and maybe the current price (subscription) as secondary. 

## Acceptance Criteria

- [ ] the pegged order view shows the reference that was used
- [ ] the pegged order view shows the offset that was used

## Tasks

- [x] Explore and sketch

GTC Buy 10.01 @ +0.01 from Mid. (currently priced 1,915.86 USD) 

or

GTC Sell 20.01 @ 1.01 from Mid. (Filled at 1,915.86 USD) 

but it could get filled a multiple prices... So maybe it should not show a price. or list all fill prices

GTC Sell 1.01 @ 0.01 from Best Bid. 
Filled at 0.50 * 1,915.86 USD
Filled at 0.50 * 1,915.87 USD
Open 0.01 * 1,915.88 USD

however:  this live data could be left out until https://github.com/vegaprotocol/frontend-monorepo/issues/2401

GTC Sell 1.01 @ 0.01 from Best Bid

- [ ] Visual Design
- [ ] Team review
- [ ] Etc.

edd:
- feat(explorer): add peg details

# Related issues 🔗

Closes #3489

# Description ℹ️

If an order TX is a pegged order, render the peg reference and offset

# Demo 📺
Screenshot 2023-06-30 at 16 51 02


edd:
Marking as draft to change the way this is displayed:

Currently: `0.35 from mid`
Next: `Mid + 0.35` / `Mid - 0.35` 

Depending on https://docs.vega.xyz/mainnet/concepts/trading-on-vega/orders#values-available-for-pegged-orders

apps/explorer/src/app/components/txs/details/order/tx-order-peg.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { addDecimalsFormatNumber } from '@vegaprotocol/utils';
+
+export interface TxDetailsOrderProps {
+  offset: string;
+  reference: VegaPeggedReference;
+  marketId: string;
+  side: Side;
+}
+
+export function getSettlementAsset(
```
dexturr, line 16: chore (non-blocking): function name does not seem to be doing as it suggests. Suggest getMarketDecimals

dexturr (approved):

apps/explorer/src/app/components/txs/details/order/tx-order-peg.tsx
edd: Good spot. This function totally changed functionality and I lazily didn't rename it

jeremyletang (approved):

pscott31 (approved):

PR #4240, opened 2023-07-03
chore: remove links from disclaimer

### LINKED ISSUE #3881: Remove links from footer in disclaimer
## The Chore

You already have connected a wallet when you see this disclaimer, so the footer links are not necessary. This chore is to remove the two links "Get a Vega wallet" and "Having trouble?" from the footer of the attached: 

Screenshot 2023-05-22 at 16 22 16

## Tasks

- [ ] Acs adjusted/created
- [ ] Dev
- [ ] Test

asiaznik:
# Related issues 🔗

Closes #3881 

# Description ℹ️

* removed links from the risk message dialog after connecting to the vega wallet


edd (approved):
👍 I was going to ask if we could just remove the footer component, but I see it's still used in one place

PR #4260, opened 2023-07-05
feat(governance): diff highlighter for update market proposals

### LINKED ISSUE #4092: Simple diff highlight for update market proposals view
The update market proposal equivalent of https://github.com/vegaprotocol/frontend-monorepo/issues/3808

## Story

As a user
I want to see the specification of the market being updated with the update highlighted
So that I can more clearly understand the impact of the proposal

## Acceptance Criteria

- [x] I can see the ~market specification in full~ fields that have been changed
- [x] There is a simple way to visually highlight / compare the difference with this market specification from the original markets specification e.g. using inline github style diff

## Background 

Breaks down into two main tasks:

### Technical Steps

- Maintain a list of the immutable fields to replicate the API
- Get the REST version of the proposal changes, as well as the original market's proposal changes, and the most recently enacted update of the market, if one exists
- Ensure the two version of the proposal to diff (the current proposal vs either the most recently enacted, or the original) contain all required fields, and that any values in the current proposal that are ignored by the API (i.e. won't change) are not reflected as things that will change in the diff

### How to display

Here we may need @ckpringle to do some design, but my thinking is there are two options:

1. Display ONLY the fields that have changed with some field / old value / new value formatting 
2. Display the entire new market with some additional annotation that shows which things have changed and their old values

~My preference is 2 because if we do this it will also solve the problem for Successor markets ... where I dont think 1 would be enough detail for that.  2 is also pretty easy given the way market updates are submitted you get the entire market definition in ... so you're just displaying that in full in the normal accordion, and adding some notation.~

We will go with option 1 here.  Successor markets is going to be handled slightly differently so worry about that later.

## Tasks

- [x] Identify possible libraries https://github.com/otakustay/react-diff-view, https://pschmucker.github.io/modiffy/
- [x] Work out the best approach 
- [x] UX (if needed)
- [x] Design (if needed)
- [x] Implementation
- [ ] Code review
- [ ] QA review

## Sketches

Governance (  Token Utility) - Simple diff

sam-keen:
# Related issues 🔗

Closes #4092

# Description ℹ️

A diff highlighter, being used to show the differences between an 'update market' governance proposal, and the most recently enacted proposal update if one exists, or the original market proposal if not.

Screenshot 2023-07-26 at 21 53 31


edd:
Visually we have a minor issue which I think should not block this, but does need to be resolved. These properties are not in the market update (`/proposals/f664002024eb4f071f946bc591b1d07798bd941f038a83a97d26802f657abe76`)

Screenshot 2023-07-10 at 12 01 48
Screenshot 2023-07-10 at 12 01 43

However, they're not being removed, they're *just not being changed*. So this view is technically incorrect. This is related to https://github.com/vegaprotocol/vega/issues/5750 - *some* fields can be omitted from an update, others can't. I'm not sure how to resolve this. The reason I think it's non-blocking is that the view is still useful, if wrong - most of the highlighted stuff IS changing and IS useful.

Full change:
Screenshot 2023-07-10 at 12 03 19

apps/governance/src/routes/proposals/components/proposal-market-changes/proposal-market-changes.tsx
```diff
@@ -0,0 +1,43 @@
...
+        dataTestId={'proposal-description-toggle'}
+      >
+        <SubHeading title={t('updatesToMarket')} />
+      </CollapsibleToggle>
+
+      {showChanges && (
+        <div className="mb-6">
+          <JsonDiff
+            left={left || {}}
+            right={right || {}}
```
edd, line 36: **nitpick (non-blocking)**: Defaulting to `{}` doesn't make a lot of sense, it would be better to force `showChanges` to false if `left`/`right` is unset.

apps/governance/src/routes/proposals/components/proposal/proposal.tsx
```diff
@@ -36,13 +37,16 @@ export interface ProposalProps {
   assetData?: AssetQuery | null;
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   restData: any;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  originalMarketProposalRestData?: any;
```
edd, line 41: **nitpick (non-blocking)**: I feel like somewhere we will have the type for a market proposal, and that if we did type it properly we might catch errors in the future (say `proposal.terms` changes). But the way the data is handled is careful enough to cope and the JSONDiff component would be fine so I think it's ok *this time*

edd (approved):
Nitpicks, and the main comment that says 'This is a bit broken but still work merging'.

sam-keen:
converted back to draft until we understand how to best deal with omitted values

asiaznik:
question (blocking?): I see that you compare the update values with the original proposal for a market, what if the market had already couple of update proposals enacted?
Wouldn't it be better to get all enacted proposals for a market, merge those values in "original" proposal json and then compare with the new update proposal?

edd (approved):
✅ Updates look good too. Looking forward to seeing this in action!

PR #4265, opened 2023-07-06
feat(trading): add stop orders table and form

bglownia:
# Related issues 🔗

Closes #[Issue number here]

# Description ℹ️

# Demo 📺

# Technical 👨‍🔧

### DIFF
--- a/libs/orders/src/lib/components/stop-orders-manager/stop-orders-manager.tsx
+++ b/libs/orders/src/lib/components/stop-orders-manager/stop-orders-manager.tsx
@@ -1,0 +1,65 @@
+import { t } from '@vegaprotocol/i18n';
+import { useCallback, useRef } from 'react';
+import type { AgGridReact } from 'ag-grid-react';
+import { StopOrdersTable } from '../stop-orders-table/stop-orders-table';
+import type { useDataGridEvents } from '@vegaprotocol/datagrid';
+import { useVegaTransactionStore } from '@vegaprotocol/wallet';
+import type { StopOrder } from '../order-data-provider/stop-orders-data-provider';
+import { useDataProvider } from '@vegaprotocol/data-provider';
+import { stopOrdersWithMarketProvider } from '../order-data-provider/stop-orders-data-provider';
+
+export interface StopOrdersManagerProps {
+  partyId: string;
+  onMarketClick?: (marketId: string, metaKey?: boolean) => void;
+  isReadOnly: boolean;
+  gridProps?: ReturnType<typeof useDataGridEvents>;
+}
+
+export const StopOrdersManager = ({
+  partyId,
+  onMarketClick,
+  isReadOnly,
+  gridProps,
+}: StopOrdersManagerProps) => {
+  const gridRef = useRef<AgGridReact | null>(null);
+  const create = useVegaTransactionStore((state) => state.create);
+  const variables = { partyId };
+
+  const { data, error } = useDataProvider({
+    dataProvider: stopOrdersWithMarketProvider,
+    variables,
+    update: ({ data }) => {
+      if (data && gridRef.current?.api) {
+        gridRef.current.api.setRowData(data);
+        return true;
+      }
+      return false;
+    },
+  });
+
+  const cancel = useCallback(
+    (order: StopOrder) => {
+      if (!order.market) return;
+      create({
+        stopOrdersCancellation: {
+          stopOrderId: order.id,
+          marketId: order.market.id,
+        },
+      });
+    },
+    [create]
+  );
+
+  return (
+    <StopOrdersTable
+      rowData={data}
+      ref={gridRef}
+      onCancel={cancel}
+      onMarketClick={onMarketClick}
+      isReadOnly={isReadOnly}
+      suppressAutoSize
+      overlayNoRowsTemplate={error ? error.message : t('No stop orders')}
+      {...gridProps}
+    />
+  );
+};
--- a/apps/trading/client-pages/market/trade-grid.tsx
+++ b/apps/trading/client-pages/market/trade-grid.tsx
@@ -87,6 +87,11 @@ const MarketBottomPanel = memo(
                   <TradingViews.orders.component marketId={marketId} />
                 </VegaWalletContainer>
               </Tab>
+              <Tab id="stop-orders" name={t('Stop Orders')}>
+                <VegaWalletContainer>
+                  <TradingViews.stopOrders.component />
+                </VegaWalletContainer>
+              </Tab>
               <Tab id="fills" name={t('Fills')}>
                 <VegaWalletContainer>
                   <TradingViews.fills.component onMarketClick={onMarketClick} />
@@ -159,6 +164,11 @@ const MarketBottomPanel = memo(
               <TradingViews.orders.component marketId={marketId} />
             </VegaWalletContainer>
           </Tab>
+          <Tab id="stop-orders" name={t('Stop Orders')}>
+            <VegaWalletContainer>
+              <TradingViews.stopOrders.component />
+            </VegaWalletContainer>
+          </Tab>
           <Tab id="fills" name={t('Fills')}>
             <VegaWalletContainer>
               <TradingViews.fills.component onMarketClick={onMarketClick} />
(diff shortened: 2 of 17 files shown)


libs/deal-ticket/src/components/deal-ticket/deal-ticket-stop-order.tsx
```diff
@@ -0,0 +1,600 @@
...
+  if (storedFormValues?.size && rawSize !== storedFormValues?.size) {
+    setValue('size', storedFormValues.size);
+  }
+  if (storedFormValues?.price && rawPrice !== storedFormValues?.price) {
+    setValue('price', storedFormValues.price);
+  }
+
+  const isPriceTrigger = triggerType === 'price';
+  const size = removeDecimal(rawSize, market.positionDecimalPlaces);
+  const price = useMemo(() => {
```
mattrussell36, line 118: I'd not use useMemo here.. just
```
const price = marketPrice && getDerivedPrice(...)
```

libs/deal-ticket/src/components/deal-ticket/deal-ticket-stop-order.tsx
```diff
@@ -0,0 +1,600 @@
...
+
+  const notionalSize = useNotionalSize(
+    price,
+    size,
+    market.decimalPlaces,
+    market.positionDecimalPlaces
+  );
+
+  useEffect(() => {
+    const subscription = watch((value, { name, type }) => {
```
mattrussell36, line 148: In the other deal ticket we have the reverse approach. Where the zustand store is the 'source of truth' for deal ticket data. I'd prefer it this way around, but perhaps we refactor in the follow up tickets

libs/deal-ticket/src/components/deal-ticket/deal-ticket-stop-order.tsx
```diff
@@ -0,0 +1,600 @@
...
+  const expire = watch('expire');
+  const triggerType = watch('triggerType');
+  const triggerPrice = watch('triggerPrice');
+  const timeInForce = watch('timeInForce');
+  const type = watch('type');
+  const rawPrice = watch('price');
+  const rawSize = watch('size');
+
+  if (storedFormValues?.size && rawSize !== storedFormValues?.size) {
+    setValue('size', storedFormValues.size);
```
mattrussell36, line 110: Its a bit of a code smell that we are setting state directly in the body of the function component

I think you could avoid this if we make the fields fully controlled by the zustand store _only_, and then update react-hook-form when they change (like how the default deal ticket works currently)

That said, I wonder if not using react-hook-form is the way to go.. seeing as we are fully controlling every field 🤔

libs/deal-ticket/src/components/deal-ticket/deal-ticket.spec.tsx
```diff
@@ -402,7 +374,7 @@ describe('DealTicket', () => {
     render(generateJsx());
 
     act(() => {
-      screen.getByTestId(`order-type-${Schema.OrderType.TYPE_MARKET}`).click();
+      screen.getByTestId('order-type-Market').click();
```
mattrussell36, line 377: userEvent?

libs/deal-ticket/src/components/deal-ticket/type-selector.tsx
```diff
@@ -1,32 +1,118 @@
...
+            className={classNames('rounded', {
+              'bg-vega-clight-500 dark:bg-vega-cdark-500': value === itemValue,
+            })}
+          >
+            {label}
+          </button>
+        </RadioGroup.Item>
+      ))}
+      {FLAGS.STOP_ORDERS && (
+        <DropdownMenu.Root>
```
mattrussell36, line 68: Can we not just use a more 'default' TradingDropdown

TradingDropdown lets you pass any trigger element now so that the styleing of the trigger button is not constricted

libs/deal-ticket/src/hooks/use-type-store.ts
```diff
@@ -0,0 +1,28 @@
...
+import { persist, subscribeWithSelector } from 'zustand/middleware';
+
+export enum DealTicketType {
+  Limit = 'Limit',
+  Market = 'Market',
+  StopLimit = 'StopLimit',
+  StopMarket = 'StopMarket',
+}
+
+export const useDealTicketTypeStore = create<{
```
mattrussell36, line 11: IMO this should just be part of the main deal ticket store, we shouldn't need a separate store just for the type

libs/orders/src/lib/components/stop-orders-manager/stop-orders-manager.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const gridRef = useRef<AgGridReact | null>(null);
+  const create = useVegaTransactionStore((state) => state.create);
+  const variables = { partyId };
+
+  const { data, error, reload } = useDataProvider({
+    dataProvider: stopOrdersWithMarketProvider,
+    variables,
+    update: ({ data }) => {
+      if (data && gridRef.current?.api) {
+        gridRef.current.api.setRowData(data);
```
mattrussell36, line 35: I don't think we need to use setRowData now that we are using colDefs? I think we can just pass data directly to the grid as a prop?

apps/trading/components/orderbook-container/orderbook-container.tsx
```diff
@@ -1,20 +1,26 @@
 import { OrderbookManager } from '@vegaprotocol/market-depth';
 import { useCreateOrderStore } from '@vegaprotocol/orders';
 import { ViewType, useSidebar } from '../sidebar';
+import { useStopOrderFormValues } from '@vegaprotocol/deal-ticket';
 
 export const OrderbookContainer = ({ marketId }: { marketId: string }) => {
   const useOrderStoreRef = useCreateOrderStore();
   const updateOrder = useOrderStoreRef((store) => store.update);
+  const updateStoredFormValues = useStopOrderFormValues(
```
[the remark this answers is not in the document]
bglownia: There is change request for that. I wanted to limit interference with exising deal ticked implementation to minimum.

libs/deal-ticket/src/components/deal-ticket/deal-ticket.spec.tsx
bglownia: This is exising test, just updating selector

libs/deal-ticket/src/hooks/use-type-store.ts
bglownia: I agree, it will be refactored in follow up PR

pscott31 (approved):

libs/deal-ticket/src/components/deal-ticket/deal-ticket-stop-order.tsx
bglownia: IMHO going using zustand store as source only makes things to complicated I prefere to write to store as side effect of change in form.

apps/trading/components/orderbook-container/orderbook-container.tsx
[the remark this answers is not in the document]
bglownia: It is planned to merge all three stores together

libs/deal-ticket/src/components/deal-ticket/deal-ticket-stop-order.tsx
bglownia: I wanted to refactor it to make to not write through zustand store

libs/deal-ticket/src/components/deal-ticket/deal-ticket-stop-order.tsx
mattrussell36: Where should the 'true' ticket state live then? I don't think it should be within the useForm hook itself (as it is now in this PR). What about in future if we want to place trades from outside of the deal ticket (eg trading from the chart), or perhaps as we do now with arming from the orderbook? Now there is also no guarantee the deal ticket form will even be showing. 

Overall I think we should probably just using a fully controlled form (controlled from the zustand store), probably completely removing useForm. Not for this PR though 😛

libs/deal-ticket/src/components/deal-ticket/deal-ticket-stop-order.tsx
mattrussell36: Lets not block over this. We'll address in the follow up tickets

libs/deal-ticket/src/components/deal-ticket/deal-ticket-stop-order.tsx
mattrussell36: Again, resolving so as not to block

PR #4329, opened 2023-07-14
fix(ui-toolkit): toolkit bugs

dexturr:
# Related issues 🔗

Minor UI toolit bugs

# Description ℹ️

- Has error did not highlight border & text with correct colors
- Checkboxes were centre aligned instead of aligned at start when overflowing
- Secondary buttons should always have white text
- Ternary buttons should always have black text

### DIFF
--- a/libs/ui-toolkit/src/components/button/button.tsx
+++ b/libs/ui-toolkit/src/components/button/button.tsx
@@ -30,14 +30,14 @@ const primary = [
   'enabled:active:bg-vega-yellow-550 enabled:active:border-vega-yellow-550',
 ];
 const secondary = [
-  'text-white dark:text-black',
+  'text-white',
   'border-vega-pink',
   'dark:bg-vega-pink bg-vega-pink-550',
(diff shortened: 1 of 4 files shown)


libs/ui-toolkit/src/components/button/button.tsx
```diff
@@ -30,14 +30,14 @@ const primary = [
   'enabled:active:bg-vega-yellow-550 enabled:active:border-vega-yellow-550',
 ];
 const secondary = [
-  'text-white dark:text-black',
```
MadalinaRaicu, line 33: ahm, are you sure this works fine on every theme?

dexturr: As per designs:

Screenshot 2023-07-14 at 12 44 32
ge per theme np?

dexturr: (Secondary is pink)

MadalinaRaicu: alright

MadalinaRaicu (approved):

PR #4330, opened 2023-07-14
feat(ci): migrate ci-cd trigger to python

mkjmdski:
closes: https://github.com/vegaprotocol/frontend-monorepo/4072


edd (approved):
👍 My branch is replaced by this python setup, which is *hopefully* more approachable to all users (ops and devs). 

A conversation me and @mkjmdski had about tests was that it should be hard to break this pipeline, to help devs get to grips with it. I did that with tests up front, whereas @mkjmdski doesn't want to favour simpler code - so we may add tests down the line.

PR #4378, opened 2023-07-25
fix(deal-ticket): 4046 wrong margin requirement in order

### LINKED ISSUE #4046: Stagnet1 - a market in opening auction has extremely high margin requirement in console
## Description

Console seems to calculate required margin incorrectly in stagnet1, see the difference when I use wallet command to send order, it works but Console seems to thinkt it should not.

## Steps to Reproduce

1. Go to a market in opening auction on stagnet1 e.g. https://trading.stagnet1.vega.rocks/#/markets/e683ab0471ebf10ca658e05b7853e587e8dc372c3803eb8b9127873e52a8cfd0
2. Click on sell order and see if you get it to work, probably says that you need extreme amount of margin.
3. Do the same order using wallet command and you will see that it is accepted
`{
    "pubKey": "69464e35bcb8e8a2900ca0f87acaf252d50cf2ab2fc73694845a16b7c8a0dc6f",
    "propagate": true,
    "orderSubmission": {
        "marketId": "e683ab0471ebf10ca658e05b7853e587e8dc372c3803eb8b9127873e52a8cfd0",
        "price": "186355",
        "size": "70",
        "side": "SIDE_SELL",
        "timeInForce": "TIME_IN_FORCE_GTC",
        "type": "TYPE_LIMIT"
    }
}
`

## Device details

- OS: MacOS
- Browser Chrome

## Video

https://github.com/vegaprotocol/frontend-monorepo/assets/100769129/7939bca6-e308-4697-a9d3-3a72851c216b

https://github.com/vegaprotocol/frontend-monorepo/assets/100769129/b0925b06-d7b1-461f-8889-764b1b13bbb9

macqbat:
# Related issues 🔗

Closes #4046 

# Description ℹ️

Add filtering of active orders being sent as parameter to `estimatePosition`

# Demo 📺

# Technical 👨‍🔧


libs/deal-ticket/src/components/deal-ticket/deal-ticket.tsx
```diff
@@ -142,7 +142,7 @@ export const DealTicket = ({
   );
   const { data: activeOrders } = useDataProvider({
     dataProvider: activeOrdersProvider,
-    variables: { partyId: pubKey || '' },
+    variables: { partyId: pubKey || '', marketId: market.id },
```
mattrussell36, line 145: Good find 👍. Something to consider though.. It may be better to not change the variable and instead filter the resulting orders (perhaps use a derived data provider, or a wrapping hook), thus avoiding keeping two stores in sync at the same time. By changing the variable I think we may end up with duplicated orders in two different stores?

@bglownia would love some input here

bglownia (approved):

libs/deal-ticket/src/components/deal-ticket/deal-ticket.tsx
mattrussell36: Looks like activeOrdersProvider is already handling the filtering (after the query resolves) internally 👍

PR #4442, opened 2023-07-31
feat(trading): 3114 pane context

### LINKED ISSUE #3114: Pane/Tab context menu
## The Chore

We need a method of allowing users to execute actions from within a pane, where those actions are specific to the content of the pane itself.

Some examples of these 'actions' could be:

- Changing the view to display a settings area for that data grid or chart
- Executing transactions such as cancelling all orders or depositing
- Changing chart overlays and studies
- Changing deal ticket types
- Changing number precisions

## Tasks

- [x] Design
- [x] Review
- [ ] Dev (ui-toolkit)
  - [ ] Adjust tabs component to allow for rendering buttons/dropdowns/anything in the right hand section
  - [ ] Add way of passing actions/state from the tab bars to the rendered view below
  - [ ] Implement for chart overlays and studies

- Open to any ideas here but my initial thoughts here are that this is one or more contextual (the context being the current selected tab) menus (possibly similar to our Dropdown component) from the top right of each tab bar. Alternatively this menu could exist within the tab itself.
- [Some examples gathered on Miro](https://miro.com/app/board/uXjVPUenJos=/?moveToWidget=3458764548123344271&cot=14)

mattrussell36:
# Related issues 🔗

Closes #3114 

# Description ℹ️

- Adds support to the tabs component for additional tab menu buttons
- Moves chart controls to tab section

# Demo 📺

Screenshot 2023-08-02 at 11 08 01

Screenshot 2023-08-02 at 11 08 49


libs/ui-toolkit/src/components/tabs/tabs.tsx
```diff
@@ -17,24 +18,40 @@ export const Tabs = ({
   onValueChange,
   ...props
 }: TabsProps) => {
+  // Bunch of refs in order to detect wrapping in side the tabs so that we
+  // can apply a bg color
+  const wrapperRef = useRef<HTMLDivElement | null>(null);
```
mattrussell36, line 23: Adding the refs and the resize observer is only needed to set the background color when the menu wraps underneath the tabs. This very much annoys me. LMK if you have a better idea

MadalinaRaicu (approved):

libs/ui-toolkit/src/components/tabs/tabs.tsx
asiaznik: the menu seems to be “wrapped” only if it’s not the direct sibling of [role=tablist], hence in this case:
```
className={classNames(
            'flex-1 p-1',
            'bg-vega-clight-700 dark:bg-vega-cdark-700',
            'peer-[[role="tablist"]]:bg-inherit dark:peer-[[role="tablist"]]:bg-inherit'
          )
```

would do the same as what you're trying to do with refs

asiaznik:
```
diff --git a/libs/ui-toolkit/src/components/tabs/tabs.tsx b/libs/ui-toolkit/src/components/tabs/tabs.tsx
index d20720556..960f07f08 100644
--- a/libs/ui-toolkit/src/components/tabs/tabs.tsx
+++ b/libs/ui-toolkit/src/components/tabs/tabs.tsx
@@ -2,11 +2,10 @@ import * as TabsPrimitive from '@radix-ui/react-tabs';
 import {
   useLocalStorageSnapshot,
   getValidItem,
-  useResizeObserver,
 } from '@vegaprotocol/react-helpers';
 import classNames from 'classnames';
 import type { ReactElement, ReactNode } from 'react';
-import { Children, isValidElement, useRef, useState } from 'react';
+import { Children, isValidElement, useState } from 'react';
 export interface TabsProps extends TabsPrimitive.TabsProps {
   children: (ReactElement<TabProps> | null)[];
 }
@@ -25,19 +24,6 @@ export const Tabs = ({
     return children.find((v) => v)?.props.id;
   });
 
-  // Bunch of refs in order to detect wrapping in side the tabs so that we
-  // can apply a bg color
-  const wrapperRef = useRef<HTMLDivElement | null>(null);
-  const tabsRef = useRef<HTMLDivElement | null>(null);
-  const menuRef = useRef<HTMLDivElement | null>(null);
-  const [wrapped, setWrapped] = useState(() =>
-    isWrapped(tabsRef.current, menuRef.current)
-  );
-
-  useResizeObserver(wrapperRef.current, () => {
-    setWrapped(isWrapped(tabsRef.current, menuRef.current));
-  });
-
   return (
     <TabsPrimitive.Root
       {...props}
@@ -45,14 +31,10 @@ export const Tabs = ({
       onValueChange={onValueChange || setActiveTab}
       className="h-full grid grid-rows-[min-content_1fr]"
     >
-      <div
-        ref={wrapperRef}
-        className="flex flex-wrap justify-between border-b border-default min-w-0"
-      >
+      <div className="flex flex-wrap justify-between border-b border-default min-w-0">
         <TabsPrimitive.List
-          className="flex flex-nowrap overflow-visible"
+          className="flex flex-nowrap overflow-visible peer"
           role="tablist"
-          ref={tabsRef}
         >
           {Children.map(children, (child) => {
             if (!isValidElement(child) || child.props.hidden) return null;
@@ -86,20 +68,18 @@ export const Tabs = ({
           })}
         </TabsPrimitive.List>
         <div
-          ref={menuRef}
-          className={classNames('flex-1 p-1', {
-            'bg-vega-clight-700 dark:bg-vega-cdark-700': wrapped,
-            '': wrapped,
-          })}
+          className={classNames(
+            'flex-1 p-1',
+            'bg-vega-clight-700 dark:bg-vega-cdark-700',
+            'peer-[[role="tablist"]]:bg-inherit dark:peer-[[role="tablist"]]:bg-inherit'
+          )}
         >
           {Children.map(children, (child) => {
             if (!isValidElement(child) || child.props.hidden) return null;
             return (
               <TabsPrimitive.Content
                 value={child.props.id}
-                className={classNames('flex flex-nowrap gap-1', {
-                  'justify-end': !wrapped,
-                })}
+                className={classNames('flex flex-nowrap gap-1', 'justify-end')}
               >
                 {child.props.menu}
               </TabsPrimitive.Content>
@@ -162,13 +142,3 @@ export const LocalStoragePersistTabs = ({
     />
   );
 };
-
-const isWrapped = (
-  tabs: HTMLDivElement | null,
-  menu: HTMLDivElement | null
-) => {
-  if (!tabs || !menu) return;
-  const listRect = tabs.getBoundingClientRect();
-  const menuRect = menu.getBoundingClientRect();
-  return menuRect.y > listRect.y;
-};

```

asiaznik: Ignore me, I've been resizing the whole window instead of a pane

mattrussell36: This won't work when you resize the pane.. window resize is handled by the switch from trade-grid to trade-panels

asiaznik (approved):

PR #4449, opened 2023-08-01
feat(governance,markets,ui-toolkit): highlight successor market changes

### LINKED ISSUE #4132: Highlight the differences for Successor Market Proposals
Should get most of this for free after the work to do #4092 

## Story

As a user
I want to see the difference between the parent market and the Successor Market being proposed 
So that I can more clearly understand the impact of the proposal

## Acceptance Criteria

- [ ] I can see the market specification with the proposed update
- [ ] There is a simple way to visually highlight / compare the difference with this market specification from the original markets specification 

## Background 

Details of an existing market via the market API:  https://docs.vega.xyz/mainnet/api/rest/data-v2/trading-data-service-get-market, combined with governance API: https://docs.vega.xyz/mainnet/api/rest/data-v2/trading-data-service-get-governance-data details of a new proposal. 

## Tasks

- [ ] UX (if needed)
- [ ] Design (if needed)
- [ ] Implementation
- [ ] Code review
- [ ] QA review

## Sketch
Governance (  Token Utility) - Successor market diff

Miro: https://miro.com/app/board/uXjVPUTU6m4=/?moveToWidget=3458764558106362635&cot=14

sam-keen:
# Related issues 🔗

Closes #4132

# Demo 📺

Screenshot 2023-08-06 at 15 50 32


asiaznik (approved):

apps/governance/src/routes/proposals/components/proposal-market-data/proposal-market-data.tsx
```diff
@@ -58,8 +60,21 @@ export const ProposalMarketData = ({
...
+  const parentTerminationData =
+    parentMarketData?.tradableInstrument.instrument?.product
+      ?.dataSourceSpecForTradingTermination?.data;
+
+  const isParentSettlementDataEqual =
+    parentSettlementData !== undefined &&
+    isEqual(settlementData, parentSettlementData);
+  const isParentTerminationDataEqual =
+    parentTerminationData !== undefined &&
+    isEqual(terminationData, parentTerminationData);
```
edd, line 77: **thought (non-blocking)**: This is getting quite hefty. I guess at a certain point it would be neater to move it out to a function that grabs the relevant properties. On the other hand, it's pretty clear what's happening - so it's definitely not important.

apps/governance/src/routes/proposals/components/proposal/proposal.tsx
```diff
@@ -32,6 +32,7 @@ export interface ProposalProps {
   proposal: ProposalFieldsFragment | ProposalQuery['proposal'];
   networkParams: Partial<NetworkParamsResult>;
   newMarketData?: MarketInfoWithData | null;
+  parentMarketData?: MarketInfoWithData | null;
   assetData?: AssetQuery | null;
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   restData: any;
```
edd, line 38: **nitpick (non-blocking)**: We probably brought this up when it was introduced, and it wasn't introduced here so it's not a change request. But it *is* possible to type the REST responses. Block explorer has types generated. The reason it's not a blocker is that it still doesn't guarantee that the underlying REST responses matches the type, so it's more the illusion of proper typing. More so than normal.

libs/markets/src/lib/components/market-info/info-key-value-table.tsx
```diff
@@ -22,17 +24,26 @@ interface RowProps {
...
   unformatted,
   assetSymbol = '',
   noBorder = true,
+  parentValue,
+  hasParentData,
 }: RowProps) => {
+  // Note: we need both 'parentValue' and 'hasParentData' to do a conditional
+  // check to differentiate between when parentData itself is missing and when
+  // a specific parentValue is missing. These values are only used when we
+  // have successor market parent data.
```
edd, line 45: **praise. (non-blocking)**: Nice useful commenting. Soundssimple, but something you could lose a lot of time to.

edd (approved):
Nice comments, clear code. I feel an urge to request tests, but almost everything is too simple for it really to be worth it.

libs/markets/src/lib/components/market-info/info-key-value-table.tsx
```diff
@@ -63,10 +78,35 @@ const Row = ({
...
+        )}
+
+        {newValueInSuccessorMarket && (
+          <Lozenge className="py-0" variant={Intent.Primary}>
+            {t('Added')}
+          </Lozenge>
+        )}
+      </div>
+      <div style={{ wordBreak: 'break-word' }}>
+        {valueDiffersFromParentMarket ? (
```
asiaznik, line 99: suggestion: I think we could use red/green diff pattern

apps/governance/src/routes/proposals/components/proposal/proposal.tsx
sam-keen: Oh, thanks, I'll deal with this separately but I'll check out the Explorer code.

libs/markets/src/lib/components/market-info/info-key-value-table.tsx
sam-keen: thanks

libs/markets/src/lib/components/market-info/info-key-value-table.tsx
sam-keen: thanks for the suggestion - we wanted to avoid the success/failure colouring as it's really just about informing the user of changes occuring

JonRay15:
One thing that isnt clear to me on this, and I cant get a preview to open to check it, is whether the "updated" tag is inherited up to the top level of the accordion?

ie.  In my initial view where all my parts of the accordion are collapsed, will I get a tag showing me there are changes in a section and therefore I might want to expand it?  Or will I have to expand each one to try to see if there are updates?

image

JonRay15:
Raised https://github.com/vegaprotocol/frontend-monorepo/issues/4517 for this

PR #4478, opened 2023-08-03
feat(trading): use instrument code instead of name in positions, fills and accounts breakdown tables

### LINKED ISSUE #4480: Use market code instead of market name in positions table
## Story

As a user
I want see market code in positions table
So that market column show same value as in other places

## Acceptance Criteria

- [ ] Code is rendered for the market in the positions table

## Tasks

- [x] ACs reviewed and/or updated
- [x] Dev
- [x] Test

bglownia:
# Related issues 🔗

Closes #4480

# Description ℹ️

Use instrument.code instead of instrument.name in positions, fills and accounts breakdown tables

# Demo 📺

image


mattrussell36:
@bglownia there is a min width applied to the market name column that should probably be reduce or removed

mattrussell36 (approved):

PR #4494, opened 2023-08-04
feat(trading): merge deal ticket stores

### LINKED ISSUE #4451: Merge deal ticket stores
## The Chore

Stores that are storing data connected to deal ticket form should be merged into one deal ticket store and removed from orders lib.

bglownia:
# Related issues 🔗

Closes #4451

# Technical 👨‍🔧

Merge three zustand stores into one. Change flow of data in deal ticket, use store as side effect.


mattrussell36:
As discussed, it would be good if the hierarchy within the store state should be markets -> type -> fields

I think that would be cleaner in the code but also give us separated persistance between the different order types in the deal ticket

libs/deal-ticket/src/components/deal-ticket/deal-ticket-stop-order.tsx
```diff
@@ -50,32 +48,40 @@ export interface StopOrderProps {
...
-
 export const StopOrder = ({ market, marketPrice, submit }: StopOrderProps) => {
   const { pubKey, isReadOnly } = useVegaWallet();
-  const setDealTicketType = useDealTicketTypeStore((state) => state.set);
-  const [, updateOrder] = useOrder(market.id);
-  const updateStoredFormValues = useStopOrderFormValues(
-    (state) => state.update
+  const setType = useDealTicketFormValues((state) => state.setType);
+  const dealTicketType = useDealTicketFormValues((state) =>
+    state.formValues[market.id]?.type === DealTicketType.StopMarket
```
mattrussell36, line 67: What if the stored ticket type is normal limit or normal market? This doesn't look right to me?

libs/deal-ticket/src/hooks/use-form-values.ts
```diff
@@ -0,0 +1,141 @@
...
+          set((state) => {
+            state.formValues[marketId] = Object.assign(
+              state.formValues[marketId] ?? {},
+              { type }
+            );
+          });
+        },
+      })),
+      {
+        name: 'deal-ticket-form-values',
```
mattrussell36, line 137: vega_deal_ticket_store for consistency

libs/deal-ticket/src/hooks/use-form-values.ts
```diff
@@ -0,0 +1,141 @@
...
+export const isStopOrderType = (dealTicketType?: DealTicketType) =>
+  dealTicketType === DealTicketType.StopLimit ||
+  dealTicketType === DealTicketType.StopMarket;
+
+export const useDealTicketFormValues = create<Store>()(
+  immer(
+    persist(
+      subscribeWithSelector((set) => ({
+        formValues: {},
+        updateStopOrder: (marketId, formValues) => {
```
mattrussell36, line 87: Its super unclear when to use updateStopOrder vs update vs updateOrder

mattrussell36 (changes requested):

libs/deal-ticket/src/hooks/use-form-values.ts
bglownia: I renamed `update` to `updateAll`.

mattrussell36 (approved):

PR #4521, opened 2023-08-09
feat(explorer): view 'Cancel Stop Order' Transactions

### LINKED ISSUE #4084: View 'Cancel Stop Order' Transactions
## Story

As a Vega user
I want to be able to view 'Cancel Stop Order' transactions on Explorer
So that I can check and verify the details of the transaction / send to someone to evidence 

## Acceptance Criteria

- [x] I can see `Cancel Stop Order` transactions in the transactions list
- [x] I can identify `Cancel Stop Order` transactions in the transactions list (by type)
- [x] I can filter the transactions list to only view `Cancel Stop Order` transactions 

Check this one with Edd
- [x] I can view details of `Cancel Stop Order` transactions

## Tasks

- [ ] UX (if needed)
- [ ] Design (if needed)
- [ ] Implementation
- [ ] Code review
- [ ] QA review

The cancel stop order TX will have the same fields as a Cancel Order Transaction, so should look very similar. It has two fields:
- Market ID
- Stop Order ID

If Market ID is set and stop order id is not, cancel all stops in a market
If stop order ID is set and market ID is not, cancel one stop order
If both are set, cancel one stop order

As with Cancel Order, we should try to render the stop order as it was at the time of cancellation.

sam-keen:
# Related issues 🔗

Closes #4084

Screenshot 2023-08-09 at 16 57 53
Screenshot 2023-08-09 at 16 54 40
Screenshot 2023-08-09 at 16 15 23

### DIFF
--- a/apps/explorer/src/app/components/txs/details/tx-stop-order-cancel.tsx
+++ b/apps/explorer/src/app/components/txs/details/tx-stop-order-cancel.tsx
@@ -26,7 +26,7 @@ export const TxDetailsStopOrderCancel = ({
   pubKey,
   blockData,
 }: TxDetailsStopOrderCancelProps) => {
-  if (!txData || !txData.command.stopOrderId) {
+  if (!txData || !txData.command) {
     return <>{t('Awaiting Block Explorer transaction details')}</>;
   }
 
--- a/apps/explorer/src/app/components/txs/details/tx-stop-order-cancel.tsx
+++ b/apps/explorer/src/app/components/txs/details/tx-stop-order-cancel.tsx
@@ -4,7 +4,6 @@ import { MarketLink } from '../../links/';
 import type { TendermintBlocksResponse } from '../../../routes/blocks/tendermint-blocks-response';
 import { TxDetailsShared } from './shared/tx-details-shared';
 import { TableCell, TableRow, TableWithTbody } from '../../table';
-import DeterministicOrderDetails from '../../order-details/deterministic-order-details';
 import { CancelSummary } from '../../order-summary/order-cancellation';
 import Hash from '../../links/hash';
 import type { components } from '../../../../types/explorer';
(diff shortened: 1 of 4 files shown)


edd (changes requested):

apps/explorer/src/app/components/txs/details/tx-stop-order-cancel.tsx
```diff
@@ -0,0 +1,63 @@
...
+  txData,
+  pubKey,
+  blockData,
+}: TxDetailsStopOrderCancelProps) => {
+  if (!txData || !txData.command.stopOrdersCancellation) {
+    return <>{t('Awaiting Block Explorer transaction details')}</>;
+  }
+
+  const marketId: string = txData.command.stopOrdersCancellation.marketId;
+  const orderId: string = txData.command.stopOrdersCancellation.orderId;
```
edd, line 30: **thought (non-blocking)**: I suspect you can drop the `: string` when you switch to using the generated types

apps/explorer/src/app/components/txs/tx-order-type.tsx
```diff
@@ -46,7 +46,7 @@ const displayString: StringMap = {
   'Batch Market Instructions': 'Batch',
   'Stop Orders Submission': 'Stop',
   StopOrdersSubmission: 'Stop',
-  StopOrdersCancellation: 'Cancel stop',
+  StopOrdersCancellation: 'Cancel stop order',
```
edd, line 49:
```suggestion
  StopOrdersCancellation: 'Cancel stop',
```

I prefer to keep these shorter. But this could go either way.

apps/explorer/src/app/components/txs/details/tx-stop-order-cancel.tsx
```diff
@@ -0,0 +1,68 @@
...
+export const TxDetailsStopOrderCancel = ({
+  txData,
+  pubKey,
+  blockData,
+}: TxDetailsStopOrderCancelProps) => {
+  if (!txData || !txData.command) {
+    return <>{t('Awaiting Block Explorer transaction details')}</>;
+  }
+
+  const command: StopOrderCancellationTransaction = txData.command;
```
edd, line 33: **chore (blocking)**: This is incorrect. It's an example of where this heavy-handed typing isn't ideal I suppose! If you take a look at 

https://github.com/vegaprotocol/frontend-monorepo/blob/39962a02465a9d80ecf1869de2a2732f32586237/apps/explorer/src/types/explorer.d.ts#L453

you'll see that the `txData.command` contains properties depending on the type of transaction. In this case it's:

```suggestion
  const command: StopOrderCancellationTransaction = txData.command.stopOrdersCancellation;
```

Part of the challenge of these transactions is often finding an example of one on a network. Often @daunatv has done one already, but if not you have to give it a go yourself so you can see what the actual shape ends up being. I never quite trust the generated types without seeing one first-hand.

edd (changes requested):

apps/explorer/src/app/components/txs/details/tx-stop-order-cancel.tsx
daunatv: Adding [a link ](https://explorer.fairground.wtf/txs/46E716FCC2D01BA8DE1E4B3A4230085B2E4C41950851A5F4D3146F293EE06812)to one I have done on Fairground

edd:
Thanks @daunatv. Yeah, this current version of the branch is broken

Screenshot 2023-08-14 at 15 03 49

apps/explorer/src/app/components/txs/details/tx-stop-order-cancel.tsx
sam-keen: thanks both!

PR #4543, opened 2023-08-11
feat(trading): distinguish between product types

### LINKED ISSUE #4319: Distinguish spot, futures, and perpetual markets
## Story

As a trader on Vega Console
I want to know what type a market is when I see it in a list
So I can better understand my overall position when browsing orders etc.

## Acceptance Criteria

- [x] I can determine the type of market wherever it is mentioned:
  - [x] "All" tab on the markets page
  - [x] market switcher "All" tab ([when implemented](https://github.com/vegaprotocol/frontend-monorepo/issues/4384))
  - [x] orders tables
  - [x] positions table
  - [x] asset usage breakdown modal table
  - [ ] ~~watchlist/favourites (when/if this is implemented)~~
  - [x] ledger entries

## Tasks

- [x] UX (if needed)
- [x] Design (if needed)
- [x] Implementation
- [x] Code review
- [x] QA review

### Example: Tags

Tags are easiest to understand as they can be read, but they are harder to scan and tell apart. Colour could be used to improve this, however.

image

- This might be evident from the name of the market, but having a clear visual indicator could assist with scanning lists.

macqbat:
# Related issues 🔗

Closes #4319 

# Description ℹ️

Add small message (Pill component) near by market name about type of product it's set of.

# Demo 📺

image
image
image
image

# Technical 👨‍🔧


libs/datagrid/src/lib/cells/market-name-cell.tsx
```diff
@@ -1,39 +1,75 @@
-import type { MouseEvent, ReactNode } from 'react';
+import type { MouseEvent } from 'react';
 import { useCallback } from 'react';
 import get from 'lodash/get';
+import { Pill } from '@vegaprotocol/ui-toolkit';
+import type { Market } from '@vegaprotocol/types';
+
+const productTypeMap: { [key in string]: string } = {
```
mattrussell36, line 7: (non blocking) It would be better to use an `as const` pattern here for stricter types. Or better yet create a type using the schema?

```
const productTypeMap = {
  Future: 'Futr',
  Spot: 'Spot',
  Perpetual: 'Perp',
} as const;
type ProductType = keyof typeof productTypeMap
```

apps/trading/client-pages/markets/successor-market-cell.spec.tsx
```diff
@@ -0,0 +1,158 @@
...
+    const successorValue = 'market1';
+    render(
+      <MockedProvider>
+        <SuccessorMarketRenderer value={successorValue} />
+      </MockedProvider>
+    );
+    expect(screen.getByTestId('market-code')).toBeInTheDocument();
+    expect(screen.getByText('code successor 1')).toBeInTheDocument();
+    expect(screen.getByText('Futr')).toBeInTheDocument();
+    await act(() => {
```
mattrussell36, line 83: There is no need for `act` here. This should be simply

```
await userEvent.click(screen.getByRole('button'));
```

apps/trading/client-pages/markets/successor-market-cell.spec.tsx
```diff
@@ -0,0 +1,158 @@
...
+}));
+
+describe('SuccessorMarketRenderer', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+  it('should properly rendered successor market', async () => {
+    const successorValue = 'market1';
+    render(
+      <MockedProvider>
```
mattrussell36, line 76: Why do you need a MockedProvider when you are mocking the entire data provider already?

For the record, I'd prefer to use the MockedProvider and actually pass mock queries to it

libs/datagrid/src/lib/cells/market-name-cell.tsx
```diff
@@ -1,39 +1,75 @@
...
-        onMarketClick(id, ev.metaKey || ev.ctrlKey);
-      }
+      id && onMarketClick?.(id, ev.metaKey || ev.ctrlKey);
     },
     [id, onMarketClick]
   );
-  if (!value || !data) return null;
-  return onMarketClick ? (
+
+  productType =
```
mattrussell36, line 52: (non blocking): The props of this component seem overly complex. Can we simplify in any way. I'd look into making this component take two values only, a product type string and a market name string

libs/proposals/src/components/proposals-list/use-column-defs.tsx
```diff
@@ -25,8 +26,40 @@ import type { ProposalListFieldsFragment } from '../../lib/proposals-data-provid
...
+  value,
+  data,
+}: VegaICellRendererParams<
+  ProposalListFieldsFragment,
+  'terms.change.instrument.code'
+>) => {
   const { VEGA_TOKEN_URL } = useEnvironment();
+  const { change } = data?.terms || {};
+  if (change?.__typename === 'NewMarket' && VEGA_TOKEN_URL) {
+    const productType =
```
mattrussell36, line 39: This seems like a sensible helper function to live in the proposals lib (with a unit test)

mattrussell36 (changes requested):

libs/datagrid/src/lib/cells/market-name-cell.tsx
macqbat: changed to enum as const

libs/datagrid/src/lib/cells/market-name-cell.tsx
macqbat: I would prefer passing props like that, because it gives possibility to register the component in ag-grid and call renderer easily as string in the columns

apps/trading/client-pages/markets/successor-market-cell.spec.tsx
macqbat: act removed

libs/proposals/src/components/proposals-list/use-column-defs.tsx
macqbat: a helper has been added to the lib utils

libs/positions/src/lib/positions-data-providers.ts
```diff
@@ -135,6 +136,8 @@ export const getMetrics = (
       totalBalance: totalBalance.multipliedBy(10 ** decimals).toFixed(),
       unrealisedPNL: position.unrealisedPNL,
       updatedAt: position.updatedAt || null,
+      productType:
+        market?.tradableInstrument.instrument.product.__typename || '',
```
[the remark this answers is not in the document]
macqbat: Not sure how it can be done, but when these changes https://github.com/vegaprotocol/vega/pull/9003 will be available on stagnet or testnet it all these types will have to be amend and adjusted to

mattrussell36 (approved):

PR #4544, opened 2023-08-11
test(explorer): successor market explorer

jtsang586:
# Related issues 🔗

- Fix failing e2e tests due to wallet update that required chain id to now be mocked
- Add e2e test for successor market display in block explorer

### DIFF
--- a/apps/explorer-e2e/src/integration/market.cy.js
+++ b/apps/explorer-e2e/src/integration/market.cy.js
@@ -1,8 +1,8 @@
+import { createSuccessorMarketProposal } from '../support/governance.functions';
+
 context('Market page', { tags: '@regression' }, function () {
   describe('Verify elements on page', function () {
     const marketHeaders = 'markets-heading';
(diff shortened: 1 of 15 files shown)


apps/explorer-e2e/src/integration/market.cy.js
```diff
@@ -106,24 +106,68 @@ context('Market page', { tags: '@regression' }, function () {
...
+        });
+    });
+
+    it('Able to go to market details page for successor market', function () {
+      const successionLineItem = 'succession-line-item';
+      const successionLineMarketId = 'succession-line-item-market-id';
+
+      createSuccessorMarketProposal(this.createdMarketId);
+      cy.navigate_to('markets');
+      cy.reload();
```
edd, line 140: **question (non-blocking)**:  Why is this required?

edd (approved):
Approved with one non-blocking question (why does something require `cy.reload()`?.

PR #4550, opened 2023-08-14
chore(trading): use action icon buttons

### LINKED ISSUE #3764: Table row action icons
## The Chore

- [x] Replace edit text button with edit icon button
- [x] Replace cancel text button with cross icon button
- [x] Replace close text button (in positions) with cross icon button

## Tasks

- [x] Design icons for each
- [x] Replace current text

MadalinaRaicu:
# Related issues 🔗

Closes #3764 

# Description ℹ️

-  Replace edit text button with edit icon button
-  Replace cancel text button with cross icon button
-  Replace close text button (in positions) with cross icon button
-  Design icons for each
-  Replace current text

# Demo 📺

Screenshot 2023-08-14 at 18 00 14

# Technical 👨‍🔧


mattrussell36 (changes requested):
Can you adjust the min/max widths in the col defs so that the pinned right column doesn't take more space than it needs

PR #4562, opened 2023-08-16
feat(markets,governance): small UX improvements to successor market panels

### LINKED ISSUE #4517: Small UX improvements to successor markets comparison
## The Chore

Two improvements to the functionality when looking at a successor market proposal.

## Tasks

- [ ] Add an "Updated" tag to the top level of the accordion view if any of the parameters in that expandable section have changed so that users know to expand it and look.

image

 
- [ ] In the old vs new display, need to emphasise the new value more.  Either put the new ones in bold, or slightly grey out the old text in addition to the strikethrough .... or both 

image

sam-keen:
# Related issues 🔗

Closes #4517

# Description ℹ️

Opens the successor market panels, and provides more visual difference between old parent values and the successor's values

# Demo 📺

Screenshot 2023-08-16 at 15 16 21


libs/markets/src/lib/components/market-info/info-key-value-table.tsx
```diff
@@ -98,7 +98,7 @@ export const Row = ({
       <div style={{ wordBreak: 'break-word' }}>
         {valueDiffersFromParentMarket ? (
           <div className="flex items-center gap-3">
-            <span className="line-through">
+            <span className="line-through dark:text-vega-dark-300">
```
asiaznik, line 101: Please, use `text-vega-cdark-300` here.

asiaznik: Also it'd be nice to have it light theme friendly, `text-vega-clight-300 dark:text-vega-cdark-300` should do the trick

sam-keen: I'll move to using the new c colours as part of a broader ticket

PR #4585, opened 2023-08-21
fix(explorer): check linking type when summing stake

edd:


edd:
Self merged, which isn't reasonable. However it's due to the fact that I pushed a version of this straight to develop, which is also poor behaviour.

PR #4613, opened 2023-08-24
fix(proposals): parsing block duration value

### LINKED ISSUE #4607: Make sure network upgrade checks for block duration units
## The Chore

On the call to the statistics API to get the block duration sometimes this is given in s (if over 1s) and sometimes in ms (if under). 

The estimator needs to account for this

## Tasks

- [ ] Do it 
- [ ] Test it

asiaznik:
# Related issues 🔗

Closes #4607 

# Description ℹ️

Block duration value is not just a number it can be presented in seconds, milliseconds, nanoseconds, microseconds, minutes and hours. This PR fixes the problem of parsing that value for time estimations.

### DIFF
--- a/libs/proposals/src/lib/protocol-upgrade-proposals/use-time-to-upgrade.ts
+++ b/libs/proposals/src/lib/protocol-upgrade-proposals/use-time-to-upgrade.ts
@@ -7,6 +7,10 @@ const DEFAULT_POLLS = 10;
 const INTERVAL = 1000;
 const durations = [] as number[];
 
+export const ERR_NO_TIME_UNITS = new Error(
+  'could not parse block duration value - no time units detected'
+);
+
 /**
  * Parses block duration value and output a number of milliseconds.
  * @param input The block duration input from the API, e.g. 4m5.001s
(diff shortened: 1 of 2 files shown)


libs/proposals/src/lib/protocol-upgrade-proposals/use-time-to-upgrade.ts
```diff
@@ -7,6 +7,43 @@ const DEFAULT_POLLS = 10;
...
+  let S = 0;
+  const lessThanSecond = /^[0-9.]+[nµm]*s$/gu.test(input);
+  const exp = /(?<hours>[0-9.]+h)?(?<minutes>[0-9.]+m)?(?<seconds>[0-9.]+s)?/gu;
+  const m = exp.exec(input);
+  const seconds = lessThanSecond ? input : m?.groups?.['seconds'];
+  if (seconds) {
+    S = parseFloat(seconds);
+    if (seconds.includes('ns')) S /= 1000 * 1000;
+    else if (seconds.includes('µs')) S /= 1000;
+    else if (seconds.includes('ms')) S *= 1;
```
sam-keen, line 33: Chore (non-blocking): line 33 looks unnecessary and can be removed

libs/proposals/src/lib/protocol-upgrade-proposals/use-time-to-upgrade.ts
```diff
@@ -7,6 +7,43 @@ const DEFAULT_POLLS = 10;
...
+  // ms -> 1
+  // µs -> 1/1000
+  // ns -> 1/1000000
+  let H = 0;
+  let M = 0;
+  let S = 0;
+  const lessThanSecond = /^[0-9.]+[nµm]*s$/gu.test(input);
+  const exp = /(?<hours>[0-9.]+h)?(?<minutes>[0-9.]+m)?(?<seconds>[0-9.]+s)?/gu;
+  const m = exp.exec(input);
+  const seconds = lessThanSecond ? input : m?.groups?.['seconds'];
```
sam-keen, line 28: Suggestion (non-blocking): error handling would be a nice improvement here, perhaps throwing an error if none of the matching groups are present in the input and lessThanSecond is false

sam-keen (approved):

edd (approved):

Issue #4603, opened 2023-08-22
Ensure that wallet connection flows can be used by the new Referral Site


JonRay15:
## The Chore

We need the new referral site to be able to make use of the full set of new connection flows available in Console.

## Tasks

- [x] Refactor to not depending on environment lib
- [x] Publish them
- [x] Check the site can use them

PR #4621, opened 2023-08-24
Feat/4510 metamask snaps

### LINKED ISSUE #4510: Metamask snaps
## Story

As a user
I want to connect with Metamask snaps
So that I can connect as quickly as possible without downloading other wallets

## Acceptance Criteria

- [x] I am prompted to install the vegaprotocol snap into metamask
- [x] I can connect using the vegaprotocol snap
- [x] I can submit transactions using the vegaprotocol snap

## Tasks

- [x] UX (if needed)
- [x] Design (if needed)
- [x] Implementation
- [x] Code review
- [x] QA review

A POC for this has already been put together for this on the branch feat/snaps, but we need to harden it up for prod

asiaznik:
# Related issues 🔗

Closes #4510 

# Description ℹ️

# Demo 📺

# Technical 👨‍🔧

### DIFF
--- a/libs/wallet/src/connectors/snap-connector.ts
+++ b/libs/wallet/src/connectors/snap-connector.ts
@@ -1,5 +1,11 @@
 import type EthereumProvider from '@walletconnect/ethereum-provider/dist/types/EthereumProvider';
-import type { PubKey, Transaction, VegaConnector } from './vega-connector';
+import type { TransactionResponse } from './vega-connector';
+import {
+  WalletError,
+  type PubKey,
+  type Transaction,
+  type VegaConnector,
+} from './vega-connector';
 import { clearConfig, setConfig } from '../storage';
 
 declare global {
@@ -27,16 +33,23 @@ type InvokeSnapRequest = {
 };
 type InvokeSnapResponse<T> = Promise<T>;
 
-type SendTransactionResponse = {
-  transactionHash: string;
-  receivedAt: string;
-  sentAt: string;
-  transaction: {
-    signature: {
-      value: string;
+type SendTransactionResponse =
+  | {
+      transactionHash: string;
+      receivedAt: string;
+      sentAt: string;
+      transaction?: {
+        signature?: {
+          value: string;
+        };
+      };
+    }
+  | {
+      error: Error & {
+        code: number;
+        data: unknown;
+      };
     };
-  };
-};
 type GetChainIdResponse = {
   chainID: string;
 };
@@ -166,12 +179,21 @@ export class SnapConnector implements VegaConnector {
       payload
     );
 
+    if ('error' in result) {
+      const { message, code, data } = result.error;
+      throw new WalletError(
+        message,
+        code,
+        typeof data === 'string' ? data : ''
+      );
+    }
+
     return {
       transactionHash: result.transactionHash,
+      signature: result?.transaction?.signature?.value,
       receivedAt: result.receivedAt,
       sentAt: result.sentAt,
-      signature: result.transaction.signature.value,
-    };
+    } as TransactionResponse;
   }
 
   async getChainId(): Promise<GetChainIdResponse> {
@@ -187,7 +209,6 @@ export class SnapConnector implements VegaConnector {
   }
 
   async disconnect() {
-    console.log('here');
     clearConfig();
   }
 }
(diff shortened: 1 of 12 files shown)


libs/wallet/src/connectors/snap-connector.ts
```diff
@@ -0,0 +1,233 @@
...
+      throw new WalletError(
+        message,
+        code,
+        typeof data === 'string' ? data : ''
+      );
+    }
+
+    return {
+      transactionHash: result.transactionHash,
+      signature: result?.transaction?.signature?.value,
```
mattrussell36, line 212: chore (blocking): can we throw `WalletError` if signature is undefined. This would be better than just type casting the response

```
if (!result?.transaction?.signature) {
  throw new WalletError(message, code)
}

return {
  transactionHash: result.transactionHash,
  signature: result.transaction.signature.value,
  receivedAt: result.receivedAt,
  sentAt: result.sentAt,
};

```

libs/wallet/src/use-eager-connect.ts
```diff
@@ -1,11 +1,14 @@
...
 import { useVegaWallet } from './';
 import { useEffect, useState } from 'react';
 import type { VegaConnector } from './connectors/vega-connector';
 import { getConfig } from './storage';
+import { useEnvironment } from '@vegaprotocol/environment';
 
 export function useEagerConnect(Connectors: {
   [connector: string]: VegaConnector;
 }) {
+  const { VEGA_URL } = useEnvironment();
```
mattrussell36, line 11: chore (blocking): lets pass in origin as an argument for the hook. We'll need to publish this (and the connect dialog) soon, so we don't want to tie in deps that don't need to be

libs/wallet/src/use-injected-connector.ts
```diff
@@ -11,28 +17,42 @@ export enum Status {
   AcknowledgeNeeded = 'AcknowledgeNeeded',
 }
 
+export const ERR_VEGA_UNDEFINED = new Error('window.vega not found');
```
mattrussell36, line 20: suggestion (non blocking): I think its nicer for these to be objects, it makes finding the right error via auto-import/auto completion a little clearer

```
export const InjectedConnectorErrors = {
  VEGA_UNDEFINED: new Error('window.vega not found'),
  ....
}
```

libs/wallet/src/use-injected-connector.ts
```diff
@@ -11,28 +17,42 @@ export enum Status {
...
+export const ERR_INVALID_CHAIN = new Error('Invalid chain');
+
 export const useInjectedConnector = (onConnect: () => void) => {
   const { connect, acknowledgeNeeded } = useVegaWallet();
   const [status, setStatus] = useState(Status.Idle);
   const [error, setError] = useState<Error | null>(null);
+  const { VEGA_URL } = useEnvironment();
```
mattrussell36, line 27: Again, can we pass in VEGA_URL into connect-dialog and then into this hook?

libs/wallet/src/use-injected-connector.ts
```diff
@@ -11,28 +17,42 @@ export enum Status {
...
 
   const attemptConnect = useCallback(
-    async (connector: InjectedConnector, appChainId: string) => {
+    async (
+      connector: InjectedConnector | SnapConnector,
```
mattrussell36, line 31: Would it be simpler to leave useInjectedConnector and create a fresh useSnapConnector hook?

mattrussell36 (changes requested):

libs/wallet/src/use-eager-connect.ts
mattrussell36: Was just taking a look at this and its a bigger job than I expected and not worth cluttering this PR. I'm working on  https://github.com/vegaprotocol/frontend-monorepo/issues/4603 which will address this

libs/wallet/src/use-injected-connector.ts
mattrussell36: Disregard

libs/wallet/src/connectors/snap-connector.ts
mattrussell36: I've updated this

mattrussell36 (approved):

PR #4629, opened 2023-08-25
fix(deal-ticket): disable iceberg for IOC and FOK

### LINKED ISSUE #4617: Stagnet console cannot submit iceberg orders
## Description

When the user sends an iceberg order, the TX sent to the wallet is not an iceberg order. I think this is a regression - it used to work. See screenshots - this is the TX that is submitted to the user's wallet, not an API problem.

## Steps to Reproduce

1. Go to https://console.stagnet1.vega.rocks 
2. Go to an active market
3. Enter an iceberg order and hit submit
4. View the raw TX in your wallet, observe the lack of iceberg fields
5. Submit the order
6. View the raw TX on the block explorer, observe the lack of iceberg fields

## Expected behavior

Submitting an iceberg order submits an iceberg order

## Screenshots
image

image

image

## Video

https://github.com/vegaprotocol/frontend-monorepo/assets/6678/1a756500-f5a3-4895-bf64-3bf48f718364

bglownia:
# Related issues 🔗

Closes #4617 

# Description ℹ️

Fixes bug which make impossible to send iceberg order. Uncheck and iceberg checkbox when  TIF is selected to FOK or IOC

### DIFF
--- a/libs/deal-ticket/src/components/deal-ticket/deal-ticket.tsx
+++ b/libs/deal-ticket/src/components/deal-ticket/deal-ticket.tsx
@@ -38,8 +38,6 @@ import {
 } from '@vegaprotocol/utils';
 import { activeOrdersProvider } from '@vegaprotocol/orders';
 import { getDerivedPrice } from '@vegaprotocol/markets';
-import type { OrderInfo } from '@vegaprotocol/types';
-
 import {
   validateExpiration,
   validateMarketState,
@@ -59,8 +57,6 @@ import {
   useMarketAccountBalance,
   useAccountBalance,
 } from '@vegaprotocol/accounts';
-
-import { OrderType } from '@vegaprotocol/types';
 import { useDataProvider } from '@vegaprotocol/data-provider';
 import {
   DealTicketType,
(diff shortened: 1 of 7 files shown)


libs/deal-ticket/src/components/deal-ticket/deal-ticket.tsx
```diff
@@ -465,7 +467,18 @@ export const DealTicket = ({
           <TimeInForceSelector
             value={field.value}
             orderType={type}
-            onSelect={field.onChange}
+            onSelect={(value) => {
+              if (
+                iceberg &&
+                [
```
edd, line 473: **chore (blocking)**: this check is repeated three times in the cdiff, so it is probably worth refactoring to its own shared function that expresses the _why_ of the check a bit more. Both IOC and FOK are non-persistent, which is why they can't be icebergs, so perhaps `isNonPersistentOrder` (or `isPersistentOrder` that returns the inverse, obviously).

Otherwise the change looks good

edd (changes requested):
Minor refactor request

libs/deal-ticket/src/components/deal-ticket/deal-ticket.tsx
mattrussell36: @edd @bglownia I'm going to make this change so I can get this in before deploying fairground

mattrussell36 (approved):

Issue #4608, opened 2023-08-23
Update deal ticket submit buttons


mattrussell36:
## The Chore

Update the deal ticket submit buttons per the design. They'll likely need a bespoke button component to show the price/size information within

## Tasks

- [ ] Adjust/Add ACs
- [x] Dev
- [ ] Test

## Acceptance Criteria

- [x] Move the notional to match new designs
- [x] Change colour and design of submit button
- [x] Any other changes?

## Additional details / background info

https://www.figma.com/file/ywcCZvXdvG28bg0AmCY9mF/Console-v3?type=design&node-id=3166%3A5839&mode=design&t=4IBlhkbpaOFLzIKm-1

PR #4635, opened 2023-08-28
feat(deal-ticket): update deal ticket submit buttons

bglownia:
# Related issues 🔗

Relates to #4608 

# Description ℹ️

Update deal ticket submit button. Show notional under price input


libs/deal-ticket/src/components/deal-ticket/deal-ticket-stop-order.tsx
```diff
@@ -181,7 +189,7 @@ const Trigger = ({
                       data-testid={`triggerPrice${oco ? '-oco' : ''}`}
                       type="number"
                       step={priceStep}
-                      appendElement={assetSymbol}
+                      appendElement={<Pill size="xs">{quoteName}</Pill>}
```
mattrussell36, line 192: The padding of the pill and the alignment of it within the input looks off to me. This can be fixed later but if its a quick win lets just do it. We should:

- Reduce the py of the pill
- Reduce the padding right within the input when the pill is present
- 
Screenshot 2023-08-29 at 21 16 28

libs/deal-ticket/src/components/deal-ticket/deal-ticket-stop-order.tsx
```diff
@@ -332,7 +344,7 @@ const Size = ({
         const { value, ...props } = field;
         const id = `order-size${oco ? '-oco' : ''}`;
         return (
-          <div className="mb-4">
+          <div className={`mb-${isLimitType ? '4' : '2'}`}>
```
mattrussell36, line 347: I'm pretty  sure this won't compile.. you'll get away with it as we definitely already use mb-2 and mb-4, but dynamically creating classes like this won't work with tailwind

```suggestion
          <div className={isLimitType ? 'mb-4' : 'mb-2'}>
```

libs/deal-ticket/src/components/deal-ticket/deal-ticket-stop-order.tsx
```diff
@@ -584,6 +705,84 @@ export const StopOrder = ({ market, marketPrice, submit }: StopOrderProps) => {
...
   });
 
+  const normalizedPrice = price && removeDecimal(price, market.decimalPlaces);
+  const normalizedSize =
+    size && removeDecimal(size, market.positionDecimalPlaces);
+
+  const risesAbove =
+    triggerDirection ===
+    Schema.StopOrderTriggerDirection.TRIGGER_DIRECTION_RISES_ABOVE;
+  const subLabel = oco ? (
```
mattrussell36, line 715: (non blocking) perhaps make this a component?

libs/deal-ticket/src/components/deal-ticket/key-value.tsx
```diff
@@ -0,0 +1,51 @@
...
+  formattedValue?: string;
+  onClick?: () => void;
+}
+
+export const KeyValue = ({
+  label,
+  value,
+  labelDescription,
+  symbol,
+  indent,
```
mattrussell36, line 20: Looks like lots of these props aren't required?

libs/deal-ticket/src/components/deal-ticket/deal-ticket-stop-order.spec.tsx
```diff
@@ -176,143 +226,164 @@ describe('StopOrder', () => {
...
-    await userEvent.type(screen.getByTestId(triggerPriceInput), '0.01');
-    expect(screen.queryByTestId(triggerPriceErrorMessage)).toBeNull();
-    expect(
-      screen.queryByTestId(triggerPriceWarningMessage)
-    ).toBeInTheDocument();
-
-    // change to correct value
-    await userEvent.type(screen.getByTestId(triggerPriceInput), '2');
-    expect(screen.queryByTestId(triggerPriceWarningMessage)).toBeNull();
+    [false, true].forEach(async (ocoValue) => {
```
mattrussell36, line 291: Perhaps use it.each, you'll be able to construct a better test name and therefore better reporting if it fails (same for the other tests)

mattrussell36 (approved):

PR #4638, opened 2023-08-29
chore(wallet): prepare wallet lib for publishing

mattrussell36:
# Related issues 🔗

#4603 

# Description ℹ️

- Adds config prop to the vega wallet
- Remove dependency of libs/environment from libs/wallet by passing env vars from the app to the required components
- Updates libs/wallet to be buildable/publishable
- Fix some circular dependencies with the lib/wallet package


libs/wallet/src/connect-dialog/connect-dialog.tsx
```diff
@@ -159,21 +140,28 @@ const ConnectDialogContainer = ({
     // Immediately connect on selection if jsonRpc is selected, we can't do this
     // for rest because we need to show an authentication form
     if (connector instanceof JsonRpcConnector) {
+      connector.url = walletUrl;
```
macqbat, line 143: set url of this connector is little bit tricky, since setter of this property has some logic, which may cause losing of auth token and thus errors during connecting

macqbat: ok, I can't reproduce this connection error, so seems to be ok with this set

macqbat (approved):

package.json
```diff
@@ -139,7 +139,9 @@
     "@storybook/react-webpack5": "7.0.24",
     "@svgr/rollup": "^8.0.1",
     "@svgr/webpack": "^6.1.2",
-    "@swc/core": "^1.2.173",
+    "@swc/cli": "~0.1.62",
```
macqbat, line 142: these packages are not needed yet, I guess?

MadalinaRaicu (approved):

Issue #4349, opened 2023-07-19
Separate out and delay telemetry opt in


ckpringle:
## The Chore

This chore is to create a standalone telemetry opt in that is not shown on first use / onboarding but at a later time. For example, once a wallet is connected and a period of time has passed i.e. we know they are actually using Vega and not just trying to get a feel for it. 

This chore would involve repurposing the existing component from Governance: 

<img width="918" alt="Screenshot 2023-07-19 at 11 45 38" src="https://github.com/vegaprotocol/frontend-monorepo/assets/46783148/3a60e5a0-4f7a-4661-8b04-40280a9b77f7">

## Tasks

- [ ] Work out best approach to manage timing
- [ ] Implement

PR #4642, opened 2023-08-29
chore(trading): 4349 delayed telemetry opt in

macqbat:
# Related issues 🔗

#4349 

# Description ℹ️

Add approval of telemetry popup on load

# Demo 📺

https://github.com/vegaprotocol/frontend-monorepo/assets/338931/8c0ca9b7-1381-49b2-9106-6a8ed3427ed3

# Technical 👨‍🔧

Detect local storage value and handle presentation logic according to it


mattrussell36:
@barnabee @MM0819 @JonRay15 

Thoughts on this? Personally, I don't like the experience here and I think we want a different treatment from governance. I would prefer to use a toast so that you don't get _another_ dialog after immediately connecting. It just feels like fighting through a sea of dialogs

barnabee:
I agree the experience sucks. I think defaulting to no telemetry and a toast would be nicer. Or just no telemetry? 

What do we actually use telemetry for? I have not seen anything we learned from it mentioned in a single ticket or Slack discussion about a bugfix OR a UX improvement… is there anything?

How many mainnet issues or serious UX problems have we found and solved with telemetry across all our apps?

macqbat:
We follow here logic from governance: it means on non-mainnet envs we are enabling by default telemetry in the moment we show the popup with these two buttons to choose from - stop sharing data or continue it. On mainnet we enable telemetry only when user click confirmation he wants it in this popup.
I can change the UX popup for the toast, if we want it.

As for the question how meny issues telemetry helped to find: I can't remember any such case, but it doesn't mean it will never help with anything. I think it have some values.

macqbat:
How the toast could look like:
image

MadalinaRaicu (approved):

barnabee:
> As for the question how meny issues telemetry helped to find: I can't remember any such case, but it doesn't mean it will never help with anything. I think it still has some values.

Seems highly tenuous that it's worth all this dev time tbh, but up to the front end team to decide ultimately, as long as the system never opts people in by default on mainnet and on testnets it **never opts them in by default without explicitly telling them**.

mattrussell36:
@macqbat lets merge this, but I have a few suggestions

- Make the toast an info state
- Use vega icons

macqbat:
> @macqbat We should merge this, but I have a few suggestions first
> 
> * Make the toast an info state
> * Use vega icons

Done

Issue #4597, opened 2023-08-22
Review buildable libraries


JonRay15:
## The Chore

Review what needs to be built and published. Simplify where possible. This will make migrating the console app out of the monorepo easier.

## Tasks

- [x] Review build-able libraries. The following must be buildable/publishable as they are in use (or will be very soon):
  - announcements
  - tailwindcss-config
  - ui-toolkit
  - types
  - wallet
  - react-helpers (we can possibly unpublish this, its not used extensively, see vega.xyz deps)
  - utils (we can possibly un-publish this, its not used extensively, see vega.xyz deps)
- [ ] Make all other libs non-buildable and unpublish from npm if needed
- [ ] Check with Governance & BE & vega.xyz
- [x] Consider migrating to SWC for remaining libs that must be built/published
- [x] Ensure everything still builds

bglownia:
![image](https://github.com/vegaprotocol/frontend-monorepo/assets/1754247/a2a77c63-b12d-43d5-bddf-8ac7fcdb1043)
![image](https://github.com/vegaprotocol/frontend-monorepo/assets/1754247/bd6c6a3c-7e77-4b42-bfbb-25c7adf1d343)

bglownia:
@mattrussell36 Do we need to deliver publishable libs in "cjs" format ? "umd" should be enough for frontend libs
https://github.com/vegaprotocol/frontend-monorepo/pull/3518

PR #4689, opened 2023-09-04
chore(trading): convert libs to swc

bglownia:
# Related issues 🔗

Issue: #4597 

# Description ℹ️

Speed up the builds

# Technical 👨‍🔧


mattrussell36:
@bglownia closing this as I think we'll approach this differently. See the newly refined ticket

Issue #4506, opened 2023-08-08
De-emphasise successor market column in market list


edd:
## Successor Markets

<img width="1269" alt="Screenshot 2023-08-08 at 12 27 47" src="https://github.com/vegaprotocol/frontend-monorepo/assets/6678/9434513b-4208-4eee-a449-912453a70bfa">

## Tasks

- [x] Remove successor market column
- [x] Choose an indicator that can go next to the name to indicate that the market is a successor
  - [x] ~~Suggest a pill with market's sequence in the succession line~~ Uses stack cell design from positions table

## High Level Requirements
Successor Markets is a new feature. It makes some sense to have a table column to display these details. However the column is quite wide (which will be fixed by another change), and pushes more useful information of the table. Most markets will not be or have successors, so we need to change how this information is displayed.

PR #4722, opened 2023-09-07
chore(trading): replace successor cols with market code cell

mattrussell36:
# Related issues 🔗

Issue: #4506 

# Description ℹ️

- Removes successor col from markets table and closed markets table
- Removes feature flags in tables
- Renders the successor or parent status using a 'short name'. Either 'PRNT' or 'SCCR', using the stack-cell that is in the positions table
- Also renders product type in the stack cell (and not within a pill)
- Moves market conmponents to the trading app to allow for easier react router navigation

# Demo 📺

Screenshot 2023-09-06 at 21 08 17


macqbat (approved):

bwallacee:
@mattrussell36 This may be incredibly dumb, but I would have thought we would show SCCR on the succcessor markets and PRNT on the suucceded markets.

It looks like we show PRNT on the successor markets and now PRNT on the parent.

image

mattrussell36:
> @mattrussell36 This may be incredibly dumb, but I would have thought we would show SCCR on the succcessor markets and PRNT on the suucceded markets.
> 
You are totally right. Will amend. if parentMarketID is present then the markt is a successor. If successorMarketID is present then market is a parent.

guoguojin (approved):

Issue #4647, opened 2023-08-29
Dedicated transfer page


mattrussell36:
## Story

As a user
I want a simplified dedicated transfer page
So that I can easily be linked to transfers from other apps

## Acceptance Criteria

- [x] I can transfer on /transfer
- [x] I can transfer with a preselected asset at /transfer?assetId=xyz

## Tasks

- [x] UX (if needed)
- [x] Design (if needed)
- [ ] Team and stakeholder review
- [x] Specs reviewed and created or adjusted
- [x] Implementation
- [x] Testing (unit and/or e2e)
- [x] Code review
- [x] QA review

bwallacee:
@macqbat Can we make the asset box to match vega key and amount when no asset has been selected?
![image](https://github.com/vegaprotocol/frontend-monorepo/assets/133853298/dee9ff0f-1fa6-4082-aaa1-bedf99acebc8)

macqbat:
> @macqbat, @mattrussell36 Can we make the asset box to match vega key and amount when no asset has been selected? ![image](https://user-images.githubusercontent.com/133853298/266964155-93098d17-3e06-4900-afa4-00acbb07ec6b.png)

this has been done in the last commit

bwallacee:
@macqbat All looks good to me.

Is this going through a team review or can we move to closed?

Issue #4646, opened 2023-08-29
Dedicated withdraw page


mattrussell36:
## Story

As a user
I want a simple dedicated withdraw page
So that I can be linked to a page to withdraw from other websites or the wallet extension

## Acceptance Criteria

- [x] I can go to /withdraw to withdraw
- [x] I can go to /withdraw?assetId=xyz and have an asset already selected

## Tasks

- [x] UX (if needed)
- [x] Design (if needed)
- [x] Team and stakeholder review
- [x] Specs reviewed and created or adjusted
- [x] Implementation
- [x] Testing (unit and/or e2e)
- [x] Code review
- [x] QA review

PR #4728, opened 2023-09-07
chore(withdraws): standalone pages for withdraw and transfer

macqbat:
# Related issues 🔗

Issue: #4646 & #4647 
# Description ℹ️

Dedicated pages fot transfer and withdrawal tickets. Contains handling of assetId pass and clear it for not breaking the params state

# Demo 📺

image
image

# Technical 👨‍🔧


apps/trading/client-pages/transfer/transfer.tsx
```diff
@@ -0,0 +1,34 @@
...
+  TransferContainer,
+  useTransferAssetIdStore,
+} from '@vegaprotocol/accounts';
+import { useSearchParams } from 'react-router-dom';
+import { useEffect } from 'react';
+
+export const Transfer = () => {
+  const [searchParams, setSearchParams] = useSearchParams();
+  const assetId = searchParams.get('assetId') || undefined;
+  const selectedAssetId = useTransferAssetIdStore((store) => store.assetId);
```
mattrussell36, line 13: I don't see any purpose for the useTransferAssetIdStore.

apps/trading/client-pages/withdraw/withdraw.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { WithdrawContainer } from '../../components/withdraw-container';
+import { useEffect } from 'react';
+
+export const Withdraw = () => {
+  const [searchParams, setSearchParams] = useSearchParams();
+  const assetId = searchParams.get('assetId') || undefined;
+  const asset = useWithdrawStore((store) => store.asset);
+  useEffect(() => {
+    if (assetId && asset?.id && assetId !== asset.id) {
+      setSearchParams({});
```
mattrussell36, line 14: I don't think we need to worry about clearing the url if the asset changes.

apps/trading/pages/client-router.tsx
```diff
@@ -35,6 +35,14 @@ const LazyDeposit = dynamic(() => import('../client-pages/deposit'), {
   ssr: false,
 });
 
+const LazyWithdraw = dynamic(() => import('../client-pages/withdraw'), {
```
mattrussell36, line 38: I'd like to move away from using next/dynamic and even React.lazy in favor of just having a single js bundle. This makes things easier with ipfs and the benefits we gain from code splitting are minimal due to the nature of this being more app-like

apps/trading/pages/client-router.tsx
```diff
@@ -104,6 +116,8 @@ const routerConfig: RouteObject[] = [
     element: <LazyDisclaimer />,
   },
   { path: Routes.DEPOSIT, element: <LazyDeposit /> },
+  { path: Routes.WITHDRAW, element: <LazyWithdraw /> },
```
mattrussell36, line 119: (non blocking): We should probably wrap these with a layout route to ensure consistency

libs/accounts/src/lib/transfer-form.tsx
```diff
@@ -392,3 +396,16 @@ export const AddressField = ({
     </>
   );
 };
+
+interface TransferAssetIdStore {
+  assetId?: string;
+  setAssetId: (assetId: string) => void;
+}
+export const useTransferAssetIdStore = create<TransferAssetIdStore>()(
```
mattrussell36, line 404: I think this store is unneccessary

mattrussell36 (changes requested):

apps/trading/client-pages/transfer/transfer.spec.tsx
```diff
@@ -0,0 +1,70 @@
...
+jest.mock('@vegaprotocol/accounts', () => ({
+  ...jest.requireActual('@vegaprotocol/accounts'),
+  TransferContainer: jest.fn(({ assetId }: { assetId?: string }) => (
+    <div data-testid="assetId">{assetId}</div>
+  )),
+}));
+
+const renderJsx = () => {
+  render(
+    <VegaWalletContext.Provider value={mockWalletContext}>
```
mattrussell36, line 29: This is not needed as the container is mocked

mattrussell36: Actually it was needed for the GetStarted component, but I think it simpler to mock it. I've pushed changes

apps/trading/pages/client-router.tsx
mattrussell36: I'll make an additional PR for this

libs/accounts/src/lib/transfer-form.tsx
mattrussell36: Having a whole new store to clear the query string is overkill.

apps/trading/pages/client-router.tsx
mattrussell36: I've changed these to just be imported in a regular fashion. I'll make an additional PR to convert the rest

mattrussell36 (approved):

bwallacee:
@macqbat On the withdraws page, if I disconnect my etherum wallet and then select connect again, It connects straight away with no messages or popups etc on my metamask wallet, is this expected?

bwallacee:
If I try to scroll down on the withdraw assets list, it flashes and brings me back up to the top. Meaning I am unable to select any assets.
https://github.com/vegaprotocol/frontend-monorepo/assets/133853298/ae89ece1-099c-4c2e-a16b-2092bddb2fa2

bwallacee:
I have raised a new issue for a withdraw scroll as it's happening on the deal ticket also:
https://github.com/vegaprotocol/frontend-monorepo/issues/4741

bwallacee:
Should we add new specs within this PR?

mattrussell36:
> @macqbat On the withdraws page, if I disconnect my etherum wallet and then select connect again, It connects straight away with no messages or popups etc on my metamask wallet, is this expected?

@bwallacee this is expected. Because the wallet already allowed connection to this website it doesn't require you to confirm the connection a second time

mattrussell36:
@macqbat @bwallacee 

I've added a single spec for each page to capture the setting the asset via the assetId query param. I don't think we need additional specs as everything else works exactly the same as when the component is opened in the sidebar.

Issue #3922, opened 2023-05-23
Margin warning is way too shouty


barnabee:
The not enough margin warning is far too big and shouty IMO. This should be like a simple line of text or small highlight or something, as it's not a massive error condition to have the ticket in a state like this.

<img width="1912" alt="image" src="https://github.com/vegaprotocol/frontend-monorepo/assets/1179257/b5257a58-3c8f-4f3b-b613-0313178c4848">

mattrussell36:
I've added this one to the [design change epic](https://github.com/vegaprotocol/frontend-monorepo/issues/4077) as I think we'll probably want to consider this along with other deal ticket changes

PR #4732, opened 2023-09-08
chore(deal-ticket): make margin warning more subtle

macqbat:
# Related issues 🔗

Issue: #3922 

# Description ℹ️

Change Notification banner for one line short text

# Demo 📺

[chore/3922-more-subtle-margin-warning](https://github.com/vegaprotocol/frontend-monorepo/tree/chore/3922-more-subtle-margin-warning)
image

# Technical 👨‍🔧

### DIFF
--- a/libs/deal-ticket/src/components/deal-ticket-validation/margin-warning.tsx
+++ b/libs/deal-ticket/src/components/deal-ticket-validation/margin-warning.tsx
@@ -1,6 +1,12 @@
 import { addDecimalsFormatNumber } from '@vegaprotocol/utils';
 import { t } from '@vegaprotocol/i18n';
-import { Notification, Intent } from '@vegaprotocol/ui-toolkit';
+import {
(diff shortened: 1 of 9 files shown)


bwallacee:
@macqbat I'm not sure what the best way would be but when tying in the size, if you dont have enough margin, the message will "flash" on every keystroke, same with removing numbers.
I wonder if we can add a little delay or something so the place order button and message doesn't move around as you type in your size

MadalinaRaicu (approved):

macqbat:
> @macqbat I'm not sure what the best way would be but when tying in the size, if you dont have enough margin, the message will "flash" on every keystroke, same with removing numbers. I wonder if we can add a little delay or something so the place order button and message doesn't move around as you type in your size

readdressed here https://github.com/vegaprotocol/frontend-monorepo/issues/4743

Issue #4747, opened 2023-09-12
Deal ticket should be open by default


JonRay15:
## The Chore

When you load a new market the deal ticket should be open from the sidebar by default.

## Tasks

- [ ] Make sure whenever you load a market the deal ticket is open by default 
- [ ] Test it

PR #4749, opened 2023-09-12
chore(trading): make deal ticket open on market change

asiaznik:
# Related issues 🔗

Issue: #4747 

# Description ℹ️

* deal ticket is opened alway by default on market page

# Demo 📺

https://github.com/vegaprotocol/frontend-monorepo/assets/1980305/82683888-c890-4b4d-a7d8-03eefd0b155b

### DIFF
--- a/apps/trading/client-pages/market/market.tsx
+++ b/apps/trading/client-pages/market/market.tsx
@@ -69,15 +69,15 @@ export const MarketPage = () => {
   useEffect(() => {
     if (data?.id && data.id !== lastMarketId) {
       update({ marketId: data.id });
+      // make sidebar open on market id change
+      setView({ type: ViewType.Order });
     }
-  }, [update, lastMarketId, data?.id]);
 
-  // Make sidebar open on deal ticket by default
-  useEffect(() => {
+    // make sidebar open on deal ticket by default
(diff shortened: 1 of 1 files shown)


MadalinaRaicu (approved):

apps/trading/client-pages/market/market.tsx
```diff
@@ -69,15 +69,15 @@ export const MarketPage = () => {
...
-  }, [update, lastMarketId, data?.id]);
 
-  // Make sidebar open on deal ticket by default
-  useEffect(() => {
+    // make sidebar open on deal ticket by default
     if (init && view === null) {
       setView({ type: ViewType.Order });
     }
-  }, [init, view, setView]);
+  }, [update, lastMarketId, data?.id, setView, init, view]);
```
macqbat, line 80: does this change change anything in the logic of operation?

macqbat (approved):

JonRay15:
If I am interpreting that video correctly ... if someone simply navigates to the "Trading" screen rather than directly to a market, the deal ticket will NOT be automatically open?

Do we want that?  It feels more consistent to just always have it default to open whenever you are on the trading screen, regardless of how you have navigated there.

Issue #4748, opened 2023-09-12
Remaining tweaks to market list screen


JonRay15:
## The Chore

Make a few changes to the market list screen as there are still some issues here. 

## Tasks

- [x] Remove successor market from this view completely, it is too niche, most ppl wont know what it is, and it takes up too much room
- [x] Add 24h volume
- [x] Reorder columns (Market, description, trading mode, status, mark price, 24h vol, settlement asset)
- [x] Remove Bid Price & Ask Price
- [x] Instead replace them with a single calculated "Spread" field (calculated as `Ask price - bid price`) shown on far right
- [x] Tweak column spacing to remove space from numbers where its redundant and provide more room in the "market" & "description" for long names if they are provided. 
- [x] Add handling for Trading Mode - Monitoring Auction .... dont show the extra info, just show "Monitoring Auction" and pull the full details into a tooltip

barnabee:
I quite like the sound of option 3 but anything other than option for works for me.

JonRay15:
Went with option 3

JonRay15:
This went out

PR #4756, opened 2023-09-12
chore(trading): market list tweaks

asiaznik:
# Related issues 🔗

Issue: #4748 

# Demo 📺

...

### DIFF
--- a/libs/markets/src/lib/components/markets-container/use-column-defs.tsx
+++ b/libs/markets/src/lib/components/markets-container/use-column-defs.tsx
@@ -1,5 +1,5 @@
 import { useMemo } from 'react';
-import type { ColDef } from 'ag-grid-community';
+import type { ColDef, ValueFormatterParams } from 'ag-grid-community';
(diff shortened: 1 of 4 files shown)


libs/markets/src/lib/components/markets-container/use-column-defs.tsx
```diff
@@ -181,6 +164,79 @@ export const useColumnDefs = ({ onMarketClick }: Props) => {
...
+            const bid =
+              data?.data?.bestBidPrice === undefined
+                ? undefined
+                : toBigNum(
+                    data?.data?.bestBidPrice,
+                    data.decimalPlaces
+                  ).toNumber();
+            return Number(bid) - Number(offer);
+          },
+          cellRenderer: ({
```
mattrussell36, line 191: The spread column should render the actual calculated spread. Not the best offer/ask next to each other

mattrussell36 (changes requested):

mattrussell36 (approved):

Issue #4743, opened 2023-09-11
re-render of deal ticket caused recreating of useEstimatePositionQuery and other query hooks


macqbat:
## Description

Each re-render of deal ticket - each change of any value in input - caused recreating of query hooks and this caused blinking of margin warning. `useEstimatePositionQuery` hook always returns firstly **data: undefined** and this caused blinking of margin warning component. This perhaps can be fixed by change fetchPolicy on cache first on by other apollo client settings.  

## Steps to Reproduce

1. Go to deal ticket
2. Set size and price for bigger margin than you have.
3. Increase size or price
4. See error: margin warning blinks (appears and hides)

## Expected behavior

Margin warning should be still visible, because we have still smaller margin than required

## Device details

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

PR #4770, opened 2023-09-13
chore(deal-ticket): 4743 preserve estimate position cache in deal ticket

macqbat:
# Related issues 🔗

Issue: #4743

# Description ℹ️

Since we can't easily cache in apollo `EstimatePositionQuery` (no-cache policy, no explicitly id, changing variables) we can as workaround preserve old results in case when current one is undefined 

# Demo 📺

-

# Technical 👨‍🔧

-


libs/deal-ticket/src/components/deal-ticket/deal-ticket.tsx
```diff
@@ -253,16 +251,14 @@ export const DealTicket = ({
...
-    },
+
+  const positionEstimate = usePositionEstimate({
+    marketId: market.id,
+    openVolume,
+    orders,
+    collateralAvailable:
+      marginAccountBalance || generalAccountBalance ? balance : undefined,
     skip: !normalizedOrder,
-    fetchPolicy: 'no-cache',
```
mattrussell36, line 265: I dont think you need the new hook if you change the fetch policy to 'cache-and-network'. This should result in the same behaviour

mattrussell36 (dismissed):

MadalinaRaicu (approved):

libs/deal-ticket/src/components/deal-ticket/deal-ticket.tsx
macqbat: I can't agree: by the any change of order amounts we send new set of variables what break the cache. We could prepare some extended cache rule for Estimates, but sice they don't have explicity id, have a rich various set of variables it would be quite complicated. I think new hook isn't too high price for robust cache in this important place of application

JonRay15:
Discussed with @macqbat this morning and he doesnt think the suggestion from Matt can work ... I want to delay unti @mattrussell36 is in to progress this so have said we should move ahead with the current implementation and revisit later if it is a problem.

PR #4781, opened 2023-09-13
fix(trading): simplify onboarding and explore button to use top traded 

### LINKED ISSUE #4773: The "explore" button on the onboarding modal shows wrong market
## Description

I am landing on DOGE when I shoudl be on ETH I imagine

## Steps to Reproduce

1. Go to console as new uers
2. Click on EXPLORE
3. You will see DOGE (UPDATE: just did this again and now got LINK ... so maybe its not reverse ... maybe its just not sorting)

## Expected behavior

It should be showing the market with top 24h volume ... I suspect it is showing the least

## Screenshots

image

mattrussell36:
# Related issues 🔗

Closes #4773 

# Description ℹ️

- Makes the explore button go to the top traded market (notional volume). I've also updated the initial default market to also use the top traded market (unless we have a stored market id). This makes clicking explore more seamless as the user will already be on the right page and we won't have to refetch anything.
- Fix incorrect icon in telemetry toast

# Technical 👨‍🔧

- Combine onboarding into single store for easier debugging and versioning
- Adds useTopTradedMarkets hook to easily get a sorted list
- Removes telemetry toast logic from the welcome dialog. We were setting state in render here also which was causing react warnings.
- Uses a proper anchor element for the explorer  link


apps/trading/components/constants.ts
```diff
@@ -1,2 +1 @@
 export const THROTTLE_UPDATE_TIME = 500;
-export const ONBOARDING_VIEWED_KEY = 'vega_onboarding_viewed';
```
macqbat, line 2: on this key in localStorage bases e2e tests as they suppress onboarding popup

apps/trading/components/welcome-dialog/get-started.tsx
```diff
@@ -94,7 +84,7 @@ export const GetStarted = ({ lead }: Props) => {
     { 'mt-8': !lead }
   );
 
-  if (getStartedNeeded) {
```
macqbat, line 97: user theoretically can fulfill all onboarding condition with no "dismiss" click. In such a case we still will show him get started banner

apps/trading/components/constants.ts
mattrussell36: Updated in both cypress. And here's a PR for console-test https://github.com/vegaprotocol/console-test/pull/120

apps/trading/components/welcome-dialog/get-started.tsx
mattrussell36: Personally I think this is fine. It gets very complicated if we conditionally show based on all the user state (accounts/deposits/connection/etc). I think the better approach here is to only track the dismissed flag. With the current flow after you deposit you are presented with 'Ready to trade' which when clicked will fully dismiss onboarding

Screenshot 2023-09-14 at 11 15 01

PR #4783, opened 2023-09-14
chore(trading): get started section for deposit

### LINKED ISSUE #4771: Remove the CTA from the onboarding deposit inlay when on the deposit screen
## The Chore

When you are in onboarding flow you click on the blue Deposit button and it opens up the deposit sidepanel for you. 

But the problem is once you're here, you still get the onboarding blue button on screen, which presents issues:

1. You now have the actual deposit button above it which will actually MAKE the deposit, so its confusing there is still the blue one
2. You might think clicking the blue one would actually now deposit but it wont ... it will just keep reloading the deposit panel, which is confusing

Screenshot 2023-09-13 at 13 17 12

Either we need to get rid of the blue deposit button completely, or maybe we could grey it out, make it unclickable, and change the text to "Complete deposit on screen" or something?

## Tasks

- [ ] Decide on exact implementation
- [ ] Build it
- [ ] Test it

mattrussell36:
# Related issues 🔗

Closes #4771 

# Description ℹ️

- Makes the deposit button go to the standalone deposit page (the whole reason we have the standalone deposit page was for onboarding)
- Adds a variant of the GetStarted component for the deposit page that only shows when onboarding is in progerss.
- Converts GetStarted CTAs to use links when they are links

# Demo 📺

Screenshot 2023-09-13 at 20 08 48

Screenshot 2023-09-13 at 20 09 13


bwallacee:
LTGM

MadalinaRaicu (approved):

bglownia (approved):

PR #4786, opened 2023-09-14
feat(ci): use ipfs daemon in dockerfiles

mkjmdski:
closes: https://github.com/vegaprotocol/devops-infra/issues/2051

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -3,8 +3,9 @@
 daemon="${1:-nginx}"
 
 if [[ "$daemon" = "nginx" ]]; then
-  nginx -s 'deamon off;'
+  nginx -g 'daemon off;'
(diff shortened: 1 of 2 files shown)


docker/entrypoint.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+daemon="${1:-nginx}"
+
+if [[ "$daemon" = "nginx" ]]; then
+  nginx -s 'deamon off;'
+elif [[ "$daemon" = "ipfs" ]]; then
+  ipfs config --json Addresses.Gateway '"/ip4/127.0.0.1/tcp/80"'
```
edd, line 8: **Question (blocking)**: do we need to run ‘IPFS add’ here or is it already added/pinned?

edd (changes requested):

PR #4797, opened 2023-09-15
fix(trading,web3): prevent deposit toasts reappearing

### LINKED ISSUE #4791: Toasts on deposit firing lots of times
## Description

We are getting many toasts during a deposit, some of which might be reasonable, some are not. 

## Steps to Reproduce

1. Go to deposit and make a deposit
2. Wait, you will see the orange "Awaiting confirmation" toast, if you dismiss this it will keep coming back.  

image

3. Once the deposit completes you will get the complete notification, but if you dismiss it ... it will keep coming back, we've been observing about 4/5 times over a period of a few minutes. 

image

4. Finally, after a few dismisses, you will suddenly get a new yellow notification telling you it is processing a deposit - which presumably it isnt.

image

## Expected behavior

I think for 2 we would ideally have some sort of timer.  It seems like it is retriggering whenever a new confirmation comes in.  I think it needs to come back because we dont have a notifications centre currently so it is right that it doesnt disappear forever, but could we tune it to come back less freuently?  Like say have it wait for 10 more confirmations before it renotifies, because currently it feels quite oppressive.  You try to dismiss it and its back in a few seconds. 

For 3 clearly this is a bug and we should only be notifying of the success once. 

For 4 again this is clearly a bug and quite a confusing one since it makes you think you might be depositing again. 

## Screenshots

Above in line 

## Device details

- OS: Window
- Browser: Chrome
- Tested this and observed it with both browser wallet and metamask snap connected ... didnt bother with desktop

mattrussell36:
# Related issues 🔗

Closes #4791 

# Description ℹ️

Fixes an issue where toasts are shown with every transaction confirmation even if the user has dismissed it. Now if a deposit toast is dismissed it will only re-appear once the deposit is complete (both the Ethereum transaction, and awaiting for it to be credited to the users account)


MadalinaRaicu (approved):

asiaznik (approved):

asiaznik:
Thing for the near future: I'd refactor this to not rely on `dialogOpen` flag anymore. It's a relict of the dialog past that we've inherited.

libs/web3/src/lib/use-ethereum-transaction-store.tsx
```diff
@@ -27,7 +27,13 @@ export interface EthStoredTxState extends EthTxState {
...
   args: string[];
   requiredConfirmations: number;
-  requiresConfirmation: boolean; // whether or not the tx needs external confirmation (IE from a subscription even)
+  // whether or not the tx needs external confirmation (IE from a subscription even)
+  requiresConfirmation: boolean;
+  // whether or not to notify via toast
+  // true = force open toast
+  // false = force close toast
+  // undefined = leave alone
+  notify: boolean | undefined;
```
bglownia, line 36: We should be able to make it work as expected without introducing new prop. dialogOpen property toast visibility value.
We could add third argument to update function `keepDismissed?: boolean`
add condition
```
if (!keepDismissed) {
  transaction.dialogOpen = true;
}
```

than just use third parameter in use ethereum transaction manager in only one place
```
if (requiresConfirmation) {
  update(transaction.id, { status: EthTxStatus.Complete, receipt }, true);
}
```

PR #4828, opened 2023-09-20
test(trading): single worker for sim tests

daro-maj:
# Description ℹ️
Due to a problem with the performance of the CI machine, we temporarily change it to one thread because then the tests work. Many threads will be restored after analysing the performance of other machines.


radoslaw-szpiech:
Closing, as we disabled console-test for now, and we'll use github-runners and keep multiple workers.

PR #4842, opened 2023-09-21
chore(trading): deal ticket test to jest

bwallacee:
Refactor cypress tests to jest where possible.

Tests left over have been moved to console-test


MadalinaRaicu (approved):

libs/deal-ticket/src/components/deal-ticket/deal-ticket.spec.tsx
```diff
@@ -654,4 +686,40 @@ describe('DealTicket', () => {
...
+  describe('market states not accepting orders', () => {
+    const states = [
+      Schema.MarketState.STATE_REJECTED,
+      Schema.MarketState.STATE_CANCELLED,
+      Schema.MarketState.STATE_CLOSED,
+      Schema.MarketState.STATE_SETTLED,
+      Schema.MarketState.STATE_TRADING_TERMINATED,
+    ];
+
+    states.forEach((marketState) => {
```
bglownia, line 699: It would be better to use` it.each(states)` instead `states.forEach`

libs/deal-ticket/src/components/deal-ticket/deal-ticket.spec.tsx
```diff
@@ -654,4 +686,40 @@ describe('DealTicket', () => {
...
+      describe(`Market State: ${marketState}`, () => {
+        beforeEach(() => {
```
bglownia, line 701: `beforeEach` and and `it`'s are not needed, they can be removed and you can leave only what they executes

PR #4854, opened 2023-09-22
feat(trading): disable apply code page when applied or referrer

asiaznik:


asiaznik:
too much fuss with rebasing, will cherry pick this to a new branch

PR #4863, opened 2023-09-23
feat(governance): show perp or future pills

### LINKED ISSUE #4385: Show perpetuals product type in proposals
## The Chore

Ensure all fields unique to perpetual markets can be browsed when inspecting related proposals.

## Tasks

- [ ] Clearly label the product type in proposals in both perpetuals AND existing futures markets
  - [ ] One for FUTR
  - [ ] One for PERP 
- [ ] Add some sort of pill to the top tier info (that matches Console)

## Sketch

In the market panels future would need to handle perps too 

Image

In the top level view 

Image

Console pill thingy

Image

sam-keen:
# Related issues 🔗

Closes #4385

# Description ℹ️

Provides details on whether a new market is perp, future or spot. Sat behind a feature flag as the GQL required isn't available in all environments.

# Demo 📺

Screenshot 2023-09-22 at 16 16 12
Screenshot 2023-09-22 at 16 16 02
Screenshot 2023-09-23 at 15 29 28
Screenshot 2023-09-23 at 15 29 20


apps/governance/.env
```diff
@@ -32,3 +32,4 @@ LC_ALL="en_US.UTF-8"
 # Cosmic elevator flags
 NX_SUCCESSOR_MARKETS=true
 NX_METAMASK_SNAPS=true
+NX_PRODUCT_PERPETUALS=false
```
jtsang586, line 35: Question: should these feature flags also be in the e2e env files or are they on by default?

sam-keen: This feature only makes extra API requests when the flag is explicitly set to true, so nothing will fall over if it's not set. I can set the flag for the e2e env files in this PR if you'd like, but equally it's fine to wait until you pick up when you write the tests

jeremyletang (approved):

PR #4870, opened 2023-09-25
feat(deal-ticket): show worst case instead of ranges in estimated values

bglownia:
# Related issues 🔗

Related  to #4870

# Description ℹ️

In deal ticket margin and liquidation price estimation show only worst case value formatted using quantum. In tooltip show range with maximum precision.


mattrussell36 (commented):
@bglownia can we re-branch from `main` and cherry pick the commit onto the new branch? Would be good to be able to deploy this without waiting for perps

MadalinaRaicu (approved):

mattrussell36 (approved):

daro-maj (approved):

PR #4872, opened 2023-09-25
chore(trading): 4733 improve tooltip configurability

### LINKED ISSUE #4733: Improve tooltip configurability
## The Chore

The current tooltip underline feature is annoying, because:

- Too many field names have some sort of tooltip associated to them and end up underlined
- Too many values now have quantum formatting and end up underlined 

Between these factors, basically everything in the app ends up underlined... which looks bad.

## Tasks

- [x] Come up with some way of doing this that is better.  Maybe have it be configurable what is underlined and what isnt? Maybe get rid of underlines and have some sort of cursor change when you move over something with a tip? Maybe there are other options.
- [x] Implement that everywhere for names
- [x] Remove the underline on values generally ... we cant underline everything ... just have the hover
- [x] Consider how we would handle values which are not numerical (eg tradign mode and status). We may still want these underlined

macqbat:
# Related issues 🔗

Closes #4733

# Description ℹ️

Remove underline from tooltip trigger by default. Make it parameterized.

# Demo 📺

Screenshot 2023-09-25 at 14 09 30

# Technical 👨‍🔧

-


mattrussell36 (commented):
@macqbat can you re-branch from `main` and cherry pick this commit onto the new branch. Would be good to be able to deploy this without waiting for perps

PR #4907, opened 2023-09-27
chore: change runner for mainnet compatible vega version

radoslaw-szpiech:
THat change would allow us to use "old" version of vega nad vegacapsule for PRs targeted to `main`. Need to remember to be cautious about it when merging with develop, or add some logic that decide which runner version should be used.

### DIFF
--- a/.github/workflows/cypress-run.yml
+++ b/.github/workflows/cypress-run.yml
@@ -19,7 +19,7 @@ jobs:
       matrix:
         project: ${{ fromJSON(inputs.projects) }}
     name: ${{ matrix.project }}
-    runs-on: self-hosted-runner
+    runs-on: mainnet-compatible-runner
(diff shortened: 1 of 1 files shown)


.github/workflows/cypress-run.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       matrix:
         project: ${{ fromJSON(inputs.projects) }}
     name: ${{ matrix.project }}
-    runs-on: self-hosted-runner
+    runs-on: mainnet-compatible-runner
```
jtsang586, line 22: We need to update the workflow so PRs on Develop are still on the self-hosted runner

radoslaw-szpiech: they will, as this change is going to main only

daro-maj (approved):

PR #4916, opened 2023-09-27
fix(trading): remove market value proxy and formatting

### LINKED ISSUE #4879: Convert time trigger oracle data to display more readable date time
## The Chore

We render the raw timestamp in seconds for time triggered oracles. We should convert this to a more readable date and time

image (1)

## Tasks

- [ ] Add/adjust ACs
- [ ] Convert time in seconds to date and time strings
- [ ] Test

https://vegaprotocol.slack.com/archives/C037STCKMGC/p1695667864311809

### LINKED ISSUE #4894: Switch liquidity supplied to use market param for triggering ratio and remove market value proxy
## The Chore

Currently we're using the network param. 

We should be using the one on the market. The network param is just used as the default when creating a market, unless the user specifies one.

## Tasks

- [ ] Use the market param
- [ ] Remove market value proxy

MadalinaRaicu:
# Related issues 🔗

Closes #4894  
Closes #4879 

# Description ℹ️

- Remove market value proxy after liquidity SLA changes.
- Don't use network triggering ratio but the one from market liq. monitoring parameters.
- Format unix timestamp to date in internal conditions termination oracle.

# Demo 

image

Screenshot 2023-09-29 at 19 49 43


mattrussell36 (commented):
Looks good, but you need to add the liquidityMonitoringParameters to some test helpers and mocks

libs/fills/src/lib/test-helpers.ts
```diff
@@ -35,13 +35,17 @@ export const generateFill = (override?: PartialDeep<Trade>) => {
       infrastructureFee: '200',
       liquidityFee: '200',
     },
+
```
macqbat, line 38: non-blocking: this line break is redundant

macqbat (approved):

daro-maj (approved):

libs/fills/src/lib/test-helpers.ts
MadalinaRaicu, line 38:
```suggestion
```

pscott31 (approved):

Issue #4775, opened 2023-09-13
Text on the connection modal changes depending on state


JonRay15:
## The Chore

NB. Might make sense to do with #4774 and #4772

We're showing two versions of the connection modal depending on what you have installed.  We think we dont want that.

With nothing installed you get all this info 
![image](https://github.com/vegaprotocol/frontend-monorepo/assets/89918817/11e5fd2c-b383-434f-9866-109af7d5bab7)

Once you have installed both things you get this 
![image](https://github.com/vegaprotocol/frontend-monorepo/assets/89918817/3f9b1298-4e56-4b58-9185-84dc334a68b2)

We think you should get exactly the same descriptive text, ie. 

<img width="304" alt="image" src="https://github.com/vegaprotocol/frontend-monorepo/assets/89918817/4dab0dfe-2c5c-4038-87b8-df4b67fa9ab5">

and

<img width="288" alt="image" src="https://github.com/vegaprotocol/frontend-monorepo/assets/89918817/c4fd1fb8-bbc2-4d5b-82d7-7c0bd26d2040">

with the only exception that once something is installed and detected it should say "Connect with" instead of "Install"

Also ... the font looks slightly different on the two ... greyer on once maybe?

## Tasks

- [ ] Adjust wording for once vega wallet is detected
- [ ] Adjust wording for once snap is detected
- [ ] Make sure both fonts are the same
- [ ] Test it

Issue #4772, opened 2023-09-13
The disclaimer / agreement modal opens behind the onboarding dialogue


JonRay15:
## Description

See title

NB. probably do with #4774 and #4775 

## Steps to Reproduce

1. Clear your storage
2. Go to console
3. Click on connect on the onboarding modal
4. The connection box will open, connect via snap (update: or via browser wallet ... same issue)
5. The legal agreement modal will appear behind the onboarding dialogue

## Expected behavior

The legal agreement shoudl be on top of the onboarding dialogue so that you can tick it off and then continue your onboarding.

## Screenshots

![image](https://github.com/vegaprotocol/frontend-monorepo/assets/89918817/b93defba-1992-40a2-83c2-8b762f947a30)

PR #4920, opened 2023-09-28
chore(trading): 4774 put connect dialog  inside onboarding dialog - main branch

### LINKED ISSUE #4774: Modals stacking and dont look nice
## The Chore

We get looks of diff size modals stacking on each other which doesnt look very clean. 

Arguably we need some Joelle magic on these to work out some uber-modal of something ... but for now I suggest we just make them all the same size so they stack neatly on top of each other?

Impacts
- Onboarding modal 
- Connection modal (changes in #4775 might make sense to do in parallel with this)
- Legal disclaimer (which is behind the onboarding modal below ... this is covered in #4772 )

## Tasks

- [x] Decide what to do .... is making the same size ok?
- [x] Do it
- [ ] Test it

image

image

image

macqbat:
# Related issues 🔗

Closes #4774 #4775 #4772

# Description ℹ️

Add connect dialog into onboarding dialog to prevent them being stacked

# Demo 📺

https://github.com/vegaprotocol/frontend-monorepo/assets/338931/09c0d7da-a2a2-4f06-8384-6aae8dfafac4

# Technical 👨‍🔧

-

### DIFF
--- a/apps/trading/components/welcome-dialog/use-get-onboarding-step.ts
+++ b/apps/trading/components/welcome-dialog/use-get-onboarding-step.ts
@@ -12,16 +12,20 @@ import { useGlobalStore } from '../../stores';
 const ONBOARDING_STORAGE_KEY = 'vega_onboarding';
 export const useOnboardingStore = create<{
   dialogOpen: boolean;
+  walletDialogOpen: boolean;
   dismissed: boolean;
   dismiss: () => void;
   setDialogOpen: (isOpen: boolean) => void;
+  setWalletDialogOpen: (isOpen: boolean) => void;
 }>()(
   persist(
     (set) => ({
       dialogOpen: true,
+      walletDialogOpen: false,
       dismissed: false,
       dismiss: () => set({ dismissed: true }),
       setDialogOpen: (isOpen) => set({ dialogOpen: isOpen }),
+      setWalletDialogOpen: (isOpen) => set({ walletDialogOpen: isOpen }),
     }),
     {
       name: ONBOARDING_STORAGE_KEY,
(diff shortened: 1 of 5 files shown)


apps/trading/components/welcome-dialog/use-get-onboarding-step.ts
```diff
@@ -12,16 +12,20 @@ import { useGlobalStore } from '../../stores';
 const ONBOARDING_STORAGE_KEY = 'vega_onboarding';
 export const useOnboardingStore = create<{
   dialogOpen: boolean;
+  walletDialogOpen: boolean;
```
mattrussell36, line 15: (non blocking): This is not a good variable name as its not controlling a dialog. Just the state within the dialog. Lets not change it though to avoid conflicts with develop

libs/wallet/src/connect-dialog/connect-dialog.tsx
```diff
@@ -50,11 +50,15 @@ export type WalletType = 'injected' | 'jsonRpc' | 'view' | 'snap';
...
   connectors: Connectors;
   riskMessage?: ReactNode;
+  contentOnly?: boolean;
+  onClose?: () => void;
 }
 
 export const VegaConnectDialog = ({
   connectors,
   riskMessage,
+  contentOnly,
```
mattrussell36, line 60: I don't like this approach. Its weird to me that you pass a prop to unwrap the component. Lets not change now though

mattrussell36 (approved):

libs/wallet/src/connect-dialog/connect-dialog.tsx
mattrussell36: Why not just import and use the inner child component directly. This is unnecessary logic

daro-maj (approved):

PR #4924, opened 2023-09-28
chore(ci): sync main to develop

radoslaw-szpiech:
# Description ℹ️

This is basically PR that merges `main` -> `develop`, but with added fix in genesis.tmpl, as develop is running on the latest core version and need this additional parameter that was removed from `main`.

### DIFF
--- a/vegacapsule/genesis.tmpl
+++ b/vegacapsule/genesis.tmpl
@@ -196,7 +196,6 @@
 		"market.fee.factors.makerFee": "0.004",
 		"market.liquidity.stakeToCcyVolume": "0.3",
 		"market.liquidity.targetstake.triggering.ratio": "0.7",
-    "market.liquidity.providersFeeCalculationTimeStep": "5s",
 		"network.checkpoint.timeElapsedBetweenCheckpoints": "10s",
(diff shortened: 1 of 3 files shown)


vegacapsule/genesis.tmpl
```diff
@@ -196,7 +196,7 @@
 		"market.fee.factors.makerFee": "0.004",
 		"market.liquidity.stakeToCcyVolume": "0.3",
 		"market.liquidity.targetstake.triggering.ratio": "0.7",
-    "market.liquidity.providersFeeCalculationTimeStep": "5s",
+		"market.liquidity.providersFeeCalculationTimeStep": "5s",
```
MadalinaRaicu, line 199: this one was breaking something at one point when it got removed

MadalinaRaicu: that was an SLA  parameter that has been removed between tags.
Reason being it was a duplicate, we had it as a network parameter and  and market parameter. It was removed from the market fields.

MadalinaRaicu: let's give it a go for main/develop consistency. 
if it breaks again on develop capsule we know we need to remove it.

jtsang586 (approved):

PR #4940, opened 2023-09-29
feat(positions): close position with max possible size

### LINKED ISSUE #4834: Make close position us max possible size
## The Chore

## Tasks

- [ ] Confirm max possible size with core
- [ ] Use that for the close position transaction (rather than open volume) so that there is no chance of fills happening whilst the tx is going through

https://vegaprotocol.slack.com/archives/C037STCKMGC/p1695104903200239

macqbat:
# Related issues 🔗

Closes #4834 

# Description ℹ️

Add max golang uint64 as size of closing position tx.

### DIFF
--- a/libs/positions/src/lib/positions-manager.tsx
+++ b/libs/positions/src/lib/positions-manager.tsx
@@ -19,6 +19,8 @@ interface PositionsManagerProps {
   showClosed?: boolean;
 }
 
+const MAXUINT64 = '18446744073709551615';
+
 export const PositionsManager = ({
   partyIds,
   onMarketClick,
@@ -46,7 +48,7 @@ export const PositionsManager = ({
               side: openVolume.startsWith('-')
                 ? Schema.Side.SIDE_BUY
                 : Schema.Side.SIDE_SELL,
-              size: openVolume.replace('-', ''),
+              size: MAXUINT64, // workaround for avoiding leftovers filled in the meantime
               reduceOnly: true,
             },
           ],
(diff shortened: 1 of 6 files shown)


libs/positions/src/lib/positions-manager.tsx
```diff
@@ -19,6 +18,8 @@ interface PositionsManagerProps {
   showClosed?: boolean;
 }
 
+const MAXINT64 = '9223372036854775807';
```
mattrussell36, line 21: We use the same constant for deposit approvals. Perhaps we can add this to utils and share the value.

Can we also add some comments saying why we use this value

mattrussell36 (changes requested):

libs/positions/src/lib/positions-manager.tsx
macqbat: Only we need there max 64 bit int for golang, and in deposit we have max 256 bit uint from ethereum lib, and I can't see there is int64 available. So I'll just move it to some constants file.

bwallacee: is it worth getting a test in for deposit approval as well if we are building this into utils?

macqbat: > is it worth getting a test in for deposit approval as well if we are building this into utils?

I think deposit will remain like it is now, with uint256 from ethereum and close position tx with int64 stored in the utilis lib

bwallacee (approved):

MadalinaRaicu (approved):

PR #4962, opened 2023-10-03
chore(trading): add accounts jest test

bwallacee:
Moving over some cypress tests to jest

### DIFF
--- a/libs/accounts/src/lib/accounts-actions-dropdown.spec.tsx
+++ b/libs/accounts/src/lib/accounts-actions-dropdown.spec.tsx
@@ -1,0 +1,75 @@
+import { render, screen, fireEvent } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+import { AccountsActionsDropdown } from './accounts-actions-dropdown';
+import userEvent from '@testing-library/user-event';
+
+describe('AccountsActionsDropdown', () => {
+  let onClickDeposit: jest.Mock;
+  let onClickWithdraw: jest.Mock;
+  let onClickBreakdown: jest.Mock;
+  let onClickTransfer: jest.Mock;
+
+  beforeEach(() => {
+    onClickDeposit = jest.fn();
+    onClickWithdraw = jest.fn();
+    onClickBreakdown = jest.fn();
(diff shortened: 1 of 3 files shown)


radoslaw-szpiech (approved):

MadalinaRaicu (approved):

libs/accounts/src/lib/accounts-actions-dropdown.spec.tsx
```diff
@@ -0,0 +1,75 @@
...
+        assetId="testAssetId"
+        assetContractAddress="testAssetContractAddress"
+        onClickDeposit={onClickDeposit}
+        onClickWithdraw={onClickWithdraw}
+        onClickBreakdown={onClickBreakdown}
+        onClickTransfer={onClickTransfer}
+      />
+    );
+
+    userEvent.click(screen.getByTestId('icon-kebab'));
```
mattrussell36, line 34: If you await the userEvent you won't need to use await screen.findByTestId and can just use screen.getByTestId below

```suggestion
    await userEvent.click(screen.getByTestId('icon-kebab'));
```

libs/accounts/src/lib/accounts-actions-dropdown.spec.tsx
```diff
@@ -0,0 +1,75 @@
...
```
mattrussell36, line 59: Better to use an awaited userEvent where possible. This will wait for the ui to update before proceeding. The following would be better

```
await userEvent.click(screen.getByTestId('icon-kebab'));
await userEvent.click(screen.getByTestId('deposit'));
expect(onClickDeposit).toHaveBeenCalledTimes(1);

```

libs/accounts/src/lib/accounts-actions-dropdown.spec.tsx
bwallacee: fixed

libs/accounts/src/lib/accounts-actions-dropdown.spec.tsx
bwallacee: fixed

mattrussell36 (approved):

PR #4976, opened 2023-10-04
chore(trading): 4858 redesign market header

### LINKED ISSUE #4858: Rejig the perps market info slightly
## The Chore

The order on the market info doesnt feel right for perps.

I personally think it is partly because its wrong for futures too, and by following the same format we've got into trouble.

**Perps**
image

**Futures**
image

## Tasks

I would 

- [x] Move the expiry to the far right of the page 
- [x] That will mean we naturally move the funding / countdown to the right too
- [x] That leaves price as the first thing which feels correct ... and we should change it to "Mark Price" so its more explicit
- [x] We then also want to add a "Index Price" (External TWAP from market data API, formatted using asset decimals) to the far right of the perp one, as this is the price that is driving the funding rate
- [x] We also want to rename "Funding" to "Funding Rate" to be more explicit
- [x] We probably then want to optimise the spacing so the words and numbers line up nicely, currently they're a bit off and will get worse with the addition of "rate" as above

macqbat:
# Related issues 🔗

Closes #4858 

# Description ℹ️

Do some adjusts and tweaks in market header view.

# Demo 📺

image

# Technical 👨‍🔧


bwallacee:
Just so we have the info here as well as the ticket:
We need to be using asset decimals to format the index price / external TWAP.

macqbat:
> Just so we have the info here as well as the ticket: We need to be using asset decimals to format the index price / external TWAP.

yeah, I know, this is actually already done :+1:

bwallacee (approved):

PR #4978, opened 2023-10-04
chore(trading): set default sort order

bwallacee:
# Description ℹ️

Sets the usage breakdown to order by market desc, as previously it would flip flop between orders (making testing hard).

# Demo 📺

image


bwallacee:
@JonRay15 Are you happy from a UX perspective for this change?

jtsang586 (approved):

MadalinaRaicu (approved):

PR #5000, opened 2023-10-09
feat(trading): add risk factors leverage and update accordion panel

### LINKED ISSUE #4931: Add slippage factors and margin scaling factors to market accordion
## The Chore

We are missing a bunch of things from the risk model section of the accordion ... and presumably always have been. 

These need to be added

## Tasks

- [ ] Add margin scaling factors to accordion
- [ ] Add slippage factors
- [ ] Remove "Risk Parameters" heading and just put sigma under the "Risk Model"
- [ ] Add max initial leverage and max leverage (both long and short) to the risk factors

Current view:

image

Margin scaling needs to be in, and formatted nicely:

image

Slippage factors are easier ... not sure why they werent always there.

image

### LINKED ISSUE #4971: Add market status to Explorer market view (accordion)
## The Chore

In the [market view](https://explorer.vega.xyz/markets/2c2ea995d7366e423be7604f63ce047aa7186eb030ecc7b77395eae2fcbffcc5) in explorer we only show the trading mode and not the status for some reason (even though it is on the same markets API).  We would like to add it.

## Tasks

- [ ] Add the status
- [ ] Test it

### LINKED ISSUE #4998: Update font on market info accordion
## The Chore

Changes to make the info more readable

1. Reduce the font size of both the field and value so it is appreciably smaller than the heading (liquidity SLA protocol)
2. Change the text of the field to be grey, this would align to how we show field / value in the deal ticket 

image

MadalinaRaicu:
# Related issues 🔗

Closes #4931
Closes #4998 
Closes #4971 

# Description ℹ️

- [x] add slippage factors and market factors to market accordion in console and details in explorer
- [x] update market accordion and details to show labels muted and values unmuted
- [x] add market status in market accordion
- [x] update market ID to be able to copy it and see it truncated for better alignment

# Demo 📺

Screenshot 2023-10-09 at 12 22 06
Screenshot 2023-10-09 at 12 21 51

# Technical 👨‍🔧


MadalinaRaicu:
> I think leverage can be rounded to integer value (ie no decimals) if it's >= 10 and we show one decimal for leverage between 0 and 10.

Screenshot 2023-10-10 at 10 58 45

I suggest we should leave more decimals. Otherwise, values might not be very helpful.  
Suggested:

Screenshot 2023-10-10 at 11 03 28

jeremyletang (approved):

asiaznik (approved):

PR #5036, opened 2023-10-12
chore(trading): deposit test to jest

bwallacee:
Moving deposit cypress tests to jest.


daro-maj (approved):

libs/accounts/src/lib/accounts-table.spec.tsx
```diff
@@ -71,6 +93,31 @@ describe('AccountsTable', () => {
...
+        onClickDeposit={onClickDepositMock}
+        pinnedAsset={{
+          decimals: 5,
+          id: '5cfa87844724df6069b94e4c8a6f03af21907d7bc251593d08e4251043ee9f7c',
+          symbol: 'tBTC',
+          name: 'tBTC',
+        }}
+      />
+    );
+    const depositButton = container.querySelector('[data-testid="deposit"');
```
mattrussell36, line 111: Use screen, and getbyTestId

```suggestion
  const depositButton = screen.getByTestId('deposit')
```

libs/accounts/src/lib/accounts-table.spec.tsx
```diff
@@ -71,6 +93,31 @@ describe('AccountsTable', () => {
...
+    expect(depositButton).toBeVisible();
+    if (depositButton) {
+      fireEvent.click(depositButton);
```
mattrussell36, line 114: Please use userEvent unless theres a good reason not to
```suggestion
    await userEvent.click(depositButton);
```

libs/accounts/src/lib/accounts-table.spec.tsx
```diff
@@ -71,6 +93,31 @@ describe('AccountsTable', () => {
...
+          decimals: 5,
+          id: '5cfa87844724df6069b94e4c8a6f03af21907d7bc251593d08e4251043ee9f7c',
+          symbol: 'tBTC',
+          name: 'tBTC',
+        }}
+      />
+    );
+    const depositButton = container.querySelector('[data-testid="deposit"');
+    expect(depositButton).toBeVisible();
+    if (depositButton) {
```
mattrussell36, line 113: We should avoid conditions in test.

screen.getByTestId will fail if the button is not found

mattrussell36 (changes requested):

mattrussell36 (approved):

PR #5042, opened 2023-10-13
fix(trading): remove additional page for closed markets

### LINKED ISSUE #4961: chore: refactor console navigation - each main tab should start from different path
## The Chore

Currently we have two other tabs in main navigation (Markets, Trading) with the same basic starting path `/markets`. It's caused some problems when we need to add some deeper routes and mark certain menu position as active in `Navlink` react router component. 
Because of that in this [PR](https://github.com/vegaprotocol/frontend-monorepo/pull/4958) was needed to add not nice looking path for closed markets, like _/markets/all/closed/:marketId_ to make as active different main tab in main navbar. If each main routes had have different starting path, such long path shouldn't be necessary and other navi configurations would be easy and seemed exactly as how it should be done. 

So the proposal of change are: 
- first navi tab Markets should have path started from `/markets`, e.g `/markets/all`, `/markets/open`, `/markets/closed` etc.
- second navi tab should have path started from `/market`, e.g.` /market/:marketId`

These would allow for greater order in the configuration.

## Tasks

- [ ] Change trading path and url where user can trade for starting from `/market`
- [ ] Change `to` parameter of NaviLink from `/markets/all` to `/markets`
- [ ] Review all links and internal redirects to make sure they not follow to old url's nor cause redirection loop

All urls of tradable markets will be affected by this, so some redirecting in `MarketPage` component would be needed

mattrussell36:
# Related issues 🔗

Closes #4961 

# Description ℹ️

- Removes the additional page and redirect for a closed market
- Changes /markets/all to just /markets

# Demo 📺

N/A

# Technical 👨‍🔧

- Refactored routing to 'flatten' the routes and make it easier to manage
- Refactored LayoutWithSidebar to not care about routing and instead take a header and sidebar content has props. This makes managing the sidebar much easier and more declarative. We can simply make a component for the sidebar for each page.

### DIFF
--- a/apps/trading/client-pages/closed-market/closed-market.tsx
+++ b/apps/trading/client-pages/closed-market/closed-market.tsx
@@ -1,5 +1,0 @@
-import MarketPage from '../market';
-
-export const ClosedMarketPage = () => {
-  return <MarketPage closed />;
-};
--- a/apps/trading/client-pages/closed-market/index.ts
+++ b/apps/trading/client-pages/closed-market/index.ts
@@ -1,1 +1,0 @@
-export { ClosedMarketPage as default } from './closed-market';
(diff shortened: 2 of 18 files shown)


bwallacee (approved):

bwallacee:
@mattrussell36 I have created this PR to update console-test with these changes: https://github.com/vegaprotocol/console-test/pull/170
We can merge in as soon as you are ready to merge this.

PR #5050, opened 2023-10-16
fix(trading): liquidation price tooltip link

### LINKED ISSUE #4900: Add link to spec to liquidation price tooltip
## The Chore

Current tooltip just says 
image

We should add a link to the specs for full details

## Tasks

- [x] Add "for full details please see https://github.com/vegaprotocol/specs/blob/master/non-protocol-specs/0012-NP-LIPE-liquidation-price-estimate.md"
- [x] Change the range part because we're swapping this to a worst case view as part of an in-flight ticket

MadalinaRaicu:
# Related issues 🔗

Closes #4900 

# Description ℹ️

- [x] add liquidation price tooltip link
- [x] remove range

# Demo 📺

image

# Technical 👨‍🔧


daro-maj (changes requested):
@MadalinaRaicu range is still visible:
image

daro-maj (approved):

PR #5051, opened 2023-10-16
chore(trading):  trades test mock timezone

### LINKED ISSUE #5041: Failing test: trades table timestamp format
## Description

This test: https://github.com/vegaprotocol/frontend-monorepo/blame/341d78e761d8e7bbe7b8bda6b8765ab8dec20695/libs/trades/src/lib/trades-table.spec.tsx#L81-L96

Fails depending on timestamps

image

## Expected behavior

Test should pass locally as well in CI, regardless of 24h timestamp configuration or not

bwallacee:
# Related issues 🔗

Closes #5041

### DIFF
--- a/libs/trades/src/lib/trades-table.spec.tsx
+++ b/libs/trades/src/lib/trades-table.spec.tsx
@@ -4,18 +4,6 @@ import { SELL_CLASS, TradesTable, BUY_CLASS } from './trades-table';
 import type { Trade } from './trades-data-provider';
 import { Side } from '@vegaprotocol/types';
 
-const timezoneMock = function (zone: string) {
-  const DateTimeFormat = Intl.DateTimeFormat;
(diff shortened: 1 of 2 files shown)


libs/trades/src/lib/trades-table.spec.tsx
```diff
@@ -4,6 +4,18 @@ import { SELL_CLASS, TradesTable, BUY_CLASS } from './trades-table';
 import type { Trade } from './trades-data-provider';
 import { Side } from '@vegaprotocol/types';
 
+const timezoneMock = function (zone: string) {
```
mattrussell36, line 7: This should probably use a beforeAll and you also should add an afterAll to clear the mock

mattrussell36 (changes requested):

daro-maj (approved):

PR #5063, opened 2023-10-17
chore(ui-toolkit): 4970 async renderer - improve error handling

### LINKED ISSUE #4970: Distinguish between "no data" and "datanode is broken"
## The Chore

It was raised that when we have no orders for a given market AND when we are connected to a screwed datanode we get the same messaging in the orderbook as below. 

image

We want to be able to differentiate these states.

## Tasks

- [ ] If there are genuinely no orders on a market, show "No open orders" instead
- [ ] Determine if we can detect if a datanode is screwed and display some message to indicate this rather than this generic sounding message

macqbat:
# Related issues 🔗

Closes #4970 

# Description ℹ️

Improve conditions under which the error is displayed

# Demo 📺

-

# Technical 👨‍🔧

-


bwallacee:
@macqbat I think you are missing the task:

>  If there are genuinely no orders on a market, show "No open orders" instead

MadalinaRaicu (approved):

PR #5076, opened 2023-10-18
chore(orders): 4970 add no rows and error info

### LINKED ISSUE #4970: Distinguish between "no data" and "datanode is broken"

macqbat:
# Related issues 🔗

Closes #4970

# Description ℹ️

Handle overlays appearing according to query or filter results

# Demo 📺

image
Screenshot 2023-10-18 at 15 45 54
Screenshot 2023-10-18 at 15 45 58
Screenshot 2023-10-18 at 15 46 02
image

# Technical 👨‍🔧

### DIFF
--- a/apps/trading/components/orders-container/orders-container.tsx
+++ b/apps/trading/components/orders-container/orders-container.tsx
@@ -10,6 +10,19 @@ import type { DataGridStore } from '../../stores/datagrid-store-slice';
 import { OrderStatus } from '@vegaprotocol/types';
 import { Links } from '../../lib/links';
 
+const resolveNoRowsMessage = (filter?: Filter) => {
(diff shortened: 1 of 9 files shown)


macqbat:
> @macqbat We are missing No open orders on the chart and the orderbook: image

Missing places have been fixed

bwallacee (approved):

PR #5092, opened 2023-10-19
feat(governance): transfers destination field can fallback to type

### LINKED ISSUE #5085: Governance transfers - 'to' field can be empty
## Story

If a transfer is set up with a dispatch strategy, it's possible and valid that the transfer 'to' field may be empty.

In that case, the destination type should be shown.

sam-keen:
# Related issues 🔗

Closes #5085

# Description ℹ️

To avoid this (valid) situation:

Screenshot 2023-10-18 at 12 22 46

We now fall back to the destination type, as we do for the source type if the source field isn't set.


edd (approved):
👍

candida-d (approved):

PR #5093, opened 2023-10-19
for preview links:  LP SLA table collapsed

MadalinaRaicu:


libs/liquidity/src/lib/liquidity-table.spec.tsx
```diff
@@ -41,16 +43,16 @@ describe('LiquidityTable', () => {
       h.querySelector('[ref="eText"]')?.textContent?.trim()
     );
     const expectedHeaders = [
+      undefined,
```
mattrussell36, line 46: Can you add a comment about why this is undefined. Otherwise its a bit odd/confusing

libs/liquidity/src/lib/liquidity-table.tsx
```diff
@@ -24,6 +31,20 @@ const percentageFormatter = ({ value }: ValueFormatterParams) => {
   return formatNumberPercentage(new BigNumber(value).times(100), 2) || '-';
 };
 
+const copyCellRenderer = ({ value }: { value?: string | null }) => {
```
mattrussell36, line 34: This is a component, so lets make it upper camel

Issue #5086, opened 2023-10-18
Order table pagination


mattrussell36:
## Story

As a trader
I want to be able to see all orders and not be limited to the 5000 order cap
So that I can see all older orders

## Acceptance Criteria

- [x] I can click to load more orders
- [x] The orders are filter as per the current filter criteria
- [x] The orders are sorted as per the current sort criteria
- [x] If you get to the last page and no more orders are available, include some wording around the fact that depending on datanode retention you _may not_ be able see the "full" history

## Tasks

- [ ] UX (if needed)
- [ ] Design (if needed)
- [ ] Team and stakeholder review
- [ ] Specs reviewed and created or adjusted
- [ ] Implementation
- [ ] Testing (unit and/or e2e)
- [ ] Code review
- [ ] QA review

## Additional details / background inf

https://vegaprotocol.slack.com/archives/C037STCKMGC/p1696518683350629

PR #5105, opened 2023-10-23
feat(trading): show full history of orders, fills and trades

### LINKED ISSUE #4898: Update no rows found messaging in datagrids
## The Chore

Some lists such as fills and orders do not show the full history when there are many entries. 

In these cases we need to indicate how many are shown and why.

## Tasks

- [x] Identify all places this happens
- [ ] Update no rows message to indicate if filters are affecting rows displayed.
- [ ] If table has 0 rows include some wording around the fact that the number of rows is capped and that depending on datanode retention you _may not_ be able see the "full" history
  - Note that there is still an edge case where if there are more than 5000 rows (for orders) you will not be 
- [ ] ~~Provide some guidance on how to see full history~~ This will be covered by #5086 

https://vegaprotocol.slack.com/archives/C037STCKMGC/p1696518683350629

bglownia:
# Related issues 🔗

Closes #4898, #5086

# Description ℹ️

Add option to load another page of trades and orders. Inform how many items where loaded. Make it clear that results are limited by filters applied.

# Demo 📺

image

image

### DIFF
--- a/libs/fills/src/lib/fills-manager.tsx
+++ b/libs/fills/src/lib/fills-manager.tsx
@@ -1,11 +1,12 @@
 import type { AgGridReact } from 'ag-grid-react';
-import { useRef, useState } from 'react';
+import { useCallback, useRef, useState } from 'react';
 import { t } from '@vegaprotocol/i18n';
 import { FillsTable } from './fills-table';
 import type { useDataGridEvents } from '@vegaprotocol/datagrid';
 import { useDataProvider } from '@vegaprotocol/data-provider';
 import type * as Schema from '@vegaprotocol/types';
 import { fillsWithMarketProvider } from './fills-data-provider';
+import { TradingButton as Button } from '@vegaprotocol/ui-toolkit';
 
 interface FillsManagerProps {
   partyId: string;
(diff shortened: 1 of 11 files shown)


libs/fills/src/lib/fills-manager.tsx
```diff
@@ -22,26 +23,68 @@ export const FillsManager = ({
...
+            : t('%s items loaded', [
+                data?.length ? data.length.toString() : ' ',
+              ])}
+          {pageInfo?.hasNextPage ? (
+            <Button size="extra-small" className="ml-1" onClick={() => load()}>
+              {t('Load more')}
+            </Button>
+          ) : null}
+        </div>
+        {data?.length && hasFilteredRows === false ? (
```
mattrussell36, line 82: Can we not use overlayNoRowsTemplate? Is it rendered if there are rows in memory but none after filtering?

libs/orders/src/lib/components/order-list-manager/order-list-manager.tsx
```diff
@@ -96,27 +80,84 @@ export const OrderListManager = ({
...
+            {...props}
+          />
+          <div className="flex justify-between border-t border-default p-1 items-center">
+            <div className="text-xs">
+              {t(
+                'Depending on data node retention you may not be able see the "full" history'
+              )}
+            </div>
+            {data ? (
+              <div className="flex text-xs items-center">
```
mattrussell36, line 137: Probably worth adding a datagrid or ui-toolkit component for the pagination component

mattrussell36 (changes requested):

libs/fills/src/lib/fills-manager.tsx
bglownia: Ag-grid is not showing no rows overlay if there are no rows to display because of filters applied.

libs/orders/src/lib/components/order-list-manager/order-list-manager.tsx
```diff
@@ -96,27 +80,84 @@ export const OrderListManager = ({
     [create]
   );
 
+  const [hasFilteredRows, setHasFilteredRowsCount] = useState<
```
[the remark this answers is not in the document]
bglownia: I used getDisplayedRowCount. But displayedRowCount need to be stored in state to rerender components and hide/show no rows matching selected filter text.

libs/orders/src/lib/components/order-list-manager/order-list-manager.tsx
bglownia: I don't think it's necessary. This is just two div elements with classes. Most probably we will be changing it's behaviour depend on place where it's used

mattrussell36 (approved):

bwallacee:
@bglownia the number of orders only updates on a grid refresh (page refresh or swapping between tabs in positions grid) Is this intentional? Or would we want it to update as orders are placed.

I am getting 1000 + orders loaded in rejected table even though i get a message saying i dont have any:
image

Also the Load more button will only appear if i reload the page after i have 1000+ orders.
If i have 1001 orders place but haven't refreshed the page the button wont appear.

guoguojin (approved):

PR #5109, opened 2023-10-23
chore(trading): withdraw tests to jest

bwallacee:
Closes: https://github.com/vegaprotocol/console-test/issues/180


MadalinaRaicu (approved):

daro-maj (approved):

libs/withdraws/src/lib/withdraw-manager.spec.tsx
```diff
@@ -70,12 +72,20 @@ describe('WithdrawManager', () => {
...
+    // 1002-WITH-008
+    // 1002-WITH-018
     render(generateJsx(props));
 
     // Set other fields to be valid
     fireEvent.change(screen.getByLabelText('Asset'), {
       target: { value: props.assets[0].id },
     });
+    expect(
+      await screen.getByTestId('connect-eth-wallet-btn')
```
mattrussell36, line 86: getByTestId does not return a promise

libs/withdraws/src/lib/withdraw-manager.spec.tsx
```diff
@@ -70,12 +72,20 @@ describe('WithdrawManager', () => {
...
 
   it('validates correctly', async () => {
+    // 1002-WITH-010
+    // 1002-WITH-005
+    // 1002-WITH-008
+    // 1002-WITH-018
     render(generateJsx(props));
 
     // Set other fields to be valid
     fireEvent.change(screen.getByLabelText('Asset'), {
```
mattrussell36, line 82: (non blocking): Might be worth updating to userEvent

libs/withdraws/src/lib/withdraw-manager.spec.tsx
```diff
@@ -104,6 +114,13 @@ describe('WithdrawManager', () => {
     ).toBeInTheDocument();
     expect(props.submit).not.toBeCalled();
   });
+  it('can set amount using use maximum button', async () => {
+    // 1002-WITH-004
+    render(generateJsx(props));
+
+    fireEvent.click(screen.getByTestId('use-maximum'));
```
mattrussell36, line 121: (non blocking): can we use userEvent

mattrussell36 (changes requested):

mattrussell36 (approved):

Issue #5155, opened 2023-10-31
Further tweaks to referee stats page


JonRay15:
## The Chore

A few observations on this view 

![image](https://github.com/vegaprotocol/frontend-monorepo/assets/89918817/467ce213-f12c-438b-a25b-c148887e5ef8)

## Tasks

- [ ] Combined volume >> Combined set volume (last <windowLength> epochs)
- [ ] Volume to next tier (3) >> I dont know what this 3 is ... guess its <windowLength> ... I think just remove it
- [ ] Epochs to next tier (3) >> again dont know why the 3 is relevant.... remove it

JonRay15:
Have split the last bit of this out into [this ticket](https://github.com/vegaprotocol/frontend-monorepo/issues/5203) and will close this as done.

PR #5160, opened 2023-10-31
chore(trading): referrals stats tweaks

mattrussell36:
# Related issues 🔗

Partially addresses #5155 

# Description ℹ️

- Removes next tier indicators
- Show windowLength value for combined set volume on referee view
- Adds tooltips for qUSD

# Technical 👨‍🔧

TODO:
- [ ] Add additional query to get referral set stats for last 30 epochs so we can display total comission from the last 30 epochs rather than the last 3 (other values should still use last 3 epochs)


MadalinaRaicu (approved):

JonRay15:
This looks good..

PR #5175, opened 2023-11-01
chore(trading): misc changes to trades and transfer

mattrussell36:
# Related issues 🔗

N/A

# Description ℹ️

- Removes pagination from trades list for a market
- Removes quotes from data node retention message
- Fixes an issue where the transfer form doesn't render if not connected to a pubkey
- Adds balances to to/from account dropdowns as otherwise you don't know how much you can transfer
- Adds vested and vesting accounts to breakdown table


JonRay15:
**Checked**

- Removes pagination from trades list for a market - DONE
- Removes quotes from data node retention message - DONE
- Fixes an issue where the transfer form doesn't render if not connected to a pubkey - DONE
- Adds balances to to/from account dropdowns as otherwise you don't know how much you can transfer - DONE (but other UX issues raised elsewhere)
- Adds vested and vesting accounts to breakdown table - Assuming this means in the transfer, DONE

JonRay15 (commented):
Looks good

PR #5199, opened 2023-11-05
chore(trading): add console-test

bwallacee:
There are a few decisions with this PR i wanted to explain:

### **Test Running:**
I have decided to keep the tests running using docker. This may not be the final solution but for now I have chosen it for the following reasons:

- Less changes to do with our github actions. (They are already using docker images to run against)
- Keeps running tests in parallel simpler as we can just kill containers and not worry about any resources left hanging around.
- We can pull or send docker images around to test against. 

The downsides are:

- Slightly more steps for local test runs (Have to create a docker image first)
- We are testing console in a way that is different to how it is in production.

### **Location of the tests**

I have put the tests within the trading app itself as I believe this has the least chance of breaking anything with NX.

I have tested the commands and running the tests but If a few people could give it a test that would be great

@daniel1302 Would you please be able to have a look at the workflow changes. Thanks.


.gitignore
```diff
@@ -52,3 +52,19 @@ cypress.env.json
 /apps/**/cypress/reports/
 /apps/**/cypress/downloads/
 /apps/**/fixtures/wallet/node**
+
+#console-test
+apps/console-test/node_modules/
+apps/console-test//test-results/
+apps/console-test//playwright-report/
```
mattrussell36, line 59: Do we need the double slash?

package.json
```diff
@@ -13,7 +13,10 @@
     "lint:all": "nx run-many --all --target=lint",
     "e2e:all": "nx run-many --all --target=e2e",
     "vegacapsule": "vegacapsule network bootstrap --config-path=../frontend-monorepo/vegacapsule/config.hcl",
-    "release": "git checkout develop ; git pull ; node scripts/make-release.js"
+    "release": "git checkout develop ; git pull ; node scripts/make-release.js",
+    "install:console-test": "cd apps/console-test && poetry install && poetry run playwright install chromium && python -m vega_sim.tools.load_binaries --force --version v0.73.1",
```
mattrussell36, line 17: Can we make these commands consistent and for now can we just name them the same as the app directory I'd suggest

Maybe
```
trading:test
trading:test:install
trading:test:all
```

mattrussell36: Also is there anyway we can check that poetry is running, perhaps we do this in the top level fixture?

.github/workflows/console-test-run.yml
```diff
@@ -154,24 +154,6 @@ jobs:
           docker load --input /tmp/console-image.tar
           docker image ls -a
 
-      #----------------------------------------------
```
daniel1302, line 157: What code are we going to build below if we are removing the checkout action?

bwallacee: I thought we just needed to build the trading app now that the console-test repo is within trading.

But also see that the building steps are export, 

```
- name: Build trading app
        run: |
          yarn env-cmd -f ./apps/trading/.env.stagnet1 yarn nx export trading
          DIST_LOCATION=dist/apps/trading/exported
          mv $DIST_LOCATION dist-result
          tree dist-result
```

So I imagine I will need to do a traditional build as well as?

apps/trading/e2e/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+1. **Install Poetry**: Follow the instructions on the [official Poetry website](https://python-poetry.org/docs/#installing-with-the-official-installer).
+
+2. **Install Docker**: Follow the instructions on the [official Docker website](https://docs.docker.com/desktop/).
+
+3. **Set Up a Poetry Environment**:
+   Execute the command below to configure the Poetry environment.
+
+   ```bash
+   nom run trading:test:install
```
asiaznik, line 20:
```suggestion
   npm run trading:test:install
```

apps/trading/e2e/README.md
```diff
@@ -0,0 +1,78 @@
...
+   ```
+
+4. **Download Necessary Binaries**:
+   Use the following command within your Python environment. The `--force` flag ensures the binaries are overwritten, and the `--version` specifies the desired version.
+   ```bash
+   python -m vega_sim.tools.load_binaries --force --version $VEGA_VERSION
```
asiaznik, line 26: This doesn't explain how to get the `vega_sim.tools.load_binaries` module.

asiaznik: It looks like it lacks of `requirements.txt`, doesn't it?

apps/trading/e2e/README.md
```diff
@@ -0,0 +1,78 @@
...
+   ```
+
+## You now have two options:
+
+### Pull the Docker Image of the Trading App
+
+Pull the desired image:
+`bash docker pull vegaprotocol/trading:develop `
```
asiaznik, line 34:
```suggestion
`docker pull vegaprotocol/trading:develop `
```

apps/trading/e2e/README.md
```diff
@@ -0,0 +1,78 @@
...
+or
+`bash docker pull vegaprotocol/trading:main `
```
asiaznik, line 36:
```suggestion
`docker pull vegaprotocol/trading:main `
```

apps/trading/e2e/README.md
asiaznik: or `yarn` instead of `npm` as we use yarn here in this repo.

apps/trading/e2e/README.md
bwallacee: We shouldn't need a requirements.txt. The poetry lock sort of does the same.

I think the npm command for doing poetry install doesn't seem to always play nice, which is why this python command may not work.

I could add the actual commands required to do it all, rather than using the npm commands

mattrussell36 (changes requested):

MadalinaRaicu (approved):

package.json
```diff
@@ -13,7 +13,9 @@
     "lint:all": "nx run-many --all --target=lint",
     "e2e:all": "nx run-many --all --target=e2e",
     "vegacapsule": "vegacapsule network bootstrap --config-path=../frontend-monorepo/vegacapsule/config.hcl",
-    "release": "git checkout develop ; git pull ; node scripts/make-release.js"
+    "release": "git checkout develop ; git pull ; node scripts/make-release.js",
+    "trading:test": "cd apps/trading/e2e && poetry run pytest -k",
```
MadalinaRaicu, line 17: Nice! 
Do we run this after:  

```bash
yarn nx build trading ./docker/prepare-dist.sh docker build -f docker/node-outside-docker.Dockerfile --build-arg APP=trading --build-arg ENV_NAME=stagnet1 -t vegaprotocol/trading:latest .
```

bwallacee: Yeah you will need to run that after.
You have two options:
`yarn trading:test -- "test_name" -s `
run this from anywhere
or
`poetry run pytest -k "test_name" -s`
run this from within apps/trading/e2e

apps/trading/e2e/README.md
```diff
@@ -0,0 +1,118 @@
+# Trading End-To-End Tests
```
dalebennett1992, line 1: This comment relates to the folder hierachy and naming. We have trading-e2e and trading/e2e, we could perhaps distinguish a bit more between the two as it is a little confusing.

apps/trading/e2e/README.md
```diff
@@ -0,0 +1,118 @@
...
+## Getting Started
+
+1. **Install Poetry**: Follow the instructions on the [official Poetry website](https://python-poetry.org/docs/#installing-with-the-official-installer).
+2. **Install Docker**: Follow the instructions on the [official Docker website](https://docs.docker.com/desktop/).
+3. **Install Python**: Follow the instructions on the [official Python website](https://www.python.org/)
+   **ensure you install a version between 3.9 and 3.11.**
+4. **Start up a Poetry environment**: Execute the commands below to configure the Poetry environment.
+
+```bash
+poetry shell
```
dalebennett1992, line 20: hello, idiot here ;) I hear that you suggested it is obvious to be in the correct folder to run this command @mattrussell36 , however I just made this mistake. I think you lose nothing from being specific and pointing out you need to be in this folder. Definitely the kinda thing that made me say 'DUH' when I realised. But still

apps/trading/e2e/README.md
```diff
@@ -0,0 +1,118 @@
...
+## Running Tests 🧪
+
+Before running make sure the docker daemon is runnign so that the app can be served.
+
+To run a specific test, use the `-k` option followed by the name of the test.
+
+Run all tests:
+
+```bash
+poetry run pytest
```
dalebennett1992, line 89: can we add the instructions to set the correct interpreter given there is a specific poetry one. My VSCode sets to my default installation.

apps/trading/e2e/README.md
```diff
@@ -0,0 +1,118 @@
...
+8. **Pull the desired Docker image**
+
+```bash
+docker pull vegaprotocol/trading:develop
+```
+
+9. **Run tests**: Poetry/Python will serve the app from docker
+
+```bash
+poetry run pytest
```
dalebennett1992, line 51: to run this command successfully on wsl ubuntu I had to explicitly pull vegaprotocol/trading:latest, otherwise it errored when attempting to create the container.

dalebennett1992 (approved):

apps/trading/e2e/README.md
```diff
@@ -0,0 +1,118 @@
...
+   Use the following command within your Python environment. The `--force` flag ensures the binaries are overwritten, and the `--version` specifies the desired version. e.g. `v0.73.1`
+
+```bash
+python -m vega_sim.tools.load_binaries --force --version $VEGA_VERSION
+```
+
+8. **Pull the desired Docker image**
+
+```bash
+docker pull vegaprotocol/trading:develop
```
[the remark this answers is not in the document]
bwallacee: I agree, but there are issues with running tests in that way in parallel. A future enhancement will be to run the tests against served html.

PR #5206, opened 2023-11-06
feat(governance,accounts): vesting and vested accound in wallet card, fix transfer form to account

### LINKED ISSUE #5167: Add rewards vesting accounts to the Governance app
## The Chore

The new vesting mechanics introduce some issues for governance. 

Mostly we are trying to steer ppl towards Console for everything rewards related but some users may only ever stake and not use console and therefore need some options in governance app. 

## Tasks

- [x] Ensure the connected wallet view shows the breakdown of 
  - [x] VESTING account 
  - [x] VESTED accounts
- [x] Add a button by vested account to send people to the Console trasnfer page where they can redeem them

### LINKED ISSUE #5201: New transfer page with vested account stuff showing incorrect balances
## The Chore

image

See key: b62eafc46e6fadc6e45ae51de8bda199f0702131e5b23665f3411223c598acc2
On fairground 

For now can we go with option 1 ..... fix pulling the balance in properly .... that way we can do this without Matt and then we can talk revised UX on a less urgent timeframe.   

## Tasks

- [x ] Decide on option 
- [ ] FIx it

asiaznik:
# Related issues 🔗

Closes #5167 
Closes #5201 

# Description ℹ️

* adds "Vesting" and "Vested" accounts to the wallet card
* sums the balances from general and vested accounts in transfer form
* adds "Redeem" button that navigates to the Console's transfer form

# Demo 📺

image

# Technical 👨‍🔧

N/A

### DIFF
--- a/libs/accounts/src/lib/transfer-form.tsx
+++ b/libs/accounts/src/lib/transfer-form.tsx
@@ -72,8 +72,26 @@ export const TransferForm = ({
 
   const assets = sortBy(
     accounts
-      .filter((a) => a.type === AccountType.ACCOUNT_TYPE_GENERAL)
+      .filter(
+        (a) =>
+          a.type === AccountType.ACCOUNT_TYPE_GENERAL ||
+          a.type === AccountType.ACCOUNT_TYPE_VESTED_REWARDS
+      )
+      .reduce((merged, account) => {
+        const existing = merged.findIndex(
+          (m) => m.asset.id === account.asset.id
+        );
+        if (existing > -1) {
+          const balance = new BigNumber(merged[existing].balance)
+            .plus(new BigNumber(account.balance))
+            .toString();
+          merged[existing] = { ...merged[existing], balance };
+          return merged;
+        }
+        return [...merged, account];
+      }, [] as typeof accounts)
       .map((account) => ({
+        key: account.asset.id,
         ...account.asset,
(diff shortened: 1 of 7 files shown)


MadalinaRaicu (approved):

libs/accounts/src/lib/transfer-form.tsx
```diff
@@ -72,8 +72,26 @@ export const TransferForm = ({
 
   const assets = sortBy(
     accounts
-      .filter((a) => a.type === AccountType.ACCOUNT_TYPE_GENERAL)
+      .filter(
+        (a) =>
+          a.type === AccountType.ACCOUNT_TYPE_GENERAL ||
+          a.type === AccountType.ACCOUNT_TYPE_VESTED_REWARDS
+      )
+      .reduce((merged, account) => {
```
mattrussell36, line 80: This doesn't fix the issue. I still get the same general balance. Taking a look at this now...

mattrussell36 (changes requested):

libs/accounts/src/lib/transfer-form.tsx
mattrussell36: Just to clarify, this does fix _a_ issue where the balance in the asset dropdown was only the general account. But not the one where the balance in the to account dropdown is incorrect

mattrussell36 (approved):

PR #5211, opened 2023-11-07
feat(trading): close position stopped and IOC behaviour update

### LINKED ISSUE #5110: When using "Close position" function I dont get a green fill notification
## Steps to Reproduce

1. Go to console
2. Open a position
3. Close it using the close position button
4. You will see an outlandishly large TX size in your wallet
5. Approve the transaction in your wallet
6. You will not get a green fill notification just an orange stopped one
7. You will see an outlandishly large size in your closed orders as per this [ticket](https://github.com/vegaprotocol/feedback/discussions/191) from a user

## Expected behavior

You should get the normal green fill toast to indicate your order has been filled and the positon is closed. 

Instead I get this "ORDER STOPPED" toast even though the order did go through and my positoin is now closed. 

We think this is due to a change @macqbat made to use a massive reduce only order for this instead of using an order matching the size of the position.   We think this was done to address some edge case where if an open order traded just before you closed the position you might not completely close out your full position. 

In my view that is an acceptable edge case ... if you are using close position and you have open orders, you should check after you've pressed the button if there's anything left and close again if something has carried over.  This minor edge case is preferable to _everyone_ seeing these really high values and / or trying to now build handling for them.  We should roll this change back.

## Agreed approch (Now challenged by Barney)

- Roll back change to use a massive reduce only order
- Go back to original implementation where we assess the size of the position and close it out
- IMO we should provide some info on the button ... the current tooltip says "Close position" ... I think we should extend this to "Close position and cancel all open orders for this market"

## New agreed approach (Barney)

- Toast should detect reduce only + size of 9223372036854775807 (might already do this)
- Toast should show "close position" - it seems to actually do this already
- Toast for an IOC should go green when it is stopped, because stopping an IOC once all available volume has filled is the correct behaviour (it is immediate or cancel), this behaviour should apply to all IOC, not just those created due to "Close position" 
- Toast for IOC should show on it how much has filled before it was stopped, just the same as a completed IOC, ie. "+0.10 @ ~ USDT"
- The order table should also detect reduce only + size of 9223372036854775807 and instead of showing the size should show "MAX"

Image

## Screenshots

image

MadalinaRaicu:
# Related issues 🔗

Closes #5110

# Description ℹ️ & Demo 📺

- [x] Toast should detect reduce only + size of 9223372036854775807 (might already do this)
- [x] Toast should show "close position" - it seems to actually do this already
- [x] Toast for an IOC should go green when it is stopped, because stopping an IOC once all available volume has filled is the correct behaviour (it is immediate or cancel), this behaviour should apply to all IOC, not just those created due to "Close position"
- [x] Toast for IOC should show on it how much has filled before it was stopped, just the same as a completed IOC, ie. "+0.10 @ ~ USDT"

Screenshot 2023-11-08 at 10 17 39

Screenshot 2023-11-07 at 17 01 23

- [x] The order table should also detect reduce only + size of 9223372036854775807 and instead of showing the size should show "MAX"

Screenshot 2023-11-07 at 16 56 37

# Technical 👨‍🔧

### DIFF
--- a/libs/deal-ticket/src/components/deal-ticket/deal-ticket.tsx
+++ b/libs/deal-ticket/src/components/deal-ticket/deal-ticket.tsx
@@ -685,7 +685,7 @@ export const DealTicket = ({
         subLabel={`${formatValue(
           normalizedOrder.size,
           market.positionDecimalPlaces
-        )} ${baseQuote} @ ${
+        )} ${baseQuote || ''} @ ${
           type === Schema.OrderType.TYPE_MARKET
             ? 'market'
             : `${formatValue(
--- a/libs/orders/src/lib/components/order-list-manager/order-list-manager.tsx
+++ b/libs/orders/src/lib/components/order-list-manager/order-list-manager.tsx
@@ -148,6 +148,7 @@ export const OrderListManager = ({
               expiresAt: editOrder.expiresAt,
               side: editOrder.side,
               marketId: editOrder.market.id,
+              remaining: editOrder.remaining,
             };
             create({ orderAmendment }, originalOrder);
             setEditOrder(null);
--- a/libs/orders/src/lib/components/order-list/order-list.tsx
+++ b/libs/orders/src/lib/components/order-list/order-list.tsx
@@ -1,4 +1,5 @@
 import {
+  MAXGOINT64,
   addDecimalsFormatNumber,
   getDateTimeFormat,
   isNumeric,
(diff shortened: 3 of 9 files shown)


mattrussell36 (approved):

jeremyletang (approved):

guoguojin (approved):

libs/deal-ticket/src/components/deal-ticket/deal-ticket.tsx
```diff
@@ -685,7 +685,7 @@ export const DealTicket = ({
         subLabel={`${formatValue(
           normalizedOrder.size,
           market.positionDecimalPlaces
-        )} ${baseQuote} @ ${
+        )} ${baseQuote || ''} @ ${
```
MadalinaRaicu, line 688: leaving this comment to prevent merging before QA review is done

MadalinaRaicu: ah, and this line fixes this issue where `undefined` was displayed when no base quote tag was extracted
Screenshot 2023-11-07 at 15 04 32

bwallacee (approved):

PR #5212, opened 2023-11-07
feat(accounts): transfer form improvements to handle transfers from vesting accounts

### LINKED ISSUE #5202: Transfer UX improvements and alignment to Core patch
## The Chore

There are a few remaining issues here:

### Min transfer validation

We dont appear to validate for min transfer size.  You can send the TX and you get a failur from Core (without a very good error code, but I think that is the wallet from what Dex has said).

We should validate size against `transfer.minTransferQuantumMultiple` from network params. 

- [x] At minimum display some warning text saying this is below minimum transfer size
- [x] Ideally block the "confirm transfer" button

### Add a "Max" button 

- [x] Add button for max next to the amount form (this appears on the standalone form but not the sidebar one!)

### Adjust fee UX when transferring from VESTED to GENERAL for same key

Need to adjust for recent core changes:

If a user is transferring from their VESTED account to the GENERAL account of the SAME key:

  - [x] The transfer fee estimate should be zero
  - [x] The button to "include transfer fee" should be disabled
  - [x] The min size validation should not apply as long as the user has selected to transfer the max amount
  - [x] If the user is trying to transfer less than the min amount and has not selected max we should warn them: "If transferring less than the min amount of X you must transfer the full amount"

### General re-ordering

I would probably re-order the options here so that the Vega Key is showing after selecting your asset and From account since to me the logical flow is 

- I'm connected to a key already 
- I want to send this asset 
- From this account 
- To THIS key 
- In THAT account

In fact ... I'm not even sure you NEED the TO Account ... because you can only send to a GENERAL account. 

So my proposed restructure is:

1. Assume key is connected already
2. Choose asset
3. Choose From account (GENERAL or VESTED)
    4. If VESTED selected then default the "To Vega Key" to your own key and disable it
    5. If GENERAL selected then "To Vega Key" can be set 
6. Optional:
    7. Do we need to show the To account when it is always General?  We could but it would always be defaulted / disabled
    8. We coudl equally just show it as text under
    9. I'm also unsure we need to show the amount on the target key (if its external should you really 
    10. Feels like some text / tooltip here indicating target account is always general would also work.

Note on 3 that technically you CAN transfer from vested to another key's general account, at which point it is treated like a normal transfer, but I dont really see why we need to support that UX in the front end.  The main use case is redeeming them for free to your general account.  If you want to move them to another key you can do that immediately afterwards via a normal transfer. 

So it ends up looking something like this:

Image

mattrussell36:
# Related issues 🔗

Closes #5202 

# Description ℹ️

- Updates validation to use transfer quantum multiple network param, with special cases to allow for transferring the max amount if transferring from a vested account
- Disables fee inclusion if transferring from a vested account
- Shows no fees for transferring from a vested account
- Disables key selection if transferring from vested account. Can only transfer on the same key

### DIFF
--- a/libs/accounts/src/lib/transfer-container.tsx
+++ b/libs/accounts/src/lib/transfer-container.tsx
@@ -1,3 +1,4 @@
+import sortBy from 'lodash/sortBy';
 import * as Schema from '@vegaprotocol/types';
 import { truncateByChars } from '@vegaprotocol/utils';
 import { t } from '@vegaprotocol/i18n';
@@ -44,6 +45,7 @@ export const TransferContainer = ({ assetId }: { assetId?: string }) => {
   const accounts = data
     ? data.filter((account) => ALLOWED_ACCOUNTS.includes(account.type))
     : [];
+  const sortedAccounts = sortBy(accounts, (a) => a.asset.symbol.toLowerCase());
 
   return (
     <>
@@ -66,7 +68,7 @@ export const TransferContainer = ({ assetId }: { assetId?: string }) => {
         feeFactor={params.transfer_fee_factor}
         minQuantumMultiple={params.transfer_minTransferQuantumMultiple}
         submitTransfer={transfer}
-        accounts={accounts}
+        accounts={sortedAccounts}
       />
     </>
   );
--- a/libs/accounts/src/lib/transfer-container.tsx
+++ b/libs/accounts/src/lib/transfer-container.tsx
@@ -3,7 +3,7 @@ import { truncateByChars } from '@vegaprotocol/utils';
 import { t } from '@vegaprotocol/i18n';
 import {
   NetworkParams,
-  useNetworkParam,
+  useNetworkParams,
 } from '@vegaprotocol/network-parameters';
 import { useDataProvider } from '@vegaprotocol/data-provider';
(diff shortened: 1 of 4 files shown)


bwallacee:
- If i have selected vested account and then use Max for transfer, I can swap back to general account and still use the Max button. Is this expected or should the use max button disappear as soon as your from account changes to general:

image

- If i select enter manually for the "To vega key" and and then back to "select from wallet" I am unable to submit my transfer when ideally i wouldn't be able to click the enter manually button:
image


@MadalinaRaicu are you able to look at these in @mattrussell36 abecense

cc: @JonRay15

MadalinaRaicu:
> * If i have selected vested account and then use Max for transfer, I can swap back to general account and still use the Max button. Is this expected or should the use max button disappear as soon as your from account changes to general:

I believe this is expected, and a user should be able to use the Max button if they ever want to empty their general account (ex. they want to transfer all general account balance from one key to another)

> * If i select enter manually for the "To vega key" and and then back to "select from wallet" I am unable to submit my transfer when ideally i wouldn't be able to click the enter manually button:

I’m addressing this issue by removing "please select" placeholder option if "from account" is disabled.

MadalinaRaicu (approved):

Issue #4532, opened 2023-08-10
Update rewards view


JonRay15:
## Story

As a user 
I want to see details of various discounts / rewards I'm getting 
So that I feel like a valued customer and can tell what I should be doing to earn / maximise my rewards

## Acceptance Criteria

- [x] I can see the sketches in the figma
- [x] For now we will put this in the top level nav (until we can rework portfolio properly)
- [ ] All the numbers work
- [ ] There is a disconnected state that is still useful
  - [ ] The tiers are still shown just with no progress
  - [ ] The reward pool would be empty
  - [x] Vesting base rate would still show
  - [x] The multpliers would be blank 
  - [x] The reward history should be shown for the network rather than for the user 
- [x] There is a way to toggle the reward history to show "Earned by me" vs "Total distributed"
- [x] There is a way to paginate / filter the reward history
  - [x] MVP is simple pagination back through epochs
  - [x] Better would be to have the ability to show specific epoch or epochs 
  - [x] There is a default sort on epoch and then by the asset alphabetically
- [x] INclude a view of active rewards 

## Tasks

- [ ] UX - Determine which parts of console you need to surface this
- [ ] UX - Create ACs for each of these
- [ ] Team and stakeholder review
- [ ] Specs reviewed and created or adjusted
- [ ] Implementation
- [ ] Testing (unit and/or e2e)
- [ ] Code review
- [ ] QA review

## Sketch

Figma 

We are going for v2.4 in the Figma

![Image](https://github.com/vegaprotocol/frontend-monorepo/assets/89918817/4481758a-83cd-4206-8b85-0e2326f9338b)

Where there are various levels as indicated in the additional sketch in the Figma

![Image](https://github.com/vegaprotocol/frontend-monorepo/assets/89918817/1a851e19-6d9a-4161-a36b-093f988eec36)

Also using v2.1 of the reward cards. 

https://www.figma.com/file/ywcCZvXdvG28bg0AmCY9mF/Console-v3?type=design&node-id=3166-5839&mode=design&t=GXj0QjK0d268FwQ1-0

### API details

- The total rewards pot comes from the party accounts API with the account type `ACCOUNT_TYPE_VESTED_REWARDS` + `ACCOUNT_TYPE_VESTING_REWARDS`
- The amount available to withdraw is the balance in `ACCOUNT_TYPE_VESTED_REWARDS`
- The vesting base rate is a network param: `rewards.vesting.baseRate`
[...]

PR #5222, opened 2023-11-08
feat(trading): rewards page

mattrussell36:
# Related issues 🔗

Partially addresses #4532 

Does not contain
- Activity streak
- Tier cards

# Description ℹ️

Adds rewards tab to the portfolio page

# Demo 📺

Connected:

Screenshot 2023-11-14 at 11 45 21

Disconnected:

Screenshot 2023-11-14 at 11 45 55

# Technical 👨‍🔧


apps/trading/components/rewards-container/rewards-container.tsx
```diff
@@ -0,0 +1,324 @@
...
+        </tr>
+        {pubKey && (
+          <>
+            <tr>
+              <CardTableTH>{t('Vesting multiplier')}</CardTableTH>
+              <CardTableTD>0%</CardTableTD>
+            </tr>
+            <tr>
+              <CardTableTH>{t('Available to withdraw next epoch')}</CardTableTH>
+              <CardTableTD>0</CardTableTD>
```
asiaznik, line 309: question: What if there's something in a `vesting` account. Shouldn't we cover that here?

asiaznik: or is it impossible now to have anything in `vesting`?

bwallacee:
@mattrussell36 should i be able to go down to -1 epoch?
image


Also i wonder if it would be useful to have a button or something to put the "to" epoch to current?

mattrussell36:
> @mattrussell36 should i be able to go down to -1 epoch? image
> 
> Also i wonder if it would be useful to have a button or something to put the "to" epoch to current?

No, this should never bee negative, @bwallacee This is fixed.

apps/trading/components/rewards-container/rewards-container.tsx
mattrussell36: When I picked this one up it was agreed to skip over the vesting card. But it now looks like the required apis are in, so I've updated it

bwallacee:
LGTM, but do we need specs added for this?

bwallacee (approved):

apps/trading/components/rewards-container/rewards-container.tsx
mattrussell36: @bwallacee  @asiaznik  looks like I was using the wrong value here, so back to draft for this one

mattrussell36: Please take another look

mattrussell36:
> @mattrussell36 retested and it LGTM, but do we need specs added for this?

@bwallacee I've added some specs

MadalinaRaicu (approved):

PR #5240, opened 2023-11-09
feat(explorer): simplify party accounts page

### LINKED ISSUE #5209: Block Explorer: balances by asset doesn't include all accounts types
### Steps

1. Navigate to the block explorer page for a party/pubkey
2. On the "assets" panel click "show all"

### Expected

* All accounts including reward vesting, bond, etc. can be viewed, along with their balances for each asset (this could be by expanding on drilling down on a higher level value such as "used" for all margin accounts or "rewards" for all reward pots (vesting and vested)
* A total is shown covering all account types

### Actual

* Only "used" and "available" values are shown
* These can combine multiple accounts (e.g. margin on several markets
* Some account types aren't included at all
* The total shown is therefore also incorrect

edd:
- chore: sort
- fix(explorer): final sort order

# Description
Closes #5209. Removes the 'Account breakdown' view of a party's assets, using a simple table of all of the party's accounts instead.

# Examples

Two examples to demonstrate how the sorting works

## Accounts for a party, sorted by asset (with General first)
Screenshot 2023-11-09 at 18 14 57

## Accounts for a party, sorted by asset, then by market
Screenshot 2023-11-09 at 18 00 39

### DIFF
--- a/apps/explorer/src/app/components/asset-balance/asset-balance.tsx
+++ b/apps/explorer/src/app/components/asset-balance/asset-balance.tsx
@@ -1,5 +1,5 @@
 import { useAssetDataProvider } from '@vegaprotocol/assets';
-import { addDecimalsFormatNumber } from '@vegaprotocol/utils';
+import { addDecimalsFixedFormatNumber } from '@vegaprotocol/utils';
 import { AssetLink } from '../links';
 
 export type AssetBalanceProps = {
@@ -23,12 +23,12 @@ const AssetBalance = ({
 
   const label =
     !loading && asset && asset.decimals
-      ? addDecimalsFormatNumber(price, asset.decimals)
+      ? addDecimalsFixedFormatNumber(price, asset.decimals)
       : price;
 
   return (
     <div className="inline-block">
-      <span>{label}</span>{' '}
+      <span className="font-mono">{label}</span>{' '}
       {showAssetLink && asset?.id ? (
         <AssetLink showAssetSymbol={showAssetSymbol} assetId={assetId} />
       ) : null}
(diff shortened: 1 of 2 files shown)


MadalinaRaicu (approved):

bwallacee:
@edd apologies in advance as I haven't used block explorer before.

How do I "use" the sorting? I would have expected clicking on the column header would apply sorting.

Also are we missing a "total" row?

> A total is shown covering all account types

Maybe one for the future, but is there pagination on anything for this page, or does it become a long scroll? As i imagine you might end up with a lot fo rows here?

Another one for the future, but would a text search for market or asset etc be useful as well?

edd:
> @edd apologies in advance as I haven't used block explorer before.
> 
> How do I "use" the sorting? I would have expected clicking on the column header would apply sorting.
There is no sorting, it's just a simple table. The sorting mentioned in the ticket is just ensuring that general accounts are the first account type of each easset the party has.


> Also are we missing a "total" row?
No, this is an overview off all of the individual accounts a user has.

> Maybe one for the future, but is there pagination on anything for this page, or does it become a long scroll? As i imagine you might end up with a lot fo rows here?
It could become a long scroll as more assets get enabled and more markets use them. I agree this is a future thing.
 
> Another one for the future, but would a text search for market or asset etc be useful as well?
This might come across as lazy, but I would just suggest using the search in your browser for this. However if it turns out to be a feature that more people request, I can turn this in to an AG Grid table again which would give us sorting AND filtering. For now I'm going to keep it simple.

bwallacee (approved):

asiaznik (approved):

PR #5265, opened 2023-11-15
chore(trading): remove wait for graphql

bwallacee:
Removing wait for graphql as it didn't actually work.

### DIFF
--- a/apps/trading/e2e/actions/utils.py
+++ b/apps/trading/e2e/actions/utils.py
@@ -1,5 +1,3 @@
-import time
-
 from collections import namedtuple
 from playwright.sync_api import Page
 from vega_sim.null_service import VegaServiceNull
--- a/apps/trading/e2e/actions/utils.py
+++ b/apps/trading/e2e/actions/utils.py
@@ -1,5 +1,6 @@
-from collections import namedtuple
+import time
 
+from collections import namedtuple
 from playwright.sync_api import Page
 from vega_sim.null_service import VegaServiceNull
 from typing import Optional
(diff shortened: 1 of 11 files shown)


asiaznik (approved):

apps/trading/e2e/actions/utils.py
```diff
@@ -35,4 +35,29 @@ def next_epoch(vega: VegaServiceNull):
...
+
+def wait_for_graphql_response(page, query_name, timeout=5000):
+    response_data = {}
+
+    def handle_response(route, request):
+        if "graphql" in request.url:
+            response = request.response()
+            if response is not None:
+                json_response = response.json()
+                if json_response and "data" in json_response:
```
dalebennett1992, line 48: same for here really. This method is potentially a black hole for errors, unless we want it to be totally passive for some reason

apps/trading/e2e/actions/utils.py
```diff
@@ -35,4 +35,29 @@ def next_epoch(vega: VegaServiceNull):
                 "Epoch not started after forwarding the duration of two epochs."
             )
     vega.wait_fn(1)
-    vega.wait_for_total_catchup()
\ No newline at end of file
+    vega.wait_for_total_catchup()
```
dalebennett1992, line 38: not addressable but I would love to see us put a more sure-fire wait mechanism in here. From what you have said yourself the wait_fn seems a little hopeful in places. Would be good to take a look at some point soon

dalebennett1992 (approved):

PR #5281, opened 2023-11-16
chore(trading): remove some redundant capsule tests

bwallacee:
removed some capsule tests that i believe are redundant.
They are currently running against the wrong version of Vega, they fail often and we cover the UI aspects in other tests.


asiaznik (approved):

MadalinaRaicu (approved):

dalebennett1992 (approved):
approving on the basis of everything still passing. You are a better judge of this removal than I am

PR #5286, opened 2023-11-16
chore(trading): migrate LP cypress tests

bwallacee:
Move cypress tests for LP to market-sim and jest


libs/liquidity/src/lib/liquidity-table.spec.tsx
```diff
@@ -65,7 +86,28 @@ describe('LiquidityTable', () => {
...
   });
+
+  it('should be able to sort', async () => {
+    await act(async () => {
+      render(
+        <LiquidityTable rowData={multiRowData} stakeToCcyVolume={'0.3'} />
+      );
+    });
+    const headers = await screen.findAllByRole('columnheader');
+    await userEvent.click(headers[7]);
```
dalebennett1992, line 101: I would like to see if there is a better way to locate this stuff than by index. And given we are in monorepo now, if it is hard to locate we should be encouraging dev to add data-testids that make this stuff easier, or adding them ourselves, this way of locating is pretty brittle and will fail the second we get a new header at any earlier index.

libs/liquidity/src/lib/liquidity-table.spec.tsx
```diff
@@ -65,7 +86,28 @@ describe('LiquidityTable', () => {
...
+
+    const rows = screen.getAllByRole('row');
+    // 5002-LIQP-003
+    // TODO update test to correctly sort.
+    expect(rows[2]).toHaveTextContent(
```
dalebennett1992, line 106: index comment applies to this too.. sorry I know this is a draft but as you linked it ;)

libs/liquidity/src/lib/liquidity-table.spec.tsx
mattrussell36: I've updated this to use col-id

libs/liquidity/src/lib/liquidity-table.spec.tsx
mattrussell36: I've updated this test to just check that AgGrid sort is enabled. Rather than checking individual rows. AgGrid will not re-arrange rows in the DOM, instead it will just arrange them visually with styles.

apps/trading/e2e/tests/liquidity_provision/test_liquidity_provision.py
```diff
@@ -46,7 +60,35 @@ def test_liquidity_provision_amendment(continuous_market, vega: VegaService, pag
...
+        market_id=continuous_market,
+        commitment_amount=0,
+        fee=0,
+        is_amendment=False,
+    )
+    vega.wait_fn(1)
+    vega.wait_for_total_catchup()
+    
+    page.reload()
+    expect()
```
MadalinaRaicu, line 94:
```suggestion
    expect()
```

do we need this empty expect()?

MadalinaRaicu (approved):

apps/trading/e2e/tests/liquidity_provision/test_liquidity_provision.py
```diff
@@ -46,7 +60,35 @@ def test_liquidity_provision_amendment(continuous_market, vega: VegaService, pag
...
+    )
+@pytest.mark.skip("Waiting for the ability to cancel LP")
+@pytest.mark.usefixtures("page", "auth", "risk_accepted")
+def test_liquidity_provision_inactive(continuous_market, vega: VegaService, page: Page):
+    # TODO Refactor asserting the grid
+    page.goto(f"/#/liquidity/{continuous_market}")
+    page.get_by_test_id("manage-vega-wallet").click()
+    #TODO Rename "mm" to "marketMaker" so that we don't have to specify where to click when switching wallets
+    # Currently the default click will click the middle of the element which will click the copy wallet key button
+    page.locator('[role="menuitemradio"] >> .mr-2.uppercase').nth(1).click(position={ "x": 0, "y": 0}, force=True)
```
dalebennett1992, line 77: Can we change this so we do not have co-ordinates? Before with the copy button there was a genuine element that did not include the copy button. This may not be true for this particular element so I will let you judge.

apps/trading/e2e/tests/liquidity_provision/test_liquidity_provision.py
```diff
@@ -46,7 +60,35 @@ def test_liquidity_provision_amendment(continuous_market, vega: VegaService, pag
...
     page.reload()
+    expect(page.get_by_test_id("supplied-stake")).to_have_text("Supplied stake1.00001 tDAI")
+    expect(page.get_by_test_id("liquidity-supplied")).to_have_text("Liquidity supplied 17.16%")
+    row = page.get_by_test_id("tab-myLP").locator(".ag-center-cols-container .ag-row").first
+    expect(row).to_contain_text(
+        "Active"
+    )
+@pytest.mark.skip("Waiting for the ability to cancel LP")
+@pytest.mark.usefixtures("page", "auth", "risk_accepted")
+def test_liquidity_provision_inactive(continuous_market, vega: VegaService, page: Page):
```
dalebennett1992, line 71: This is an example where if I read the code I literally do not know what is happening without going to find the non-specific css locators. For instance 'menuitemradio', totally generic and I would wrap this stuff personally. This is an area when I think about lowering the barrier to entry, if this stuff is abstracted it is easier for someone else to pick up

apps/trading/e2e/tests/liquidity_provision/test_liquidity_provision.py
bwallacee: Removed

apps/trading/e2e/tests/liquidity_provision/test_liquidity_provision.py
bwallacee: Created a helper

apps/trading/e2e/tests/liquidity_provision/test_liquidity_provision.py
bwallacee: Created a helper for swapping keys

dalebennett1992 (approved):

PR #5290, opened 2023-11-16
fix(proposals): perpetual proposed markets

MadalinaRaicu:
# Related issues 🔗

Closes N/A

# Description ℹ️

Proposed markets do not display the right product type. 

# Demo 📺

After: 
Screenshot 2023-11-16 at 14 03 50
Before: 
image

# Technical 👨‍🔧

We were still using `futureProduct` instead of `product` that should be deprecated.


libs/deal-ticket/src/utils/validate-expiration.ts
```diff
@@ -1,7 +1,8 @@
 import { t } from '@vegaprotocol/i18n';
 import type { Validate } from 'react-hook-form';
 
-export const validateExpiration: Validate<string | undefined> = (
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const validateExpiration: Validate<string | undefined, any> = (
```
MadalinaRaicu, line 5:
```suggestion
export const validateExpiration: Validate<string | undefined> = (
```

libs/deal-ticket/src/utils/validate-expiration.ts
```diff
@@ -1,6 +1,7 @@
 import { t } from '@vegaprotocol/i18n';
 import type { Validate } from 'react-hook-form';
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
MadalinaRaicu, line 4:
```suggestion
```

jeremyletang (approved):

edd (approved):

bwallacee (approved):

dalebennett1992:
Market displayed correctly when in a proposed state, the voting column also removed. Paired with @bwallacee , as he said in slack it will need some fixes to the automation when this goes into develop.

dalebennett1992 (approved):

edd:
Did some manual exploratory testing on Governance & Explorer and haven't seen any issues

PR #5320, opened 2023-11-21
feat(trading): i18n language switcher

### LINKED ISSUE #4719: i18n language switcher
## Story

As a user
I want to be able to switch language
So that I can use the website

### Main Change

- [ ] I can switch to a different language in the top nav
- [ ]  The selected language is persisted
- [ ] Detect language (window.navigator.langauge or something else) and use it as the default
- [ ] If users language is not supported fall back to English
- [ ] Selected locale should change the formatting of dates and numbers. getUserLocale will need to be updated to return the selected language code

### Late addendum

- [ ] I can also switch the light / dark mode in the top nav (for consistency with website)
- [ ] No need to add a chevron, just two buttons

## Tasks

- [ ] UX (if needed)
- [ ] Design (if needed)
- [ ] Implementation
- [ ] Code review
- [ ] QA review

## Sketch

Assumption is we will go with a pretty simple UX similar to the approach on the website but we'll use the sidebar

So in the top nav we can have a simple icon like:

Image

This should also cover light / dark mode switcher while we're at it.

bglownia:
# Related issues 🔗

Closes #4719

# Description ℹ️

Add language switcher to navbar


apps/trading/lib/i18n/index.ts
```diff
@@ -6,8 +6,10 @@ import type { HttpBackendOptions, RequestCallback } from 'i18next-http-backend';
 import LanguageDetector from 'i18next-browser-languagedetector';
 import { initReactI18next } from 'react-i18next';
 
+export const supportedLngs = ['en', 'zh'];
```
bglownia, line 9: supportedLngs and useLocize should be rolled back before merge!

mattrussell36 (approved):

mattrussell36:
@bglownia I've pushed some changes to set up the icons properly in ui-toolkit and also to make sure the size and spacing looks correct in the navbar

PR #5327, opened 2023-11-22
chore(ci): create traces dir

bwallacee:


MadalinaRaicu (approved):

apps/trading/e2e/conftest.py
```diff
@@ -136,8 +136,8 @@ def init_page(vega: VegaServiceNull, browser: Browser, request: pytest.FixtureRe
             page.add_init_script(script=window_env)
             yield page
         finally:
-            if not os.path.exists("traces"):
-                os.makedirs("traces")
+            if not os.path.exists("apps/trading/e2e/traces"):
```
dalebennett1992, line 139:
```
try:
    if not os.path.exists(directory_path):
        os.makedirs(directory_path)
except OSError as e:
    print(f"Failed to create directory '{directory_path}': {e}")
```

personal preference here. I normally like to add some user friendly context in the event of an exception for things like this. I am not sure on the specifics of the python exception thrown or how clear it is, but generally I do not think wrapping things like this is a bad thing

dalebennett1992 (approved):

PR #5331, opened 2023-11-22
feat(trading): upgrade i18n, fix plurals

bglownia:


MadalinaRaicu (approved):

bwallacee (approved):

mattrussell36 (approved):

mattrussell36:
@bglownia Fixed rebased to fix the conflict here, but now getting test failures (possible due to upgrade i18n-next??)

In some tests some translation keys don't appear to work as I'd expect. For example in the positions table tests:

We have this translation
 `{ "Realised PNL: {{value}}": "Realised PNL: {{value}}" }`
 
 and we use it like:
`t("Realised PNL: {{value}}", { value: '123' })`

But in jest only the value '1.23' gets rendered.

Any ideas? The i18n instance is being set up as I'd expect in setup-tests.

PR #5348, opened 2023-11-27
feat(trading): trading view

### LINKED ISSUE #5335: Trading view datafeed
Build the datafeed object required for trading view to work with our APIs

### LINKED ISSUE #5336: Trading view import
Dynamically import trading bundle if required checks are successful

mattrussell36:
# Related issues 🔗

Closes #5335 
Closes #5336 

# Description ℹ️

- Loads trading view from external repo
- Adds a trading view compatible datafeed

# Demo 📺

Screenshot 2023-11-26 at 21 09 03

# Technical

- Added basic useScript hook for injecting scripts onto the page
- Added ChartContainer component to switch between Pennant and TradingView charts, also co-located the useChartSettings hook here
- Add trading-view lib to contain set up code for the chart


bwallacee:
@mattrussell36 Just a thought, with you wanted to have a feature toggle around this, can we also wrap this in a try catch or something similar so that if trading view breaks for whatever reason, we log it out to sentry and it won't have the ability to kill console.

mattrussell36:
> @mattrussell36 Just a thought, with you wanted to have a feature toggle around this, can we also wrap this in a try catch or something similar so that if trading view breaks for whatever reason, we log it out to sentry and it won't have the ability to kill console.

@bwallacee  I've added a ticket to address this for all panes in the trade grid
https://github.com/vegaprotocol/frontend-monorepo/issues/5410

MadalinaRaicu (approved):

mattrussell36:
@bwallacee 

Pushed some updates:
- Updated trading view link to open in a new tab (we need an additional attribution outside of the chart as part of the agreement with TV)
- Indicators should now be persisted (there is sometimes a short delay on the autosave function used by TV)

Also:
- No depth chart
- Errors in the console on page load are unrelated
- Lets leave saving and loading templates for now
- The bug with the text and when drawing on the chart is due to there being no candles on the chart. Looking for a way to mitigate this edge case

bwallacee (approved):

bwallacee:
@mattrussell36 the end column seems to take up a lot  more width than it needs: 
image

PR #5349, opened 2023-11-27
test(trading): referral E2E

dalebennett1992:
# Description ℹ️

Referrals E2E tests. Not quite ready for review, some of the numbers do not quite makes sense so need to verify with art. Feel free to check over the code in the mean time


apps/trading/e2e/fixtures/market.py
```diff
@@ -129,12 +137,19 @@ def setup_opening_auction_market(vega: VegaService, market_id: str = None, **kwa
...
+    all_markets = vega.all_markets() 
+    market_ids = [market.id for market in all_markets]
+    return market_id in market_ids
 
-def setup_continuous_market(vega: VegaService, market_id: str = None, **kwargs):
-    if market_id is None or market_id not in vega.all_markets():
-        market_id = setup_opening_auction_market(vega, **kwargs)
 
-    submit_order(vega, "Key 1", market_id, "SIDE_BUY", 1, 110)
+def setup_continuous_market(vega: VegaService, market_id: str = None, buy_orders = default_buy_orders, sell_orders = default_sell_orders, add_liquidity = True, **kwargs): #Add sell orders and buy orders to put on the book
```
dalebennett1992, line 148: not a fan of all of these params but in a separate PR I would like to refactor this so that we can pass market objects around to these methods to make this nicer to configure.

apps/trading/e2e/actions/utils.py
```diff
@@ -46,3 +51,22 @@ def change_keys(page: Page, vega:VegaServiceNull, key_name):
     page.get_by_test_id("key-" + vega.wallet.public_key(key_name)).click()
     page.click(f'data-testid=key-{vega.wallet.public_key(key_name)} >> .inline-flex')
     page.reload()
+
+def element_contains_text(locator: Locator, expected_pattern, timeout=5):
```
mattrussell36, line 55: @dalebennett1992 why exactly do we need this function? Does playwright not have some method for waiting for elements to appear?

mattrussell36: Quick search in the docs and found a `wait_for_selector` method can we use that here?

apps/trading/e2e/fixtures/market.py
mattrussell36: Sounds sensible, just converting to a dictionary would be a good start. Fire up an additional PR if you have the capacity

apps/trading/e2e/fixtures/market.py
```diff
@@ -8,12 +8,16 @@
...
+default_sell_orders = [[1, 110], [1, 105]]
+default_buy_orders = [[1, 90], [1, 95]]
+
 def setup_simple_market(
     vega: VegaService,
     approve_proposal=True,
     custom_market_name=market_name,
     custom_asset_name="tDAI",
     custom_asset_symbol="tDAI",
+    custom_quantum=1
```
bwallacee, line 20: Not one for this PR but I think we should default to a more sensible quantum at some point. But it will likely break a lot of tests.

apps/trading/e2e/tests/referrals/test_referrals.py
```diff
@@ -0,0 +1,131 @@
...
+    current_tier_element = page.get_by_test_id('current-tier-value')
+
+    volume_matched = element_contains_text(combined_volume_element, expected_volume)
+    tier_matched = element_contains_text(current_tier_element, expected_tier)
+
+    return volume_matched, tier_matched
+
+@pytest.mark.usefixtures("page", "auth", "risk_accepted")
+def test_referral_scenario(continuous_market, vega: VegaService, page: Page):
+    page.goto(f"/#/markets/{continuous_market}")
```
bwallacee, line 43: I would move the page goto, after doing everything vega side, just so it's clear when UI testing has started.

apps/trading/e2e/tests/referrals/test_referrals.py
```diff
@@ -0,0 +1,131 @@
...
+                "minimum_running_notional_taker_volume": 300,
+                "minimum_epochs": 3,
+                "referral_reward_factor": 0.03,
+                "referral_discount_factor": 0.03,
+            },
+        ],
+        staking_tiers=[
+            {"minimum_staked_tokens": 100, "referral_reward_multiplier": 1},
+            {"minimum_staked_tokens": 1000, "referral_reward_multiplier": 2},
+            {"minimum_staked_tokens": 10000, "referral_reward_multiplier": 2},
```
bwallacee, line 73:
```suggestion
            {"minimum_staked_tokens": 10000, "referral_reward_multiplier": 3},
```

apps/trading/e2e/tests/referrals/test_referrals.py
```diff
@@ -0,0 +1,131 @@
...
+        ],
+        window_length=1,
+    )
+    next_epoch(vega=vega)
+    vega.wait_for_total_catchup()
```
bwallacee, line 78:
```suggestion
```
We can delete this as next_epoch includes the wait

apps/trading/e2e/tests/referrals/test_referrals.py
```diff
@@ -0,0 +1,131 @@
...
+    vega.create_referral_set(key_name=PARTY_A.name)
+    vega.wait_fn(10)
+    vega.wait_for_total_catchup()
+    print(vega.list_referral_sets())
```
bwallacee, line 82: Not sure if we need these prints or if they are the right ones.

apps/trading/e2e/tests/referrals/test_referrals.py
```diff
@@ -0,0 +1,131 @@
...
+    if "referrals" in page.url:
+         page.reload()
+    else:
+         page.goto("/#/referrals/")
+
+    combined_volume_element = page.get_by_test_id('combined-volume-value')
+    current_tier_element = page.get_by_test_id('current-tier-value')
+
+    volume_matched = element_contains_text(combined_volume_element, expected_volume)
+    tier_matched = element_contains_text(current_tier_element, expected_tier)
```
bwallacee, line 37: I think we can just use expect here for contains text, it will include a wait and means we don't need to use the assert.

apps/trading/e2e/tests/referrals/test_referrals.py
```diff
@@ -0,0 +1,131 @@
...
+from playwright.sync_api import Page, expect
+from vega_sim.service import VegaService
+from conftest import init_vega
+from fixtures.market import setup_continuous_market, setup_simple_market
+from actions.utils import WalletConfig, change_keys, create_and_faucet_wallet, element_contains_text, next_epoch
+from actions.vega import submit_order, submit_liquidity
+from wallet_config import MM_WALLET, MM_WALLET2, TERMINATE_WALLET
+
+PARTY_A = WalletConfig("party_a", "party_a")
+PARTY_B = WalletConfig("party_b", "party_b")
```
bwallacee, line 12: Is it worth moving this into the wallet_config file

apps/trading/e2e/tests/referrals/test_referrals.py
```diff
@@ -0,0 +1,131 @@
...
+    submit_liquidity(vega, MM_WALLET.name, continuous_market, 100, 100)
+
+    vega.wait_fn(1)
+    vega.wait_for_total_catchup()
+
+    change_keys(page, vega, PARTY_B.name)
+    submit_order(vega, PARTY_B.name, continuous_market, "SIDE_BUY", 1, 115)
+    vega.wait_for_total_catchup()
+    next_epoch(vega=vega)
+    volume_matched, tier_matched = check_volume_and_tier(page, r'^1\d{2}$', "3")
```
bwallacee, line 109: We shouldn't need regex because we know the exact volume?
also in the other checks you are creating a pattern but you haven't in this one.

bwallacee (commented):
I'm not sure if you're splitting it into other tests, but we are missing the referer  (Party A) checks here.
I think we should be checking the other tiles as well, discount, epochs in set, epochs to next there, combined volume.

Other things missing but maybe are future tests:

- Discounts dropped when our volume dropped for the epoch (Maybe a good one to add here as it's already mostly done we just need go forward another epoch i think?)

- Checking the discounts are actually applied. (probs a future test but it will pretty much be a copy and paste of this one for the setup)

apps/trading/e2e/fixtures/market.py
dalebennett1992: sure, I just set this to match the default in the lower level method for now

apps/trading/e2e/actions/utils.py
dalebennett1992: so this was because the updated figures are not immediately available. The element is available and easy to locate but I wanted to wait for it contain the updated figures. I could have waited for the element containing the given text, but the playwright would return the element handle. I wanted the bool for the sake of having something nice and testable instead of a timeout or something. I can change though you do not agree

apps/trading/e2e/fixtures/market.py
dalebennett1992: sweet.

apps/trading/e2e/tests/referrals/test_referrals.py
dalebennett1992: Couple reasons for this.
1. I do not totally trust doing this because there were magic values buried everywhere
2. I am pretty sure the default is that the values are pegged with the liquidity submitted

Long story short with those above points is that I would want to refactor this so I can be sure of my own values because right now it feels brittle to me. I would also hope that the unit tests check the calculations. So this _shouldn't_ be a gap. I just wanted to reflect moving through the tiers... let me know if you do not agree with that logic though

apps/trading/e2e/tests/referrals/test_referrals.py
dalebennett1992: will have a ganders

apps/trading/e2e/fixtures/market.py
dalebennett1992: I also think a lot of these default params should be stored in a central .py file somewhere so we have a single source of truth

apps/trading/e2e/tests/referrals/test_referrals.py
```diff
@@ -0,0 +1,151 @@
...
+def setup_market_and_referral_scheme(vega: VegaService, continuous_market: str, page: Page):
+    page.goto(f"/#/markets/{continuous_market}")
+
+    create_and_faucet_wallet(vega=vega, wallet=PARTY_A)
+    create_and_faucet_wallet(vega=vega, wallet=PARTY_B)
+    forward_time(vega)
+
+    benefit_tiers = []
+    staking_tiers = []
+    for i in range(1, 4):
```
bwallacee, line 84: I prefer keeping it as the "JSON" as that's how proposals would look when actually created and it makes it super obvious and easier to read.

apps/trading/e2e/tests/referrals/test_referrals.py
```diff
@@ -0,0 +1,151 @@
...
+    return {
+        '[data-testid=combined-volume-value]': expected_volume,
+        '[data-testid=current-tier-value]': expected_tier,
+        '[data-testid=discount-value]': expected_discount,
+        '[data-testid=epochs-in-set-value]': expected_epochs,
+        '[data-testid=epochs-to-next-tier-value]': expected_epochs_to_next_tier
+    }
+
+
+def check_tile_values(page: Page, expected_results: dict):
```
bwallacee, line 48: We need to confirm it's "ok" to refresh the page in order to see the data, It might be an issue that we have to refresh. As a general rule in console the most up to date data should always be displayed ( no refresh required).

Doesn't block this pr though.

apps/trading/e2e/actions/utils.py
```diff
@@ -46,3 +48,19 @@ def change_keys(page: Page, vega:VegaServiceNull, key_name):
...
     page.reload()
+
+def forward_time(vega: VegaService, forward_epoch: bool = False):
+    vega.wait_fn(1)
+    vega.wait_for_total_catchup()
+
+    if forward_epoch:
+        next_epoch(vega)
+
+def selector_contains_text(page: Page, selector, expected_text, timeout=5000):
```
bwallacee, line 59: Why aren't we just using the playwright expect?

apps/trading/e2e/tests/referrals/test_referrals.py
```diff
@@ -0,0 +1,151 @@
...
+        asset=tdai_id,
+        amount=10e6,
+    )
+
+    submit_liquidity(vega, MM_WALLET.name, continuous_market, 100, 100)
+    forward_time(vega)
+
+
+@pytest.mark.usefixtures("page", "auth", "risk_accepted")
+def test_referral_scenario(continuous_market, vega: VegaService, page: Page):
```
bwallacee, line 121: No blocking, but pulling this test out into separate tests could be good for the future.

bwallacee (commented):
Can you add some TODOs or create extra tickets for the tests that we don't have covered.

apps/trading/e2e/actions/utils.py
dalebennett1992: remember when we spoke about the initial values being incorrect? It fails because it finds the element, and the element initially has the incorrect text so fails the assertion. This ensures it gives the text chance to update.

apps/trading/e2e/tests/referrals/test_referrals.py
dalebennett1992: This makes it more scalable. You can make 5000 tiers if you want. You can see what the json would be if you ctrl click into the method

dalebennett1992: I genuinely think it is a bad practice to ctrl c + ctrl v json every time we want to configure referrals. If the schema changes you can just fix in one place this way. If you have one key change, and you have 5 tiers and you use the json in 10 tests, that is 50 places you have to update

apps/trading/e2e/actions/utils.py
bwallacee: ahh, fuck that's annoying. I remember.

apps/trading/e2e/tests/referrals/test_referrals.py
bwallacee: it's cool, lets go with it.

I'll update my tests as well to match.

apps/trading/e2e/actions/utils.py
```diff
@@ -1,5 +1,7 @@
 from collections import namedtuple
 from playwright.sync_api import Page
+from vega_sim.service import VegaService
```
bwallacee, line 3: I dont think you need to import this as VegaServiceNull is already imported?

apps/trading/e2e/tests/referrals/test_referrals.py
dalebennett1992: This is as expected. Confirmed with @asiaznik

bwallacee (approved):

PR #5353, opened 2023-11-27
feat(deal-ticket): enhance for fee discounts

### LINKED ISSUE #5244: Enhance deal ticket info for fee discounts
The way discounts are displayed in the deal ticket & tooltip have led to some confusion.

1. The -40% next to fees, at a glance, looks like 40% fees, but even -40% fees doesn't imply "saving 40% on fees".
2. Tooltip isn't clear that the "volume discount" is subtracted from the total, so it actually looks like another fee. This could be better expressed, perhaps with a minus. 
3. Rename "Volume discount":
* Either rename to Fee discount (volume tier X) 
* OR lump all 'fee discounts' together in one line as a collective % & value, and let people see the breakdown on the fees/discounts tab.
4. How the 40% is calculated has also caused confusion. Per @wwestgarth: "this fee bit on stagnet for me seems odd to say the volume discount fee is 40% when 0.0072/22.5 is more like 0.032%". But the maths are more like: 
> Total fee (discounted) is 0.0108
> Discount is 0.0072
> Fee before discounts = 0.0072+0.0108 = 0.018
> Discount = 0.018 * 0. 4 = 0.0072
Not sure how best to handle this but it's not an obvious one. 

image

## Acceptance criteria
- [x] Table rows are structured as outlined in [this comment](https://github.com/vegaprotocol/frontend-monorepo/issues/5244#issuecomment-1807695420)
- [x] Prefix with '-' for discounted percentages and amounts
- [x] Check that the maths is correct for discount calculations
- [x] Combine all discounts together into one 'Discount' row. This is either the volume discount or the referral discount or the combined discount value if both are applied

bglownia:
# Related issues 🔗

Closes #5244

# Demo 📺

With discounts

image

Without discounts

image


bwallacee:
@bglownia I'm not sure if the total fee in your screenshot is correct?
3.020065 - 19% is 2.44625265 ( and this would round to 2.446253? )
not 2.446254

bglownia:
> @bglownia I'm not sure if the total fee in your screenshot is correct? 3.020065 - 19% is 2.44625265 ( and this would round to 2.446253? ) not 2.446254

Calculations are correct. When counting discount values floor function is used. Because  of rounding we get slightly different value, and there is no way to correct it. We can try to calculate factor base on values but this will give factor with lot of decimals

fee
3020065

referral
floor(3020065*0.1)=302006

volume 
3020065-302006=2718059
floor(2718059*0.2)=543611

total discount
543611+302006=845618

discount factor calculated from value
845618/3020065=0.2799999337762598

mattrussell36:
@bwallacee I think in this case using the value directly from the formula is fine, we might get a little rounding but in this case its fine

MadalinaRaicu (approved):

mattrussell36 (approved):

bwallacee (approved):

PR #5358, opened 2023-11-27
chore(accounts): no assets avaiable in transfer form

### LINKED ISSUE #4997: Improve transfers UX when key has no assets
## The Chore

Currently we have seen the following when you have no assets:

- Dropdown cant be clicked 
- Dropdown can be clicked but the list is empty and just shows some lines 

## Tasks

- [x] Make it work nicer when you have no assets, should say "No assets availble" and grey out the selector for instance =

image

asiaznik:
# Related issues 🔗

Closes #4997

# Description ℹ️

* adds "No assets available" / "No accounts available" message in transfer form when there are no assets / accounts

# Demo 📺

image

# Technical 👨‍🔧

N/A


mattrussell36 (approved):
@asiaznik really I think the RichSelect component and the normal Select component should handle an empty state so you don't need to wrap it with a check for length. If you have capacity to change this then do, otherwise I think this is probably fine for now as is

MadalinaRaicu (approved):

asiaznik:
> @asiaznik really I think the RichSelect component and the normal Select component should handle an empty state so you don't need to wrap it with a check for length. If you have capacity to change this then do, otherwise I think this is probably fine for now as is

We plan to change the rich selector anyway, so I'd leave this as is for now.

dalebennett1992 (approved):
@asiaznik tells me ben looked and was all good but forgot to tick. @bwallacee if this is not the case then you and art can fight each other when you are back :D

PR #5359, opened 2023-11-27
chore(trading): don't redirect to inactive markets

### LINKED ISSUE #5193: Do not redirect to last market id if that market is not active
## The Chore

Its possible if you bookmark the home page and then don't use Vega for a while the last market you used will be used as the default market even if its settled or expired. This means you may be greeted with a market showing the 'market expired' banner or worse the 'market does not exist' state.

## Tasks

- [ ] When landing on the home route, check that the market is active before redirecting
- [ ] If market is active go to that market
- [ ] If market is inactive use top traded

https://vegaprotocol.slack.com/archives/C037STCKMGC/p1699036552276329

asiaznik:
# Related issues 🔗

Closes #5193 

# Description ℹ️

N/A

# Demo 📺

N/A

# Technical 👨‍🔧

N/A

### DIFF
--- a/apps/trading/lib/hooks/use-navigate-to-last-market.spec.ts
+++ b/apps/trading/lib/hooks/use-navigate-to-last-market.spec.ts
@@ -1,0 +1,60 @@
+import * as router from 'react-router';
+import { useNavigateToLastMarket } from './use-navigate-to-last-market';
+import { useGlobalStore } from '../../stores';
+import { renderHook } from '@testing-library/react';
+import { useTopTradedMarkets } from './use-top-traded-markets';
+import { Links } from '../links';
+
+const mockLastMarketId = 'LAST';
+
+jest.mock('../../stores', () => {
+  const original = jest.requireActual('../../stores');
+  return {
+    ...original,
+    useGlobalStore: jest.fn(),
+  };
+});
+
+jest.mock('./use-top-traded-markets', () => {
+  return {
+    useTopTradedMarkets: jest.fn(),
+  };
+});
+
+describe('useNavigateToLastMarket', () => {
+  const navigate = jest.fn();
+  beforeAll(() => {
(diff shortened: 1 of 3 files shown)


apps/trading/lib/hooks/use-navigate-to-last-market.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+jest.mock('./use-top-traded-markets', () => {
+  return {
+    useTopTradedMarkets: jest.fn(),
+  };
+});
+
+describe('useNavigateToLastMarket', () => {
+  const navigate = jest.fn();
+  beforeAll(() => {
+    jest.spyOn(router, 'useNavigate').mockImplementation(() => navigate);
```
mattrussell36, line 27: I think its cleaner/simpler to just wrap the rendering of the hook with a memory router

```
const setup = () => {
  const wrapper = ({ children }: { children: ReactNode }) => (
    <MemoryRouter>{children}</MemoryRouter>
  );
  return renderHook(() => useNavigateToLastMarket(), { wrapper });
}
```

apps/trading/lib/hooks/use-navigate-to-last-market.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+  });
+
+  it('navigates to the last market when it is active', () => {
+    (useGlobalStore as unknown as jest.Mock).mockReturnValue(mockLastMarketId);
```
mattrussell36, line 31: I've been enjoying just calling setState directly before rendering for some tests. Feels simpler as you can avoid setting up all the mocks above (also useGlobalStore, gets inherently tested too )

You do need to be careful about the zustand store state leaking between tests though

useGlobalStore.setState({ marketId: mockLastMarketId })

mattrussell36 (approved):
Couple suggestions but both non blocking

bwallacee (approved):

PR #5367, opened 2023-11-28
chore(trading): migrate cypress tests to python

bwallacee:
To Note: the route handler with the assert is a simpler way to get around having to use asyncio and other methods to capture the response but carry on with playwright functions.


apps/trading/e2e/tests/market_lifecycle/test_market_lifecycle.py
```diff
@@ -10,6 +11,8 @@
 
 @pytest.mark.usefixtures("vega", "page", "proposed_market", "risk_accepted")
 def test_market_lifecycle(proposed_market, vega: VegaService, page: Page):
```
dalebennett1992, line 13: The characteristics of the checks in this tests are something I would expect to see in a test of a particular module, not e2e. Am I missing something? How comes this has to be e2e?

Normally I would expect to mock a state for the given ui module, and then check the default values as we do here. Is this not something we do atm?

apps/trading/e2e/tests/wallet/test_wallet.py
```diff
@@ -0,0 +1,116 @@
...
+def vega(request):
+    with init_vega(request) as vega:
+        yield vega
+
+
+@pytest.fixture(scope="module")
+def continuous_market(vega):
+    return setup_continuous_market(vega)
+
+def handle_route_connection_lost(route: Route, request):
```
dalebennett1992, line 28: So if my assumption is correct, here we are just handling incoming API requests. I would personally put these into a specific handler file and would not bundle it in with the tests.

apps/trading/e2e/tests/wallet/test_wallet.py
```diff
@@ -0,0 +1,116 @@
...
+        if request.method == "POST" and re.match(r"http://localhost:\d+/api/v2/requests", request.url):
```
dalebennett1992, line 29: could the regex pattern be stored in a variable, it is referenced three times here

apps/trading/e2e/tests/wallet/test_wallet.py
```diff
@@ -0,0 +1,116 @@
...
+def test_wallet_connection_rejected(continuous_market, page: Page):
+    # 0002-WCON-002
+    # 0002-WCON-005
+    # 0002-WCON-007
+    # 0002-WCON-015
+    page.goto(f"/#/markets/{continuous_market}")
+    page.route("**/*", handle_route_connection_rejected)
+    page.get_by_test_id("connect-vega-wallet").click()
+    page.get_by_test_id("connector-jsonRpc").click()
+    expect(page.get_by_test_id("dialog-content").nth(1)).to_have_text("User errorthe user rejected the wallet connectionTry againAbout the Vega wallet  | Supported browsers ")
```
dalebennett1992, line 81: oooh this expected text is odd... I guess it passes though? 

dialog-content, what does this element look like sorry? Only reason I say this is because locating things by nth child is not particularly robust in my experience and I would normally advise against it as a rule

apps/trading/e2e/tests/wallet/test_wallet.py
```diff
@@ -0,0 +1,116 @@
...
+
+def assert_connection_approve(route: Route, request, page:Page):
+        if request.method == "POST" and re.match(r"http://localhost:\d+/api/v2/requests", request.url):
+            expect(page.get_by_test_id("toast-content")).to_have_text("Please go to your Vega wallet application and approve or reject the transaction.")
+        else:
+            route.continue_()
+
+@pytest.mark.usefixtures("page", "auth", "risk_accepted")
+def test_wallet_connection_error(continuous_market, page: Page):
+    page.goto(f"/#/markets/{continuous_market}")
```
dalebennett1992, line 65: Personal preference and I know it is a one liner so will leave with you, I literally see this line everywhere and I would just have this url declared in one place. Or enclose it in a method where you can pass in any market. If this url changes for any reason (although I am assuming unlikely) I would not want to change in 56 tests.

apps/trading/e2e/tests/market_lifecycle/test_market_lifecycle.py
bwallacee: To get these tests into jest would be quite difficult to mock as there are multiple different data sources. It would be possible but it's a lot easier to do it here

apps/trading/e2e/tests/wallet/test_wallet.py
bwallacee: I'm not sure im sold on pulling into a separate file, as currently these tests are the only place we do anything like this, and not sure we gain a lot from pulling into a separate file vs keeping here?

apps/trading/e2e/tests/wallet/test_wallet.py
bwallacee: yeah good idea

apps/trading/e2e/tests/wallet/test_wallet.py
bwallacee: image
Basically the test id is duplicated I can look fixing the test ids but this will break quite a few tests so should maybe sit in a separate pr

apps/trading/e2e/tests/market_lifecycle/test_market_lifecycle.py
dalebennett1992: sure

apps/trading/e2e/tests/wallet/test_wallet.py
dalebennett1992: Yeah cool. I normally do that so it is more visibly reusable if it is _likely_ to be useful to someone else or reused. However if those things aren't true then no issue with it being left

PR #5405, opened 2023-11-30
fix(trading): fills fees fixes for maker

### LINKED ISSUE #5401: Fees and discounts for makers are not calculating correctly
## The Chore

We seem to be inferring the maker / taker role based on the `aggressor` and whether the connected party is the buyer or seller. 

Then if we are taker things are fine but if we're maker we're flipping the sign on the maker fee to show that you are being paid.  In isolation thats OK... but then the total fee ends up being infra + liq - maker which is wrong. 

We also have a problem that we cant tell from the trades API if a market is in auction, when the fees should change.  We've asked Core for an enhancement for that ... but there's a short term workaround below.

These issues impact both the total fee / discount in the table AND the values in the tooltip.

image

### Fees

**In Continuous trading**

if Taker:
makerFee = makerFee
infraFee = infraFee
liqFee = liqFee
Total fee = makerFee + infraFee + liqFee

If Maker:
makerFee = - makerFee
infraFee = 0
liqFee = 0
Total fee = - makerFee

**If in Auction (which we can detect by checking both buyer AND seller makerFee = 0)**

makerFee = 0
infraFee = infraFee / 2
liqFee = liqFee / 2
Total fee = (infraFee + liqFee)/2

### Discounts (these apply to both volume and referal discounts)

**In Continuous trading**

if Taker:
makerFeeDiscount = makerFeeDiscount
infraFeeDiscount = infraFeeDiscount
liqFeeDiscount = liqFeeDiscount
Total Discount= makerFeeDiscount + infraFeeDiscount + liqFeeDiscount

If Maker:
makerFeeDiscount = makerFeeDiscount
infraFeeDiscount = 0
liqFeeDiscount = 0
Total discount = as above

**If in Auction (which we can detect by checking both buyer AND seller makerFee = 0)**

makerFeeDiscount = 0
infraFeeDiscount = infraFeeDiscount / 2
liqFeeDiscount = liqFeeDiscount / 2
Total fee Discount = (infraFeeDiscount + liqFeeDiscount)/2

## Tasks

- [ ]
- [ ]

MadalinaRaicu:
# Related issues 🔗

Closes #5401 

# Description ℹ️

Fees in the fills table were not calculated correctly for the maker.

# Demo 📺

Screenshot 2023-11-30 at 16 02 49
Screenshot 2023-11-30 at 16 02 44

# Technical 👨‍🔧

### DIFF
--- a/libs/fills/src/lib/fills-table.tsx
+++ b/libs/fills/src/lib/fills-table.tsx
@@ -348,7 +348,7 @@ const FeesBreakdownTooltip = ({
       className="z-20 max-w-sm px-4 py-2 text-sm text-black border rounded bg-vega-light-100 dark:bg-vega-dark-100 border-vega-light-200 dark:border-vega-dark-200 break-word dark:text-white"
     >
       <p className="mb-1 italic">
-        {t('If the market is %s', [
+        {t('If the market was %s', [
           Schema.MarketStateMapping[marketState].toLowerCase(),
         ])}
       </p>
--- a/libs/fills/src/lib/fills-table.tsx
+++ b/libs/fills/src/lib/fills-table.tsx
@@ -345,7 +345,7 @@ const FeesBreakdownTooltip = ({
   return (
     <div
       data-testid="fee-breakdown-tooltip"
-      className="z-20 max-w-sm px-4 py-2 text-sm text-black border rounded bg-vega-light-100 dark:bg-vega-dark-100 border-vega-light-200 dark:border-vega-dark-200 break-word dark:text-white"
+      className="z-20 max-w-sm px-4 py-2 text-xs text-black border rounded bg-vega-light-100 dark:bg-vega-dark-100 border-vega-light-200 dark:border-vega-dark-200 break-word dark:text-white"
     >
       <p className="mb-1 italic">
         {t('If the market was %s', [
@@ -406,8 +406,8 @@ const FeesDiscountBreakdownTooltipItem = ({
 }) =>
   value && value !== '0' ? (
     <>
-      <dt className="col-span-1">{label}</dt>
-      <dd className="text-right col-span-1">
+      <dt className="col-span-2">{label}</dt>
+      <dd className="text-right col-span-2">
         {addDecimalsFormatNumber(value, asset.decimals)} {asset.symbol}
       </dd>
     </>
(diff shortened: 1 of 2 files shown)


libs/fills/src/lib/fills-table.tsx
```diff
@@ -315,7 +314,16 @@ export const getRoleAndFees = ({
   } else {
     return { role: '-', fees: undefined };
   }
-  return { role, fees };
+
+  // We make the assumption that the market state is active if the maker fee is zero on both sides
```
mattrussell36, line 318: This is a very helpful comment, is it true that its when the market is active? I thought it was more due to trading mode?

libs/fills/src/lib/fills-table.tsx
```diff
@@ -479,20 +492,35 @@ export const getTotalFeesDiscounts = (fees: TradeFeeFieldsFragment) => {
 
 export const getFeesBreakdown = (
   role: Role,
-  feesObj: TradeFeeFieldsFragment
+  feesObj: TradeFeeFieldsFragment,
```
mattrussell36, line 495: It would be sensible to update the getFeesBreakdown tests to include a test checking for market state

mattrussell36 (approved):

libs/fills/src/lib/fills-table.tsx
MadalinaRaicu: The market is active only if the trading mode is "trading continuously"

libs/fills/src/lib/fills-table.tsx
MadalinaRaicu: Added for each case!

libs/fills/src/lib/fills-table.tsx
```diff
@@ -1,10 +1,8 @@
 import { useMemo } from 'react';
-import type {
-  AgGridReact,
-  AgGridReactProps,
-  AgReactUiProps,
-} from 'ag-grid-react';
-import type { ITooltipParams, ColDef } from 'ag-grid-community';
+import type { AgGridReact } from 'ag-grid-react';
```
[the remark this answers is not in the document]
MadalinaRaicu: This is against main and it uses node 16. 
But I can re-group them anyway! More files will have that issue but for now will only update fills-table.tsx

PR #5411, opened 2023-12-01
fix(trading): use discount stats only from previous epoch

### LINKED ISSUE #5400: Volume discount tier incorrect on fees page
## The Chore

Ben has observed that he is seeing a discount in the total ... but not seeing the indicator in the correct tier.  BOTH are wrong since there should be NO discount at all.

Needs to be fixed, and we need to work out if it will be impacting mainnet or not too and fix it there if needed. 

image

We think this is because of not checking ffor the epooch

```
        },
        {
          "node": {
            "atEpoch": 115612,
            "partyId": "a91faad4fdd52f49659579c877225d984c3e26cc3fae8d2d7c6a84647d74da80",
            "discountFactor": "0.4",
            "runningVolume": "14793050"
          }
```

The last volume discount was in atEpoch 115612 ... when current epoch is 115651... so no discount should be applying at all. 

The code should be looking for an entry in the volumeStats API which is for current epoch - 1 and applying that value.  If there is no value for the current epoch -1 epoch then no discount should apply.

Potentially the "current tier" part is an issue of the fact the lowest tier is STILL a discount.  So maybe we need to include a "tier 0" so that if you arent getting any discount at all we show the "your tier" lozenh there?  Either that or we need to not show the lozenge or your volume if you arent in at least tier 1.

## Tasks

- [x] Make sure tiers work
- [x] Make sure the deal ticket works

bglownia:
# Related issues 🔗

Closes #5400
# Description ℹ️

Use discount stats only from previous epoch. Identify tier base on discount value


guoguojin (approved):

asiaznik (approved):

dalebennett1992:
This looks good to me. Checked over on a call with Bart.

dalebennett1992 (approved):

PR #5425, opened 2023-12-04
feat(environment): users controlled feature flags

### LINKED ISSUE #5372: Allow users to locally flag features on / off to use bleeding edge features if desired
## The Chore

Could do this using the console in the browser, or could add it to settings.

Barney is keen on this so seems like adding it to settings would be a nicer way of doing it. 

Marked with some appropriate "these features are still in development and are used at your own risk" type warning.

## Tasks

- [ ]
- [ ]

bglownia:
# Related issues 🔗

Closes #5372
# Description ℹ️

# Demo 📺

# Technical 👨‍🔧


libs/environment/src/hooks/use-environment.ts
```diff
@@ -269,11 +144,16 @@ const testSubscription = (client: Client) => {
   });
 };
 
+export const userControllableFeatureFlags: (keyof FeatureFlags)[] = [
```
bglownia, line 147: reset to empty array before merge

apps/trading/components/settings/settings.tsx
```diff
@@ -98,6 +124,11 @@ export const Settings = () => {
           <dd className="break-words">{process.env.GIT_COMMIT}</dd>
         </dl>
       </SettingsGroup>
+      <SettingsGroup inline={false} label={t('Experimental features')}>
```
bglownia, line 127: remove before merge

apps/trading/.env
```diff
@@ -20,8 +20,8 @@ NX_WALLETCONNECT_PROJECT_ID=fe8091dc35738863e509fc4947525c72
 
 # Cosmic elevator flags
 NX_SUCCESSOR_MARKETS=true
-NX_STOP_ORDERS=true
+NX_STOP_ORDERS=false
```
bglownia, line 23: rollback before merge

bwallacee (commented):
Can we add a unit test for when feature flag is set to false as well?
Also should we log out to sentry what flags are getting enabled.
e.g. if we get an error in console and log to sentry we should also log what flags were enabled (to help debugging etc)

mattrussell36 (approved):
Looks good. One comment about tidying up the dupcliate mock, also I agree using the switch would be nicer.

bwallacee (approved):

PR #5431, opened 2023-12-04
chore(trading): update generic data provider

MadalinaRaicu:
# Related issues 🔗

Closes N/A

- [x] update generic data provider pagination object to contain `last` field
- [x] when using `last` we can get the most recent records from the API
Note: this should be used since there is a limit between 1k and 5k records that can be retrieved (i.e. candles query)

### DIFF
--- a/libs/trades/src/lib/trades-data-provider.ts
+++ b/libs/trades/src/lib/trades-data-provider.ts
@@ -98,7 +98,6 @@ export const tradesProvider = makeDataProvider<
   pagination: {
     getPageInfo,
     append,
-    // first: MAX_TRADES,
     last: MAX_TRADES,
   },
   fetchPolicy: 'no-cache',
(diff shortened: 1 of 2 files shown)


libs/trades/src/lib/trades-data-provider.ts
```diff
@@ -98,7 +98,8 @@ export const tradesProvider = makeDataProvider<
   pagination: {
     getPageInfo,
     append,
-    first: MAX_TRADES,
+    // first: MAX_TRADES,
```
mattrussell36, line 101: Delete  this

mattrussell36 (changes requested):

libs/trades/src/lib/trades-data-provider.ts
MadalinaRaicu, line 101:
```suggestion
```

mattrussell36 (approved):

PR #5434, opened 2023-12-04
fix(governance): disable penalty column in standby validator table

### LINKED ISSUE #5191: The validator penalties dont disntinguish between NULL and actually zero
## The Chore

The validator scoring relies quite heavily on the rewardScore part of the epochs API. 

But if you are a candidate (pending) validator this part is null as below.

image

This means the scores for Candidate validators are always misleading 
image

And it means in the first epoch after promotion the scores for Standby validators are misleading
image

## Tasks

- [ ] For Candidate validators
  - [ ] Set total penalties to "-" as rewardScore will always be null so there is no validatorScore to use
  - [ ] Set overstaking penalty to  "-" because it requires the rawValidatorScore from rewardScore so will always be null
  - [ ] Performance penalty should be ok because it is taken from rankingScore in current epoch and this is never null

- [ ] For Standby validators 
  - [ ] If rewardScore is null then apply treatment as above
  - [ ] Else calculate scores as usual

edd:
# Related issues 🔗

Closes #5191

# Description ℹ️

Fixing this in the specific way outlined in the ticket was hard, as so many of these values are calculated by munging
together bits of data from different APIs. So I undid all those changes and just hid the 'Penalties' column from the 
standby validators table.

# Demo 📺
Screenshot 2023-12-04 at 16 09 11

### DIFF
--- a/apps/governance/src/routes/staking/home/validator-tables/standby-pending-validators-table.tsx
+++ b/apps/governance/src/routes/staking/home/validator-tables/standby-pending-validators-table.tsx
@@ -16,7 +16,6 @@ import {
   stakedTotalPercentage,
   ValidatorFields,
   ValidatorRenderer,
-  TotalPenaltiesRenderer,
   TotalStakeRenderer,
   StakeShareRenderer,
   PendingStakeRenderer,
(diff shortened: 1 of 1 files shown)


edd:
Waiting for @JonRay15 to check if he's happy with this solution..

PR #5437, opened 2023-12-04
fix(trading): rewards page updates

MadalinaRaicu:
# Related issues 🔗

Closes N/A

# Description ℹ️

- Show infra fees in rewards table 
- Only shows USDT (also USDT on testnet) if connected party as a positive balance in the vested rewards account
- Handles empty activity streak data (renders '-' instead of 1x)
- Fixes an issue when if assets weren't loaded before the reward history was loaded we wouldn't have quantum/decimals available for formatting

# Demo 📺

Screenshot 2023-12-04 at 19 40 15

# Technical 👨‍🔧


apps/trading/components/rewards-container/use-reward-row-data.spec.ts
```diff
@@ -0,0 +1,176 @@
+import type { AssetFieldsFragment } from '@vegaprotocol/assets';
+import { getRewards } from './use-reward-row-data';
+import * as Schema from '@vegaprotocol/types';
+
+const testData = {
+  rewards: [
+    {
+      rewardType: Schema.AccountType.ACCOUNT_TYPE_FEES_INFRASTRUCTURE,
+      assetId:
+        'ff3df1224e159f8619d9d66c56c21a11e10422dfa99dc45937fdff364de5e546',
```
mattrussell36, line 10: can we change these keys to something easier to read liket asset-1, asset-2 etc

apps/trading/components/rewards-container/use-reward-row-data.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+      symbol: 'USD-KRW',
+      decimals: 6,
+      quantum: '1000000',
+      source: {
+        __typename: 'ERC20',
+        contractAddress: '0xcFB5538B6235531BEbF09a04c25912e1327C5c52',
+        lifetimeLimit: '100000000000000000000',
+        withdrawThreshold: '1',
+      },
+      status: Schema.AssetStatus.STATUS_ENABLED,
```
mattrussell36, line 39: If you are typecasting to keep the test succint and not have to include irrelevant fields, why not remove the other irrelevant fields such as `source` or anything else that this function doesn' require

apps/trading/components/rewards-container/use-reward-row-data.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+    },
+    {
+      rewardType: Schema.AccountType.ACCOUNT_TYPE_FEES_INFRASTRUCTURE,
+      assetId:
+        '8ba0b10971f0c4747746cd01ff05a53ae75ca91eba1d4d050b527910c983e27e',
+      amount: '5501',
+    },
+  ],
+  assets: {
+    ff3df1224e159f8619d9d66c56c21a11e10422dfa99dc45937fdff364de5e546: {
```
mattrussell36, line 27: If you pull each asset out into a variable, you can make this test more maintainable as you can use them in your assertions later

apps/trading/components/rewards-container/use-reward-row-data.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+        lifetimeLimit: '3000000000000000000000000000000000000000000',
+        withdrawThreshold: '1',
+      },
+      status: Schema.AssetStatus.STATUS_ENABLED,
+    } as AssetFieldsFragment,
+  },
+};
+
+describe('getRewards', () => {
+  it('should return the correct rewards when infra fees are included', () => {
```
mattrussell36, line 87: Seeing as this is the first test coverage here I think its probably worthwhile to make this test more fields, that way we'll gain coverage on the code that handles summing of all the reward values.

mattrussell36 (changes requested):

apps/trading/components/rewards-container/use-reward-row-data.spec.ts
MadalinaRaicu: updated

apps/trading/components/rewards-container/use-reward-row-data.spec.ts
MadalinaRaicu: updated. still need the cast tho

apps/trading/components/rewards-container/use-reward-row-data.spec.ts
MadalinaRaicu: I added more fields

mattrussell36 (approved):

JonRay15:
So ... as discussed with 
[@Matt](https://vegaprotocol.slack.com/team/UD9SXJ1E0)
... I have just been digging around in the APIs looking at user rewards for some whales (both a whale staker and a whale LP).

I think we are safe to put the entire [Rewards ](https://github.com/vegaprotocol/frontend-monorepo/pull/5437)page live IF we hardcode the LOCKED & VESTING balance for USDT to 0.  If we dont do that then the Core bug which is impacting the vestingDetails API is going to trip us up and show vesting USDT that does not really exist.  In reality all the USDT that was incorrectly in the vesting account was there because of the previous Core bug that put infra fees through vesting.  That was fixed quite a while ago and as a result all that USDT should've now vested, and I have verified that for the biggest whale I know on the network (who logically should've had the most vesting USDT and therefore taken the longest to VEST).

We should still show USDT if the user has some in their VESTED account since that is the money from that bug that users may want to redeem.  But that should be fine because that "Available to withdraw" section should be coming from the accounts API anyway which is not impacted by the bug.

I have tested the vestingDetails API for VEGA on a LP and verified that the locked / vesting balance is working properly on it so I think it is safe to rely on that properly for the locked / vesting balance for Vega which will satisfy some of the users that 
[@barney](https://vegaprotocol.slack.com/team/U8W4R5HC5)
 has mentioned were complaining about not being able to see the locked.

PR #5459, opened 2023-12-06
fix(governance): handle null validator scores properly

### LINKED ISSUE #5191: The validator penalties dont disntinguish between NULL and actually zero

edd:
# Related issues 🔗

Closes #5191

# Description ℹ️

Properly handle null values when calculating total and other penalties. It's not totally clear to me why the value in the test changes, but it seems correct to me.

# Demo 📺

Screenshot 2023-12-06 at 12 28 02

### DIFF
--- a/apps/governance/src/routes/staking/home/validator-tables/standby-pending-validators-table.tsx
+++ b/apps/governance/src/routes/staking/home/validator-tables/standby-pending-validators-table.tsx
@@ -124,6 +124,15 @@ export const StandbyPendingValidatorsTable = ({
             }
           }
 
+          const overstakingPenalty = calculateOverallPenalty(
+            id,
+            allNodesInPreviousEpoch
+          );
+          const totalPenalty = calculateOverstakedPenalty(
+            id,
+            allNodesInPreviousEpoch
+          );
+
           return {
             id,
             [ValidatorFields.RANKING_INDEX]: stakedTotalRanking,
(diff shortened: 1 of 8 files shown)


MadalinaRaicu (approved):

apps/governance/src/routes/staking/home/validator-tables/standby-pending-validators-table.tsx
```diff
@@ -153,14 +162,12 @@ export const StandbyPendingValidatorsTable = ({
...
-              2
-            ),
-            [ValidatorFields.TOTAL_PENALTIES]: formatNumberPercentage(
-              calculateOverallPenalty(id, allNodesInPreviousEpoch),
-              2
-            ),
+            [ValidatorFields.OVERSTAKING_PENALTY]: overstakingPenalty
+              ? formatNumberPercentage(overstakingPenalty, 2)
+              : '-',
+            [ValidatorFields.TOTAL_PENALTIES]: totalPenalty
```
asiaznik, line 168: **nitpick**: all of this could be handled inside `formatNumberPercentage`

apps/governance/src/routes/staking/shared.ts
```diff
@@ -37,19 +41,30 @@ const calculateTheoreticalStakeScore = (nodeId: string, nodes: Node[]) => {
 };
 
 /**
- * Calculates overall penalty for a given node
+ * Calculates overall penalty for a given nodsadase
```
asiaznik, line 44: **new word warning** 🥲 Say `nodsadase` one more time, I dare you. @#$% 🔫

apps/governance/src/routes/staking/home/validator-tables/standby-pending-validators-table.tsx
edd: Making this change was the right move, but revealed some failing tests in develop! I blame you!

MadalinaRaicu (approved):

MadalinaRaicu (approved):

MadalinaRaicu (approved):

edd:
@vegaprotocol/frontend-qa my memory is that we agreed that this is hard to test, so we should just merge it? I don't mind being told no - but if that isn't the case, how can we make this mergeable?

bwallacee (approved):

edd:
Waiting for #5646 to be merged before this can be merged. Out of politeness, not a technical reason.

PR #5467, opened 2023-12-07
feat(trading): show commission earned on referrer stats from last 30 days

### LINKED ISSUE #5203: Commission earned on referer stats should be over 30 days
## The Chore

On referrer stats

Image

## Tasks

- [x] Confirm we want to use 30 epochs here
- [x] Change text to "Commission earned in qUSD (last 30 epochs)"
- [x] Check on the query on referralSetReferees ... ensure we're setting aggregationEpochs to 30 for the commission (but still using `windowLength` for the volume)
- [x] Add a tooltip over the 30 saying "Depending on data node retention you may not be able see the full 30 days"

bglownia:
# Related issues 🔗

Closes #5203 

# Description ℹ️

Show commission earned on referrer stats from last 30 days


bwallacee:
@bglownia I think we should get a market-sim test made for this.

@dalebennett1992 Will have his referrals PR in soon that you can use as the base for the tests.

It might be worth speaking to dale before he leaves as he is set up on a windows machine.

apps/trading/client-pages/referrals/tile.tsx
```diff
@@ -55,7 +55,7 @@ export const StatTile = ({
       >
         {title}
       </h3>
-      <div data-testid={`${testId}-value`} className="text-5xl text-left">
```
bwallacee, line 58: Removing the value is what has broken the tests:
image

`'[data-testid=combined-volume-value]': expected_volume,`
Are you happy for it to be added back in.

bwallacee (approved):

asiaznik (approved):

PR #5499, opened 2023-12-13
chore(trading): avoid portfolio and market page re-rendering on sidebar view change

bglownia:
# Description ℹ️

When sidebar view was changed it caused MarketPage and Portfolio page to re-render. It caused to re-render all child components and some of them were remounted. This was noticed in liquidation price rendered by positions table when active sidebar section was changed


bwallacee:
@bglownia Maybe i've misunderstood the ticket, but if i click to open withdraws on the sidebar the whole of console is re rendered for me.

bwallacee (approved):

PR #5501, opened 2023-12-13
feat(wallet): disconnect if wallet unreachable

### LINKED ISSUE #4717: Apps do not recognise a wallet being disconnected
## Chore

Disconnect the user when the connection is closed in Desltop Wallet or CLI Wallet. This could be done by polling or possibly using window blur and focus events.

Note. This ticket has been edited to only concern json rpc connections

## Tasks

- [ ] ACs
- [ ] Dev
- [ ] Test

### OG bug report

From Dex:

"I think there is a bug in both console and governance that when a wallet is disconnected the apps do not realise it. This is a hark back to the desktop/cli wallet days, but IMO:

- A CLI/DW wallet should be polling some easy endpoint (version) and when the connection drops should handle a disconnected state
- A BW should (but does not yet) emit an event when a disconnect takes place (same as eth) so that when a wallet is disconnected the app can handle the disconnected state"

## Acceptance Criteria

- [ ] If Console / Governance attempt to send a transaction of any sort and they get back a 401 / 404 from CLI wallet they should treat that as being disconnected and clear any existing connections from the UI
- [ ] If Console / Governance attempt to send a transaction of any sort and they get back a disconnected status in the JSON RPC they should treat that as being disconnected and clear any existing connections from the UI.

## Additional Info 
There will be some other disconnect event stuff on a later ticket

asiaznik:
# Related issues 🔗

Closes #4717

# Description ℹ️

* closes connection to the wallet if it's unreachable (closed inside wallet, closed process, crashed app, etc.)

# Demo 📺

https://github.com/vegaprotocol/frontend-monorepo/assets/1980305/a5b03c3e-6828-4fcc-bd01-9bda92aa0846

# Technical 👨‍🔧

It's way easier to check if the connection is alive rather than responding to the failed transactions (for whatever reason) as it automatically resets the connection if wallet is unreachable.

### DIFF
--- a/apps/trading/components/vega-wallet-connect-button/vega-wallet-connect-button.tsx
+++ b/apps/trading/components/vega-wallet-connect-button/vega-wallet-connect-button.tsx
@@ -24,7 +24,13 @@ import classNames from 'classnames';
 import { useGetCurrentRouteId } from '../../lib/hooks/use-get-current-route-id';
 import { useT } from '../../lib/use-t';
 
-export const VegaWalletConnectButton = () => {
+export const VegaWalletConnectButton = ({
+  intent = Intent.None,
+  onClick,
+}: {
+  intent?: Intent;
+  onClick?: () => void;
+}) => {
   const t = useT();
   const [dropdownOpen, setDropdownOpen] = useState(false);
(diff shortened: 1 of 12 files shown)


bwallacee (approved):
Looks good.

Issue #5245, opened 2023-11-10
Manage AgGrid table state


mattrussell36:
## Story

As a user
I want to be able to reset or control AgGrid state
So that I can get back to the default state or better control table setup

## Acceptance Criteria

- [x] I can reset column state removing all applied filter, sort and columns rules
- [x] I can switch view to a settings view which provides a space for a reset table button

## Tasks

- [x] Verify there isn't a built in AgGrid UI to handle this
- [ ] UX (if needed)
- [ ] Design (if needed)
- [ ] Team and stakeholder review
- [ ] Specs reviewed and created or adjusted
- [x] Implementation
- [x] Testing (unit and/or e2e)
- [ ] Code review
- [x] QA review

## Additional details / background info

https://vegaprotocol.slack.com/archives/C037STCKMGC/p1699553640171639

Column and filter state from AgGrid gets persisted in local storage. For some actions, for example if you drag a column off the table, there is no way to reverse it and get that column back, so we need a way to reset table state or better, control it via some UI.

mattrussell36:
New issue here: https://github.com/vegaprotocol/frontend-monorepo/issues/5394

PR #5509, opened 2023-12-18
feat(trading): pane settings view

### LINKED ISSUE #5394: Pane settings view
## The Chore

Each pane might contain lots of different settings (for example column config, decimal precision). This ticket is to create a component or extend the components we have to allow each pane view to be toggled between the main view and the settings view. Ideally the toggle would exist in the pane context menu

Note that we don't need to populate the settings controls in this ticket, just provide the functionality to switch between views

## Tasks

- [x] Create pane settings ui and toggle
- [x] Dev
- [ ] Test

## Sketch

Screenshot 2023-11-29 at 18 30 46

bglownia:
# Related issues 🔗

Closes #5394 #5245

# Description ℹ️

Add option to reset tables filters and sorting in console app

# Demo 📺

image


libs/ui-toolkit/src/components/tabs/tabs.tsx
```diff
@@ -122,6 +140,23 @@ export const Tabs = ({
           );
         })}
       </div>
+      {settingsOpened && !!settings ? (
```
mattrussell36, line 143: For me this UI doesn't look tidy. I don't like the pane overlaying the tabs bar and how the button has to dropdown.

mattrussell36: It needs a left border too.

libs/ui-toolkit/src/components/tabs/tabs.tsx
```diff
@@ -38,12 +38,22 @@ export const Tabs = ({
     setWrapped(isWrapped(tabsRef.current, menuRef.current));
   });
 
+  const settings = children.find((c) => c?.props.id === activeTab)?.props
```
mattrussell36, line 41: Why not just render the settings within the map of children, then you can access the settings render prop directly

libs/ui-toolkit/src/components/tabs/tabs.tsx
```diff
@@ -18,12 +20,10 @@ export const Tabs = ({
   onValueChange,
   ...props
 }: TabsProps) => {
-  const [activeTab, setActiveTab] = useState<string | undefined>(() => {
-    if (defaultValue) {
-      return defaultValue;
-    }
-    return children.find((v) => v)?.props.id;
-  });
+  const [settingsOpened, setSettingsOpened] = useState(false);
```
mattrussell36, line 23: I'm wondering if each pane store (the ones that use the `createDataGridSlice`) should have a `createSettingsSlice`, which includes a simple boolean field for if the settings are open and a setter function. This has some advantages such as being open the settings for a specific pane from anywhere. It also means we don't have to tie the tabs component to our 'pane' logic

mattrussell36 (changes requested):

libs/ui-toolkit/src/components/tabs/tabs.tsx
mattrussell36: Pr to merge here: https://github.com/vegaprotocol/frontend-monorepo/pull/5532

mattrussell36: Resolving this as we are taking a different approach

libs/ui-toolkit/src/components/tabs/tabs.tsx
bglownia: Settings are in popover now.

daro-maj (approved):

mattrussell36 (approved):

PR #5515, opened 2023-12-18
chore(trading): fix wallet test

bwallacee:
I found this test was just proceeding and not actually doing any asserts.

The fix is a bit crappy, I can't poll specifailly for the expect to return. Inside the routing any expects won't actually fail the test.
So I have to have an expect inside and outside.

If I remove the expect inside then the routing won't catch the request properly which then allows me to expect.

### DIFF
--- a/apps/trading/e2e/tests/wallet/test_wallet.py
+++ b/apps/trading/e2e/tests/wallet/test_wallet.py
@@ -16,6 +16,7 @@ expire = "expire"
 api_request_match = r"http://localhost:\d+/api/v2/requests"
 
 
+
 @pytest.fixture(scope="module")
 def vega(request):
     with init_vega(request) as vega:
@@ -25,6 +26,7 @@ def vega(request):
 @pytest.fixture(scope="module")
 def continuous_market(vega):
     return setup_continuous_market(vega)
+
 
 
 def handle_route_connection_lost(route: Route, request):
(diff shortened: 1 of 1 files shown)


apps/trading/e2e/tests/wallet/test_wallet.py
```diff
@@ -91,6 +102,9 @@ def test_wallet_connection_rejected(continuous_market, page: Page):
     expect(page.get_by_test_id("dialog-content").nth(1)).to_have_text(
         "User errorthe user rejected the wallet connectionTry againAbout the Vega wallet  | Supported browsers "
     )
+    expect(page.get_by_test_id("dialog-content").nth(1)).to_have_text(
```
mattrussell36, line 105: This is a duplication of the line above

mattrussell36 (changes requested):

apps/trading/e2e/tests/wallet/test_wallet.py
bwallacee: Yeah so this is the crappy "fix"

Playwright doesn't acknowledge errors thrown within routing, so the test will "error" but the test result will be pass, 

But duplicating the expect the test is then failing correctly.

There must be a better solution but this one works.

I'm more than happy for someone else to have a look to see if they can solve better.

PR #5520, opened 2023-12-19
chore(trading): static vega port

bwallacee:
If we use local-server when running tests it will use a static port for graphql to make debugging tests locally easier.

### DIFF
--- a/apps/trading/e2e/conftest.py
+++ b/apps/trading/e2e/conftest.py
@@ -6,7 +6,8 @@ import requests
 import time
 import docker
 import http.server
-
+import sys
+from dotenv import load_dotenv
 from contextlib import contextmanager
 from vega_sim.null_service import VegaServiceNull, Ports
 from playwright.sync_api import Browser, Page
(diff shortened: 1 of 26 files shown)


apps/trading/e2e/conftest.py
```diff
@@ -58,7 +58,12 @@ def do_GET(self):
 
 # Start VegaServiceNull
 @contextmanager
-def init_vega(request=None):
+def init_vega(request=None, local_server: bool = False):
```
edd, line 61:
```suggestion
def init_vega(request=None):
```

**thought (blocking)**:  It feels to me like this could be inverted and the whole diff could be simplified. If the environment variable was not set, it could just use a default of 8001? The 5 lines below seem like they cover the logic, much more simply

edd (changes requested):

apps/trading/e2e/conftest.py
bwallacee: I think I've come up with a good solution with using the .env file. Please let me know what you think

edd (approved):

mattrussell36 (approved):

PR #5526, opened 2023-12-20
feat(trading): poll market info, and last fundingPeriod

### LINKED ISSUE #5409: Funding countdown can be out of sync after market update proposal
## Description

Its possible for the funding countdown to get out of sync if a market proposal is enacted which changes the funding interval time.

We fetch the market's data source once on page load, if a market update is enacted this data source can be stale and incorrect.

## Tasks
- [ ] Add interval to refetch 'static' market data every 15 seconds (so we can get the updated data source with funding payment interval)
- [ ] Get and use startTime from market data subscription for funding payment countdown calculation

## Additional context

Concluded in refinement yesterday that if this can impact this and other market info we should just be polling all market info on a regular interval to ensure nothing can get out of sync.  Specific interval TBC pending performance considerations.

https://vegaprotocol.slack.com/archives/C8WS7M50E/p1701352681026319

bglownia:
Related issues 🔗

Closes #5409 

# Description ℹ️

Update market info, and last fundingPeriod every 5 seconds to ensure that data is up to date


apps/trading/client-pages/market/market-header-stats.tsx
```diff
@@ -212,12 +212,22 @@ const useEvery = (marketId: string) => {
...
+    data: fundingPeriods,
+    startPolling,
+    stopPolling,
+  } = useFundingPeriodsQuery({
     variables: {
       marketId: marketId,
       pagination: { first: 1 },
     },
   });
+  useEffect(() => {
```
mattrussell36, line 225: you can just pass `pollInterval: 5000` to the query directly

mattrussell36 (changes requested):

daro-maj (approved):

PR #5619, opened 2024-01-13
chore(trading, governance, explorer): 0.74.0 type regen

mattrussell36:
# Related issues 🔗

N/A

# Description ℹ️

Fix changed types for 0.74

- New interval types
- ProposalNode is not a union of Proposal and BatchProposal. This caused lots of typescript errors but none of these should result in run time errors
- New OrderRejectionReason enum fields
- New TransferType enum fields

# Demo 📺

N/A


libs/deal-ticket/src/components/deal-ticket/deal-ticket.tsx
```diff
@@ -263,8 +263,11 @@ export const DealTicket = ({
     marketId: market.id,
     openVolume,
     orders,
-    collateralAvailable:
-      marginAccountBalance || generalAccountBalance ? balance : undefined,
+    marginAccountBalance: marginAccountBalance,
+    generalAccountBalance: generalAccountBalance,
+    orderMarginAccountBalance: '0', // TODO: Get real balance
+    marginMode: Schema.MarginMode.MARGIN_MODE_CROSS_MARGIN, // TODO: unhardcode this and get users margin mode for the market
```
mattrussell36, line 269: @bglownia You'll need to update these in your estimatePosition changes work

apps/governance/src/routes/proposals/types.ts
```diff
@@ -0,0 +1,11 @@
+import type { ProposalQuery } from './proposal/__generated__/Proposal';
+
+/**
+ * The default Proposal type needs extracting from the ProposalNode union type
+ * as lots of fields on the original type don't exist on BatchProposal. Eventually
+ * we will support BatchProposal but for now we don't
+ */
+export type Proposal = Extract<
```
mattrussell36, line 8: @edd @asiaznik  I've added this type to work around the fact that ProposalNode is now a union of Proposal and BatchProposal. Use this type for proposals to avoid type narrowing the union as right now we only have normal Proposals

PR #5627, opened 2024-01-16
chore(trading): store all chart state

mattrussell36:
# Related issues 🔗

N/A

# Description ℹ️

- Updates to store the entire serialized chart state object, so that the entire chart including studies, annotations, emojis are all persisted
- Copies in trading view types and uses them
- Ensures chart doesn't have to remount when the market/symbol changes


bwallacee:
@mattrussell36 It might be a stagnet thing,(it appears to be very slow at the moment)
But when i swapped from TV to pennant and back to TV charts I got a symbol error 
image

bwallacee (approved):

asiaznik (approved):

PR #5633, opened 2024-01-17
chore(trading): delete and move some tests to jest

bwallacee:


apps/trading/components/grid-settings/grid-settings.spec.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { GridSettings } from './grid-settings';
+import '@testing-library/jest-dom';
+
+jest.mock('../../lib/use-t', () => ({
```
mattrussell36, line 7: You shouldn't need to mock use-t like this. Its handled in setup-tests

bwallacee: removed, thanks.

daro-maj (approved):

mattrussell36 (approved):

PR #5640, opened 2024-01-18
feat(trading): filter out suspended transfers

### LINKED ISSUE #5587: Filter out non-paying rewards from the rewards view
## The Chore

The rewards view is currently still showing rewards for markets that have been terminated.

Now ... yes ... this is because the transfers to fund them havent yet been cancelled so it is an admin task that needs to be sorted, BUT i think it would still be preferable to hide them long term so we're not confusing users with bad admin (and I there is a devops task now to add an alert for that admin).

## Acceptance Criteria

- [x] If the reward specifies a list in `markets`:
  - [x] If there is at least one market in status ACTIVE show the card as usual
  - [x] If there is at least one market in status SUSPENDED grey out the card and show a tooltip saying "Eligible market(s) currently suspended"
  - [x] If neither of the above is true and ALL markets are either terminated or settled then do not show the card at all
- [x] If a reward is for all markets in a given settlement asset check all markets in that asset:
  - [x] if there is at least one market in status ACTIVE show the card as usual
  - [x] Otherwise grey out the card and show the tooltip "Currently no markets eligible for reward" (ie. for this type we dont hide it completely because someone creating a market might want to know that if they create it settling in asset X it will be eligible for a reward)

MadalinaRaicu:
# Related issues 🔗

Closes #5587 

# Description ℹ️

Gray out reward cards if they're related to suspended markets. 
Filter transfers related to terminated or settled markets.

# Demo 📺

Screenshot 2024-01-18 at 15 07 16

# Technical 👨‍🔧

Removed redundant queries.
Refactor active rewards variable names.


apps/trading/components/rewards-container/active-rewards.tsx
```diff
@@ -85,6 +83,22 @@ export const applyFilter = (
   ) {
     return false;
   }
+
+  if (
```
bwallacee, line 87: Are we missing some of the other suspended market states?
And would it be easier to reverse this if statement if we only want to show colour for active?

bwallacee (approved):

PR #5653, opened 2024-01-23
fix(governance): handle an expected error

### LINKED ISSUE #5634: Error: Something went wrong: failed to get party for ID on the governance rewards page
## Description

Error: Something went wrong: failed to get party for ID on the governance rewards page 

## Steps to Reproduce

1. Go to https://governance.fairground.wtf/rewards
2. Connect vega browser wallet
3. Switch from total distributed to entered by me 
4. See error: Something went wrong: failed to get party for ID

## Screenshots
Screenshot 2024-01-17 at 11 59 47

edd:
# Related issues 🔗

Closes #5634

Also closes a similar issue with Voting - if a user had no accoutns and went to a single proposal, they would see an error rather than a suggestion to connect a wallet.

# Description ℹ️
If a party has never deposited or traded, they have no accounts and the `parties` API will return an error that the party
does not exit. This was handled poorly, and rendered a 'party does not exist' error instead of simply 'the key you are
looking at does not have any rewards'. This PR filters out the Party Does Not Exist error so that the view renders more
usefully for the user.

# Demo 📺

Screenshot 2024-01-23 at 14 33 20

### DIFF
--- a/apps/governance/src/routes/rewards/epoch-individual-rewards/epoch-individual-rewards.tsx
+++ b/apps/governance/src/routes/rewards/epoch-individual-rewards/epoch-individual-rewards.tsx
@@ -10,7 +10,7 @@ import { EpochIndividualRewardsTable } from './epoch-individual-rewards-table';
 import { generateEpochIndividualRewardsList } from './generate-epoch-individual-rewards-list';
 import { calculateEpochOffset } from '../../../lib/epoch-pagination';
 import { useNetworkParam } from '@vegaprotocol/network-parameters';
-import { type ApolloError } from '@apollo/client';
+import { filterAcceptableGraphqlErrors } from '../../../lib/party';
 
 const EPOCHS_PAGE_SIZE = 10;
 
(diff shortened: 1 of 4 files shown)


apps/governance/src/routes/rewards/epoch-individual-rewards/epoch-individual-rewards.tsx
```diff
@@ -99,17 +100,24 @@ export const EpochIndividualRewards = ({
...
+  // Workarounds for the error handling of AsyncRenderer
+  const filteredErrors = filterExpectedErrors(error);
+  const filteredData = data || [];
+
   return (
     <AsyncRenderer
       loading={loading}
-      error={error}
-      data={data}
+      error={filteredErrors}
```
MadalinaRaicu, line 110: the same issue I've seen when I was trying to cast a vote but there were no associated tokens on my account. 
could you please check `usePoolForDelegations` where `isPartyNotFoundError` gets triggered?

edd: Done

PR #5654, opened 2024-01-23
chore(trading): merge main back in develop

MadalinaRaicu:


libs/markets/src/lib/components/market-info/market-info-panels.tsx
```diff
@@ -698,17 +698,13 @@ export const PriceMonitoringBoundsInfoPanel = ({
...
   const quoteUnit = getQuoteName(market);
 
-  const trigger =
-    market.priceMonitoringSettings?.parameters?.triggers?.[triggerIndex];
-
   const bounds = data?.priceMonitoringBounds?.[triggerIndex];
+  const trigger = bounds?.trigger;
 
   if (!trigger) {
-    console.error(
```
MadalinaRaicu, line 707: this is a redundant error log

PR #5663, opened 2024-01-24
chore(trading): reuse vega instance WIP DO NOT MERGE

bwallacee:


apps/trading/e2e/pyproject.toml
```diff
@@ -26,4 +26,7 @@ log_cli_date_format = "%Y-%m-%d %H:%M:%S"
 log_cli_level = "INFO"
 log_file_format = "%(asctime)s - %(name)s - %(levelname)s: %(message)s"
 log_file_date_format = "%Y-%m-%d %H:%M:%S"
-log_file_level = "INFO"
\ No newline at end of file
+log_file_level = "INFO"
+markers = [
```
mattrussell36, line 30: What does this do?

apps/trading/e2e/tests/fees/test_fees.py
```diff
@@ -645,6 +656,7 @@ def test_fills_maker_fee_tooltip_discount_program(
     )
 
 
+@pytest.mark.xdist_group(name="test_fees")
```
mattrussell36, line 659: Nice, presumably this groups for parallelization?

bwallacee: Yeah, if you use groups to specifiy parralelisation, anything that isn't given a group will just be run randomly (not grouped by file etc). So this is required to make sure tests under the same group name are run in the same thread

apps/trading/e2e/pyproject.toml
bwallacee: It works very simmilar to the groups, annoying with pytest you can't manually kick off runs of groups, but you can use markers.
So I'm using this to run the shared vega tests only, locally.

PR #5668, opened 2024-01-25
feat(trading): gas fee estimation for withdraw transaction

### LINKED ISSUE #5510: Add an estimate of gas fee's for withdrawals when releasing funds
## Story

So when you prepare to withdraw and release funds from Vega you will not know the gas cost until your funds are ready for withdrawal and at that point you can't change your mind.
We have had users trying to withdraw 25 USD but at the time they realised gas fee's were so high that it basically cost them 25 USD to withdraw it was too late, so now they leave them at pending.

Would be good to give an estimate of the fee so that someone does not try to withdraw a tiny amount that does not end up being financially viable.

Screenshot 2023-12-18 at 11 21 11
Screenshot 2023-12-18 at 11 15 38

## Tasks

- [x] Add the gas estimate to the table above with the linits / delays etc
- [x] If the cost of gas is "high" show a warning
  - [ ] ~~Ideally we could look at historic withdrawals and compare to an average and~~ show it as green / amber / red relative to this
  - [x] An easier solution might be to just compare the gas cost to the size of your withdrawal, if your're paying more in gas than you're withdrawing ~~that could be a warning~~
  - [ ] ~~Or we could simply look at historic data and set some thresholds manually, that would be the easiest (LETS DO THIS cos its easy)~~

## Current implementation:

 It compares the gas fee to the withdrawal amount and colours the gas fee estimation:
   - red - if the amount is lower than the base gas fee
   - amber - if the amount is lower than the maximum gas fee
   - white/black - otherwise

## Additional notes

Check ethers methods and / or metamask API to have a look at this stuff.

asiaznik:
# Related issues 🔗

Closes #5510

# Description ℹ️

This PR adds the gas fee estimation for the `<WitdrawLimits>` component. It colours the prices orange if the amount it between base and max price, and red if it's below the base fee.

# Demo 📺

Screenshot 2024-01-25 at 19 40 14
Screenshot 2024-01-25 at 19 40 23

Shows in ETH when price is >1e6 gwei:
Screenshot 2024-01-30 at 15 13 16

# Technical 👨‍🔧

The withdrawal amount (any asset) is only somewhat comparable to the estimated gas (ETH) fee only if they're both equalised to qUSD. This PR also adds the `useWETH` hook which looks for the `WETH` configuration (quantum is really what's needed) in the Vega's assets (it returns hardcoded values if not found, 1 WETH ~= 2000 qUSD).

### DIFF
--- a/libs/utils/src/lib/format/ether.ts
+++ b/libs/utils/src/lib/format/ether.ts
@@ -1,0 +1,85 @@
+// eslint-disable-next-line no-restricted-imports
+import { formatNumber, toBigNum } from './number';
+import type BigNumber from 'bignumber.js';
+
+export enum EtherUnit {
+  /** 1 wei = 10^-18 ETH */
+  wei = '0',
+  /** 1 kwei = 1000 wei */
+  kwei = '3',
+  /** 1 mwei = 1000 kwei */
+  mwei = '6',
+  /** 1 gwei = 1000 kwei */
+  gwei = '9',
+
+  // other denominations:
+  // microether = '12', // aka szabo, µETH
+  // milliether = '15', // aka finney, mETH
+
+  /** 1 ETH = 1B gwei = 10^18 wei */
+  ether = '18',
+}
+
+export const etherUnitMapping: Record<EtherUnit, string> = {
+  [EtherUnit.wei]: 'wei',
+  [EtherUnit.kwei]: 'kwei',
+  [EtherUnit.mwei]: 'mwei',
+  [EtherUnit.gwei]: 'gwei',
+  // [EtherUnit.microether]: 'µETH', // szabo
+  // [EtherUnit.milliether]: 'mETH', // finney
+  [EtherUnit.ether]: 'ETH',
+};
+
+type InputValue = string | number | BigNumber;
+type UnitisedTuple = [value: BigNumber, unit: EtherUnit];
+
+/**
+ * Converts given raw value to the unitised tuple of amount and unit
+ */
+export const unitiseEther = (
+  input: InputValue,
+  forceUnit?: EtherUnit
+): UnitisedTuple => {
+  const units = Object.values(EtherUnit).reverse();
+
+  let value = toBigNum(input, Number(forceUnit || EtherUnit.ether));
+  let unit = forceUnit || EtherUnit.ether;
+
+  if (!forceUnit) {
+    for (const u of units) {
+      const v = toBigNum(input, Number(u));
+      value = v;
+      unit = u;
+      if (v.isGreaterThanOrEqualTo(1)) break;
+    }
+  }
+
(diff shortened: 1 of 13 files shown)


MadalinaRaicu (approved):

libs/web3/src/lib/use-gas-price.ts
```diff
@@ -0,0 +1,105 @@
+import { useEffect, useState } from 'react';
+import { useWeb3React } from '@web3-react/core';
+import { useEthereumConfig } from './use-ethereum-config';
+import BigNumber from 'bignumber.js';
+
+const DEFAULT_INTERVAL = 15000; // 15 seconds
+
+export enum ContractMethod {
```
edd, line 8: **nitpick (non-blocking)**:  Another minor documentation thing, but it would be nice to know where these values come fromm if things change (or other bridges get developed). A comment is probably overkill, but maybe add it to the PR description? 

I assume the easy method is by checking the contract source on Etherscan... 🤷

libs/withdraws/src/lib/withdraw-limits.tsx
```diff
@@ -91,3 +121,117 @@ export const WithdrawLimits = ({
...
+    };
+
+    const uBasePricePerGas = unitiseEther(basePricePerGas);
+    const uMaxPricePerGas = unitiseEther(
+      maxPricePerGas,
+      uBasePricePerGas[1] // forces the same unit as min price
+    );
+
+    const uBasePrice = unitiseEther(basePrice);
+    const uMaxPrice = unitiseEther(maxPrice, uBasePrice[1]);
```
edd, line 168: **praise (non-blocking )**: this is nice and readable for quite a complex set of interactions, nicely done

edd (approved):
**praise (non-blocking)**: Good test coverage for the important bits, and it's nicely broken up to make a complex thing understandable.

libs/utils/src/lib/format/ether.ts
```diff
@@ -0,0 +1,85 @@
+// eslint-disable-next-line no-restricted-imports
```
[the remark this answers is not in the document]
asiaznik: Good catch, it shouldn't be here, my first approach was relying on the `ethers` hence this comment. It's not needed anymore, will remove.

libs/web3/src/lib/use-gas-price.ts
asiaznik: Yup, it's straight outta compt... I mean contract (on etherscan) 😉

libs/withdraws/src/lib/withdraw-limits.tsx
asiaznik: ❤️

daro-maj (approved):

PR #5718, opened 2024-02-01
feat(trading): update mobile layout

### LINKED ISSUE #5625: Console Mobile Review LAYOUT UPDATES
## User story

_As a_ user
_I want to_ use console on mobile 
_So that_ I can connect my wallet, trade on my phone, and view my positions

## Acceptance criteria

- [x] Split the main view into market data on top and user data below
- [ ] Add/modify our bottom actions (sidebar when horizontal) to have clear CTAs to trade (IE make it super clear how to get to the deal ticket)
- [x] Change the market data (or MarketHeader) to be a popover or collapsible section, so that the default view only shows the most important info.. EG price/change

## Sketch

mobile

```[tasklist]
### Tasks
- [ ] https://github.com/vegaprotocol/frontend-monorepo/issues/5741
```

MadalinaRaicu:
# Related issues 🔗

Partially closes #5625 

# Description ℹ️

- [x] two views:  one for charts & order book etc., second one for positions, orders etc. 
- [x] mobile market header
- [ ] sidebar transformed to larger buttons (may come in a new PR)  https://github.com/vegaprotocol/frontend-monorepo/pull/5751

# Demo 📺
Screenshot 2024-02-02 at 14 44 27

Still in progress: replacing sidebar with footer buttons
Screenshot 2024-01-30 at 23 21 04

# Technical 👨‍🔧


mattrussell36:
Another thing we should look at changing is how the pane action buttons get dropped onto a new row. For mobile we should render all pane actions in the pane settings dropdown. This would save quite a bit of vertical space

Screenshot 2024-02-01 at 11 11 45

asiaznik (approved):

bwallacee:
@MadalinaRaicu Have a few thoughts:
image
I wonder if we could move the highlighted green bit down below. Maybe move the dark mode switch into the settings. We could then get rid of the vega icon (or move it) and gain a nice bit of vertical space.
For the highlighted purple bit i wonder if we can also move this into settings as well and gain another row of space.

jeremyletang:
I'm being told by @edd that if you guys were to update the manifest.json appropriately, one could see the application in full screen on their smartphones (e.g: when adding it to the home page on iOS) and saving a lot of wasted space. Any chance that could be added / experiemented with @MadalinaRaicu as part of this PR / improvements to the mobile UI.

MadalinaRaicu:
> I'm being told by @edd that if you guys were to update the manifest.json appropriately, one could see the application in full screen on their smartphones (e.g: when adding it to the home page on iOS) and saving a lot of wasted space. Any chance that could be added / experiemented with @MadalinaRaicu as part of this PR / improvements to the mobile UI.

sure I can try it out!

bwallacee:
@MadalinaRaicu I've added the page to my homescreen but it still has the URL bar when i open it.

MadalinaRaicu:
> Yes, I am still trying out with it. For Governance and Explorer works fine, however Console uses Next.JS and our custom Head component doesn't seem to import either links or meta tags right.

@mattrussell36 I think we don't even serve the app as Next.js app, so...is it time for us to re-think and remove next altogether? (that should be another PR)

mattrussell36:
@MadalinaRaicu  We aren't really leveraging anything nextjs (asign from a tiny bit of pre-generated html) so it would be good to move off it to something more client side centric (probably vite). Had a quick go today and its a fairly chunky task

Also here are a few things I found:

Its possible to scroll the main section:

Screenshot 2024-02-06 at 19 24 21

The alignment/padding of the new control bar is all a bit off, the chevrons don't align for example

Screenshot 2024-02-06 at 19 27 38

mattrussell36:
@bwallacee Don't agree with moving the nav items down, too much of a divergence from a normal layout and a much bigger change if we did want to do it. So something to reconsider later.

What you've highlighted in the purple box though, I agree with (though it should be a separate PR). Any 'pane controls' that we show inline on desktop should be hidden and only shown in the pane settings dropdown on mobile

apps/trading/client-pages/market/trade-panels.tsx
```diff
@@ -12,21 +11,21 @@ import {
...
-
 interface TradePanelsProps {
   market: Market;
   pinnedAsset?: PinnedAsset;
 }
 
 export const TradePanels = ({ market, pinnedAsset }: TradePanelsProps) => {
-  const [view, setView] = useState<TradingView>('chart');
-  const viewCfg = TradingViews[view];
+  const [view1, setView1] = useState<TradingView>('chart');
```
mattrussell36, line 23: Perhaps call these `topView` and `bottomView`

apps/trading/client-pages/market/trade-panels.tsx
```diff
@@ -69,56 +69,83 @@ export const TradePanels = ({ market, pinnedAsset }: TradePanelsProps) => {
...
+              );
+            })}
+        </div>
+        <div className="h-[376px] sm:h-[460px] lg:h-full relative">
+          <div>{renderMenu(viewCfg1)}</div>
+          <div className="overflow-auto h-full">{renderView(view1)}</div>
+        </div>
       </div>
+
+      {
```
mattrussell36, line 116: I don't think these braces are needed

apps/trading/client-pages/market/trade-panels.tsx
```diff
@@ -69,56 +69,83 @@ export const TradePanels = ({ market, pinnedAsset }: TradePanelsProps) => {
...
+                  view={key}
+                  isActive={isActive}
+                  onClick={() => {
+                    setView1(key);
+                  }}
+                />
+              );
+            })}
+        </div>
+        <div className="h-[376px] sm:h-[460px] lg:h-full relative">
```
mattrussell36, line 110: (non blocking): I think we can probably make a better experience by
A) not define heights here and actually use the grid in the wrapping element
B) use percentage or viewport units

libs/markets/src/lib/components/last-24h-price-change/last-24h-price-change.tsx
```diff
@@ -18,12 +18,15 @@ interface Props {
...
   noUpdate?: boolean;
+  // to render nothing instead of '-' when there is no price change
+  hideZero?: boolean;
 }
 
 export const Last24hPriceChange = ({
   marketId,
   decimalPlaces,
   initialValue,
+  hideZero,
```
mattrussell36, line 29: This would be better as a render prop so this component is more flexible/declaritive

mattrussell36 (changes requested):

apps/trading/client-pages/market/trade-panels.tsx
MadalinaRaicu: updated

apps/trading/client-pages/market/trade-panels.tsx
MadalinaRaicu: updated

libs/markets/src/lib/components/last-24h-price-change/last-24h-price-change.tsx
MadalinaRaicu: updated

apps/trading/client-pages/market/trade-panels.tsx
MadalinaRaicu: updated!

MadalinaRaicu:
> The alignment/padding of the new control bar is all a bit off. The chevrons aren't aligned vertically (you might need to ensure the rotation has a centered transform origin). Also the padding on the left and right doesn't align with the nav above 

@mattrussell36  I have addressed the alignment and the scroll, if you could please review again.

apps/trading/components/market-header/mobile-market-header.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+  return (
+    <div className="p-2 flex justify-between gap-2 items-center h-10 border-b border-default bg-vega-clight-700 dark:bg-vega-cdark-700">
+      <FullScreenPopover
+        open={openMarket}
+        onOpenChange={(x) => {
+          setOpenMarket(x);
+        }}
+        trigger={
+          <h1 className="flex gap-1 sm:gap-2 md:gap-4 items-center text-lg md:text-lg whitespace-nowrap xl:pr-4 xl:border-r border-default">
```
mattrussell36, line 39: these xl styles will never show because this component will never render at that size

mattrussell36 (changes requested):

bwallacee (approved):

MadalinaRaicu:
It still shows changes requested by Matt, but @mattrussell36 pushed the last commit. 
I will be merging this one and any other updates can be done in the PR https://github.com/vegaprotocol/frontend-monorepo/pull/5751 or any other.

PR #5735, opened 2024-02-02
feat(governance): batch proposals

### LINKED ISSUE #5561: Support batch proposals in Governance app
## The Chore

Need to support these, and it will require some UX to work out how to show this.

Note that this will also require changes in the GraphQL due to this change:
https://github.com/vegaprotocol/vega/blob/develop/datanode/gateway/graphql/schema.graphql#L6115-L6125

## Tasks

- [x] Design the UX
- [x] Build it
- [ ] ~~Add a form to handle multiple proposals within a batch~~
- [x] Adjust proposal list to show that its a batch
  - [x] Show sub proposal lozenges across the top
- [x] Adjust proposal page to show that its a batch and include each proposal data in the batch
  - [ ] Add an accordion for each sub proposal type
- [ ] Share design changes with team for signoff

API ticket for Core: https://github.com/vegaprotocol/vega/issues/10069

Also dependent on this bug: https://github.com/vegaprotocol/vega/issues/10525

Be aware there's come complicated voting shenanigans on this because the way batches work is that each propsal is evaluated against ITS OWN THRESHOLDS and then if ALL pass then the batch passes.

That means:

- You will have overall vote majority and participation
- To get the status and details for each proposal you will have to compare those values to the thresholds for that specific proposal type to determine if it is passing of failing currently
- If it is a market update it is even worse because you need to also check whether you have enough LP votes for that market 
  - in theory that info is on the API and is given by market ... which you need because you might have more than one market update and those markets have different LP breakdown so the vote counts for different amounts against them! 
  - The text generation will need to be smarter on those too ... and its painful, and its late ... so for now write "Market updates are complex, please come back later" because its 7:15pm and I want to go home
- Once you know the pass / fail of each individual proposal and have generated the relevant status and text you can generate the summary figure at the top to say whether the batch as a whole is currently passing

## Sketch 

Image

mattrussell36:
# Related issues 🔗

Closes #5561

# Description ℹ️

Updates proposals UI to support batch proposals

# Demo 📺

Screenshot 2024-02-05 at 21 39 10
Screenshot 2024-02-05 at 21 38 56
Screenshot 2024-02-05 at 21 38 46

# Technical 👨‍🔧

- Combined Proposals.graphql and Proposal.graphql into single file for easier sharing of fragments
- Updates proposals list and proposal page to handle batch proposals


apps/governance/src/routes/proposals/components/vote-details/vote-buttons.tsx
```diff
@@ -35,20 +39,97 @@ interface VoteButtonsContainerProps {
...
+        minVoterBalance =
+          networkParams.governance_proposal_updateNetParam_minVoterBalance;
+        break;
+      case 'NewFreeform':
+        minVoterBalance =
+          networkParams.governance_proposal_freeform_minVoterBalance;
+        break;
+      case 'NewTransfer':
+        // TODO: check minVoterBalance for 'NewTransfer'
+        minVoterBalance =
```
edd, line 104: **question (non-blocking)**: Do you still need to todo this?

apps/governance/src/routes/proposals/components/proposal/proposal-change-details.tsx
```diff
@@ -0,0 +1,114 @@
...
+      return <ProposalUpdateMarketState change={terms.change} />;
+    }
+    case 'UpdateReferralProgram': {
+      return <ProposalReferralProgramDetails change={terms.change} />;
+    }
+    case 'UpdateVolumeDiscountProgram': {
+      return <ProposalVolumeDiscountProgramDetails change={terms.change} />;
+    }
+    case 'UpdateNetworkParameter': {
+      if (terms.change.networkParameter.key.slice(-13) === '.benefitTiers') {
```
edd, line 99: **nitpick (non-blocking)**:  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith would be neater... but then this file can only be so neat.

edd (approved):
After fixing the lint error, this is worth merging. I've tried it out with some batches on stagnet.

apps/governance/src/routes/proposals/components/proposal/proposal-change-details.tsx
mattrussell36: I've decided to be more verbose and actually check the full network param key. I think its much clearer

apps/governance/src/routes/proposals/components/vote-details/vote-buttons.tsx
mattrussell36: Yup! Just noticed that this is a bug that is currently on develop (I just moved the code closer to where its used), should be fixed now as correctly referencing the transfer governance proposal params

asiaznik (approved):

bwallacee (approved):

daro-maj (approved):

PR #5752, opened 2024-02-05
fix(environment): pick best node

asiaznik:
# Related issues 🔗

Closes: #5757 

# Description ℹ️

This PR fixes the observed issue and changes the mechanics of the initial node picking in a way that it tests all available nodes, orders them by block height and picks the fastest.

# Demo 📺

N/A

# Technical 👨‍🔧

Moved time measurement from `row-data` component to a util `time.ts`.

### DIFF
--- a/libs/environment/src/components/node-switcher/row-data.spec.tsx
+++ b/libs/environment/src/components/node-switcher/row-data.spec.tsx
@@ -23,7 +23,6 @@ import {
   SUBSCRIPTION_TIMEOUT,
   useNodeBasicStatus,
   useNodeSubscriptionStatus,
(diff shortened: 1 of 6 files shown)


edd (approved):
👍 

**nitpick (non-blocking)**: I have a slight feeling that this should be ordered by response speed, then block height - or something that adds a little diversity to try and give users the *fastest* node. Like if one node consistently responds with the highest height it might get congested and be selected for people on the other side of the world. 

But that’s all a feeling and this PR solves the current issue so complex & theoretically nicer shouldn’t be the enemy of not broken.

PR #5756, opened 2024-02-06
chore(wallet, governance): reuse enums from @vegaprotocol/rest-clients

dexturr:
# Related issues 🔗

N/A

# Description ℹ️

We are hand writing and duplicating types in the wallet lib that are specified in the openapi files.

This has the added advantage of meaning that we use strings rather than swithcing between number enums and string enums.


edd (approved):

daro-maj (changes requested):
Isolated margin  returned error
image

`{jsonrpc: "2.0", error: {code: -32602, message: "Invalid params",…}, id: "DGfaflzxVoK2IJIijM1Y0"}
error: {code: -32602, message: "Invalid params",…}
code: -32602
data: "the transaction does not use a valid Vega command: unknown value \"\\\"MARGIN_MODE_ISOLATED_MARGIN\\\"\" for enum vega.commands.v1.UpdateMarginMode.Mode"
message: "Invalid params"
id: "DGfaflzxVoK2IJIijM1Y0"
jsonrpc: "2.0"`

package.json
```diff
@@ -42,6 +42,7 @@
     "@sentry/nextjs": "^6.19.3",
     "@sentry/react": "^6.19.2",
     "@sentry/tracing": "^6.19.2",
+    "@vegaprotocol/rest-clients": "^0.0.5",
```
mattrussell36, line 45: @dexturr  Rather than using the enums from rest-clients (or enums from graphql generation), would it not be better to import and use the enums that are actually derived from the wallet?

dexturr:
Spoken about on slack.

Would be better to have this coming from the enums from the protos or have a wallet lib with the types of transactions and the enums they use included.

Closing this PR as the newer is lib for enums would accomplish this more easily and effectively.

PR #5761, opened 2024-02-06
chore(utils): improve formatNumber to keep precision

bglownia:


libs/utils/src/lib/format/number.ts
```diff
@@ -90,7 +105,29 @@ export const formatNumber = (
...
+    return formatted;
+  }
+  const minimumFractionDigits = Math.min(
+    Math.max(0, formatDecimals),
+    MIN_FRACTION_DIGITS
+  );
+  if (!minimumFractionDigits) {
+    return formatted;
+  }
+  const parts = formatted.split(decimalSeparator);
```
mattrussell36, line 126: Can you add a comment why this is needed, eg. By default we show all values with 2 dp

libs/utils/src/lib/format/number.ts
```diff
@@ -90,7 +105,29 @@ export const formatNumber = (
   rawValue: string | number | BigNumber,
   formatDecimals = 0
 ) => {
-  return getNumberFormat(formatDecimals).format(Number(rawValue));
+  const decimalSeparator = getDecimalSeparator() || '.';
```
mattrussell36, line 108: I think the js docs for formatNumber and formatNumberFixed need updating, they both indicate the same thing.. we need to clarify

mattrussell36 (approved):

asiaznik (approved):

PR #5772, opened 2024-02-08
chore(trading): team page refactor, games with specified epoch from

### LINKED ISSUE #5762: Game details API calls need to specify epochFrom
## The Chore

The games view in the team profile screen seems to only be pulling back the latest epoch - which is default behaviour if an epochFrom is not specified in the query.

image

## Tasks

- [ ] Always set the `epochFrom` to `currentEpoch - 30` so that we get approx a month's worth of data
- [ ] Include a note somewhere indicating that you are seeing the last 30 days

Tan seems to think that if you set epochTo to current epoch the epochFrom will automatically set to 30 days prior to it ... which is exciting ... so maybe try that if you'd rather not do current minus 30.

If this works we can close Core bug here: https://github.com/vegaprotocol/vega/issues/10569

asiaznik:
# Related issues 🔗

Closes #5762

# Description ℹ️

* exported games query to it's own hook
* previous `useGames` is now `useGameCards`
* added `epochFrom` to the `games` query
* failure of `games` query won't crash the team page

# Demo 📺

N/A

# Technical 👨‍🔧

N/A


asiaznik:
This doesn't miraculously fix issues of games api, but prevent the page from crashing (`page not found`). It also extends timeout for the games query although it tends to fail or return no data - this has to be addressed by core

bwallacee:
image (20)
image (19)

@asiaznik The teams page is missing the games

asiaznik:
> image (20) image (19)
> 
> @asiaznik The teams page is missing the games

As I mentioned before: This doesn't miraculously fix issues of games api.

🤷🏻

daro-maj (approved):

PR #5785, opened 2024-02-10
chore(trading): misc fixes

mattrussell36:
# Related issues 🔗

N/A

# Description ℹ️

- Fix spacing of fees/rewards titles on mobile
- Use page title hook to reduce number of imports


bwallacee:
Looks good on rewards:
image


It looks a little close on Fees:
image

bwallacee:
I'm approving cos its better than before but it feels like it could be better still.

Fees:
image
Rewards:
image

I think there is a bigger gap on the fees than the rewards. (Rewards looks perfect IMO)

bwallacee (approved):

PR #5788, opened 2024-02-12
fix(trading): add missing funding panel, fix 24 hr price/volume change

mattrussell36:
# Related issues 🔗

Closes N/A

# Description ℹ️

- Adds missing funding panel in mobile view
- Fixes overflow breaking layout
- Fixes incorrect 24 hour price values
- Fixes incorrect 24 hour volume values


bwallacee:
@mattrussell36 
The right hand side of the funding payments chart is cut off. Also the actual tab itself is cut off. (I tried scrolling further right but it bounces back to this position) (This was tested on my actual phone not emulator).

You can see the closed markets button at the bottom is simmilar.

IMG_7852

bwallacee (approved):

PR #5815, opened 2024-02-19
(trading, governance): wallet provider rewrite

### LINKED ISSUE #5734: Simplify the connection box
## The Chore

Apparently users have commented that this box is too complicated 

image

Reading is hard clearly.

The preferance seems to be to get right down to the fairly standard box that is more like this.

image

So it would literally just have a box for the Vega Wallet with either the V logo or the Chrome / FF logo or both somehow, and then the Metamask fox, and then a box for advanced options which woiuld offer you the Desktop, CLI, or View as user. Maybe we could incude the existing text as a hover tooltip?
 
## Tasks

- [ ] Agree how it should look
- [ ] Work out what to do with text/

mattrussell36:
# Related issues 🔗

Closes #5734 

# Description ℹ️

- Refactors wallet lib to have no dependency on react
- Adds wallet-react lib for context, components and hooks
- Rewrites redesigns connection dialog to be less busy and include additional icons

# Demo 📺

**Normal connection**
Screenshot 2024-02-18 at 21 12 31

**Onboarding connection**
Screenshot 2024-02-18 at 21 12 57


bwallacee:
@mattrussell36 quick review:

mobile allignment isn't great:
image


I wonder if it would be better as a list on mobile?

There appears to be a regression with not being able to get to the connect wallet screen on mobile:
image

Should there be a message at all if the wallet connection is denied:
image

only the CLI connection works.

libs/wallet/src/types.ts
```diff
@@ -0,0 +1,130 @@
...
+
+export interface Connector {
+  readonly id: ConnectorType;
+  readonly name: string;
+  readonly description: string;
+
+  bindStore(state: StoreApi<Store>): void;
+  connectWallet(chainId?: string): Promise<{ success: boolean }>;
+  disconnectWallet(): Promise<{ success: boolean }>;
+  getChainId(): Promise<{ chainId: string }>;
```
dexturr, line 41: What does the return value of success add here out of interest? Can this not just be `void` if we are throwing errors now?

libs/wallet/src/connectors/snap-connector.ts
```diff
@@ -1,247 +1,282 @@
...
+  }
 
-    const result = await invokeSnap<SendTransactionResponse>(
-      this.snapId,
-      payload
+  off(event: VegaWalletEvent, callback?: () => void) {
+    this.pollListeners[event] = this.pollListeners[event].filter(
+      (cb) => cb !== callback
     );
+  }
```
dexturr, line 178: Suggest extending the `EventEmitter` class rather than hand rolling as it's already bundled?

This is also duplicated across classes, so by extending EventEmitter we can remove this duplication.

libs/wallet/src/connectors/snap-connector.ts
```diff
@@ -1,247 +1,282 @@
...
+    params?: object
+  ): Promise<TResult> {
+    if (
+      'ethereum' in window &&
+      typeof window.ethereum === 'object' &&
+      window.ethereum &&
+      'request' in window.ethereum &&
+      'isMetaMask' in window.ethereum &&
+      window.ethereum.isMetaMask &&
+      typeof window.ethereum.request === 'function'
```
dexturr, line 264: Can this not just be done with simple optional chaining?

```suggestion
      window.ethereum?.isMetaMask &&
      typeof window.ethereum?.request === 'function'
```

libs/wallet/src/connectors/snap-connector.ts
```diff
@@ -1,247 +1,282 @@
...
-      receivedAt: result.receivedAt,
-      sentAt: result.sentAt,
-    };
+  /**
+   * Requests permission for a website to communicate with the specified snaps
+   * and attempts to install them if they're not already installed.
+   * If the installation of any snap fails, returns the error that caused the failure.
+   * More informations here: https://docs.metamask.io/snaps/reference/rpc-api/#wallet_requestsnaps
+   */
+  async requestSnap() {
```
dexturr, line 211: These utility methods appear to only be used internally to this class.

Suggest making all these methods private so this implementation only exposes public methods that conform to the interface.

libs/wallet/src/errors.ts
```diff
@@ -0,0 +1,22 @@
...
+  noConnector: new ConnectorError('no connector', 1),
+  connect: new ConnectorError('failed to connect', 2),
+  disconnect: new ConnectorError('failed to disconnect', 3),
+  chainId: new ConnectorError('incorrect chain', 4),
+  listKeys: new ConnectorError('failed to list keys', 5),
+  isConnected: new ConnectorError('failed to check connection', 6),
+  sendTransaction: new ConnectorError('failed to send transaction', 7),
+  unknown: new ConnectorError('unknown error', 8),
+  noWallet: new ConnectorError('no wallet running', 9),
+};
```
dexturr, line 22: Creating the errors here means we always throw the same instance of the error. Might be better to store the message + code and create the errors before throwing. I've never seen this before, not sure what the result is possibly incorrect stack?

libs/wallet/src/chains.ts
```diff
@@ -0,0 +1,29 @@
+export type Chain = {
+  id: string;
+  testnet: boolean;
+  name: string;
+};
+
+export const mainnet = {
+  id: 'vega-mainnet-0011',
+  testnet: true,
+  name: 'Fairground',
```
asiaznik, line 10: 🤔 this should be mainnet, shouldn't it?

libs/wallet/src/errors.ts
mattrussell36: Addressed this in this commit: https://github.com/vegaprotocol/frontend-monorepo/pull/5815/commits/99e5efbd88ca92f4b10774b46f72cf4a04b5b870

libs/wallet/src/connectors/snap-connector.ts
mattrussell36: Addressed in https://github.com/vegaprotocol/frontend-monorepo/pull/5815/commits/e0192f9780f78108cd4d09ee16557ecac93cc512

apps/trading/components/navbar/navbar.tsx
```diff
@@ -88,7 +84,7 @@ export const Navbar = ({ theme = 'system' }: { theme?: Theme }) => {
         </div>
         <NavbarMobileButton
           onClick={() => {
-            if (isConnected) {
+            if (status === 'connected') {
```
MadalinaRaicu, line 87: (non-blocking): it'd be great to have here some sort of status type to prevent any typos

apps/trading/components/vega-wallet-connect-button/vega-wallet-connect-button.tsx
```diff
@@ -33,46 +32,50 @@ export const VegaWalletConnectButton = ({
...
+
   const walletInstalled = isBrowserWalletInstalled();
-  const activeKey = useMemo(() => {
-    return pubKeys?.find((pk) => pk.publicKey === pubKey);
-  }, [pubKey, pubKeys]);
 
-  if (isConnected && pubKeys) {
+  const activeKey = pubKeys?.find((pk) => pk.publicKey === pubKey);
+
+  if (status === 'connected') {
```
MadalinaRaicu, line 52: same here

apps/trading/components/vega-wallet-connect-button/vega-wallet-connect-button.tsx
```diff
@@ -86,7 +89,7 @@ export const VegaWalletConnectButton = ({
         >
           <div className="min-w-[340px]" data-testid="keypair-list">
             <TradingDropdownRadioGroup
-              value={pubKey}
+              value={pubKey || undefined}
```
MadalinaRaicu, line 92: can pubkey be null here?

apps/trading/components/welcome-dialog/welcome-dialog.tsx
```diff
@@ -39,46 +40,132 @@ export const WelcomeDialog = () => {
...
+  }
+};
+
+const OnboardConnectionOptions = ({ onConnect }: { onConnect: () => void }) => {
+  const t = useT();
+  const error = useWallet((store) => store.error);
+  const { connect, connectors } = useConnect();
+  const status = useWallet((store) => store.status);
+
+  if (status === 'disconnected') {
```
MadalinaRaicu, line 116: same here

libs/deal-ticket/src/components/deal-ticket/margin-mode-selector.tsx
```diff
@@ -39,7 +40,7 @@ export const MarginChange = ({
   marginMode,
   marginFactor,
 }: {
-  partyId: string | null;
+  partyId: string | undefined;
```
MadalinaRaicu, line 43: (non-blocking):`string | null` made sense however `partyId: string | undefined` is the same with `partyId?: string`

MadalinaRaicu (approved):

apps/trading/components/navbar/navbar.tsx
mattrussell36: I'm not 100% convinced we should always use enums for this sort of thing. I quite like not to have to import an enum for a simple type comparison. The string will autocomplete because status is correctly typed. I think @dexturr feels strongly about using enums and I'm happy to make the change

apps/trading/components/vega-wallet-connect-button/vega-wallet-connect-button.tsx
mattrussell36: No pubKey can never be null. The default state for an unconnected wallet is to have pubKey as undefined. 

This made it much easier to work with a persisted zustand store, when I needed to partialize what was put into local storage

libs/deal-ticket/src/components/deal-ticket/margin-mode-selector.tsx
mattrussell36: This change is only to support the fact that by default pubKey is now undefined and not null.

PR #5816, opened 2024-02-19
fix(explorer): sizedelta in ammends displayed incorrectly'

### LINKED ISSUE #5774: Order amendment bug in Explorer
Core now have the ability to set a new value in the order amendment rather than supply a delta.

### Explorer issue 

There is an issue on the Explorer transaction view because it displays that delta as the "new size" which it is not.  So we should fix that to show both the new size (whatever that is called in the new core build) and / or the sizeDelta if provided.

image

I am not clear if BOTH can be provided - and if so is that validated - or if it is one or the other.

## Tasks

- [x] Check how it works with Core
- [x] Fix Explorer

Old method with a delta

image

edd:
- **fix(explorer): fix display of amends with sizedelta**

# Related issues 🔗

Closes #5774

# Description ℹ️

`sizeDelta` displayed as the new price, rather than a change to the price. This change changes the label, and displays the actual size more clearly.

# Demo 📺

Screenshot 2024-02-19 at 11 48 54

# To test
- You need an order 
- Then you need to amend the order using the sizeDelta field
- Example: https://explorer.fairground.wtf/txs/0x2ABAE5BCDF7F45BE24CD9599E77AEEED0D281B1F0548429D9041FDC1F15670A1 was size `7000`, then ammended using sizeDelta `-2000` to size `5000`

### DIFF
--- a/apps/explorer/src/app/components/order-details/amend-order-details.spec.tsx
+++ b/apps/explorer/src/app/components/order-details/amend-order-details.spec.tsx
@@ -175,6 +175,7 @@ describe('Amend order details', () => {
 
     const res = renderExistingAmend('123', 1, amend);
     expect(await res.findByText('New size')).toBeInTheDocument();
+    expect(await res.findByText('Size ±')).toBeInTheDocument();
   });
 
   it('Renders Reference if provided', async () => {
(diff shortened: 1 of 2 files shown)


daro-maj:
new price is updated but size not 
image

mattrussell36 (approved):

bwallacee (approved):

PR #5817, opened 2024-02-19
feat(explorer): add epoch to block pages

### LINKED ISSUE #5550: Add epoch to block/:id page
## The Chore

https://docs.vega.xyz/mainnet/api/rest/data-v2/trading-data-service-get-epoch#get-epoch will give us an epoch for a block height. That would be useful info

## Tasks

- [ ] Add it in a row on the block by ID page

edd:
- **feat(explorer): add epoch by block query and use it in block by id and shareddetails**

# Related issues 🔗

Closes #5550

# Description ℹ️

Adds a component and query to get the epoch number for a block. This PR is marked as draft because the query returns an error in 0.73.x, 0.74.1 & 0.74.2, which is every network at the moment. Some tidying will be done with there's a working environment

# Demo 📺
## block/:id page: epoch rendered below time
Screenshot 2024-03-04 at 12 38 50

## txs/:id page: epoch shown on block page
Screenshot 2024-03-04 at 12 38 44

## Hovering an epoch gets epoch details
Screenshot 2024-03-04 at 12 38 54
 
Bonus (required during development): transfer types are now differentiated in TX overview
Screenshot 2024-03-04 at 12 12 54
Screenshot 2024-03-04 at 12 13 30

### DIFF
--- a/apps/explorer/src/app/components/epoch-overview/epoch.tsx
+++ b/apps/explorer/src/app/components/epoch-overview/epoch.tsx
@@ -8,12 +8,14 @@ import EpochMissingOverview from './epoch-missing';
 import { Icon, Tooltip } from '@vegaprotocol/ui-toolkit';
 import type { IconProps } from '@vegaprotocol/ui-toolkit';
 import isPast from 'date-fns/isPast';
+import { EpochSymbol } from '../links/block-link/block-link';
 
 const borderClass =
   'border-solid border-2 border-vega-dark-200 border-collapse';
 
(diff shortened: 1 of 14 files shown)


bwallacee (approved):
Nit picky, but the E is pretty small, and maybe a tooltip or something to explain what the e is would be a nice to have. Non blocking. Approved:
image

mattrussell36 (approved):

ze97286 (approved):

PR #5819, opened 2024-02-20
feat(trading): key aliases

### LINKED ISSUE #5717: Set party profile alias in Console somewhere
## The Chore

To support team leaderboards you can now set a party alias for your key.

We need somewhere that we can actually set this in Console....not sure where.

Maybe in the key dropdown in the top right?

Maybe directly in the team profile page?

Maybe we introduce another "individual banner" which records your individual stats if you have them (though we dont have leaderboards for individuals yet).

## Tasks

- [ ] Move the copy button to the far right 
- [x] Add a pen button to each row in the key dropdown to allow you to edit
- [x] If clicked it can just bring up a dialogue allowing you to edit
- [ ] Submitting should send a TX to `UpdatePartyProfile`
- [x] Show the party profile in the dropdown
- [ ] Look at multiline options and some labelling to make sure that the wallet name / private name is distinguished from the public name

image

```
message UpdatePartyProfile {
  // Alias given to the party. It must be unique network-wide.
  string alias = 1;
  // Freeform data to associate to the party.
  // Support a maximum of 10 entries.
  repeated vega.Metadata metadata = 2;
}
```

mattrussell36:
# Related issues 🔗

Closes #5717

# Description ℹ️

- Adds key aliases to key list dropdown
- Adds a dialog for updating key aliases
- Hides key in dropdown trigger

# Demo 📺

Screenshot 2024-03-04 at 15 48 12

Screenshot 2024-03-04 at 15 48 35


bwallacee:
@mattrussell36 because you get the hover affect over the whole key row it semi feels like the pen icon isn't actually clickable if that makes sense?
image

also on mobile i'm not able to add alias's etc, 

(Mobile in general looks like it needs a bit of love)
image

There is no success message state:
image

there is also no failure message state.

bwallacee:
@mattrussell36 I have to refresh the page in order to see my new alias after updating it:
image

For another ticket, but it might be nice for these aliases to exist elsewhere e.g. transfer form:
image

Non blocking as I think it would be good to get this in, so maybe needs another ticket:
image
Theres no way to set alias through mobile view.  Also I think we should use white text for the key names

bwallacee (approved):

EVODelavega (approved):

PR #5830, opened 2024-02-20
fix(governance): market name component not showing correct instrument code

### LINKED ISSUE #5821: Batch governance proposal summary / titles
### Proposals in batch

For updates we are allowing a copy of the market ID and giving a link to console ... for resume market we arent .. we should follow same pattern for both market updates and market lifecycle governance.

image

- [x] Ideally we would look up the market name from the ID and show this here rather than the ID
- [x] Then we would show the copy button which would show a tooltip "Copy market ID" since users are far more likely to want to copy an ID than a name
- [x] Then we can include the link to console as usual
- [x] We should follow that pattern for both market updates and market lifecycle governance 

### Proposal details 

- [ ] As above we're just showiing the proposal type ... we need the full title of the update here so you can tell what you're looking at.  I would expect this to match the top of the page

image

## Tasks

- [ ]
- [ ]

https://governance.stagnet1.vega.rocks/proposals/e2bd2f75e07001b19aae12ce9b7dfb0613802c85264a3ff61fb347c23e0a0dce

mattrussell36:
# Related issues 🔗

Closes #5821

# Description ℹ️

Instrument does not have an ID so it needs a custom key field so it can be looked up in the cache

### DIFF
--- a/apps/governance/src/app.tsx
+++ b/apps/governance/src/app.tsx
@@ -62,6 +62,9 @@ const cache: InMemoryCacheConfig = {
     Account: {
       keyFields: false,
     },
+    Instrument: {
+      keyFields: ['code'],
+    },
     Delegation: {
       keyFields: false,
       // Only get full updates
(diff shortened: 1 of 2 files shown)


daro-maj (approved):

edd (approved):
💅

PR #5849, opened 2024-02-22
fix(trading): banner for resume trading

### LINKED ISSUE #5810: Proposal to resume a market in suspended mode shows termination banner
## Description

When a market is suspended by governance and a proposal is made to resume it, the market termination banner shows

## Steps to Reproduce

1. Suspend a market by governance
2. Make a proposal to resume
3. Visit the market page
4. See the termination banner shown

## Expected behavior

The termination banner should not show. A banner with different content, including the resume (enactment) date should show

## Screenshots

image

## Device details

All

## Additional context

https://vegaprotocol.slack.com/archives/C8WS7M50E/p1708104591348549

MadalinaRaicu:
# Related issues 🔗

Closes #5810 

# Description ℹ️

Update banner for markets that will be resumed. 

# Demo 📺

Screenshot 2024-02-26 at 10 14 55

# Technical 👨‍🔧


edd (approved):

apps/trading/components/market-banner/market-update-state-banner.tsx
```diff
@@ -24,6 +24,16 @@ export const MarketUpdateStateBanner = ({
   const t = useT();
   const governanceLink = useLinks(DApp.Governance);
 
+  const openTradingProposals = sortBy(
+    proposals.filter(
+      (p) =>
+        p.state === ProposalState.STATE_OPEN &&
+        p.terms.change.__typename === 'UpdateMarketState' &&
+        market.state === MarketState.STATE_SUSPENDED_VIA_GOVERNANCE
```
asiaznik, line 32: In order to get the actual resuming proposals it'd need to have additional condition:
```suggestion
        market.state === MarketState.STATE_SUSPENDED_VIA_GOVERNANCE &&
        p.terms.change.updateType === MarketUpdateType.MARKET_STATE_UPDATE_TYPE_RESUME
```

asiaznik: Additionally I'd recheck if `useUpdateMarketStateProposals` which is used one level up from here is returning the actual proposal data or the whole **batch** proposal (if the update market state is inside a batch).

If it is batch then we'd need to change the hook to handle proposal picking from `proposal.proposals` which is an array of proposals (they have its own id and `batchId` which should be used for the link in this case)

MadalinaRaicu: from what I can see `useUpdateMarketStateProposals` returns `MarketViewProposalFieldsFragment[]`  which is the actual proposal data and there's no `batchId`

MadalinaRaicu: I have updated the queries to flatten the proposals

libs/proposals/src/lib/proposals-data-provider/proposals-data-provider.tsx
```diff
@@ -104,7 +115,25 @@ export const update: Update<
...
+    if (edge?.proposalNode) {
+      if (edge.proposalNode.__typename === 'Proposal') {
+        proposals.push(edge.proposalNode as MarketViewProposalFieldsFragment);
+      } else if (
+        edge.proposalNode.__typename === 'BatchProposal' &&
+        edge.proposalNode.subProposals
+      ) {
+        proposals.push(
+          ...(edge.proposalNode
+            .subProposals as MarketViewProposalFieldsFragment[])
```
mattrussell36, line 130: Rather than type casting here, which could be problematic why not defined the type of the data provider correctly? Right now when used in a component you will expect to only receive MarketViewProposalFieldsFragment, which is not correct, as you might actually have a sub proposal.

What I would suggest is to adjust the fragments slight, and then chagne the type of the data provider and helper functions. Something like this:

```
fragment SubProposal on ProposalDetail {
  ...
}

fragment MarketViewProposalFields {
  ...
}

query MarketViewProposals(
  $proposalType: ProposalType
  $inState: ProposalState
) {
  proposalsConnection(proposalType: $proposalType, inState: $inState) {
    edges {
      proposalNode {
        ... on BatchProposal {
          id
          subProposals {
             ...SubProposal
          }
        }
        ... on Proposal {
          ...MarketViewProposalFields
        }
      }
    }
  }
}

```

Then in data dataprovider we would need to update the types to reflect a combined array of 'normal' proposals and sub proposals
```
type Proposals = Array<MarketViewProposalFieldsFragment | SubProposalFieldsFragment>;

export const marketViewProposalsDataProvider = makeDataProvider<
  MarketViewProposalsQuery,
  Proposals,
  MarketViewLiveProposalsSubscription,
  MarketViewProposalFieldsFragment,
  MarketViewProposalsQueryVariables,
  MarketViewLiveProposalsSubscriptionVariables
>({
  query: MarketViewProposalsDocument,
  subscriptionQuery: MarketViewLiveProposalsDocument,
  update,
  getDelta: (subscriptionData: MarketViewLiveProposalsSubscription) =>
    subscriptionData.proposals,
  getData: getMarketProposalsData,
  getSubscriptionVariables: () => subscriptionVariables,
});

```

mattrussell36 (changes requested):

libs/proposals/src/lib/proposals-data-provider/proposals-data-provider.tsx
MadalinaRaicu: I have updated

MadalinaRaicu:
> Looks like issue still exist image

I have fixed this issue @daro-maj  if you could please review again

guoguojin (approved):

mattrussell36 (approved):

PR #5902, opened 2024-03-04
feat(trading): take profit and stop loss

### LINKED ISSUE #5091: Add TP / SL to the deal ticket for market / limit / stop orders
## Story

As a user
I want to be able to attach a TP / SL to another order that I am making
So that if the original order trades the TP / SL will also be set up

## Acceptance Criteria

- [x] I can add a TP / SL to a market order
- [x] I can add a TP / SL to a limit order
- [x] I have a tick box which when clicked causes a section to expand on the deal ticket
- [x] The expanded section allows me to set a price for a take profit and / or stop loss
- [x] When I submit the order BOTH my main order AND a STOP order is submitted as a batch
  - [x] Order type = stop market
  - [x] Order ref = some derived text string indicating that this is a position level stop order (`partyId` + `timeStamp` + `uniqueRef`)
  - [x] Time in force = FoK
  - [x] Size = the full size of the original order
  - [x] Price = Use the value set in the input box
  - [x] For direction it needs to be implied
    - [x] If position is LONG
      - [x] TP is SHORT and trigger is RISES ABOVE 
      - [x] SL is SHORT and trigger is FALLS BELOW
    - [x] If position is SHORT
      - [x] TP is LONG and trigger is FALLS BELOW
      - [x] SL is LONG and trigger is RISES ABOVE
- [x] If I have provided both a TP AND a SL it should be an OCO
- [x] Consider if we can put a reference on the orders that links the two orders together future use, we'd have to define a string that could uniquely link the two.

## Sketch

Image

There is no special handling or cleanup on the stop orders ... they just go into the stop order table as usual. 

If the original order was a limit order and it never actually trades that means there may be stop orders sitting around ... this is fine.

MadalinaRaicu:
# Related issues 🔗

Closes #5091 

# Description ℹ️

Add TP / SL to the deal ticket for market / limit 

# Demo 📺

image

# Technical 👨‍🔧


libs/deal-ticket/src/components/deal-ticket/deal-ticket-price-tp-sl.tsx
```diff
@@ -0,0 +1,172 @@
...
+}
+
+export const DealTicketPriceTakeProfitStopLoss = ({
+  control,
+  market,
+  takeProfitError,
+  stopLossError,
+  quoteName,
+}: DealTicketPriceTakeProfitStopLossProps) => {
+  const t = useT();
```
bglownia, line 31: It looks like none of labels used in DealTicketPriceTakeProfitStopLoss were added to i18n/src/locales/en/deal-ticket.json

libs/deal-ticket/src/components/deal-ticket/deal-ticket.tsx
```diff
@@ -674,40 +691,63 @@ export const DealTicket = ({
...
-                    <p>
-                      {t(
-                        'ICEBERG_TOOLTIP',
-                        'Trade only a fraction of the order size at once. After the peak size of the order has traded, the size is reset. This is repeated until the order is cancelled, expires, or its full volume trades away. For example, an iceberg order with a size of 1000 and a peak size of 100 will effectively be split into 10 orders with a size of 100 each. Note that the full volume of the order is not hidden and is still reflected in the order book.'
-                      )}{' '}
-                      <ExternalLink href={DocsLinks?.ICEBERG_ORDERS}>
-                        {t('Find out more')}
-                      </ExternalLink>{' '}
-                    </p>
+                    <p>{t('TP_SL_TOOLTIP', 'Take profit / Stop loss')}</p>
```
bglownia, line 734: 'TP_SL_TOOLTIP' not added to i18n/src/locales/en/deal-ticket.json

libs/deal-ticket/src/components/deal-ticket/deal-ticket.tsx
```diff
@@ -674,40 +691,63 @@ export const DealTicket = ({
...
                   <div>
                     <Checkbox
-                      name="iceberg"
+                      name="tpSl"
                       checked={field.value}
                       onCheckedChange={field.onChange}
-                      disabled={disableIcebergCheckbox}
-                      label={t('Iceberg')}
+                      disabled={false}
+                      label={t('TP / SL')}
```
bglownia, line 743: 'TP / SL' not added to i18n/src/locales/en/deal-ticket.json

MadalinaRaicu: updated: added translations

libs/deal-ticket/src/components/deal-ticket/deal-ticket.tsx
MadalinaRaicu: updated: added translations

bwallacee (commented):
I think we need some unit and e2e tests for this.

MadalinaRaicu:
> I think we need some unit and e2e tests for this.

I have added unit tests for the order submission body.

bwallacee (approved):

mattrussell36 (approved):

PR #5909, opened 2024-03-04
fix(trading): pick the best node to connect, slow notification notification, other node picking improvements

### LINKED ISSUE #5759: Show Node switcher on Loading screen of console
### Fix the loading spinner thingy

- [ ] It seems one of these issues is caused specifically by the app (both Governance & Console) thinking that it has a token from a wallet connection and then the wallet wants the connection to be approved via the wallet being unlocked to load the page

### Show node selector on timeout

- [ ] If it is NOT the above problem and it is just datanode taking ages so console is taking too long to load, show the option to open the manual node switcher and manually choose a node.

After x time (3 seconds maybe)
Show a message 

> “It looks like you’re connection is slow, try switching to another node… ”

with a CTA to open the manual node switcher.

Can we also detect the conditions of that Lovali node?  Ie. if its just "no data" can we detect that and prompt the node switcher too?

### Add copy button to node selector

- [ ] Its infuriatingly hard to copy a node address from this screen ... can we add a copy button

Image

asiaznik:
# Related issues 🔗

Closes #5759

# Description ℹ️

* the app already picks the best node (on reload), but this time it will pick the best node if the stored url (previously connected node) is no longer healthy
* adds "it appears your connection is slow" notification to the loader
* adds initialisation error screen if it is really broken
* adds copy buttons to the node switcher

# Demo 📺

Screenshot 2024-03-04 at 17 48 29
Screenshot 2024-03-04 at 17 48 56
Screenshot 2024-03-04 at 17 49 05

# Technical 👨‍🔧

N/A


bwallacee (approved):

apps/trading/components/bootstrapper/bootstrapper.tsx
```diff
@@ -1,63 +1,123 @@
 import type { InMemoryCacheConfig } from '@apollo/client';
 import {
-  AppFailure,
```
mattrussell36, line 3: AppFailure isn't used anywhere anmore

libs/ui-toolkit/src/components/trading-radio-group/radio-group.tsx
```diff
@@ -86,25 +88,36 @@ export const TradingRadio = ({
...
+            disabled
+              ? 'text-vega-clight-200 dark:text-vega-cdark-200'
+              : 'cursor-pointer'
+          }
+        >
+          {label}
+        </span>
+      </label>
+      {value && value !== 'custom' && (
+        <span className="text-muted">
```
mattrussell36, line 113: I wouldn't add the copy functionality. This should be passed in as children/render prop

mattrussell36 (changes requested):

apps/trading/components/bootstrapper/bootstrapper.tsx
```diff
@@ -1,63 +1,123 @@
...
+    </Splash>
+  );
+};
+
+const Loading = () => {
+  const [showSlowNotification, setShowSlowNotification] = useState(false);
+  const t = useT();
+  const setNodeSwitcher = useNodeSwitcherStore((store) => store.setDialogOpen);
+  useEffect(() => {
+    setTimeout(() => {
```
mattrussell36, line 45: Should probably clear this timeout with a clean up function

libs/environment/src/hooks/use-environment.ts
```diff
@@ -600,35 +618,31 @@ export const useEnvironment = create<EnvStore>()((set, get) => ({
...
-      if (isValidUrl(storedUrl)) {
-        set({ VEGA_URL: storedUrl, status: 'success' });
-        return;
-      } else {
-        LocalStorage.removeItem(STORAGE_KEY);
-      }
-    }
-
-    // VEGA_URL env var is set and is a valid url no need to proceed
-    if (state.VEGA_URL) {
```
mattrussell36, line 627: @asiaznik @edd  gov and explorer have set VEGA_URLs at build. The idea here was that if VEGA_URL is set at build time we just use that node immediately, without checking other nodes.

So this change I believe will make gov and explorer start searching for nodes at startup. Which is probably fine, but just wanted to highlight this

edd: @mattrussell36 this sounds fine, thanks for highlighting it

asiaznik: It's a delibarate feature :feelsgood:

mattrussell36 (approved):

PR #5959, opened 2024-03-11
fix(trading,types): fix default build and type generation commands

mattrussell36:
# Related issues 🔗

N/A

# Description ℹ️

- Re adds the types:generate command
- Removes root project.json and includedScripts


edd (approved):
Taking your word for this one (rather than actually running it)

bwallacee (approved):

PR #5980, opened 2024-03-12
fix(trading): key details panel throwing error on batch created market

mattrussell36:
# Related issues 🔗

N/A

# Description ℹ️

Updates key details panel to account for batch proposals


daro-maj:
works good 
image

PR #5986, opened 2024-03-12
feat(explorer): add market specific oracle page

edd:
# Related issues 🔗

Closes #5689

# Description ℹ️

Adds a page underneath a market that lists the orcles specificlly for that market

# Demo 📺


apps/explorer/src/app/components/oracle-table/index.tsx
```diff
@@ -0,0 +1,159 @@
+import compact from 'lodash/compact';
```
edd, line 1: Note: This component is not new, it has just been pulled out of `Oracles /home / index.tsx` so it can be reused

bwallacee (approved):

mattrussell36 (approved):

PR #5987, opened 2024-03-12
feat(trading): update validation and rules on take profit and stop loss

### LINKED ISSUE #5966: TP/SL unable to leave TP blank
When using TP/SL I am unable to leave TP blank and just use a SL.
image

MadalinaRaicu:
# Related issues 🔗

Closes #5966 

# Description ℹ️

- [x] Remove rules on take profit and stop loss.
- [x] Adds feature flag - disabled on mainnet

# Demo 📺
Prevent this: 
image

Now:
Screenshot 2024-03-12 at 17 16 42

# Technical 👨‍🔧


asiaznik:
Why? This should also be validated against the same rules, shouldn't it? Perhaps allowing empty values would be better rather that removing the whole validation?

JonRay15:
Can we also add a feature flag to hide the stop loss toggle button?

bwallacee (approved):

PR #6013, opened 2024-03-14
feat(emblem): create example emblem component

edd:
# Description ℹ️
Creates an example component for embedding asset icons in a vega frontend. This is for *discussion* at first.

# Demo 📺

With a simple added component:
Screenshot 2024-03-14 at 16 23 59

You get:
Screenshot 2024-03-14 at 16 23 29
 
# Considerations
- You need to know the vega chain ID you're currently looking at to use  the`EmblemByAsset` component. It defaults to `mainnet` (i.e. `vega-mainnet-0011`)
- You need to know the external chain ID you're currently looking at to use the `EmblemByContract` component. It defaults to `Ethereum mainnet` (i.e. `1`)
- In future this component might embed the logos, and only fall back to requesting it from `icon.vega.xyz`. Right now it always uses the remote URL.
- Sizes are hardcoded, but match the Tailwind size CSS properties. Overriding the classes will not change the width, but given that we're a single page app I think this is fine.

### DIFF
--- a/libs/emblem/src/components/emblem.tsx
+++ b/libs/emblem/src/components/emblem.tsx
@@ -1,0 +1,101 @@
+import type { ImgHTMLAttributes } from 'react';
+import {
+  DEFAULT_CHAIN,
+  DEFAULT_VEGA_CHAIN,
+  FALLBACK_URL,
+  FILENAME,
+  URL_BASE,
+} from '../config/index';
+
+export type ImgProps = ImgHTMLAttributes<HTMLImageElement>;
+
+export type EmblemByContractProps = {
+  contract: string;
+  chainId?: string;
+  asset?: never;
+};
+
+export type EmblemByAssetProps = {
+  asset: string;
+  vegaChain?: string;
+  contract?: never;
+};
+
+export type EmblemProps = ImgProps &
+  (EmblemByAssetProps | EmblemByContractProps);
+
+/**
+ * Type guard for generic Emblem component, which ends up rendering either an EmblemByContract
+ * or EmblemByAsset depending on the arguments
+ */
+export function isEmblemByAsset(
+  args: EmblemByAssetProps | EmblemByContractProps
+): args is EmblemByAssetProps {
+  return (args as EmblemByAssetProps).asset !== undefined;
+}
+
+/**
+ * A generic component that will render an emblem for a Vega asset or a contract, depending on the props
+ * @returns React.Node
+ */
+export function Emblem(props: EmblemProps) {
+  if (isEmblemByAsset(props)) {
+    return <EmblemByAsset {...props} />;
+  } else {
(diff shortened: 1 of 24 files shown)


libs/emblem/tsconfig.json
```diff
@@ -0,0 +1,22 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "module": "commonjs",
```
mattrussell36, line 4: We need "jsx": "react-jsx"

libs/emblem/src/components/emblem.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+export type EmblemProps = EmblemByAssetProps | EmblemByContractProps;
+
+/**
+ * A generic component that will render an emblem for a Vega asset or a contract, depending on the props
+ * @returns React.Node
+ */
+export function Emblem(props: EmblemProps) {
+  if ('asset' in props) {
+    const emblemByAssetProps = props as EmblemByAssetProps;
```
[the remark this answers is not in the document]
edd: Aha, that last bit was what I couldn't work out - thank you!

tsconfig.base.json
```diff
@@ -61,7 +61,8 @@
         "libs/wallet-react/src/testing.ts"
       ],
       "@vegaprotocol/web3": ["libs/web3/src/index.ts"],
-      "@vegaprotocol/withdraws": ["libs/withdraws/src/index.ts"]
+      "@vegaprotocol/withdraws": ["libs/withdraws/src/index.ts"],
+      "emblem": ["libs/emblem/src/index.ts"]
```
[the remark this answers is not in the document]
edd: No, will fix this

libs/emblem/tsconfig.json
edd: Good catch

libs/emblem/src/components/emblem.tsx
```diff
@@ -0,0 +1,101 @@
...
+ * @returns React.Node
+ */
+export function EmblemByAsset(p: EmblemByAssetProps) {
+  const url = `${URL_BASE}/vega/${
+    p.vegaChain ? p.vegaChain : DEFAULT_VEGA_CHAIN
+  }/asset/${p.asset}/${FILENAME}`;
+  return <EmblemBase url={url} {...p} />;
+}
+
+export interface EmblemBaseProps extends ImgProps {
```
mattrussell36, line 76: Very minor.. but I feel like this could just be be ImgProps, then you'll get alt and we could just use `src` instead of `url`

edd: True!

mattrussell36 (approved):

asiaznik (approved):

PR #6057, opened 2024-03-20
feat(deal-ticket): use order price if markPrice is not available in useMaxSize

### LINKED ISSUE #6047: Size slider does not work for markets in price auction
## Description

Using the size slider does not working for markets in price auction, nothing happens.

## Steps to Reproduce

1. Go to a market in price auction
2. Try to use size slider 
3. See error

## Expected behavior

Slider should work for price auction as well, I have chosen a price on limit order.

## Screenshots

https://github.com/vegaprotocol/frontend-monorepo/assets/100769129/7cc54c8b-230e-4d37-898b-5803737d95f0

bglownia:
# Related issues 🔗

Closes #6047

# Description ℹ️

If market is in action markPrice is not available. In that case when calculating max order size use order price for cross margin limit orders

### DIFF
--- a/libs/deal-ticket/src/hooks/use-max-size.ts
+++ b/libs/deal-ticket/src/hooks/use-max-size.ts
@@ -13,7 +13,7 @@ interface UseMaxSizeProps {
   marginAccountBalance: string;
   marginFactor?: string;
   marginMode?: MarginMode;
-  marketPrice?: string;
+  markPrice?: string;
   openVolume: string;
   positionDecimalPlaces: number;
   price?: string;
@@ -43,7 +43,7 @@ export const useMaxSize = ({
   activeOrders,
   riskFactors,
   scalingFactors,
-  marketPrice,
+  markPrice,
   marketIsInAuction,
 }: UseMaxSizeProps) =>
   useMemo(() => {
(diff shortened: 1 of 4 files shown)


libs/deal-ticket/src/hooks/use-max-size.ts
```diff
@@ -66,10 +66,12 @@ export const useMaxSize = ({
         .div(marginFactor)
         .div(toBigNum(price, decimalPlaces));
     } else {
+      const effectivePrice =
+        type === OrderType.TYPE_LIMIT ? marketPrice || price : marketPrice;
```
mattrussell36, line 70: Something seems wrong here. The price value passed into the hook is already derived using order type in getDerivedPrice?

mattrussell36 (changes requested):

libs/deal-ticket/src/hooks/use-max-size.ts
mattrussell36: So I think just using `price` is correct here, and there is no need to check order type

bglownia: In cross mode orders mark price is used to calculate required margin for both limit and market orders. If market is in auction there is no market price and estimate is calculated using price from deal ticket.

Use always marketPrice until order type is limit and there is no marketPrice.

mattrussell36 (approved):

daro-maj (approved):

PR #6070, opened 2024-03-21
fix(positions): rework of the liquidation tooltip

### LINKED ISSUE #6019: Rework liquidation notification in Console
## Story

As a user
I want to know i've been closed out, at what price, and for what loss
So that I dont label Vega a scam and stop trading there 

## Tasks

- [x] Check the latest trade on the market for that key
- [x] If it is of type: `TYPE_NETWORK_CLOSE_OUT_BAD` then you have been closed out
- [x] Do not look at the position API for positionStatus at all anymore because it is rubbish
- [x] If last trade shows a close out then 
  - [x] Show the tooltip indicating you have been closed out
  - [x] The price from that trade can be taken as the `closeoutPrice` 
  - [x] Display message below

NB. We have the tradestream subscription to monitor the last trade on that market for this pubkey.  This logic should be built off that.

## Tasks

- [ ] UX (if needed)
- [ ] Design (if needed)
- [ ] Implementation
- [ ] Code review
- [ ] QA review

### Current view 

image

```
Your position was closed. 

You did not have enough `settlementAsset` to meet the margin required for your position, so it was liquidated by the network at `closeoutPrice`.

Read more about position resolution(link)
```

asiaznik:
# Related issues 🔗

Closes #6019 

# Description ℹ️

* removed dependency on positions status
* checks the latest trades for specific market and party to deduce if a positions has been liquidated
* changes words
* fixes randomly failing tests

# Demo 📺

image

# Technical 👨‍🔧

N/A


bwallacee:
@asiaznik I have to refresh the page in order to see the warning icon.

Also I wonder if we should change the 0 from being green:
image

bwallacee (approved):

libs/positions/src/hooks/use-latest-trades.ts
```diff
@@ -0,0 +1,78 @@
...
+        ...updatedData[index],
+      };
+    } else if (!data?.length || tradeUpdate.createdAt >= data[0].createdAt) {
+      updatedData.unshift(omit(tradeUpdate, '__typename') as LatestTrades);
+    }
+  });
+  return updatedData.slice(0, MAX_TRADES);
+};
+
+const latestTradesDataProvider = makeDataProvider<
```
mattrussell36, line 59: Why not use the trades data provider we already have? Why not add partyIds/marketIds and pagination as variables for the data provider?

Then we can only get the single last trade for the user?

mattrussell36 (changes requested):

mattrussell36 (approved):

PR #6113, opened 2024-03-26
feat(trading): show oracle spec in index price stats

### LINKED ISSUE #5942: Tooltip on Index price to link to explorer for that oracle
## The Chore

## Tasks

- [ ]
- [ ]

MadalinaRaicu:
# Related issues 🔗

Closes #5942 

# Description ℹ️

Show Oracle spec in index price stats. 

# Demo 📺

image

# Technical 👨‍🔧


libs/orders/src/lib/components/order-data-provider/order-data-provider.ts
```diff
@@ -131,7 +131,8 @@ export const update = <T extends Omit<OrderFieldsFragment, 'market'> & Cursor>(
 ): T[] => {
   const updatedData = data ? [...data] : ([] as T[]);
   delta.forEach((orderUpdate) => {
-    const index = data?.findIndex((order) => order.id === orderUpdate.id) ?? -1;
+    const index =
+      data?.findIndex((order) => order?.id === orderUpdate.id) ?? -1;
```
mattrussell36, line 135: Why these changes the seem completely unrelated to the rest of this PR? Are the types incorrect here and order can be undefined?

MadalinaRaicu: weirdly while I was testing i got some exceptions on stagnet1 and order was undefined

MadalinaRaicu: image

MadalinaRaicu: added that "?." since it is an extra check that does not hurt and prevents runtime exceptions... but yeah, spotted that on staging only

daro-maj (approved):

libs/orders/src/lib/components/order-data-provider/order-data-provider.ts
mattrussell36: Its very annoying to have to clutter the code with syntax to account for poorly typed objects. Its fine for now. @bglownia we probably want to take a look at this

mattrussell36 (approved):

PR #6122, opened 2024-03-27
fix(trading): recurring governance transfers after 0.75.1

### LINKED ISSUE #6048: Add governance transfers to rewards cards
## The Chore

Filter for reward cards needs to include governance transfers

These can be identified by filter `RecurringGoveranceTransfer`

## Tasks

- [x] update queries and conditions

MadalinaRaicu:
# Related issues 🔗

Closes #6048 

# Description ℹ️

Display recurring governance transfers. 

# Demo 📺
https://governance.fairground.wtf/proposals/04e9a7d0b20dc723d497930afdf1d776fb8249069e4c6d724d14a97c10bd1283
Screenshot 2024-03-27 at 12 57 10

# Technical 👨‍🔧


apps/trading/lib/hooks/Rewards.graphql
```diff
@@ -75,6 +75,29 @@ query ActiveRewards(
           timestamp
           gameId
           kind {
+            ... on RecurringGovernanceTransfer {
+              startEpoch
```
mattrussell36, line 79: Would probably be better to make a fragment for these fields. It will help ensure consistency between RecurringGovernanceTransfer and RecurringTransfer and also create a type for us that can represent both

daro-maj (approved):

apps/trading/lib/hooks/Rewards.graphql
MadalinaRaicu: I added some TransferKind fields. GraphQL won't let you make a fragment between 2 types unless you specify into the schema that those types are sharing the same properties.

mattrussell36: Ah yeah I see there are different types in the schema, even though the fields are (almost) entirely the same

wwestgarth (approved):

mattrussell36 (approved):

PR #6124, opened 2024-03-27
feat(trading): show current epoch progress bar on rewards page trading

### LINKED ISSUE #5813: Add an epoch counter to the reward page
## The Chore

I have to keep going to the Governance site reward page to see what epoch we're in and when the next one ends. 

This is rubbish.

We need to find a home for it on the rewards / competitions page somewhere.

We have a component in governance we could steal 

Image

Maybe play with finding a home for it

## Tasks

- [ ]
- [ ]

MadalinaRaicu:
# Related issues 🔗

Closes #5813 

# Description ℹ️

- [x] Add an epoch counter to the reward page
- [x] Re-use component from governance, added to ui-toolkit

# Demo 📺

Screenshot 2024-04-04 at 11 16 13

# Technical 👨‍🔧


mattrussell36:
I'm not sure this is the best spot for it. I think it needs to be more prominent, perhaps right under the main page header with 1/3 the width so it aligns with the grid below

MadalinaRaicu:
> @MadalinaRaicu I'm not sure this is the best spot for it. I think it needs to be more prominent, perhaps right under the main page header with 1/3 the width so it aligns with the grid below

This is how it looks with full width and 1/3 width, let me know what you prefer. 

Screenshot 2024-03-28 at 13 57 16

image

wwestgarth (approved):

bwallacee (approved):

JonRay15:
Can we just simplify this to something like this?

This whole component just seems entirely overengineered from ages ago.  Its an epoch, and a time remaining.  It doesnt need a giant progress bar! 😂 

image

JonRay15:
Something like this would be another option .... but that also feels too much cos it introduces a new tile outside of structure. 

So I prefer the first one where its just text, no tiles to muddy the water?

image

MadalinaRaicu:
> Something like this would be another option .... but that also feels too much cos it introduces a new tile outside of structure.
> 
> So I prefer the first one where its just text, no tiles to muddy the water?
> 
> image

updated 

Screenshot 2024-04-04 at 11 16 13

PR #6131, opened 2024-03-28
fix(explorer): fix timeout out explorer page for oracles by market

edd:
# Description ℹ️

Pagination reversed a few versions ago, and this query was still using `last` instead of `first`. On some fairground markets, this could timeout


ValentinTrinque (approved):

daro-maj:
Looks good
image

daro-maj (approved):

PR #6148, opened 2024-04-02
chore(governance): allow batch proposals

### LINKED ISSUE #5890: Allow batch propsal via json on governance site
## The Chore

Currently when proposing on the governance site, you can paste in json for your proposal. However, this is not the full transaction, and behind the scenes we wrap the propsoal json in an object specifying that its a normal proposalSubmission. We need to change this so that the full payload is pasted in which will allow users to create either a normal proposal or a batch proposal

## Tasks

- [ ] Change propose form to send a full transaction payload, allowing specifying the transaction type
- [ ] Update the templates in the docs
- [ ] Also scrolling doesnt work in this JSON input box, fix that

mattrussell36:
# Related issues 🔗

Closes #5890 

# Description ℹ️

- Unwrap proposal transaction to allow for batch proposals

### DIFF
--- a/apps/governance/src/routes/proposals/components/propose/proposal-form-transaction-dialog.tsx
+++ b/apps/governance/src/routes/proposals/components/propose/proposal-form-transaction-dialog.tsx
@@ -8,6 +8,7 @@ import type {
   ProposalEventFieldsFragment,
   VegaTxState,
 } from '@vegaprotocol/proposals';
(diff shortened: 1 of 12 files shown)


MadalinaRaicu (approved):

daro-maj:
error message is not visible 
image

daro-maj (approved):

PR #6152, opened 2024-04-02
fix(trading): increase market selector medium width

### LINKED ISSUE #6028: Allow more space for market codes in the market selector
## The Chore

We need to allow more space in the market selector to accomadate longer codes for points markets

## Tasks

- [ ]
- [ ]

image (2)

MadalinaRaicu:
# Related issues 🔗

Closes #6028 

# Description ℹ️

The best way to **allow more space for market codes in the market selector** is to increase the overall medium width. 

# Demo 📺
Screenshot 2024-04-02 at 16 11 32

vs 
Screenshot 2024-04-02 at 15 55 08

# Technical 👨‍🔧


mattrussell36 (changes requested):
The column header not longer aligns for 24h volume. It might just need text align right

bwallacee:
On mobile we still can't see if its a points market. Are we ok with this?
image

MadalinaRaicu:
> The column header not longer aligns for 24h volume. It might just need text align right



> The column header not longer aligns for 24h volume. It might just need text align right

fixed 
Screenshot 2024-04-04 at 12 11 48
Screenshot 2024-04-04 at 12 12 15

bwallacee (approved):

mattrussell36 (approved):

PR #6186, opened 2024-04-08
chore(trading): create spot market for sim tests

daro-maj:
# Related issues 🔗

Closes #[Issue number here]

# Description ℹ️

Create spot market for sim tests and add market info tests.


apps/trading/e2e/fixtures/market.py
```diff
@@ -282,6 +282,122 @@ def setup_perps_market(
...
+    vega.mint(
+        MM_WALLET.name,
+        asset=vega.find_asset_id(symbol="VOTE", enabled=True),
+        amount=mint_amount,
+    )
+
+    vega.update_network_parameter(
+        MM_WALLET.name, parameter="market.fee.factors.makerFee", new_value="0.1"
+    )
+    vega.forward("10s")
```
bwallacee, line 305: Do we have to go forward 10s here?

apps/trading/e2e/fixtures/market.py
```diff
@@ -282,6 +282,122 @@ def setup_perps_market(
...
+
+    market_id = vega.create_simple_spot_market(
+        proposal_key_name=MM_WALLET.name,
+        market_name="Bitcoin / Tether USD (Spot)",
+        market_code="BTC/USDT-SPOT",
+        base_asset_id=vega.find_asset_id(symbol=custom_base_asset_symbol),
+        quote_asset_id=vega.find_asset_id(symbol=custom_quote_asset_symbol),
+        market_decimal_places=1,
+        position_decimal_places=2,
+        tick_size=10,
```
bwallacee, line 381: Are these values what we would expect to see on mainnet?

apps/trading/e2e/fixtures/market.py
```diff
@@ -282,6 +282,122 @@ def setup_perps_market(
...
+    )
+
+    vega.forward("10s")
```
bwallacee, line 384: Do we need to move forward 10s here?

apps/trading/e2e/fixtures/market.py
```diff
@@ -282,6 +282,122 @@ def setup_perps_market(
...
+
+    submit_liquidity(vega, MM_WALLET.name, market_id)
+    submit_multiple_orders(
+        vega, MM_WALLET.name, market_id, "SIDE_SELL", [[1, 110], [1, 105]]
+    )
+    submit_multiple_orders(
+        vega, MM_WALLET2.name, market_id, "SIDE_BUY", [[1, 90], [1, 95]]
+    )
+    submit_order(vega, "Key 1", market_id, "SIDE_BUY", 1, 110)
+    vega.forward("10s")
```
bwallacee, line 396: Do we need to move forward 10s here?

apps/trading/e2e/tests/spot_market/test_spot.py
```diff
@@ -0,0 +1,222 @@
...
+        ["Quantum", "0.00001"],
+        ["Status", "Enabled"],
+        ["Max faucet amount", "10,000,000,000.00"],
+        ["Infrastructure fee account balance", "0.003"],
+        ["Global reward pool account balance", "0.00"],
+    ]
+    validate_info_section(page, fields)
+
+def test_market_spot_info_metadata(page: Page):
+    page.get_by_test_id(market_title_test_id).get_by_text("Metadata").click()
```
bwallacee, line 150: Should there be any asserts here?

apps/trading/e2e/tests/spot_market/test_spot.py
```diff
@@ -0,0 +1,222 @@
...
+        "Price monitoring bounds"
+    ).click()
+    expect(page.get_by_test_id(market_accordion_content).locator(
+        ".w-full").nth(1)).to_contain_text("90.001")
+    expect(page.get_by_test_id(market_accordion_content).locator(".w-full").nth(2)
+           ).to_contain_text("within 15m")
+    expect(page.get_by_test_id(market_accordion_content).locator(".text-left").nth(0)
+           ).to_contain_text("104.7")
+    expect(page.get_by_test_id(market_accordion_content).locator(".text-right").nth(0)
+           ).to_contain_text("110.3")
```
bwallacee, line 163: Complicates things, but do you think its worth adding the test ids we need into the frontend so that we dont need to use .w-full etc to grab what we need.

apps/trading/e2e/tests/spot_market/test_spot.py
```diff
@@ -0,0 +1,222 @@
...
+        ["Scaling Factor", "1"],
+    ]
+    validate_info_section(page, fields)
+
+
+# Liquidity resolves to 3 results
+def test_market_spot_info_liquidity(page: Page):
+    page.get_by_test_id(market_title_test_id).get_by_text(
+        "Liquidity", exact=True
+    ).click()
```
bwallacee, line 181: Simmilar as above, I wonder if we should add test ids to the element we need.

apps/trading/e2e/fixtures/market.py
daro-maj: it is stagnet1 but those parameters are in market setup so should be the same for mainnet.

apps/trading/e2e/fixtures/market.py
daro-maj: fixed

apps/trading/e2e/tests/spot_market/test_spot.py
daro-maj: it is empty so validation is by clicking

bwallacee (approved):

PR #6215, opened 2024-04-12
feat(emblem): add market logo component

edd:
- **feat(emblem): basic hacky market component**
- **feat(emblem): use emblem in markets table**
- **feat(emblem): allow markets to only show one icon**

### DIFF
--- a/libs/emblem/src/components/hooks/use-market-info.ts
+++ b/libs/emblem/src/components/hooks/use-market-info.ts
@@ -1,0 +1,58 @@
+import { useEffect, useState } from 'react';
+import { URL_BASE } from '../../config';
+
+export type MarketInfo = {
+  quoteLogo: string;
+  settlementLogo?: string;
+  baseLogo: string;
+};
+
+export type MarketInfoState = {
+  loading: boolean;
+  data?: MarketInfo | null;
+  error: Error | null | undefined;
+};
+
+export async function fetchMarketInfo(path: string) {
+  return (await fetch(path)).json();
+}
+
+export const useMarketInfo = (vegaChain: string, marketId: string) => {
+  const [state, setState] = useState<MarketInfoState>({
+    loading: true,
+    data: null,
+    error: null,
+  });
+
+  useEffect(() => {
+    let mounted = true;
+
+    const path = `${URL_BASE}/vega/${vegaChain}/market/${marketId}/index.json`;
+
+    fetchMarketInfo(path)
+      .then((market) => {
+        if (mounted) {
+          setState({
+            loading: false,
+            data: market,
+            error: null,
+          });
+        }
+      })
+      .catch((err) => {
+        if (mounted) {
+          setState({
+            loading: false,
+            data: null,
+            error: err,
+          });
+        }
+      });
+
+    return () => {
+      mounted = false;
+    };
+  }, [vegaChain, marketId]);
+
+  return state;
+};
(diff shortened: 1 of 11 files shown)


libs/emblem/src/components/market-emblem.tsx
```diff
@@ -0,0 +1,57 @@
...
+ * Given a Vega Market ID...
+ *
+ * @param market string the market ID
+ * @param vegaChain string the vega chain ID (default: Vega Mainnet)
+ * @returns React.Node
+ */
+export function EmblemByMarket(props: EmblemByMarketProps) {
+  const chain = getVegaChain(props.vegaChain);
+  const data = useMarketInfo(chain, props.market);
+  const showBase = props.showBase ?? true;
```
mattrussell36, line 26: non blocking: destructuring with a default property value tends to be more common place for default property values

```
export function EmblemByMarket({
  ...
  showBase = true,
  showQuote = true
}: EmblemByMarketProps) {
  ...
}
```

libs/emblem/src/components/market-emblem.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { EmblemBase } from './emblem-base';
+import { useMarketInfo } from './hooks/use-market-info';
+import { getVegaChain } from './lib/get-chain';
+
+export type EmblemByMarketProps = {
+  market: string;
+  vegaChain?: string;
+  showBase?: boolean;
+  showQuote?: boolean;
+  contract?: never;
```
mattrussell36, line 11: To me there doesn't seem to be any reason to include contract and asset if they are typed as `never`. You should get a warning if you provide the prop and its not declared in props

libs/emblem/src/components/hooks/use-market-info.ts
```diff
@@ -0,0 +1,58 @@
...
+    data: null,
+    error: null,
+  });
+
+  useEffect(() => {
+    let mounted = true;
+
+    const path = `${URL_BASE}/vega/${vegaChain}/market/${marketId}/index.json`;
+
+    fetchMarketInfo(path)
```
mattrussell36, line 32: It seems sensible to have this only fetched and cached once? I think we do something similar for the oracle proofs json. Perhaps a helper hook for fetching and caching the result in memory for these types of things would be good

libs/emblem/src/components/market-emblem.tsx
```diff
@@ -0,0 +1,57 @@
...
+ * @param market string the market ID
+ * @param vegaChain string the vega chain ID (default: Vega Mainnet)
+ * @returns React.Node
+ */
+export function EmblemByMarket(props: EmblemByMarketProps) {
+  const chain = getVegaChain(props.vegaChain);
+  const data = useMarketInfo(chain, props.market);
+  const showBase = props.showBase ?? true;
+  const showQuote = props.showQuote ?? true;
+  const wrapperClass = props.wrapperClass ?? 'mr-2';
```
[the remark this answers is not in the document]
edd: Fixed

libs/emblem/src/components/market-emblem.tsx
edd: I've used partial destructuring in a follow up. My reason for doing it this way was to pass `props` down to the underlying `Emblem` component directly, so they can also be controlled.

But now I'm not sure if that's actually a good idea or not. It feels a bit spaghetti like. I vote I just remove that ability. How does that feel?

libs/emblem/src/components/market-emblem.tsx
edd: I've used partial destructuring in a follow up. My reason for doing it this way was to pass props down to the underlying Emblem component directly, so they can also be controlled.

But now I'm not sure if that's actually a good idea or not. It feels a bit spaghetti like. I vote I just remove that ability. How does that feel?

asiaznik (approved):

MadalinaRaicu (approved):

PR #6216, opened 2024-04-12
feat(trading): reward cards grouping

### LINKED ISSUE #5580: Improve summary views of rewards because there are likely to be many
## The Chore

Mainnet currently has 36 reward "cards" showing.

This is because there are dedicated per-market rewards for every reward type, so effectively every time there is a new market 4 new rewards are funded. 

The view needs to somehow make this more digestible or it will become even more unwieldy as it grows.

## Acceptance Criteria

- [ ] THIS ONLY APPLIES TO REWARD CARDS .... NOT TO GAME CARDS .... SO WE WILL NEED TO FORK THAT COMPONENT
- [ ] As a result of this we should now NOT show game cards in the reward page or it will be very confusing so add that filter
- [ ] Identify "groups" of rewards based on matching reward, type and eligibility criteria, ie:
  - [ ] Reward Asset 
  - [ ] Reward type
  - [ ] Eligibility criteria (Entity scope, staked amount, average position)
  - [ ] Pro rata vs rank
  - [ ] Assessed over
- [ ] On the summary card will need to display following aggregates:
  - [ ] Sum the reward amount across all underlying rewards
  - [ ] Display a range of durations (ie. 90 - 120 epochs)
  - [ ] The title string just becomes `rewardType` + "Rewarding markets" which is clickable and leads to the page/ modal
  - [ ] Add a link to the reward transaction in explorer so that users can see rank table
- [ ] The other details can be moved into a modal / popup / reward page to show a table of the information that may differ
  - [ ] Market Name (or market list)
  - [ ] Settlement asset
  - [ ] Reward duration
  - [ ] Lock period
  - [ ] Reward amount
  - [ ] Reward cap
  - [ ] Plus all the aggregate info along the top
- [ ] Make a new page for the reward detail
  - [ ] If I click on the aggregate reward card I get taken to a reward detail page
  - [ ] The URL could be a query string with the 5 group defining fields above
  - [ ] It might get expanded later with stats on what you've earned (you earned X or the Y available last epoch)
  - [ ] It would show the headline numbers along the top plus an AG Grid of the other details beneath
- [ ] Need handling for the greying out
  - [ ] The group would only go grey if ALL rewards it were inactive
  - [ ] I guess some rows in the details might go grey individually too
  - [ ] Add a tick box for "Hide inactive" on both the main view and the details

 
## Sketch

image

### LINKED ISSUE #6182: Finalise reward filter logic
## The Chore

At the moment we show:

- Only entityScope = Team  on Competitions page
- Everything on Rewards page

Do we want to change this so they're mutually exclusive?  I feel like at some point we said we did but it clearly never got ticketed, and maybe never got agreed.

## Tasks

- [x] Decide 
- [ ] Filter the reward cards on the reward page to exclude entityScore = Team

asiaznik:
# Related issues 🔗

Closes #5580 
Closes #6182 

# Description ℹ️

* removed `StakingRewardCard`, added special `dispatchStrategy` for this type of reward,
* added groups (`GroupRewardCard`) of reward cards based on:
  - reward asset (usually VEGA),
  - reward for (dispatch metric),
  - reward scope (teams vs individuals),
  - reward distribution strategy
* added dialogs with cards from a group
* excludes scoped to teams rewards from 'Rewards' page

# Demo 📺

https://github.com/vegaprotocol/frontend-monorepo/assets/1980305/bd49af42-ec18-4a90-b18a-bbc4891e713c

# Technical 👨‍🔧

Button colours:

Screenshot 2024-04-12 at 17 32 54

### DIFF
--- a/apps/trading/client-pages/competitions/competitions-team.tsx
+++ b/apps/trading/client-pages/competitions/competitions-team.tsx
@@ -14,7 +14,12 @@ import {
   Intent,
   CopyWithTooltip,
 } from '@vegaprotocol/ui-toolkit';
-import { TransferStatus, type Asset } from '@vegaprotocol/types';
+import {
+  type DispatchStrategy,
+  type StakingDispatchStrategy,
+  TransferStatus,
+  type Asset,
+} from '@vegaprotocol/types';
 import classNames from 'classnames';
 import { useT } from '../../lib/use-t';
 import { Table } from '../../components/table';
@@ -139,7 +144,9 @@ const TeamPage = ({
   members?: Member[];
   games?: TeamGame[];
   gamesLoading?: boolean;
-  transfers?: EnrichedRewardTransfer[];
+  transfers?: EnrichedRewardTransfer<
+    DispatchStrategy | StakingDispatchStrategy
+  >[];
   transfersLoading?: boolean;
   allMarkets?: MarketMap;
   refetch: () => void;
@@ -259,7 +266,9 @@ const Games = ({
 }: {
   games?: TeamGame[];
   gamesLoading?: boolean;
-  transfers?: EnrichedRewardTransfer[];
+  transfers?: EnrichedRewardTransfer<
+    DispatchStrategy | StakingDispatchStrategy
+  >[];
   transfersLoading?: boolean;
   allMarkets?: MarketMap;
 }) => {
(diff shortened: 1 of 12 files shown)


bglownia (approved):

edd (commented):
Late review. I like the pattern for the 'base' STAKING_DISPATCH_STRATEGY, it's nice and clear

PR #6239, opened 2024-04-17
feat(trading): WIP new markets page list view 

### LINKED ISSUE #5095: New markets page list view
## Story

As a user
I want a much nicer looking market summary page
So that I have a good place to look at all markets and choose one

## Acceptance Criteria

- [x] There is a list of all active markets as shown in the sketch
- [x] The list shows
  - [x] Code and name
  - [x] Price (use mark for perp / fut and last traded for spot)
  - [x] 24h change in value and % plus a spark line
  - [x] 24h volume in contracts and in USD terms (assume we will only settle in USD equivalents for now)
  - [x] Open interest in contracts and USD terms (as above)
- [x] There is a panel at the top shown the top network wide 24h volume.  Need to make an API request for this, but can loop over markets for now if needed
- [x] There is a panel at the top showing the total TVL ... that should be easier .. its on DefILlama I think and probably is elsewhere too, might need to run some read function on the asset pool contract and do it for each active asset?
- [x] There is a panel showing the top gainers, ie. sorted by biggest 24h change
- [x] There is a panel showing new listsing which is just the 3 most recent opening timestamps on the network
- [ ] You can leave the toggle for tile view for now
- [ ] If the market is in status other than continuous trading indicate this with icons to follow shortly

## Sketch

https://www.figma.com/file/ywcCZvXdvG28bg0AmCY9mF/Console-v3?type=design&node-id=6030-34216&mode=design&t=6l6qTRkC2tzvPOLv-0

### Filters
- Filter list to exclude all settled or terminated markets
- A "quick filter" for these will be added in the next ticket but for now remove them

### Status Icons
- A list of icons to express the following statuses should be displayed when they are appropriate:
  - Proposed
  - Opening auction
  - Monitoring auction
  - Suspended
  - Terminated
  - Settled

MadalinaRaicu:
# Related issues 🔗

Closes #5095 

# Description ℹ️

New markets page list view 

# Demo 📺

**WIP**

Screenshot 2024-04-16 at 22 57 02

# Technical 👨‍🔧


apps/trading/client-pages/markets/use-markets-stats.tsx
```diff
@@ -0,0 +1,96 @@
...
+};
+
+/**
+ * useTVL returns the total value locked in the network
+ * by summing the balances of all insurance accounts
+ *
+ * @param activeMarkets
+ * @returns MarketMaybeWithCandles[] | null
+ */
+export const useTVL = (activeMarkets: MarketMaybeWithCandles[] | null) => {
```
MadalinaRaicu, line 85: this is calculated by adding the Insurance Pool Balance for each market. 
how can I convert this to quantum?


note: this works well for perps/futures, not sure for spots.

apps/trading/client-pages/markets/use-column-defs.tsx
```diff
@@ -208,60 +161,39 @@ export const useMarketsColumnDefs = () => {
...
+          const openInterest =
+            data?.data?.openInterest === undefined
+              ? '-'
+              : addDecimalsFormatNumber(
+                  data?.data?.openInterest,
+                  data?.positionDecimalPlaces
+                );
+          // TODO: calculate USDT value
+          // const markPrice = data?.data?.markPrice;
+          const openInterestValue = openInterest;
```
MadalinaRaicu, line 193: unclear how I can calculate the open interest in quantum

edd (approved):
Emblem stuff looks good

JonRay15:
A few comments:

1. I am not getting icons .... dont know if this is because I'm running the preview link, or if its because I'm connecting to mainnet datanode maybe?
2. There is a weird slight white outline on the icons that i think Ben has also mentioned
3. The two panels at the top should be displaying the market code instead of the name as we likely wont have space for the name there, also the font of the "top gainers" title looks to be the same as the market names, I think the codes should be smaller, they currently look bigger than the corresponding codes in the main table?
4. I think its weird that the right two panels have clear titles and the ones on the left dont ... I would probably favour consistently using the top left of the panel for the title but I think that was a conscious design choice so lets wait for further feedback.... if we do leave it like this for now then i thin the "24h volume" and "TVL" should be in the grey text?
5. You wouldnt have known this but in the last design session we agreed it is fine to use $ instead of USDT so we dont keep repeating that all over the place so can you change that please 
6. The horizontal default sizing isnt _quite_ right for my 13in monitor, I lose the open interest to the right slightly.  Lets adjust some of the widths to avoid it.
7. Think the default size of the Market column should be wider so we dont lose the text on some of the descriptions
8. Dont think we want the vertical lin to the right of the name, and dont think we want the little burger menu column on the far right ... both of these make it feel a bit too "AG griddy" which we wanted to avoid

image

bwallacee:
A few more comments sorry
- Can We improve the size of the reset columns in settings:

image

- the markets in new listings aren't equally spaced apart (probs because of the bracketed 0.00)

- Sui/Tether is missing the sparkline

image

- Could we use the pills instead of the (market type)
image
-No mouse hover icon when over the markets in the table.
-Sort on 24h change doesn't work

wwestgarth (approved):

edd:
Two emblem related things on this branch that I did not spot earlier:

1. On preview links, all market icons are black. Black is 'missing.svg'. They're missing because the vega chain ID is not always provided correctly. https://github.com/vegaprotocol/frontend-monorepo/blob/2c9bec7697b025a2c2228b8add3c66b10b479b24/libs/wallet-react/src/hooks/use-chain-id.ts#L4 relies on `useWallet` to fetch a chainId, but it does not always pass one. In Explorer, I set it in .env, and all the other `.env`s
2. Since editing the positioning of the icons to both be vertically aligned, the two icons are too wide. This line needs to change to `w-9`. https://github.com/vegaprotocol/frontend-monorepo/blob/2c9bec7697b025a2c2228b8add3c66b10b479b24/libs/emblem/src/components/market-emblem.tsx#L46

libs/emblem/src/components/market-emblem.tsx
```diff
@@ -42,6 +42,7 @@ export function EmblemByMarket(props: EmblemByMarketProps) {
   // Widths are calculated here as they are required for using absolute positioning to
   // render the logos as overlapping. Moving to blocks with negative margins should work
   // and could remove this calculation, but was not working reliably cross platform.
+  // const wrapperClassString = classNames('relative flex items-center', {
   const wrapperClassString = classNames('relative inline-block h-6', {
     'w-8': logoCount === 2,
```
edd, line 47:
```suggestion
    'w-9': logoCount === 2,
```

`w-8` is too narrow for the two logos side by side, which can cause wrapping in some places

apps/trading/client-pages/markets/top-market-list.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const { chainId } = useChainId();
+  return (
+    <div className="grid auto-rows-min grid-cols-6 gap-3 text-sm">
+      {markets?.map((market) => {
+        return (
+          <>
+            <span className="col-span-3">
+              <Link to={`/markets/${market.id}`}>
+                <span className="pr-1">
+                  <EmblemByMarket market={market.id} vegaChain={chainId} />
```
edd, line 21: **chore (blocking)** : `chainId` is defaulting to `mainnet` here, presumably because `useChainId` is not populated when the wallet is not connected. Or something like that. Either way, the chain ID needs to be set properly for all users, which it isn't - this is why the logos in the deploy previews are black. In explorer, this is set in `.env` (to stagnet) and then more explicitly in every other `.env.*`.

edd (changes requested):
Two changes required to make this branch mergeable (from an emblem perspective)

libs/emblem/src/components/market-emblem.tsx
```diff
@@ -42,6 +42,7 @@ export function EmblemByMarket(props: EmblemByMarketProps) {
   // Widths are calculated here as they are required for using absolute positioning to
   // render the logos as overlapping. Moving to blocks with negative margins should work
   // and could remove this calculation, but was not working reliably cross platform.
+  // const wrapperClassString = classNames('relative flex items-center', {
```
edd, line 45:
```suggestion
```

We don't need to keep this around

apps/trading/client-pages/markets/top-market-list.tsx
MadalinaRaicu: I have updated all `.env.*`

apps/trading/client-pages/markets/use-markets-stats.tsx
mattrussell36, line 85: I'd prefer to avoid abbreviate here. Can we make this `useTotalVolumeLocked`

apps/trading/client-pages/markets/markets-page.tsx
```diff
@@ -13,58 +9,213 @@ import {
...
+    error,
+    reload,
+  } = useDataProvider({
+    dataProvider: activeMarketsWithCandlesProvider,
+    variables: {
+      since: new Date(yesterday).toISOString(),
+      interval: Interval.INTERVAL_I1H,
+    },
+  });
+  useEffect(() => {
```
mattrussell36, line 49: Are we not able to set a pollInterval on the activeMarketsDataProvider?

CC @bglownia

apps/trading/client-pages/markets/markets-page.tsx
```diff
@@ -13,58 +9,213 @@ import {
...
+  }, 0);
+
+  return (
+    <ErrorBoundary feature="rewards">
+      <TinyScroll className="p-2 max-h-full overflow-auto">
+        <div className="flex flex-col w-full gap-3 p-3 xxl:px-[5.5rem]">
+          <div className="grid auto-rows-min grid-cols-3 lg:grid-cols-5 xl:grid-cols-3 gap-3 w-full">
+            <div className="flex flex-col gap-2 col-span-full lg:col-span-1">
+              <Card key="24h-vol" className="grow">
+                <div className="flex items-center justify-center w-full">
```
mattrussell36, line 89: Seems like some unnecessary markup and classes here. This could be cleaner

- The first element inside Card is unnecessary
- w-full is not required on the next element (block elements are full width by default)

apps/trading/client-pages/markets/top-market-list.tsx
```diff
@@ -0,0 +1,51 @@
...
+  return (
+    <>
+      {markets?.map((market) => {
+        return (
+          <div
+            className="grid auto-rows-min grid-cols-7 gap-3 text-xs"
+            key={market.id}
+          >
+            <span className="col-span-3 overflow-hidden">
+              <Link to={`/markets/${market.id}`}>
```
mattrussell36, line 27: Please use links object

```suggestion
              <Link to={Links.MARKET(market.id)}>
```

apps/trading/client-pages/markets/markets-page.tsx
```diff
@@ -13,58 +9,213 @@ import {
...
+              target="_blank"
+              rel="noreferrer"
+            >
+              {t('Propose a new market')}
+            </a>
+          )}
+        </div>
+      </div>
+      <div className="h-full my-1 border rounded border-default">
+        {activeTab === 'open-markets' && (
```
mattrussell36, line 204: Perhaps this has already been discussed. But my expectation was that this would be a single table with ALL markets that would be programmatically filtered. 

Currently the columns are different between each state and we may still want that. If that is the case then this is probably suitable.

BUT its worth saying that in the current implementation, there won't be a way to see all markets with the same settlement asset (for example), regardless of its open/closed/proposed state.

For me, I think that kind of filtering is needed for this 'master' list. 

@JonRay15

apps/trading/client-pages/markets/top-market-list.tsx
```diff
@@ -0,0 +1,51 @@
...
+}: {
+  markets?: MarketMaybeWithCandles[];
+}) => {
+  const { chainId } = useChainId();
+  return (
+    <>
+      {markets?.map((market) => {
+        return (
+          <div
+            className="grid auto-rows-min grid-cols-7 gap-3 text-xs"
```
mattrussell36, line 23: Couple things

- The layout of the card looks too tight
- Font size looks too small

Screenshot 2024-04-25 at 14 28 55

apps/trading/client-pages/markets/use-markets-stats.tsx
```diff
@@ -0,0 +1,96 @@
...
+  if (!activeMarkets || activeMarkets.length === 0) return [];
+  for (let i = 0; i < 24; i++) {
+    const totalVolume24hr = activeMarkets.reduce((acc, market) => {
+      const c = market.candles?.[i];
+      if (!c) return acc;
+      return (
+        acc +
+        Number(
+          addDecimal(
+            c.notional,
```
mattrussell36, line 26: We need to convert with quantum first to get something close to the actual usd amount

apps/trading/client-pages/markets/use-markets-stats.tsx
```diff
@@ -0,0 +1,96 @@
...
+  return candles;
+};
+
+/**
+ * useTopGainers returns the top 3 markets with highest gains, i.e. sorted by biggest 24h change
+ *
+ * @param activeMarkets
+ * @returns MarketMaybeWithCandles[]
+ */
+export const useTopGainers = (
```
mattrussell36, line 43: Non blocking: It may not be suitable but the use-market-selector-list hook does this already. You may want to use that

Same for new listings.

libs/markets/src/lib/components/last-24h-price-change/last-24h-price-change.tsx
```diff
@@ -37,6 +39,28 @@ export const Last24hPriceChange = ({
   const change = priceChange(candles);
   const changePercentage = priceChangePercentage(candles);
 
+  if (orientation === 'vertical') {
```
mattrussell36, line 42: Dont love this. Could probably be done better by just flipping between 'flex-row' and 'flex-col'? You might need to re-arrange the markup though

apps/trading/client-pages/markets/use-markets-stats.tsx
MadalinaRaicu: sure!

apps/trading/client-pages/markets/top-market-list.tsx
MadalinaRaicu: ok so made few changes
use the market code and used text-sm 

image


not sure it looks better imo but the cards are really to tight because we try to fit a lot inside.
didn't change sparkline component to make it smaller or with adjustable width... 

i even added a breakpoint to make those at the beginning smaller after a certain breakpoint 
image

apps/trading/client-pages/markets/markets-page.tsx
MadalinaRaicu: yes, the columns are different for open, closed and proposed markets. 
totally different tables because of that and that's why I made these tabs like this for now 

i think we can revisit this when we add more filters

apps/trading/client-pages/markets/markets-page.tsx
MadalinaRaicu: simplified it in latest commit

apps/trading/client-pages/markets/use-markets-stats.tsx
MadalinaRaicu: image
we also have this, can we use it for this

libs/markets/src/lib/components/last-24h-price-change/last-24h-price-change.tsx
MadalinaRaicu: they do look slightly different, it's not just the orientation

I can make another component 

Screenshot 2024-04-26 at 12 57 40

Screenshot 2024-04-26 at 12 57 31

edd (approved):

bwallacee (approved):

bglownia (approved):

PR #6274, opened 2024-04-22
feat(utils): round to smallest meaningful number  

### LINKED ISSUE #6271: Rounding the amount in collateral is inaccurate
## Description

After selling the entire amount of the asset that was from collateral tab, there may be some very small amounts left that and are visible only in the usage breakdown

## Steps to Reproduce

1. Go to trading collateral
2. Sell all amount of visible asset 
4. Error: in usage breakdown tab we have small amount and 0.00 in the collateral row

## Expected behavior

Should be 0.00 in  usage breakdown and in the collateral row

## Screenshots
image

bglownia:
# Related issues 🔗

Closes #6271

# Description ℹ️

Show more decimal places if to avoid showing 0 when number is not 0.

# Demo 📺

image


daro-maj:
please move this pr to the main

jeremyletang (approved):

Issue #5746, opened 2024-02-05
Add referral benefits to team profile


JonRay15:
## The Chore

We need to see this info on another tab so that users can see the referral benefits of a team they are consideraing joining

## Tasks

- [ ]
- [ ]

## Additional details / background info

Original ticket: https://github.com/vegaprotocol/frontend-monorepo/issues/5486

#### Sketch 2 Referral benefit: 

![Image](https://github.com/vegaprotocol/frontend-monorepo/assets/89918817/cf9f2254-131c-4939-88ad-be34c6684af3)

#### Sketch 3 Already in referral set 

![Image](https://github.com/vegaprotocol/frontend-monorepo/assets/89918817/72dca3f4-dccb-4400-857e-599a54b47f44)

JonRay15:
Adding a small task labbel to this as I cant imagine its not small ... its adding a tab and reusing an existing component.

JonRay15:
Seems like we have decided not to do this.

Closing

PR #6276, opened 2024-04-23
fix(trading): referee tier statistic

asiaznik:
# Related issues 🔗

Related #5746 

# Description ℹ️

* fixes tier statistics

# Demo 📺

N/A

# Technical 👨‍🔧

N/A

### DIFF
--- a/apps/trading/client-pages/competitions/competitions-team.tsx
+++ b/apps/trading/client-pages/competitions/competitions-team.tsx
@@ -245,13 +245,6 @@ const TeamPage = ({
               count: members ? members.length : 0,
             })}
           </ToggleButton>
-          <ToggleButton
-            active={tab === Tab.Benefits}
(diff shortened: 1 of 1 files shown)


JonRay15:
Sorry bad news on this ... I dont think we want to merge it. 

We just noticed some behaviour in mainnet that it is possible to join a team without getting its referral benefits....and apparently we think that is correct and desirable.... therefore we think that showing the referral benefits associated to a team is not the right thing to do.  Sorry @asiaznik ... 🙈

JonRay15 (changes requested):
See comment .... based on discussion literally just now with Barney and Charlie we dont think we should do this ... sorry.

asiaznik:
Remove the part that displayed the tiles on a team page and left the fix in.

mattrussell36 (approved):

JonRay15 (approved):

PR #6281, opened 2024-04-23
feat(trading): market icons in console

### LINKED ISSUE #5096: Add icons for markets
## Story

As a user
I want to be able to see nice icons in Console for markets and assets
So that I think "wow ... Console looks cool"

## Acceptance Criteria

- [ ] Show in market header for current market
- [ ] Show in the market selector
- [x] Show in the markets page

## Tasks

- [ ] UX (if needed)
- [ ] Design (if needed)
- [ ] Implementation
- [ ] Code review
- [ ] QA review

MadalinaRaicu:
# Related issues 🔗

Closes #5096 

# Description ℹ️

Add icons for markets in console. 

# Demo 📺

Screenshot 2024-04-23 at 11 09 30
Screenshot 2024-04-23 at 11 07 30
Screenshot 2024-04-23 at 11 03 22
Screenshot 2024-04-23 at 11 04 46

# Technical 👨‍🔧

### DIFF
--- a/apps/trading/components/liquidity-header/liquidity-header.tsx
+++ b/apps/trading/components/liquidity-header/liquidity-header.tsx
@@ -61,7 +61,7 @@ export const LiquidityHeader = () => {
         marketId && (
           <HeaderTitle>
             {marketId && (
-              <span className="pr-1">
+              <span className="mr-2">
                 <EmblemByMarket market={marketId} vegaChain={chainId} />
               </span>
             )}
--- a/apps/trading/components/liquidity-header/liquidity-header.tsx
+++ b/apps/trading/components/liquidity-header/liquidity-header.tsx
@@ -24,6 +24,8 @@ import {
 } from '@vegaprotocol/liquidity';
 import { useParams } from 'react-router-dom';
 import { useT } from '../../lib/use-t';
+import { EmblemByMarket } from '@vegaprotocol/emblem';
+import { useChainId } from '@vegaprotocol/wallet-react';
 
 export const LiquidityHeader = () => {
   const t = useT();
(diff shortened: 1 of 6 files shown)


edd (approved):
Two notes, neither blocking or requiring changes:

1. It's possible to use `Emblem` and have that component determine that it's a market component, and then render `EmblemByMarket`. There's no advantage to that, it's just the pattern I've used. 
2. `explorer` uses the point above to make another wrapper component, `EmblemWithChain`, that avoids the repeated need to import `Emblem` as well as `useVegaChain`. Again, though, that's just convenience. What you've done is also valid

One last point - the fact that the two icons are not aligned vertically is more of a mistake than an intentional design fix. I guess we might change this separately when/if someone requests it. On the other hand: it looks fine.

JonRay15:
I dont love the vertical offset here.

I feel like there needs to be a less horizontal overlap and they should be at the same height.

image

MadalinaRaicu:
> I dont love the vertical offset here.
> I feel like there needs to be a less horizontal overlap and they should be at the same height.
> Also when we start adding that third icon to denote the chain it will be a problem if we dont do the above.

> One last point - the fact that the two icons are not aligned vertically is more of a mistake than an intentional design fix. I guess we might change this separately when/if someone requests it. On the other hand: it looks fine.


I will fix the vertical offset in the emblem icon component, that should fix it everywhere. Initially, I thought that was done on purpose, but I will change that. 

The third icon (chain) will be added in another PR because we don't know yet how to get the proper `chain icon`  and it is in prog here: https://vegaprotocol.slack.com/archives/C8WS7M50E/p1713540865169819

bwallacee (approved):
We are missing the white outlines on the Icons, but approved so we can get in for the incentive.

PR #6294, opened 2024-04-24
fix(governance): update vegacaspule 76.preview6

daro-maj:
# Description ℹ️

Update vegacaspule to 76.preview3


daniel1302:
I would suggest to rename it to v0.76.0-preview.6 and use this runner version: https://github.com/vegaprotocol/docker/pull/136/files

If you feel it is not required, please revert my commit :) I have tested it locally and it worked well:

image

asiaznik (approved):

PR #6330, opened 2024-04-30
feat(trading): add intervals to the slider

### LINKED ISSUE #6093: Add intervals to the slider
## The Chore

Would be nice to add a simple 0%, 25%, 50%, 75%, 100% to the slider so users can click and set it quickly.

Image

Or could have a little % input box somewhere too?

## Tasks

- [x] Add 0, 25, 50, 75, 100 % clickable intervals to the slider
- [x] When you hover over these the icon should change to a hand pointer to show you can click
- [ ] Adding % tooltip when you're dragging

Follow the same style guide as the ;leverage slider intervals

image

bglownia:
# Related issues 🔗

Closes #6093 

# Description ℹ️

Add a simple 0%, 25%, 50%, 75%, 100% to the slider so users can click and set it quickly.
# Demo 📺

image

### DIFF
--- a/libs/deal-ticket/src/components/deal-ticket/deal-ticket.tsx
+++ b/libs/deal-ticket/src/components/deal-ticket/deal-ticket.tsx
@@ -25,7 +25,7 @@ import {
   TradingButton as Button,
   Pill,
   ExternalLink,
-  Slider,
+  PercentageSlider as Slider,
 } from '@vegaprotocol/ui-toolkit';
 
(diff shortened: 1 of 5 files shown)


bwallacee:
@bglownia We are missing this task:

> Adding % tooltip when you're dragging

Also I don't think there's much we can do, but on mobile, you obviously don't get any hover states, so it's not the clearest that the 0,25 etc is clickable.

bwallacee (approved):

PR #6333, opened 2024-04-30
fix(trading): fix quantum formatting tvl

MadalinaRaicu:
# Related issues 🔗

Closes N/A 

# Description ℹ️

Quantum formatting on balance messes up the calculation

# Demo 📺

d2

# Technical 👨‍🔧

### DIFF
--- a/apps/trading/client-pages/markets/use-markets-stats.tsx
+++ b/apps/trading/client-pages/markets/use-markets-stats.tsx
@@ -88,8 +88,7 @@ export const useTotalVolumeLocked = (
       ?.filter((e) => e?.node?.type === AccountType.ACCOUNT_TYPE_INSURANCE)
       .map((e) => e?.node);
     const balance = accounts?.reduce((acc, a) => {
-      const balance = toQUSD(a?.balance || 0, a?.asset.quantum || 1);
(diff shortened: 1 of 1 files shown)


apps/trading/client-pages/markets/use-markets-stats.tsx
```diff
@@ -88,8 +88,7 @@ export const useTotalVolumeLocked = (
       ?.filter((e) => e?.node?.type === AccountType.ACCOUNT_TYPE_INSURANCE)
       .map((e) => e?.node);
     const balance = accounts?.reduce((acc, a) => {
-      const balance = toQUSD(a?.balance || 0, a?.asset.quantum || 1);
-      return toBigNum(balance || 0, a?.asset.decimals || 0).plus(acc);
+      return toBigNum(a?.balance || 0, a?.asset.decimals || 0).plus(acc);
```
asiaznik, line 91: I'm not sure that adding balances this way is correct, the $3 + €4 + PLN 1 != 8.
So the previous quantum was ok IMO, but loose the `toBigNum` it's not needed.

```javascript
const balance = toQUSD(a?.balance || 0, a?.asset.quantum || 1);
return balance.plus(acc);
```

PR #6413, opened 2024-05-13
feat(trading): market page snags

### LINKED ISSUE #6316: New markets page snags
## The Chore

Feedback for further improvements to the markets page

### Sorting issue

- [ ] 24h volume is sorting by the raw number (correction ... it is sorting by .... I dont know what ... it seems pretty random!) when ideally it would sort by the USDT equivalent given below
- [ ] Same issue for open interest 
- [ ] 24h Change is sorting by lord knows what ... it seems to think zero (no change) is the lowest number possible and its sorting the other stuff seemingly at random - weirdly the "Top gainers" section is sorting just fine

### DP issues

- [ ] The 24h volume and open interest as expressed in _contracts_ should be given to the marketDp of the market 
- [ ] The 24h volume and open interest as expressed in _USD_ should be given to ZERO DP

### Column width

I have lost the edge of the "open interest" column on a 13in monitor with 90% zoom on Chrome.  If I go to 100% I lose it almost completely.  Therefore in both cases I have a horizontal scroll bar. 

image

- [ ] Adjust default column widths so we fit all columns on a 13in monitor at 100% zoom

### Column widths on Proposed & Closed

- [ ] On these panes it looks like we need wider columns because theres just dead space on the far right?

image

### Spacing / padding

- [x] In my opinion the rows still need more top and bottom padding... it feels a bit dense for this "websitey" vibe we are going for.
- [ ] I think the icons could be a little bigger too 
- [ ] I think the font of the "primary" info in the table (market code, price etc) could be bigger to distinguish it more from the "secondary" grey info

image

### 24h change section

It seems like the spark line graph doesnt fill the entire height of the row? The chart starts from maybe a third of the way up? But then it goes to the very top? 

image

I'm conscious this is a share component with the market dropdown so anythign we do here will impact that too.

image

Its particularly jarring on things in auction where they get this weird grey bar.

image

Not sure exactly is the best course of action ... I can think of two:

1. Fill the full row top to bottom, but given we have the fill under the line that will mean a lot of filled in space.  We would likely need to cut it off somewhere so maybe we have a margin top and bottom but lets have _the same_ margin since at the moment it is only really occupying the top 2/3 of the row
2. If we think that is going to lead to too much filled space, could we do that but not have the gradient fill under the line?  It seems to me that it is that gradient - and trying to work out where it stops - that is causing the issue here? If I look at other examples most dont have that gradient under the line so dont face this problem

image

- [ ] Decide what to do about the spark line and do it

MadalinaRaicu:
# Related issues 🔗

Closes #6316 

# Description ℹ️

- [x] increase row height
- [x] fix sorting to use percentage / notional values
- [x] adjust market state icons' background
- [x] remove grid store
- [x] adjust font size

# Demo 📺

image

# Technical 👨‍🔧


libs/ui-toolkit/src/components/icon/vega-icons/svg-icons/icon-closed.tsx
```diff
@@ -16,7 +16,7 @@ export const IconClosed = ({
       >
         <path
           d="M0 10C0 4.47715 4.47715 0 10 0C15.5228 0 20 4.47715 20 10C20 15.5228 15.5228 20 10 20C4.47715 20 0 15.5228 0 10Z"
-          fill="#323339"
+          fill="#65676b"
```
mattrussell36, line 19: we shouldn't be using fill properties as these icons should rely on currentColor. See how the other icons are set up.

If this is for the background circle that should be outside of the svg and created/styled with a div/span

apps/trading/client-pages/markets/use-column-defs.tsx
```diff
@@ -29,29 +29,40 @@ import { EmblemByMarket } from '@vegaprotocol/emblem';
...
   ).multipliedBy(toBigNum(markPrice, decimalPlaces));
 
-  const openInterestNotional =
+  return {
+    openInterest: openInterestPosition,
+    openInterestNotional,
+  };
+};
+
+const openInterestRenderer = (data: MarketMaybeWithData | undefined) => {
```
mattrussell36, line 53: You are using this as a component but not capitalizing and calling it as JSX. Something that returns JSX should be written as a component IMO

for example:

`const OpenInterestCell = () => ...`
and
`<OpenInterestCell  />`

apps/trading/client-pages/markets/closed.tsx
```diff
@@ -250,17 +251,25 @@ const ClosedMarketsDataGrid = ({
     ];
   }, [chainId, t]);
 
+  const gridStore = useMarketsStore((store) => store.gridStore);
```
mattrussell36, line 254: I dont see loads of value of persisting grid state for these grids and would prefer to leave it out unless someone has specifically argued a case to have it

libs/ui-toolkit/src/components/icon/vega-icons/svg-icons/icon-closed.tsx
MadalinaRaicu: this is not the current color, it is for the background, and they seemed as different icons altogether in the designs. 
will see what i can do to just reuse ones we have with the background added

MadalinaRaicu:
> @MadalinaRaicu Open interest is still cut off for 13inch mac 100% zoom: image
> 
> I don't think this has been done:
> 
> > I think the font of the "primary" info in the table (market code, price etc) could be bigger to distinguish it more from the "secondary" grey info
> 
> This also hasn't been done:
> 
> > It seems like the spark line graph doesn't fill the entire height of the row? The chart starts from maybe a third of the way up? But then it goes to the very top?

1. i previously enabled autosize on all the tables but it also saves the column widths if you dragged them. 
i think the open interest column width was saved previously on your computer so resetting console would make it work...but .now I have removed storing column preferences based on Matt's comment 

2. the font was slightly larger on market names, i don't think it looks very well on numerical fields. i have pushed now a change to make it slightly larger. 

3. the sparkline doesn't start from the third height of the row but from the middle and it is aligned with the 24h change  values. i sort of left the sparkline changes out of this PR

apps/trading/client-pages/markets/closed.tsx
MadalinaRaicu: cool. removed them for now.

MadalinaRaicu:
Screenshot 2024-05-15 at 10 44 07

libs/ui-toolkit/src/components/icon/vega-icons/svg-icons/icon-closed.tsx
mattrussell36: The icon should not contain the circle background, only the icon itself. None of the icons in the set use a background color right now (I think), we should keep this consistent. If you need a background then it should be style on the wrapping element

bwallacee (approved):

libs/ui-toolkit/src/components/icon/vega-icons/svg-icons/icon-closed.tsx
MadalinaRaicu: updated!

bglownia (approved):

PR #6423, opened 2024-05-14
fix(trading): trading button break word

### LINKED ISSUE #6332: Size overflow on deal ticket
## The Chore

If the user is some kind of moron and types hundreds of numbers into the size field the submit button overflows off the side of the deal ticket and creates a horizontal scroll situation. 

image

I cant be bothered to try to include a max size / max decimals check in the actual box ... so....

## Tasks

- [ ] If the number in size gets bigger than the default screen width will allow put a ... on the end of the number shown on the submit button

MadalinaRaicu:
# Related issues 🔗

Closes #6332 

# Description ℹ️

Fixes size overflow on the deal ticket by adding break word on trading button.

# Demo 📺

Fixes: 
Screenshot 2024-05-14 at 15 53 07

To be:

Screenshot 2024-05-14 at 15 52 09

# Technical 👨‍🔧

### DIFF
--- a/libs/ui-toolkit/src/components/trading-button/trading-button.tsx
+++ b/libs/ui-toolkit/src/components/trading-button/trading-button.tsx
@@ -97,7 +97,7 @@ const Content = ({
     {subLabel && (
       <span
         data-sub-label
-        className="mt-0.5 font-mono text-xs leading-tight"
+        className="mt-0.5 font-mono text-xs leading-tight [word-break:break-word]"
         key="trading-button-sub-label"
       >
         {subLabel}
--- a/libs/deal-ticket/src/components/deal-ticket/key-value.tsx
+++ b/libs/deal-ticket/src/components/deal-ticket/key-value.tsx
@@ -28,11 +28,17 @@ export const KeyValue = ({
   );
 
   const valueElement = onClick ? (
-    <button type="button" onClick={onClick} className="font-mono ml-auto">
+    <button
+      type="button"
+      onClick={onClick}
+      className="font-mono ml-auto [word-break:break-word]"
+    >
       {displayValue}
     </button>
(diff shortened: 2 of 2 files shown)


libs/ui-toolkit/src/components/trading-button/trading-button.tsx
```diff
@@ -97,7 +97,7 @@ const Content = ({
     {subLabel && (
       <span
         data-sub-label
-        className="mt-0.5 font-mono text-xs leading-tight"
+        className="mt-0.5 font-mono text-xs leading-tight [word-break:break-word]"
```
mattrussell36, line 100: Pretty sure the default TW class `break-all` should work here

mattrussell36 (approved):

bwallacee:
@MadalinaRaicu on the side of silly but if the size is too much it then reverts to 0

image

Also should we fix the notional wordwrap in this as well or would you rather I raise a new ticket:
image

libs/ui-toolkit/src/components/trading-button/trading-button.tsx
MadalinaRaicu: i can use `break-all` here because it is a number, but i thought it is safer to use `break-word` inside the trading-button content for words as well

MadalinaRaicu:
> @MadalinaRaicu on the side of silly but if the size is too much it then reverts to 0
> 
> image
> 
> Also should we fix the notional wordwrap in this as well or would you rather I raise a new ticket: image

I pushed a fix for the notional and every key value pair from that info beneath the button in this PR

I haven't changed anything to reset the size here, just styling changes, but I guess that is the size slider logic

bwallacee (approved):

Issue #6110, opened 2024-03-26
Asset breakdown in portfolio page & sidebar


JonRay15:
## Story

As a user
I want to see the breakdown of my assets and how they're used in portolfio screen
So that I know where my assets are allocated

## Acceptance Criteria

- [x] There is a view on portfolio page per asset giving the split by account
- [x] This can for now be shown as another tab in the asset mgmt panel until we find a better long term home for it ... it should be called "Assets"
- [x] This work is common to this and the sidebar work ... it should be the same everywhere like everything in this asset mgmt component

## Sketch

Contained in here in teh ASSETS section.  We will use the expanded full version for now.

https://www.figma.com/file/ywcCZvXdvG28bg0AmCY9mF/Console-v3?type=design&node-id=6158-40256&mode=design&t=ABCsSihKRELNzJxI-0

JonRay15:
I'm closing this because the PR is already merged on Matt's stuff and in FG ... and we now have other tickets for snags etc.

PR #6427, opened 2024-05-15
feat(trading): enhanced trading sidebar

### LINKED ISSUE #6106: Enhanced trading screen sidebar
## Story

As a user
I want a sidebar on te trade screen that makes the deal ticket more obvious and ensures I always know my asset balance
So that I dont get stuckk trying to find these things or have to navigate away from trading to find them

## Acceptance Criteria

- [x] The minibar is removed from this page and all other pages (users will use the new portfolio screen in #6098 if they want to manage assets)
- [ ] The settings will need to move to the header as part of this
- [x] The default view on the trading screen is TRADE sketch, showing:
  - [x] An expanded deal ticket
  - [x] An market info tab which expands if clicked, causing the deal ticket to shrink
  - [x] A mini-asset view which expands if clicked, causing the deal ticket to shrink
  - [x] A datanode health indicator which launches the existing datanode selector if clicked
- [ ] The mini asset view should show: 
  - [x] The asset for the current market
  - [x] The icon from the icon store
  - [ ] Your balance of that asset
  - [ ] How much is allocated across any margin accounts or spot swap orders
  - [ ] How much is available collateral
  - [ ] A 3-dot menu to quick launch deposit / transfer / withdrawal actions
  - [ ] If those buttons are pressed it sends you to the portfolio page and opens the relevant tab
  - [ ] If the market is SPOT it will need to show both assets, condensed and maybe simplified side by side cards, Joelle is working on this now and should have something by the time someone gets to this
- [ ] If the user clicks on the asset pane it should expand to all assets and their breakdownas shown in ASSETS
  - [ ] If the user clicks on deposit / transfer / withdraw button here te same behaviour as above should happen
  - [ ] If the user clicks "manage" it should send the user to the new full portfolio page with the asset mgmt panel set to the same asset breakdown as we use in the sidebar #6110 
- [x] If the user clicks on the **market info** tab it should expand and show the full market info accordion inside the component, with the other cards shrinking to their minimised state as in MARKETS  (poorly named sketch) ... the preferred design is the second one (3.2)
- [x] In each case there is no "minimise" button on the open pane... you minimise by clicking on something else
- [x] The size of each section should be predefined and locked, with the expanded one always getting the same max size
- [x] Scrolling should scroll WITHIN each section if required, the sidebar should not scroll as a whole since the parts should be locked in place as above
- [ ] As a result of this change we can now remove the "Collateral" pane on the bottom table
- [ ] Ignore the sketches related to SWAP for now ... these will be dealt with seperately

## Sketch

All sketches now finalised in Figma: https://www.figma.com/file/ywcCZvXdvG28bg0AmCY9mF/Console-v3?type=design&node-id=6158-40256&mode=design&t=mj8twKYmzekRYkP3-0

mattrussell36:
# Related issues 🔗

Closes #6106 , #6110

# Description ℹ️

- Removes mini sidebar from all pages
- Adds new trading only sidebar.
- Adds a drawer to handle the current mobile experience
- Moves settings to top nav
- update accounts / collateral section

# Demo 📺

Screenshot 2024-05-20 at 11 37 16

https://github.com/vegaprotocol/frontend-monorepo/assets/6803987/346f626b-1fed-4f50-89cd-9f202947f340

# Technical details

- vaul added to handle the drawer component
- allotment downgraded to 1.17.1 as any later version and a three layer nested allotment grid does not respect the preferredSizes. Issue here: https://github.com/johnwalley/allotment/issues/583
- For now setting the sidebar view to withdraw/deposit/transfer has been replaced with navigating to the bespoke withdraw/deposit/transfer page


bwallacee:
Just incase its useful here is some early feedback:

1. The intial view needs to be configured so the chart takes up more of the space (currently the orderbook is too large)
image

2. Scroll bar for the trading sidebar looks out of place (the colour (looks like it needs a dark mode))
image
3. Changing the selected "tab" colour would be nice
image

4. Unsure if this is ok or if we would want to change
but the stop order dropdown, might look better right aligned and made so it depend go over the scrollbar?
image

5. Similar with tool tips
image

bwallacee:
@mattrussell36 

1. Should the font size match across these options?

image

2. The white scrollbar is now black

image

3. Bump for this 

> 3. Changing the selected "tab" colour would be nice

4. Why do we have a dark mode switcher if we have a button in the main nav as well
image
5. When the trading section is expanded the drop down for stop order doesn't allign:
image
6. I'm not getting anything in this list:
image

libs/ui-toolkit/src/components/bottom-drawer/bottom-drawer.tsx
```diff
@@ -0,0 +1,67 @@
+import {
+  type ComponentPropsWithoutRef,
+  type ElementRef,
+  forwardRef,
+} from 'react';
+import { Drawer as DrawerPrimitive } from 'vaul';
```
asiaznik, line 6: praise: Great for choosing this one, it works great!

bwallacee (approved):

JonRay15:
Need a solution for which assets you show on the asset breakdown .... because showing everything seems a little clunky.

We could 

1. Only show assets you have PLUS the current market asset 
2. Do somethign more fancy

Castellana9:
In a year filled with uncertainty and financial struggles for many, I'm thrilled to share a beacon of hope and triumph! Despite the market's unpredictability, I've achieved a remarkable milestone - earning over $900,000 in profits! This journey wasn't without its setbacks, though. I faced discouraging losses in 2020, but perseverance and the right guidance turned everything around.

Thanks to the expert advice of Veronica Tolan, an exceptional investment adviser I discovered on Facebook, I regained confidence and navigated the market with clarity. Her user-friendly platform and diverse trading options empowered me to make informed decisions and thrive.

If you're seeking a trusted partner to help you achieve financial freedom, I wholeheartedly recommend VERONICA TOLAN on Facebook. Reach out to her directly on WhatsApp at +44 7465283150 and discover a new path to financial success. Don't let market uncertainty hold you back - take control of your financial future today!

edd (approved):
👍 Code-wise I took a scan, but it's hard to review such a chunky PR. Trying it on a phone, it's working nicely. As @asiaznik said vaul seems to handle this nicely.

PR #6439, opened 2024-05-20
fix(trading): stacked cell rendering incorrectly

MadalinaRaicu:
# Related issues 🔗

Closes N/A

# Description ℹ️

Fixes stacked cell issue. 

Attempt for 24h price change. 

# Demo 📺

Screenshot 2024-05-20 at 12 34 33

# Technical 👨‍🔧

### DIFF
--- a/libs/datagrid/src/lib/cells/stacked-cell.tsx
+++ b/libs/datagrid/src/lib/cells/stacked-cell.tsx
@@ -1,33 +1,5 @@
 import type { ReactNode } from 'react';
 
-export const StackedCellWithIcon = ({
-  primary,
-  secondary,
-  primaryIcon,
-}: {
-  primary: ReactNode;
-  secondary: ReactNode;
-  primaryIcon?: ReactNode;
-}) => {
-  return (
-    <div className="leading-4">
-      <div
-        className="text-ellipsis whitespace-nowrap overflow-hidden text-sm flex"
-        data-testid="stack-cell-primary"
-      >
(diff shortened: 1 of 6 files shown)


bwallacee (approved):

libs/datagrid/src/lib/cells/stacked-cell.tsx
```diff
@@ -27,3 +27,28 @@ export const StackedCell = ({
...
+  primary,
+  secondary,
+}: {
+  primary: ReactNode;
+  secondary: ReactNode;
+}) => {
+  return (
+    <div className="leading-4">
+      <div
+        className="text-ellipsis whitespace-nowrap overflow-hidden text-sm "
```
mattrussell36, line 41: non-blocking: Whitespace character at the end of this string.

Also, rather than duplicating the component with some minor changes or adding a primaryIcon prop. IMO passing in a different element to handle the icon layout is preferable. Something like this

```
<StackedCell
  primary={
    <span className="whatever-you-need">
       {value}
        <span>
          <MarketProductPill productType={productType} />
        </span>
        <span className="ml-0.5">
          <MarketIcon data={data} />
        </span>
     </span>
  }
/>
```

mattrussell36 (approved):

libs/datagrid/src/lib/cells/stacked-cell.tsx
MadalinaRaicu: agree! updated

PR #6452, opened 2024-05-21
feat(governance): passed with errors batch proposal status

### LINKED ISSUE #6415: Enhance batch proposal status info
## The Chore

It seems that we need to be a bit more sophisticated in how we deal with the final status of batch proposals. 

The issue is that in some cases like the image below the overall batch can pass -  because the vote was successful - but then some parts of the batch can fail in the execution step. 

image

In these cases the core status will remain on STATE_PASSED because the overall batch has indeed PASSED ... but if we just show this alone in the governance app it is quite misleading to users. 

## Accceptance Criteria

- [ ] When a batch proposal has an overall status of STATE_PASSED 
  - [ ] Check all subproposals for their individual statuses 
  - [ ] If all proposals have STATE_PASSED then overall status shoudl remain "Passed"
  - [ ] If any proposal has <> STATE_PASSED 
    - [ ] Display "Passed with errors" on the overall status in top right of card
    - [ ] Next to the overall status (Proposal passed: conditions met for X of X proposals) add text indicating "but some proposals failed in execution, expand for details"
    - [ ] In the vote breakdown display additional red text indicating that proposal failed in execution along with the `errorDetails` given by Core
- [ ] In addition to this I think we should improve the colours on the overall status pill in the top right as I think currently they arent very intuitive:
  - [ ] Open proposals should be white 
  - [ ] Passed & enacted proposals should be green
  - [ ] Declined or rejected proposals should be red
  - [ ] "Passed with errors" proposals as defined above should be yellow
  - [ ] If there are any other statuses found in implementation let me know

asiaznik:
# Related issues 🔗

Closes #6415 

# Description ℹ️

* adds new overall proposal state: "Passed with errors" 
* changes colours as per 🎫 
* merges New spot market with New market and Update spot market with Update market
* small filter UI improvements

# Demo 📺

Screenshot 2024-05-21 at 17 40 05

image

image

# Technical 👨‍🔧

N/A


apps/governance/src/routes/proposals/components/proposal-detail-header/proposal-header.tsx
```diff
@@ -584,6 +586,24 @@ export const ProposalHeader = ({
...
+  let overallProposalState: ProposalState | AdditionalProposalState =
+    proposal.state;
+
+  // Mark proposal as "passed with errors" if batch is passed but any of the
+  // sub-proposal errored.
+  if (
+    proposal.state === ProposalState.STATE_PASSED &&
+    proposal.__typename === 'BatchProposal'
+  ) {
+    const subStates = compact(proposal.subProposals?.map((sub) => sub?.state));
```
mattrussell36, line 598: non blocking: possibly be clearer with Array.every?

```
const allSubPassed = compact(proposal.subProposals).every(p => p.state === ProposalState.STATE_ENACTED)
if (!allSubPassed) {
  overallProposalState = AddtionalProposalState.PASSED_WITH_ERRORS
}
```

apps/governance/src/routes/proposals/components/vote-breakdown/vote-breakdown.tsx
```diff
@@ -248,23 +251,45 @@ const VoteBreakdownBatch = ({
...
+    let description = t(
+      'Proposal passed: conditions met for {{count}} of {{total}} proposals',
+      {
+        count: passingCount['true'] || 0,
+        total: voteInfo.length,
+      }
+    );
+    let descriptionIconColor = 'text-vega-green';
+    if (proposal.__typename === 'BatchProposal') {
+      const subStates = compact(
```
mattrussell36, line 263: non blocking:  worth de-duping with a helper something like: `isBatchProposalFullyPassed`. I don't love this function name though

mattrussell36 (approved):

daro-maj (approved):

jeremyletang (approved):

PR #6456, opened 2024-05-22
feat(trading): asset breakdown in sidebar

### LINKED ISSUE #6106: Enhanced trading screen sidebar

bglownia:
# Related issues 🔗

Closes #6106

# Description ℹ️

Update accounts / collateral section


apps/trading/client-pages/market/trade-panels.tsx
```diff
@@ -211,7 +211,7 @@ const useViewLabel = (view: TradingView) => {
     activeOrders: t('Open'),
     inactiveOrders: t('Order history'),
     stopOrders: t('Advanced orders'),
-    collateral: t('Collateral'),
+    assets: t('Assets'),
```
mattrussell36, line 214: Wondering if we want to go ahead and add the asset management UI to the new BottomDrawer. Will get back to you

apps/trading/components/asset-card/asset-card.tsx
```diff
@@ -1,23 +1,66 @@
...
-
   return (
-    <span>
-      <span className="flex items-center gap-2 text-base">
-        <EmblemByAsset asset={asset.id} />
-        {asset.symbol}
-      </span>
-    </span>
+    <div className="p-3 border-b border-default text-left">
+      <div className="flex items-center mb-3">
```
mattrussell36, line 32: HTML could be a little better here. At a minimum the top level parent could be `<section>`, and this element should be `<header>` and perhaps a `<dl>` for available/allocation

libs/accounts/src/lib/accounts-table.tsx
```diff
@@ -1,340 +1,187 @@
...
+            onClick={() => actions.onClickWithdraw?.(asset.id)}
+            label={t('Withdraw')}
+            icon={VegaIconNames.WITHDRAW}
+          />
+        </div>
+      ) : (
+        <button
+          className="absolute inset-0"
+          onClick={() => setExpanded(true)}
+        ></button>
```
mattrussell36, line 165: Lets add some content inside the button fir a11y `<span className="sr-only">Show asset actions</div>`

libs/accounts/src/lib/accounts-table.tsx
```diff
@@ -1,340 +1,187 @@
...
-        // consider the pinned row when determining whether to show the no
-        // rows template or not so we need to override it
-        rowData?.length ? '<span />' : props.overlayNoRowsTemplate
-      }
-      getRowHeight={getPinnedAssetRowHeight}
-      pinnedTopRowData={pinnedRow ? [pinnedRow] : undefined}
-    />
+    rowData &&
+    (pinnedAssets && pinnedAssets.length
+      ? [
```
mattrussell36, line 178: I believe this is just for sorting purposes?

This would show intention more explicitly
```
 ? sortBy(rowData, (asset) => {
      if (pinnedAssets.includes(asset.id)) return -1
      return 1
    })
    : rowData
```

mattrussell36 (changes requested):

apps/trading/client-pages/portfolio/portfolio.tsx
```diff
@@ -160,6 +160,13 @@ const PortfolioActionTabs = () => {
           </div>
         </ErrorBoundary>
       </Tab>
+      <Tab id="assets" name={t('Assets')}>
+        <ErrorBoundary feature="portfolio-assets">
+          <TinyScroll className="overflow-y-auto overflow-x-hidden h-full">
```
mattrussell36, line 165: tiny scroll no longer requires these classNames

apps/trading/components/sidebar/sidebar-accordion.tsx
```diff
@@ -68,7 +68,7 @@ const SidebarAccordionContent = forwardRef<
     className="flex-1 text-sm min-h-0"
     {...props}
   >
-    <TinyScroll className="overflow-y-auto overflow-x-hidden h-full p-2">
+    <TinyScroll className="overflow-y-auto overflow-x-hidden h-full">
```
mattrussell36, line 71: Can remove overflow classes and h-full

PR #6460, opened 2024-05-22
feat(trading): simple swap 

### LINKED ISSUE #6033: Simple swap UI in asset mgmt component
## Story

As a user
I want a simple swap UI screen 
So that I can easily exchange my Ethereum USDT for Arbitrum USDT and vice versa once the Axellar bridge is done, and can use it for other assets in the future

## Acceptance Criteria

- [x] There is a simple screen modelled on unswap or similar for swapping spot assets
- [x] This would be displayed in both the portfolio page and the trading sidebar
- [x] The form sends an order to the relevant spot market to trade for you
  - [x] Market ID (will need to be found from the asset ID of the asset you're swapping to)
  - [x] If paying the QUOTE asset to receive the BASE asset use BUY and use the BESK ASK price to calculate received amount
  - [x] If paying the BASE asset to receive the QUOTE asset use SELL and use the BESK BID price to calculate the received amount
  - [x] Size is just taken from the form
  - [x] The form should allow you to enter a price impact tolerance
  - [x] Send the order as a LIMIT FOK where the price is determine by the tolerance, eg. 
    - [x] If order type is BUY use BEST ASK + TOLERANCE
    - [x] If order type is SELL use BEST BID - TOLERANCE
- [x] The form gives you a estimate of the average price of the asset you will receive
- [x] The form sends a TX to the wallet to be confirmed
- [x] Once confirmed the form tells you how much of that asset you have received and links you to explorer 
- [x] There is a link to the spot market where this swap is being executed
- [x] If the TX fails because the FOK cant execute the toast needs to show this failure AND we will display red text beneath the swap button to let the user know

## Tasks

- [x] UX (if needed)
- [x] Design (if needed)
- [x] Implementation
- [x] Code review
- [x] QA review

## Sketch

See "SWAP" drawing here:

https://www.figma.com/file/ywcCZvXdvG28bg0AmCY9mF/Console-v3?type=design&node-id=6158-40256&mode=design&t=ABCsSihKRELNzJxI-0

Current thinking is this will either live in the sidebar or if that is too busy with all the options we're now ending up with we could have a standalone page similar to what we have fo deposits etc ... I think building that out into an "Assets" page is the best approach here.

MadalinaRaicu:
# Related issues 🔗

Closes #6033 

# Description ℹ️

- [x] There is a simple screen modelled for swapping spot assets
- [x] This would be displayed in both the portfolio page and the trading sidebar
- [x] The form sends an order to the relevant spot market to trade for you
  - [x] Market ID (will need to be found from the asset ID of the asset you're swapping to)
  - [x] Based on that price impact calculate the price of the limit order that will adhere to that and send a LIMIT FOK order, ie. If price was $10 and we said price impact max is 10% we would send a FOK limit order at $11 and it would fail if that cant be executed
  - [x] Size is just taken from the form
- [x] The form allows you to set price impact tolerance 
- [x] The form gives you a estimate of the average price of the asset you will receive
- [x] The form sends a TX to the wallet to be confirmed
- [x] There is a link to the spot market where this swap is being executed
- [x] If the TX fails because the FOK cant execute the toast needs to show this failure 
- [x] Feature flag added
- [x] Tick size validation

# Demo 📺

Screenshot 2024-06-05 at 21 51 25

# Technical 👨‍🔧


JonRay15:
Need tick size validation in here. 

We saw an issue in the demo where with the price impact tolerance applied the derived value seemingly didnt meet the tick size validation.

JonRay15:
The auto / custom pill is a bit misleading ... we're thinking maybe get rid of it and instead show teh "Custom" label next to / in the input field for the custom tolerance?

Then we shoudl just highlight the button selected or the input field in a lighter colour to show that is the one you've selected?

image

JonRay15:
Can we turn the swap button blue?

apps/trading/client-pages/markets/markets-sidebar.tsx
```diff
@@ -49,6 +49,12 @@ export const MarketsSidebar = () => {
         tooltip={t('Transfer')}
         routeId={currentRouteId}
       />
+      <SidebarButton
```
MadalinaRaicu, line 52: I will remove this from here

apps/trading/components/sidebar/sidebar.tsx
```diff
@@ -17,7 +19,7 @@ import { useConnect, usePubKeys } from '@vegaprotocol/wallet-react';
 import { useGetCurrentRouteId } from '../../lib/hooks/use-get-current-route-id';
 import { useT } from '../../lib/use-t';
 import { ErrorBoundary } from '../error-boundary';
-import { useScreenDimensions } from '@vegaprotocol/react-helpers';
```
mattrussell36, line 20: SwapContainer is used in the sidebar and also on the swap page so it should not be contained in client-pages/swap but in components. Any components in a client-pages/mypage should be only used on that page anything shared should be in /components

apps/trading/client-pages/swap/swap-form.tsx
mattrussell36, line 1: I don't like the naming/structure of the files.

This file name indicates that the form and therefore the core form logic is within it, when its actually just mostly presentational form components. Can we rename it swap-form-elements or perhaps split each out into separate files swap-asset-input etc

Additionally the container component contains both form logic and fetching logic for the component. Its best to split these

I would refactor this so its something more like

- swap-container: fetch required data
- swap-form: manage form logic
- swap-form-elements: any presentational components/form elements

apps/trading/client-pages/swap/swap-form.tsx
```diff
@@ -0,0 +1,201 @@
...
+  accountWarning = true,
+  pubKey,
+}: {
+  label: string;
+  amount: string;
+  asset?: AssetFieldsFragment;
+  balance?: string;
+  accountAssetIds?: string[];
+  assets?: Record<string, AssetFieldsFragment>;
+  onAmountChange: (e: React.ChangeEvent<HTMLInputElement>) => void;
```
mattrussell36, line 36: The codebase doesn't use the top level react object. Can you `import type { ChangeEvent } from 'react'`

apps/trading/client-pages/swap/swap-container.tsx
```diff
@@ -0,0 +1,307 @@
...
+  baseAmount: string;
+  quoteAmount: string;
+  priceImpactTolerance: string;
+}
+
+const derivePrice = (
+  marketData: MarketDataFieldsFragment | null,
+  side: Side,
+  toleranceFactor: number,
+  market: MarketFieldsFragment
```
mattrussell36, line 57: The `market` argument doesn't appear to be used within `derivePrice`

apps/trading/client-pages/swap/swap-container.tsx
```diff
@@ -0,0 +1,307 @@
...
+
+  const { data: marketData, reload: reloadMarketData } = useDataProvider({
+    dataProvider: marketDataProvider,
+    variables: { marketId },
+    skip: !marketId,
+  });
+  const { data: accounts, reload: reloadAccounts } = useAccounts(pubKey);
+
+  useEffect(() => {
+    const intervalAccounts = setInterval(reloadAccounts, POLLING_TIME);
```
mattrussell36, line 93: We shouldn't need to poll it accounts, the data provider has a subscription. If its not always got the latest data something is broken elsewhere

apps/trading/client-pages/swap/swap-container.tsx
```diff
@@ -0,0 +1,307 @@
...
+    <form
+      onSubmit={!isReadOnly && pubKey ? handleSubmit(onSubmit) : noop}
+      noValidate
+      data-testid="swap-form"
+    >
+      <div className="flex justify-between items-center mb-4">
+        <h3 className="text-lg">{t('Swap')}</h3>
+        {marketId && (
+          <Link
+            href={`/#/markets/${marketId}`}
```
mattrussell36, line 240: Use the Links object functions

apps/trading/client-pages/swap/swap-container.tsx
```diff
@@ -0,0 +1,307 @@
...
+    setValue('quoteAmount', baseAmount);
+  };
+
+  const onSubmit = () => {
+    if (!marketId || !side || !market || !orderSubmission) return;
+    create({ orderSubmission });
+  };
+
+  useEffect(() => {
+    if (!orderSubmission || !market) return;
```
mattrussell36, line 202: Setting state from within an effect like this is a code smell. Its extremely hard to know how and when this should get triggered. 

It looks like you are setting a price value based on a number of different inputs. I would much rather:

- just derive the value based on other state and/or
- calculate it once on the user clicking submit

apps/trading/client-pages/swap/swap-form.tsx
```diff
@@ -0,0 +1,201 @@
...
+        </DropdownMenuTrigger>
+      }
+    >
+      <DropdownMenuContent className="bg-gray-700 rounded-md mt-2">
+        {assets &&
+          Object.values(assets).map((asset) => (
+            <DropdownMenuItem
+              onClick={(e) => {
+                e.preventDefault();
+                e.stopPropagation();
```
mattrussell36, line 188: I'd much prefer if the dropdown closed on selection of the asset. Its quite annoying it stays open

mattrussell36 (changes requested):

apps/trading/client-pages/swap/swap-container.tsx
MadalinaRaicu: i will add back the tick size logic that uses it

apps/trading/client-pages/swap/swap-form.tsx
MadalinaRaicu: updated

apps/trading/client-pages/swap/swap-container.tsx
MadalinaRaicu: updated. removed polling altogether

apps/trading/client-pages/swap/swap-container.tsx
MadalinaRaicu: updated

apps/trading/components/sidebar/sidebar.tsx
MadalinaRaicu: updated the structure

apps/trading/client-pages/swap/swap-container.tsx
MadalinaRaicu: added back, for the tick size logic

apps/trading/client-pages/swap/swap-form.tsx
MadalinaRaicu: of course, updated.

apps/trading/client-pages/swap/swap-container.tsx
MadalinaRaicu: I went for the first option. Could you please review again?

apps/trading/components/swap/swap-container.tsx
```diff
@@ -0,0 +1,101 @@
...
+import { SwapForm } from './swap-form';
+import { Side } from '@vegaprotocol/types';
+
+export const SwapContainer = () => {
+  const { pubKey } = useVegaWallet();
+  const { data: markets } = useMarketsMapProvider();
+  const { data: assetsData } = useAssetsMapProvider();
+  const { data: accounts } = useAccounts(pubKey);
+
+  const [quoteAsset, setQuoteAsset] = useState<AssetFieldsFragment>();
```
mattrussell36, line 27: I don't understand why there is so much state requried. As far as I can see everything derives off the two selected assets? This could be much more succinct

apps/trading/components/swap/swap-container.tsx
```diff
@@ -0,0 +1,101 @@
...
+  }, [assetsData, markets]);
+
+  const chooseMarket = useCallback(() => {
+    if (!spotMarkets || !baseAsset || !quoteAsset) return;
+    return spotMarkets.find((m) => {
+      const mBaseAsset = getBaseAsset(m);
+      const mQuoteAsset = getQuoteAsset(m);
+      if (!mBaseAsset || !mQuoteAsset) return false;
+      if (mBaseAsset.id === baseAsset.id && mQuoteAsset.id === quoteAsset.id) {
+        setSide(Side.SIDE_BUY);
```
mattrussell36, line 67: Side should be derived from the selected assets -> swap market

apps/trading/components/swap/swap-container.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+  useEffect(() => {
+    const market = chooseMarket();
+    setMarketId(market?.id || '');
+    setMarket(market);
+  }, [chooseMarket, setMarketId, setMarket]);
+
+  return (
+    <SwapForm
+      marketId={marketId}
```
mattrussell36, line 86: marketId seems unnecessary as the entire market object is being passed as a prop

apps/trading/components/swap/swap-container.tsx
```diff
@@ -0,0 +1,101 @@
...
+      marketData={marketData}
+      baseAsset={baseAsset}
+      setBaseAsset={setBaseAsset}
+      quoteAsset={quoteAsset}
+      setQuoteAsset={setQuoteAsset}
+      side={side}
+      setSide={setSide}
+      market={market}
+      navigate={navigate}
```
mattrussell36, line 95: navigate unnecessary if you use the Link component correctly

mattrussell36 (changes requested):

apps/trading/components/swap/swap-form.tsx
```diff
@@ -0,0 +1,252 @@
...
+          pubKey={pubKey}
+          testId="you-receive"
+        />
+      </div>
+      <PriceImpactInput
+        value={priceImpactTolerance || ''}
+        onValueChange={(e) => setValue('priceImpactTolerance', e)}
+        data-testid="price-impact-input"
+      />
+      <button
```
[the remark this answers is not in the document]
MadalinaRaicu: This is our current Intent.None. So we can directly use that?

apps/trading/components/swap/swap-form.tsx
```diff
@@ -0,0 +1,386 @@
...
+  side: Side,
+  toleranceFactor: number,
+  market: MarketFieldsFragment
+) => {
+  if (!marketData) return;
+  const isBid = side === Side.SIDE_SELL;
+  const price = isBid
+    ? new BigNumber(marketData.bestBidPrice).times(1 - toleranceFactor)
+    : new BigNumber(marketData.bestOfferPrice).times(1 + toleranceFactor);
+  return roundUpToTickSize(price, market.tickSize, isBid);
```
asiaznik, line 54: Note, that `tickSize` is always associated with the QUOTE asset, and BASE is always associated with `positionDecimalPlaces` regardless of the side of the order.

asiaznik: In other words, `tickSize` should be takes into consideration when changing the amount of QUOTE (regardless of "You pay" or "You receive")

And `positionDecimalPlaces` when changing the amount of BASE

apps/trading/components/swap/swap-form.tsx
```diff
@@ -0,0 +1,386 @@
...
+
+  const quoteAssetBalance = getQuoteAssetBalance(quoteAsset, accounts);
+  const baseAssetBalance = getQuoteAssetBalance(baseAsset, accounts);
+
+  const side = useSide({ market, baseAsset, quoteAsset });
+
+  const marketPrice = useMarketPrice({ marketData, side });
+  const prevMarketPrice = usePrevious(marketPrice);
+
+  const handleSwapAssets = () => {
```
asiaznik, line 93: I don't like that it's confusing. BASE and QUOTE are always the same for given asset pair (unless there's a market with flipped assets).

This should only change the SIDE of the order BUY / SELL

apps/trading/components/swap/swap-form.tsx
```diff
@@ -0,0 +1,386 @@
...
+
+  const onSubmit = (e: FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+
+    // Check users is connected
+    if (isReadOnly || !pubKey) return;
+    if (!Number(quoteAmount)) return;
+
+    if (!market) {
+      throw new Error('could not dervie spot market for swap');
```
asiaznik, line 111: This should also be caught earlier

apps/trading/components/swap/swap-form.tsx
```diff
@@ -0,0 +1,386 @@
...
+      marketData?.bestOfferPrice || '0',
+      market.decimalPlaces
+    );
+    return new BigNumber(amount).dividedBy(price).toString();
+  }
+
+  return '';
+};
+
+const deriveSide = ({
```
asiaznik, line 310: This is also confusing, the side of the order should only be changes via the "arrow" button.

apps/trading/components/swap/swap-form.tsx
```diff
@@ -0,0 +1,386 @@
...
+          testId="you-receive"
+        />
+      </div>
+      <PriceImpactInput
+        value={tolerance}
+        onValueChange={(value) => setTolerance(value)}
+        data-testid="price-impact-input"
+      />
+
+      <TradingButton
```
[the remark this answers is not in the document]
MadalinaRaicu: updated to have them disabled

apps/trading/components/swap/swap-form.tsx
MadalinaRaicu: I agree. Actually I don't think we need to throw an error here

apps/trading/components/swap/swap-form.tsx
MadalinaRaicu: tickSize will be applied only on the market price

apps/trading/components/swap/swap-form.tsx
MadalinaRaicu: yes, but it's not just that. 
the is also the initial state where the user selects the assets and then we need to find out if there is a market matching those two selected assets, and then we need to derive if it is a buy or a sell.

MadalinaRaicu: but the side changes only via the arrow button, and in an initial version we were keeping state

bwallacee (approved):

apps/trading/components/swap/swap-form.tsx
MadalinaRaicu: we have another issue here though that I am about to resolve. 

the size set up is not correct and we need to fix this.

MadalinaRaicu: changes incoming

MadalinaRaicu: updated based on your feedback

mattrussell36 (approved):

PR #6462, opened 2024-05-23
chore(trading): use custom genesis

bwallacee:
Moved some of the updates we do to the network param in tests to our base genesis file. This should simplify the tests and make them more robust.

I have also removed some tests that I think were basically duplication in our rewards tests.


apps/trading/e2e/tests/market_lifecycle/test_market_lifecycle.py
```diff
@@ -74,9 +74,13 @@ def test_market_lifecycle(proposed_market, vega: VegaServiceNull, page: Page):
         wallet=vega.wallet,
     )
 
+    vega.wait_fn(1)
+    vega.wait_for_total_catchup()
     # "wait" for market to be approved and enacted
     next_epoch(vega=vega)
+    next_epoch(vega=vega)
     # check that market is in pending state
+    page.pause()
```
mattrussell36, line 83: should we leave these in?

mattrussell36 (approved):

apps/trading/e2e/tests/market_lifecycle/test_market_lifecycle.py
bwallacee: removed

daro-maj (approved):

PR #6476, opened 2024-05-28
feat(trading): asset activity

### LINKED ISSUE #6101: Transaction list in portfolio
## Story

As a user
I want to see a full list of transactions I have made
So that I know what I have done

## Acceptance Criteria

Note; this ticket is a stepping stone to a large rework of the portfolio page, but for now we will aim to create the transaction list simply in teh current AG grid structure.  After this we can move it into the new design Joelle is working on.

- [x] Modify the Collateral / Deposits / Withdrawals table in Portfolio to be Collateral & Transaction List, where this list combines all of deposits, transfers and withdrawals into a single structure
- [x] Need a way to fetxh some number of deposits, withdrawals and transfers for the connected key and order them by timestamp descending, such that we can load more if needed and dont lose any rows
- [x] A list of all transactions showing the following columns
  - [x] Type (deposit, transfer in, transfer out, withdrawal)
  - [x] Asset code (looked up from assets API)
  - [x] Datetime
  - [x] Amount (converted with asset decimals from assets API)
  - [x] To / From
  - [x] Status 
- [x] To / From field will be conditional
  - [x] For deposit it will always be the source ERC20 address
  - [x] For transfer TO the current key show the sender Vega key
  - [x] For transfer FROM the current key show the recipient Vega key
  - [x] For withdrawal show the ERC20 address
- [ ] Status will also be conditional
  - [x] For deposit: 
    - [x] Show "Pending (X / Y Confirmations)" 
    - [x] Then show "Complete"
  - [x] For withdrawal
    - [ ] ~~Show "Pending - approve in wallet" while waiting for wallet~~ N/A
    - [x] Then "Pending - sent from network" once a txHash is available
    - [x] Then "Complete - received on target network" once a ref is available (i think)
  - [x] For transfer 
    - [ ] ~~Show "Pending - approve in wallet" while waiting for wallet~~ N/A
    - [x] Then "Complete" once done ... dont think it is worth an interim message when its so fast
- [x] Should be possible to filter on the type to view only deposits / transfer / withdrawals
- [x] There is currently a consditional "deposit" and "make withdrawal" button .... remove these since there will be a new panel due to #6267 
- [x] Make sure we are showing all withdrawals correctly; there is a suspision in mainnet currently that we're doing something wrong here and hence not showing failed withdrawals ... but with extra Console validation now its very hard to verify this 
  - [x] Must ensure if there are STATUS_REJECTED withdrawals we are showing them... this might mean we need to be able to sign withdrawals under the required minimum and send them so Core fails them instead of Console / VegaWallet
  - [x] Must ensure we're showing STATUS_OPEN withdrawals when they appear ... this will be very brief since it only lasts the time it takes the validators to process, but we need to show it
  - [x] If it is STATUS_FINALISED and there is no txHash that means it has not been finalised on ethereum and we need to show a status for this, and a prompt to "Complete withdrawal"
  - [x] IF it is STATUS_FINALISED and has a txHash then it is fully complete
- [x] Ensure that transfers both TO and FROM are shown
- [x] Ensure that pending and failed / rejected withdrawals are shown
- [x] Show the chain name in brackets after the asset code
- [x] If there are incomplete withdrawals 
  - [x] We should pin these to the top of the list and potentially highlight them in some way
  - [x] We should add some sort of indicator to the menu to highlight the Portfolio page in the nav so they know to go there and complete it.

## Sketch

Suggest the easiest way to initially build this thing is to enhance the current table which shows collateral / deposits / withdrawals

That way we can do the backend work and just dump it into new tabe here. 

Image

Then in the long term Joelle is doing a redesign of this page to land on somethign more like the below.

image

Summary of logic:

image

mattrussell36:
# Related issues 🔗

Closes #6101 

# Description ℹ️

Adds an asset activity view which shows a time ordered list of deposits, withdrawals and transfers

# Demo 📺

Screenshot 2024-05-28 at 15 39 13
ctionality if applicable

# Technical 👨‍🔧

### DIFF
--- a/libs/withdraws/src/lib/test-helpers.ts
+++ b/libs/withdraws/src/lib/test-helpers.ts
@@ -59,6 +59,7 @@ export const generateWithdrawal = (
         contractAddress: '0x123',
         lifetimeLimit: '1',
         withdrawThreshold: '1',
+        chainId: '1',
       },
     },
     party: {
--- a/libs/withdraws/src/lib/use-ready-to-complete-withdrawals-toast.spec.ts
+++ b/libs/withdraws/src/lib/use-ready-to-complete-withdrawals-toast.spec.ts
@@ -20,9 +20,13 @@ const NO_THRESHOLD_ASSET: Asset = {
   symbol: 'NTA',
   decimals: 1,
   status: Types.AssetStatus.STATUS_ENABLED,
+  quantum: '1',
   source: {
     __typename: 'ERC20',
     contractAddress: '0xnta',
+    chainId: '1',
+    lifetimeLimit: '1',
+    withdrawThreshold: '1',
   },
 };
 
@@ -33,9 +37,13 @@ const LOW_THRESHOLD_ASSET: Asset = {
   symbol: 'LTA',
   decimals: 1,
   status: Types.AssetStatus.STATUS_ENABLED,
+  quantum: '1',
   source: {
     __typename: 'ERC20',
     contractAddress: '0xlta',
+    chainId: '1',
+    lifetimeLimit: '1',
+    withdrawThreshold: '1',
   },
 };
 
@@ -46,9 +54,13 @@ const HIGH_THRESHOLD_ASSET: Asset = {
   symbol: 'HTA',
   decimals: 1,
   status: Types.AssetStatus.STATUS_ENABLED,
+  quantum: '1',
   source: {
     __typename: 'ERC20',
     contractAddress: '0xhta',
+    chainId: '1',
+    lifetimeLimit: '1',
+    withdrawThreshold: '1',
   },
 };
 
@@ -73,6 +85,9 @@ const mockIncompleteW1: Withdrawal = {
   withdrawnTimestamp: null,
   txHash: null,
   asset: NO_THRESHOLD_ASSET,
+  party: {
+    id: 'party-id',
+  },
 };
 
 // delay (10 + 1000 < 5000), ready
@@ -85,6 +100,9 @@ const mockIncompleteW2: Withdrawal = {
   withdrawnTimestamp: null,
   txHash: null,
   asset: LOW_THRESHOLD_ASSET,
+  party: {
+    id: 'party-id',
+  },
 };
 // delay (4500 + 1000 > 5000), below threshold, ready
 const mockIncompleteW3: Withdrawal = {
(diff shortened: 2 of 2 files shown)


JonRay15:
Need to remove the deposit & withdrawal tabs.

Think we need something to replace the notification on the withdrawals tab that you still have open ones. Coudl be that we include that button on the main list, or maybe we have another quickl button to filter to show only those pending ones?

Do we need a button to launch deposit / withdrawal?   I am not sold on this since they are literally in the asset mgmt panel to the left anyway?

apps/trading/components/asset-activity/asset-activity.tsx
```diff
@@ -0,0 +1,355 @@
...
+    transfersPageInfo?.hasNextPage,
+  ].some((hasNextPage) => hasNextPage);
+
+  const rowData = useMemo(() => {
+    return compact(
+      [...(deposits || []), ...(withdrawals || []), ...(transfers || [])].map(
+        normalizeItem
+      )
+    );
+  }, [deposits, withdrawals, transfers]);
```
edd, line 108: **praise (non-blocking)**:  this is nicely encapsulated for the amount of data munging it's doing

apps/trading/components/asset-activity/withdrawal-status-cell.tsx
```diff
@@ -0,0 +1,123 @@
...
+        }
+      }
+    };
+
+    checkStatus();
+
+    return () => {
+      clearTimeout(timeoutRef.current);
+    };
+  }, [getThreshold, getDelay, withdrawal]);
```
edd, line 120: **praise (non-blocking)**: Also nice and neat for something quite fiddly

libs/deposits/src/lib/deposits-provider.ts
```diff
@@ -41,4 +60,25 @@ export const depositsProvider = makeDataProvider<
     );
     return uniqBy([...incoming, ...(data || [])], 'id');
   },
+  pagination: {
+    getPageInfo,
+    append,
+    first: 1000,
```
edd, line 66: **question (non-blocking)* *: is 1k a sensible default? It feels high but I'm not very experienced with this api

edd (approved):
Approved, with 1 non-blocking question

PR #6512, opened 2024-06-04
feat(trading): deposits and withdrawals via arbitrum chain

### LINKED ISSUE #6396: Same chain deposit & withdrawals (Arbitrum)
**NOTE** This ticket is no longer about using Squid SDK to do the same chain deposits ad withdrawals. We've done in natively (see linked PR).

## The Chore

## Acceptance Criteria

- [ ] There should now be two deposit modes which will appear as tabs in the asset mgmt panel:
  - [ ] "Deposit"
    - [x] This mode covers all same chain deposits and should use the existing UI with some enhancements
    - [ ] Add a chain selector that can be either Ethereum or Arbitrum (both with icons)
    - [ ] Instead of simply "Asset" we will now need "To Asset" and "From Asset" (again with icons)
    - [ ] If the Ethereum chain was selected 
      - [ ] "From Asset" should be disabled and we should set it to match whatever "To Asset" is selected (ie. we dont support swaps in the legacy flow) 
      - [ ] Process the deposit using the legacy flow on the backend
    - [ ] If the Arbitrum chain was selected then 
      - [ ] User should be able to set the To Asset based on a dropdown of assets on Vega
      - [ ] I guess we will need to look into the user's connected wallet to pull out a list of available assets for From Asset
      - [ ] If the From and To assets are the same then interact directly with the Arbitrum bridge to make the deposit
      - [ ] If the From and To assets are different then use the Squid Router SDK to swap the asset and deposit it on Arbitrum bridge
- [ ] Bonus points: add a search box to the asset selectors which filters down the asset list as you type
- [ ] Support withdrawals on same chain / cross chain too
## Not in scope

There was discussion that maybe if there was a relevant spot market on Vega that it might be cheaper to deposit same asset and THEN swap it on vega and could we build that detection into the deposit flow??  This is a nice idea, but not in scope for now so has been raised as #6397 

Will look like this, but with some enhancements to that deposit pane

image

asiaznik:
# Related issues 🔗

Closes #6396 

# Description ℹ️

* adds support for the EVM chains (Arbitrum)
* allows deposits from Arbitrum
* allows withdrawals to Arbitrum
* adds icons to AssetOption component
* adds chain icons to Deposit / Withdrawal forms

# Demo 📺

Screenshot 2024-06-07 at 14 29 57
Screenshot 2024-06-07 at 14 30 06
Screenshot 2024-06-07 at 14 30 16
Screenshot 2024-06-07 at 14 30 23

# Technical 👨‍🔧

N/A


libs/web3/src/lib/use-get-withdraw-delay.ts
```diff
@@ -10,37 +10,66 @@ import { localLoggerFactory } from '@vegaprotocol/logger';
...
 const MAX_AGE = 5 * 60 * 1000; // 5 minutes
 
-type TimestampedDelay = { value: number | undefined; ts: number };
-const DELAY: TimestampedDelay = { value: undefined, ts: 0 };
+type TimestampedDelay = {
+  chainId: number;
+  value: number | undefined;
+  ts: number;
+};
+const DELAYS: TimestampedDelay[] = [];
```
mattrussell36, line 18: non blocking: If you're indexing by chainId, wouldn't using a map make it easier to manage the cache

libs/web3/src/lib/web3-connectors.ts
```diff
@@ -118,7 +114,31 @@ export const createConnectors = (
...
+  walletMnemonic?: string
+) => {
+  if (Object.values(transports).length === 0) {
+    throw new Error('ERR_CONNECTORS: could not create connectors');
+  }
+
+  return [
+    initializeUrlConnector(localProviderUrl, walletMnemonic),
+    initializeMetaMaskConnector(),
+    // TODO: This needs to be tested out as it only allows single rpc connection
```
mattrussell36, line 138: @daro-maj can you check this

libs/withdraws/src/lib/withdraw-form.tsx
```diff
@@ -325,31 +408,55 @@ const UseButton = (props: UseButtonProps) => {
...
       data-testid="disconnect-ethereum-wallet"
     >
       {t('Disconnect')}
-    </UseButton>
+    </ButtonLink>
   );
 };
+
+// const EthereumButton = ({ clearAddress }: { clearAddress: () => void }) => {
+//   const t = useT();
```
mattrussell36, line 436: Remove

mattrussell36 (changes requested):

libs/assets/src/lib/asset-option.tsx
```diff
@@ -16,39 +15,74 @@ export const Balance = ({
...
+      -
     </div>
   );
 };
 
 export const AssetOption = ({ asset, balance }: AssetOptionProps) => {
+  const chainId = asset.source.__typename === 'ERC20' && asset.source.chainId;
+  const assetSource =
+    asset.source.__typename === 'ERC20' && asset.source.contractAddress;
+  let assetLogo = 'https://icon.vega.xyz/missing.svg';
```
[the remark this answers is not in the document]
asiaznik: Not at the moment

libs/web3/src/lib/use-bridge-contract.ts
```diff
@@ -39,3 +52,165 @@ export const useBridgeContract = (allowDefaultProvider = false) => {
...
+    return undefined;
+  }, [
+    activeChainId,
+    activeProvider,
+    allowDefaultProvider,
+    chainId,
+    defaultProvider,
+  ]);
+
+  useEffect(() => {
```
mattrussell36, line 159: Non blocking: Might save some render cycles if you just derive an error?

```
  let error
  if (chainId && !isSupportedChainId(chainId)) {
    error = ERR_CHAIN_NOT_SUPPORTED
  }
  else if (activeChainId !== chainId) {
    error = ERR_WRONG_CHAIN
  }

```

asiaznik: removed that code, not needed anymore

libs/web3/src/lib/web3-connectors.ts
asiaznik: Tested it with coinbase and it works fine for me

daro-maj (approved):

PR #6521, opened 2024-06-10
feat(trading): distinguish between eth and arb assets

### LINKED ISSUE #6464: Ensure clarity around USDT Ethereum vs Arbitrum
## The Chore

- Needs to be clear the difference between the two; anywhere that code is used more info might be needed to make sure it was obvious
- If a user is on a market which needs USDT Eth and they have USDT Arb make it clear that they can swap them or vice versa

## Tasks

- [ ]
- [ ]

mattrussell36:
# Related issues 🔗

Closes #6464 

# Description ℹ️

- Removes dupcliated asset field fragments so that chainId is always present
- Adds ChainIdMapShort for 'Arb' and 'Eth'
- Uses the short chain names in various places such as the collateral table
- Updates the pills that swap between notional/quote values with short chain names
- Shows short chain name in market headers

# Demo 📺

Screenshot 2024-06-10 at 15 31 04

# Technical 👨‍🔧


JonRay15:
We seem to have some recurrence of that weird "no accounts" bug appearing in the collateral table?  Dont know if its related or just we havent noticed. 

image

JonRay15:
Made various new comments on the origial ticket ... can we address these before merging please.

JonRay15:
It is long and ugly on testnet ... but in mainnet it will just say "Arb" or "Eth" so I think wil be ok!

image

bwallacee:
Should the asset details Source chain say Eth Sepolia (like it does for Arb Sepolia):
image

Should we have a space after tUSDC?:
image
Missing chain on deposits and withdraws tables in portfolio:
image

Should we have something to distinguish the chain the markets in order history are? (same for the other positions tabs)
image

daro-maj:
Do we need to specify chain also here ?
image

daro-maj:
Do we need to specify chain also here ?
image

daro-maj:
Is the governance is in the scope ?
image

PR #6530, opened 2024-06-11
fix(trading): fix telemetry prompt translation name of dApp

### LINKED ISSUE #6444: Refactor Emblem to take `size` instead of `width` and `height`
## The Chore

`Emblem` currently provides `width` and `height` properties. There are no instances where `width` and `height` have been different. It's also *slightly* weird that for a market icon you can provide `width` and `height`, but that is the `width`/`height` of each icon individually, rather than the containing element. `size` feels a little clearer

## Tasks

- [ ] Remove `width` and `height` prop from all Emblem components
- [ ] Add `size` prop
- [ ] Refactor all usages to provide `size` instead of `width`/`height`

MadalinaRaicu:
# Related issues 🔗

Closes #6444

# Description ℹ️

- [x] Fix telemetry prompt translation name of dApp
- [x] Fix the order of the buttons

# Demo 📺

image

# Technical 👨‍🔧


bwallacee (approved):

libs/i18n/src/locales/en/trading.json
```diff
@@ -182,7 +182,7 @@
   "Governance vote passed to close the market": "Governance vote passed to close the market",
   "Governance": "Governance",
   "Help identify bugs and improve the service by sharing anonymous usage data.": "Help identify bugs and improve the service by sharing anonymous usage data.",
-  "Help us identify bugs and improve Vega Governance by sharing anonymous usage data.": "Help us identify bugs and improve Vega Governance by sharing anonymous usage data.",
+  "TelemetryModalIntro": "Help us identify bugs and improve Vega Console by sharing anonymous usage data.",
```
mattrussell36, line 185: Alphabetical order please

mattrussell36 (changes requested):

mattrussell36 (approved):

PR #6532, opened 2024-06-12
fix(trading,withdraws,deposits): arb transactions

mattrussell36:
# Related issues 🔗

N/A

# Description ℹ️

- Prompt to switch wallet on withdraw ui
- Prompt to switch wallet on complete withdrawal toast
- Remove red 'switch chain' message from the bottom of the withdraw and deposit
- Fix to/from ethereum address labels on withdraw and deposit form
- Show chain name on the change chain toast

# Demo 📺

# Technical 👨‍🔧


libs/assets/src/lib/utils.ts
```diff
@@ -10,9 +10,8 @@ export const getAssetSymbol = (asset: AssetFieldsFragment) => {
   let symbol = asset.symbol;
 
   if (asset.source.__typename === 'ERC20') {
-    symbol = `${asset.symbol}(${
-      SUPPORTED_CHAIN_SHORT_LABELS[asset.source.chainId]
-    })`;
+    const chainLabel = getExternalChainShortLabel(asset.source.chainId);
+    symbol = `${asset.symbol}(${chainLabel})`;
```
edd, line 14: **nitpick (non-blocking) **: `getExternalChainShortLabel` will return `''` if the chain is unknown so this could end up being, say (`USDT()`) which would be a bit confusing. Better would be just returning the chain ID. Nitpick as it's unlikely to occur.

libs/environment/src/external-chain.ts
```diff
@@ -53,3 +53,11 @@ export function getExternalChainLabel(chainId?: string) {
     return 'Custom Chain';
   }
 }
+
+export function getExternalChainShortLabel(chainId?: string) {
+  if (chainId && SUPPORTED_CHAIN_IDS.includes(chainId)) {
+    return SUPPORTED_CHAIN_SHORT_LABELS[chainId];
+  } else {
+    return '';
```
edd, line 61: **nitpick (non-blocking)**:  as per the previous comment, I'd just return `chainId` rather than blank string, as it's better than nothing.

libs/withdraws/src/lib/use-ready-to-complete-withdrawals-toast.tsx
```diff
@@ -276,11 +278,29 @@ const SingleReadyToWithdrawToastContent = ({
...
+            });
+          }
+
+          // without this timeout the the complete tx does not trigger, dont know why
+          setTimeout(() => {
+            createEthWithdrawalApproval(
+              withdrawal,
+              approval?.erc20WithdrawalApproval
+            );
+          }, 500);
```
edd, line 303: **nitpick (non-blocking)**: This is a pretty big, unwieldy `onClick` that would be better outside the JSX. But it's not worth blocking.

libs/withdraws/src/lib/withdraw-form.tsx
```diff
@@ -197,10 +196,10 @@ export const WithdrawForm = ({
         data-testid="withdraw-form"
       >
         <TradingFormGroup
-          label={t('To ({{chainName}} address)', {
+          label={t('To {{chainName}} address', {
             chainName: selectedAsset?.chainId
               ? getChainName(selectedAsset.chainId)
-              : ETHEREUM_CHAIN_ID,
+              : '',
```
edd, line 202: **nitpick (non-blocking)**: Again, better then blank would be showing the chainId. Again again, not a blocker.

edd (approved):
Approved with some non-blocking nitpicks. tl;dr: showing chainID is better than the fallback blank string (but unlikely to occur), big onClick event handlers are a bit ugly.

libs/withdraws/src/lib/withdraw-form.tsx
mattrussell36: In this case there `selectedAsset.chainId` is falsey so we don't have a chainId on line 202. I would rather it fall back to just 'To address' or 'From address', rather than defaulting to '1'

libs/withdraws/src/lib/use-ready-to-complete-withdrawals-toast.tsx
mattrussell36: Agree, but we'll have to come back to this.

libs/environment/src/external-chain.ts
mattrussell36: done

libs/assets/src/lib/utils.ts
mattrussell36: Agree, and done

MadalinaRaicu (approved):

bwallacee (approved):

PR #6560, opened 2024-06-14
fix(trading): add arbiscan links

### LINKED ISSUE #6553: Approve asset TX toast fires off to ethereum wrongly
## The Chore

The toast for an approve TX on Arbitrum gives you a link to Etherscan

## Tasks

- [ ]
- [ ]

MadalinaRaicu:
# Related issues 🔗

Closes #6553 

# Description ℹ️

- [x] add Arbiscan links
- [x] fix Arbiscan links on approval 

# Demo 📺

image

# Technical 👨‍🔧

### DIFF
--- a/apps/governance/.env.mainnet
+++ b/apps/governance/.env.mainnet
@@ -5,6 +5,7 @@ NX_VEGA_URL=https://api.vega.community/graphql
 NX_VEGA_NETWORKS='{"DEVNET":"https://dev.governance.vega.xyz","TESTNET":"https://governance.fairground.wtf","MAINNET":"https://governance.vega.xyz"}'
 NX_ETHEREUM_PROVIDER_URL=https://mainnet.infura.io/v3/3ba145ccc2884bcd91213d8dc989ca76
 NX_ETHERSCAN_URL=https://etherscan.io
+NX_ARBISCAN_URL=https://arbiscan.io
 NX_GITHUB_FEEDBACK_URL=https://github.com/vegaprotocol/feedback/discussions
 NX_SENTRY_DSN=https://4b8c8a8ba07742648aa4dfe1b8d17e40@o286262.ingest.sentry.io/5882996
 NX_VEGA_EXPLORER_URL=https://explorer.vega.xyz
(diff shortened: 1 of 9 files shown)


libs/deposits/src/lib/approve-notification.tsx
```diff
@@ -166,11 +171,19 @@ const ApprovalTxFeedback = ({
...
-  const txLink = tx.txHash && (
-    <EtherscanLink sourceChainId={chainId} tx={tx.txHash}>
-      {t('View on Etherscan')}
-    </EtherscanLink>
-  );
+  const chain = getChainName(chainId);
+
+  const txLink =
+    tx.txHash &&
+    (chain.includes('Ethereum') ? (
```
edd, line 178: **nitpick (non-blocking)**:  Use a chain ID lookup would probably be more robust than relying on the string. Not worth changing as bridges are not being added regularly.

edd (approved):
Approved, with a comment about unifying this with the explorer implementation in future

apps/governance/.env.mainnet
```diff
@@ -5,6 +5,7 @@ NX_VEGA_URL=https://api.vega.community/graphql
 NX_VEGA_NETWORKS='{"DEVNET":"https://dev.governance.vega.xyz","TESTNET":"https://governance.fairground.wtf","MAINNET":"https://governance.vega.xyz"}'
 NX_ETHEREUM_PROVIDER_URL=https://mainnet.infura.io/v3/3ba145ccc2884bcd91213d8dc989ca76
 NX_ETHERSCAN_URL=https://etherscan.io
+NX_ARBISCAN_URL=https://arbiscan.io
```
mattrussell36, line 8: We already have two places defining an arbiscan root url. Now this adds a third. Probably not something to fix right now but I'd like to unify this in a .ts file which can be indexed into with a chain id

libs/environment/src/components/arbitrum-link.tsx
```diff
@@ -0,0 +1,48 @@
+import { ExternalLink } from '@vegaprotocol/ui-toolkit';
```
mattrussell36, line 1: This component is not required. EtherscanLink has already been modified to handle other chains. (We should rename it later though)

mattrussell36 (changes requested):

apps/governance/.env.mainnet
mattrussell36: Based on the fact that the EtherscanLink component already handles other chains we don't need this new ENV var at all

daro-maj:
I can not start the app
image

PR #6565, opened 2024-06-18
feat(trading): wagmi

### LINKED ISSUE #6467: Migrate Console to use WAGMI
## The Chore

## Tasks

- [ ]
- [ ]

mattrussell36:
# Related issues 🔗

Closes #6467

# Description ℹ️

TBD

# Demo 📺

TBD


apps/explorer/src/app/components/txs/details/tx-proposal.tsx
```diff
@@ -57,6 +57,7 @@ export function proposalTypeLabel(terms?: ProposalTerms): string {
   } else if (has(terms, 'updateAsset')) {
     return t('Update asset proposal');
   } else if (has(terms, 'newMarket')) {
+    // @ts-ignore TODO: figure out this type error, changes will be present
```
dexturr, line 60: chore (non-blocking): this is optionally chained so based on this message the ts-ignore can be removed

```suggestion
    // TODO: figure out this type error, changes will be present
```

apps/trading/lib/utils/get-erc20-abi.ts
```diff
@@ -0,0 +1,21 @@
+import { type Address, erc20Abi } from 'viem';
+import { ERC20_TEHTER_ABI } from '@vegaprotocol/smart-contracts';
+
+const TETHER_ADDRESS = '0xdAC17F958D2ee523a2206206994597C13D831ec7';
```
dexturr, line 4: chore (non-blocking): think this would make more sense as an array present in an env var rather than hard coded into the app.

apps/trading/components/deposit-container/deposit-container.tsx
```diff
@@ -0,0 +1,333 @@
...
+      initialAssetId={asset?.id || ''}
+      configs={allConfigs}
+    />
+  );
+};
+
+const depositSchema = z.object({
+  fromAddress: z.string().min(1, 'Connect wallet'),
+  assetId: z.string().min(1, 'Required'),
+  toPubKey: z.string().regex(/^[A-Fa-f0-9]{64}$/i, 'Invalid key'),
```
dexturr, line 76: minor (non-blocking): should we use VEGA_ID_REGEX here rather than duplicating

apps/trading/components/withdraw-container/withdraw-container.tsx
```diff
@@ -1,27 +1,426 @@
...
+      accounts={accounts || []}
+      initialAssetId={asset?.id || ''}
+      configs={allConfigs}
     />
   );
 };
+
+const withdrawSchema = z.object({
+  assetId: z.string().min(1, 'Required'),
+  fromPubKey: z.string().regex(/^[A-Fa-f0-9]{64}$/i, 'Connect Vega wallet'),
```
dexturr, line 93: minor (non-blocking): should we use VEGA_ID_REGEX here rather than duplicating

apps/trading/components/deposit-container/deposit-container.tsx
```diff
@@ -0,0 +1,333 @@
...
+      assets={assets}
+      initialAssetId={asset?.id || ''}
+      configs={allConfigs}
+    />
+  );
+};
+
+const depositSchema = z.object({
+  fromAddress: z.string().min(1, 'Connect wallet'),
+  assetId: z.string().min(1, 'Required'),
```
dexturr, line 75: fix (blocking): validation for valid eth/vega address here

apps/trading/components/withdraw-container/withdraw-container.tsx
```diff
@@ -1,27 +1,426 @@
...
+  toAddress: z.string().min(1, 'Enter address or connect wallet'),
```
dexturr, line 94: fix (blocking): we should validate the to address is valid eth address if this can be user input

apps/trading/components/withdraw-container/withdraw-container.tsx
```diff
@@ -1,27 +1,426 @@
...
+    <WithdrawForm
+      accounts={accounts || []}
+      initialAssetId={asset?.id || ''}
+      configs={allConfigs}
     />
   );
 };
+
+const withdrawSchema = z.object({
+  assetId: z.string().min(1, 'Required'),
```
dexturr, line 92: question (non-blocking): should we validate this? Or not required because selects?

apps/trading/lib/utils/get-erc20-abi.ts
mattrussell36: Prefer keeping things out of the env file and in code where possible. Ideally each env file could have a single vega chain id and ethereum chain id and everything can be keyed from that

apps/explorer/src/app/components/txs/details/tx-proposal.tsx
mattrussell36: Yes the optional chain should protect when the code runs but running build-spec will fail

dexturr (approved):

PR #6571, opened 2024-06-19
chore(trading): update vega

bwallacee:


apps/trading/e2e/.env.develop
```diff
@@ -1,4 +1,4 @@
 CONSOLE_IMAGE_NAME=vegaprotocol/trading:develop
-VEGA_VERSION=v0.76.8
+VEGA_VERSION=v0.77.0-preview.3
```
mattrussell36, line 2: develop should stay on whatever is on mainnet. If/when we use a `next` branch we will upgrade with an .env.next

mattrussell36 (changes requested):

PR #6575, opened 2024-06-19
feat(trading): price tolerance info

### LINKED ISSUE #6548: Simple swap UI price tolerance issue
## The Chore

Because we can only submit orders with a `size` and that size is always denoted in `base asset` terms ... it means that the behaviour of the swap is slightly different depending on whether you are paying the `base` or the `quote`.

The _ideal_ behaviour would be:

- I want to pay X amount of asset A for asset B and I have a price tolerance of Y
- That means I **always** pay X of asset A and if the price has moved against me I might end up with _less_ of asset B up to a min that reflects my price tolerance

This works fine if you are paying the base (ie. selling), because we set the size and if the book has some slippage you'll get somewhere between the amount purchased at your best bid price and the amount you can can purchase at your price tolerance.

But if you're paying the quote (ie. buying) it doesnt work, because we have to derive the size of the order in base terms before submitting it ... so you're a bit stuck, either you:

- Derive that size based on the best ask and if the price moves against you, you will end up overpaying up to your price tolerance ... not ideal
- Derive that size based on the price tolerance, then you will never overpay but you will ALWAYS end up with less size than you wanted

Neither option is really idea ... so I think until such time as Core gives us the ability to submit an order with the notional in quote asset terms the best we can do is some messaging.

### If paying the quote and receiving the base asset

You are BUYING ... and your SIZE is DERIVED

"You will pay between `quote asset amount based on best ask` and `quote asset amount at price tolerance` to receive `size in base asset` based on your price tolerance of `price tolerance`"

eg.  
"You will pay between 695.51 USDT and 702.4 USDT to receive 0.01 BTC based on your price tolerance of 1%"

### If paying the base and receiving the quote asset

You are SELLING ... and your SIZE is SET in the form

"You will pay `size in base asset` to receive between `quote asset amount based on best bid` and `quote asset amount at price tolerance` based on your price tolerance of `price tolerance"

eg. 
"You will pay 0.01 BTC to receive between 695.51 USDT and 688.55 USDT based on your price tolerance of 1%" 

NB. Both examples above for simplicity assume that best bid = best ask = 695.51 USDT, with price tolernance applied appropriately based on whether it is a buy or a sell.

## Acceptance criteria

- [x] Add text under the swap button providing all this information 
- [x] As a result it might make sense to remove some of the information that is already there to avoid duplication

It is probably worth considering as part of this ticket whether traversing the book properly instead of just using best bid / ask and adding price impact tolerance would help with this.  My assumption is it will not ... but if we wanted to move to using slippage properly we should at least do it AND this at the same time to avoid making life difficult and reworking things later.

MadalinaRaicu:
# Related issues 🔗

Closes #6548 

# Description ℹ️

Add price tolerance message. 

# Demo 📺
1. No price tolerance
Screenshot 2024-06-19 at 18 33 30

2. 1.0% price tolerance - buy 
Screenshot 2024-06-19 at 18 33 35

4. 1.0% price tolerance - sell
Screenshot 2024-06-19 at 18 33 47
 

# Technical 👨‍🔧


apps/trading/components/swap/spot-data.tsx
```diff
@@ -1,83 +1,108 @@
 import type { AssetFieldsFragment } from '@vegaprotocol/assets';
 import { Side } from '@vegaprotocol/types';
-import { addDecimalsFormatNumber } from '@vegaprotocol/utils';
 import { useT } from '../../lib/use-t';
+import { Intent, Notification } from '@vegaprotocol/ui-toolkit'; // Import the Notification component
```
MadalinaRaicu, line 4:
```suggestion
import { Intent, Notification } from '@vegaprotocol/ui-toolkit';
```

bwallacee (approved):

mattrussell36 (approved):

PR #6597, opened 2024-06-25
feat(trading): sidebar tweaks and actions

### LINKED ISSUE #6459: Support deposit / transfer / swap etc from within the trading sidebar
## The Chore

As an intermediate step we have directed users to Portfolio, but ideally we would embed the functions as a second layer of the sidebar.

## Acceptance Criteria

  - [ ] If you click on any of the quick launch buttons in the asset breakdown view for deposit / transfer / withdrawal / swap it should launch those flows inside the sidebar
  - [ ] Will need a breadcrumb / back button to get back to the asset breakdown from this view
  - [ ] If you use the kebab on the mini asset view to do this the asset pan should expand the same thing should happen
  - [ ] If you hit deposit give a little dropdown allowing you to choose between same chain / cross chain

### Example of the back button from sketches

image

### Icons

We only have these icons ... do we need something to distinguish between same / cross chain.

image

### LINKED ISSUE #6572: Trading sidebar feedback
## The Chore

A few things to add

## Tasks

- [ ] When you have clicked on say "Market info" and it expands that view, you should be able to click on "Market info" again to minimise it.  Doing this should just expand the "Trade" view again (which effectivly should be the default).
- [ ] Doesnt look like we've reformatted the market info to match the new styling, including removing the captilisation on the headers  .... see pictures in additional info
- [ ] We should be sorting asset balances so ones you have are at the top ... with no USD conversion we might have to do this with a conversion to quantum and a sort for now 
- [ ] The mini asset view is showing me the asset CODE but when i expand it I am shown the asset NAME ... as youc an see in pictures this made it pretty confusing whether that was actually the same asset.  Mini asset view should be showing just the code and the chain (which it is) ... expanded view probably needs to go across two lines so it can show me code, name, and chain in some combination 
- [ ] The numbers feel a bit big in the asset dropdown.  They're probably fine for the mini asset view ... but I think in the expanded one only the TOTAL should be that large ... the breakdown should be a size smaller. 
- [ ] Rename "Allocation" to "Deployed" and add a % in brackets after the number, ie.. 150 (24%)
- [ ] We're still showing assets with zero balance in the expanded asset view.  We should not do this.  
- [ ] Obviously if you're ON a market which settles in an asset you dont have the mini-asset view should still show this, and if you expand it to show the full breakdown it should still show it and in highlighted form so that users can deposit.

### Market info in sketch

image

### Market info on app

image

### Mini asset view 

image

### Expanded asset view 

image

### Number size

image

asiaznik:
# Related issues 🔗

Closes #6572 
Closes #6459

# Description ℹ️

* adds support to asset actions inside the new sidebar
* UX improvements

# Demo 📺

image

# Technical 👨‍🔧

N/A


apps/trading/components/accounts-container/sidebar-accounts-container.tsx
```diff
@@ -0,0 +1,103 @@
...
+  setInnerView?: (innerView: InnerView | undefined) => void;
+}) => {
+  const [view, assetId] = innerView;
+  switch (view) {
+    case View.Deposit:
+      return <DepositContainer assetId={assetId} />;
+
+    case View.Swap:
+      return (
+        <SwapContainer
```
mattrussell36, line 88: Swap container renders a title, so we end up with an awkward double title. Can you remove the title from the container, and then add the title on the standalone swap page.

It would be good to check all the titles rendered within these components and make sure everything looks consistent

apps/trading/components/accounts-container/sidebar-accounts-container.tsx
```diff
@@ -0,0 +1,103 @@
+import { useState } from 'react';
+import { useT } from '../../lib/use-t';
+import { SwapContainer } from '../swap';
+import { WithdrawContainer } from '../withdraw-container';
+import {
+  AccountsContainer,
+  type AccountsContainerProps,
+} from '../accounts-container/accounts-container';
+import React from 'react';
+import { DepositContainer } from '@vegaprotocol/deposits';
```
mattrussell36, line 10: This should be the deposit-container from apps/trading/components. (you may need to rebase)

apps/trading/components/sidebar/sidebar.spec.tsx
```diff
@@ -35,11 +37,15 @@ describe('Sidebar', () => {
...
     const user = userEvent.setup();
     const result = render(
-      <MemoryRouter initialEntries={[`/markets/${marketId}`]}>
-        <Routes>
-          <Route path="/markets/:marketId" element={<Sidebar />} />
-        </Routes>
-      </MemoryRouter>
+      <MockedProvider>
+        <MockedWalletProvider>
+          <MemoryRouter initialEntries={[`/markets/${marketId}`]}>
```
mattrussell36, line 42: There are no longer any Route elements within the sidebar so initialEntries can be removed

apps/trading/components/sidebar/sidebar.tsx
```diff
@@ -34,7 +34,12 @@ export const Sidebar = ({ pinnedAssets }: { pinnedAssets?: string[] }) => {
       <SidebarAccordion
         type="single"
         value={view}
-        onValueChange={(x: ViewType) => setView(x)}
+        onValueChange={(view: ViewType) => {
```
mattrussell36, line 37: This doesn't look correct based on the types. Should `view` be optional?

mattrussell36 (changes requested):

apps/trading/components/sidebar/sidebar.tsx
asiaznik: Nope, the argument of `onValueChange` is not optional.

Type signature: `onValueChange?(value: string): void`

apps/trading/components/accounts-container/sidebar-accounts-container.tsx
asiaznik: Removed the title completely, we don't show the additional title for any of the containers so it doesn't make sense IMO to have it only for swap. Also moved the "Go to market" link to the bottom of the form.

mattrussell36 (approved):

PR #6598, opened 2024-06-26
chore(environment): speed up start time of app

mattrussell36:
# Related issues 🔗

N/A

# Description ℹ️

- Sets default VEGA_URL for testnet and mainnet in env files
- Grabs the url from localStorage if set, otherwise falls back to the default VEGA_URL
- Does not await fetching of node config or to find the healthiest node before rendering. Instead this is now done in the background, unless there is no default node set

### DIFF
--- a/libs/environment/src/hooks/use-environment.ts
+++ b/libs/environment/src/hooks/use-environment.ts
@@ -224,8 +224,15 @@ const compileEnvVars = () => {
     process.env['NX_VEGA_ENV']
   ) as Networks;
 
+  let vegaUrl = LocalStorage.getItem(STORAGE_KEY);
+
+  if (!isValidUrl(vegaUrl)) {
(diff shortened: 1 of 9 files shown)


libs/environment/src/hooks/use-environment.ts
```diff
@@ -224,8 +224,15 @@ const compileEnvVars = () => {
     process.env['NX_VEGA_ENV']
   ) as Networks;
 
+  let vegaUrl = LocalStorage.getItem(STORAGE_KEY);
+
+  if (!isValidUrl(vegaUrl)) {
+    vegaUrl = windowOrDefault('VEGA_URL', process.env['NX_VEGA_URL']) as string;
+    LocalStorage.removeItem(STORAGE_KEY);
+  }
```
dexturr, line 232: question (blocking): should local storage not take precedence over the

asiaznik (approved):

PR #6608, opened 2024-06-27
fix(trading,explorer): show unlimited in asset details for large numbers

### LINKED ISSUE #6508: Text overflow on asset lifetime limit
## The Chore

The limit for the new Arbitrum Tether is set to be the max integer thing.

It is overflowing

image

## Acceptance Criteria

- [ ] It should not overflow .... I guess a dot dot dot and a hover state?  I dont think we want to wrap

edd:
# Related issues 🔗

Closes #6508

# Description ℹ️

USDTT has a lifetime deposit limit of 115792089237316195423570985008687907853269984665640564039457584007913129639935, which when formatted is a very very long number. This broke the table. This very large number is *effectively* unlimited, so this fix changes it to show that for very very large numbers. It is quite a dumb fix that could break, say if the asset was precise to a very large number of decimal places, but this is not true for any current asset.

# Demo 📺
Screenshot 2024-06-27 at 12 41 33

### DIFF
--- a/libs/assets/src/lib/asset-details-table.spec.tsx
+++ b/libs/assets/src/lib/asset-details-table.spec.tsx
@@ -6,8 +6,18 @@ import {
   AssetDetailsTable,
   useRows,
   testId,
+  num,
 } from './asset-details-table';
 import { generateBuiltinAsset, generateERC20Asset } from './test-helpers';
+
+describe('num formatting', () => {
+  it('show unlimited for large values', () => {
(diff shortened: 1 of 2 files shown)


mattrussell36 (approved):
There is a formatNumberRounded function that you could use to render `>100t` which I guess is a little more truthful and accurate, but I'm not sure I prefer that to saying 'Unlimited' which is clearer

PR #6614, opened 2024-06-28
fix(trading,governance,explorer): graphql version resolution

mattrussell36:
# Related issues 🔗

N/A

# Description ℹ️

Fixes type generation


edd (approved):
This works for me, thanks. It might be that yarn `1.22.22` is acting differently to `1.22.19`. I was getting different conflicting versions of `graphql`.

Issue #6574, opened 2024-06-19
Save some space on the deal ticket


JonRay15:
## The Chore

We need to find some ways to do this with Joelle.

## Acceptance criteria

- [x] Implement designs from Joelle
- [x] Once done we need to be able to see the submit button easily on a 13in monitor on 100% zoom

### Figma

https://www.figma.com/design/ywcCZvXdvG28bg0AmCY9mF/Console-v3?node-id=6902-8315&t=VsGQ7tcUW64dwfNq-0

### Sketch

![image](https://github.com/vegaprotocol/frontend-monorepo/assets/89918817/d70f1847-8378-4c49-8aa1-330a1241df98)

### Original brainstorm

https://miro.com/app/board/uXjVOs_PBKk=/

JonRay15:
Raised a few tickets for some minor things I found, but by and large this looks great.  Closing.

PR #6620, opened 2024-06-28
feat(trading): deal ticket type selector

mattrussell36:
# Related issues 🔗

#6574 

# Description ℹ️

- Changes type selector ui
- Makes a specific deal ticket submit button so it can be shared between stop and normal orders
- Adjusts components to not create spacing with margin (that should be the responsibility of the parent)

# Demo 📺

Screenshot 2024-06-28 at 16 48 38

### DIFF
--- a/libs/deal-ticket/src/components/deal-ticket/submit-button.tsx
+++ b/libs/deal-ticket/src/components/deal-ticket/submit-button.tsx
@@ -1,0 +1,38 @@
+import { type ReactNode } from 'react';
+import classNames from 'classnames';
+
+import { Side } from '@vegaprotocol/types';
+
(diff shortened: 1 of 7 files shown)


libs/deal-ticket/src/components/deal-ticket/submit-button.tsx
```diff
@@ -0,0 +1,38 @@
+import { type ReactNode } from 'react';
+import classNames from 'classnames';
+
+import { Side } from '@vegaprotocol/types';
+
+export const SubmitButton = ({
```
dexturr, line 6: Yay deal tickets being more componentised whoop whoop

dexturr (approved):

PR #6666, opened 2024-07-09
fix(trading): display of node health on mobile

### LINKED ISSUE #6465: Show the node health indicator on other non-trading pages
## The Chore

Dependent on trading siderbar work .... so if we do this soon it would need to be targeted to Matt's branch.

## Acceptace Criteria

- [x] In the same location as the indicator on the sidebar we should show something on ALL other pages
- [ ] It should show the same node inicator
  - [ ] Maybe the indicator and the text hovering is fine 
  - [x] If that doesnt look good then we could show ONLY the indicator and on hover show the name of the node
- [x] The page should scroll under it (its locked in place)
- [x] You should be able to launch the node switcher as always 
- [ ] The solution needs to work on mobile too (currently it is impossible to view or change your node on mobile, which is a signficant bug)

mattrussell36:
# Related issues 🔗

Closes #6465 

# Description ℹ️

Fixes node health indicator not showing on Safari on ios

### DIFF
--- a/apps/trading/components/navbar/navbar.tsx
+++ b/apps/trading/components/navbar/navbar.tsx
@@ -123,7 +123,7 @@ export const Navbar = ({ theme = 'system' }: { theme?: Theme }) => {
           <D.Content
             className={classNames(
               'lg:hidden',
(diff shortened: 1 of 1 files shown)


JonRay15:
This works now 

image

JonRay15 (approved):

PR #6671, opened 2024-07-09
feat(trading): single market list

### LINKED ISSUE #6086: Move to single master market list w/filters
## User Story 

As a user 
I want to have a single master view on markets with filters and tabs on my market page
So that I can switch easily to what I need to see

Ideally as part of this also deliver:

- #6588 6588

## Acceptance Criteria

- [x] Get to single view containing all of Open / Proposed / Closed which is basically the OPEN view now ... we are optimising for the OPEN markets first
- [x] Anywhere we dont have appropriate info for proposed / closed just show a dash instead
- [x] Have a dropdown to allow filtering for status (Open / Proposed / Closed) which is **always defaulted to Open**
- [x] Have a dropdown to allow filtering for Asset to allow filtering for settlement asset 
  - [x] Will need to include the chain name in brackets
  - [x] For Spot markets will need to include any market where the asset is either the base or the quote
- [x] Have a search box that supports standard text search
- [x] If any results are being excluded from the search result by filters display some text beneath the results saying "X results excluded due to applied filters, click to remove filters"
- [x] Have a series of quick filter buttons to support
  - [x] All
  - [x] Perp
  - [x] Future
  - [x] Spot
- [x] Re-order the market selector dropdown product filter so that it follows the same order
 

## Sketch

[Figma Link ↳ ](https://www.figma.com/design/ywcCZvXdvG28bg0AmCY9mF/Console-v3?node-id=6813-1688&t=BnlzWamJO7nBeWPD-1)

**Default State**
01 Markets@2x

---

**Excluded Results**
02 Excluded Results@2x

---

**Dropdown Filters & Search**

Showing:
- Default, Hover and Active states
- Single and Multi Select 

Market Filter States@2x

---

[Market Switcher with the updated tab order ↳](https://www.figma.com/design/ywcCZvXdvG28bg0AmCY9mF/Console-v3?node-id=6818-29185&t=BnlzWamJO7nBeWPD-1)

### LINKED ISSUE #6588: Minor tweaks to market switcher
## The Chore

A handful of changes to the market switcher dropdown component to align to the new markets page.

Must be done after 

- #6086 

## Current 

image

## Acceptance criteria 

- [x] Asset list in the dropdown needs to include the chain ID in brackets
- [x] Add a checkbox to the asset dropdown so users can select many assets if desired
- [x] Re-order the tabs to mirror market list, ie. All, Perps, Futures, Spot

https://www.figma.com/design/ywcCZvXdvG28bg0AmCY9mF/Console-v3?node-id=6813-1688&t=ZUanFLTsY8vBgpgo-0

### Asset multiselect

image

### New order

image

asiaznik:
# Related issues 🔗

Closes #6086 
Closes #6588 

# Description ℹ️

* adds filters to markets

# Demo 📺

https://github.com/vegaprotocol/frontend-monorepo/assets/1980305/c4f99ae9-2ca5-4815-9ad4-e0fded7d7d9d

# Technical 👨‍🔧

N/A


JonRay15:
**Feedback** 

1. We seem to have lost the page margins.  The mainnet version is intentionally quite narrow to look more "websitey" but the preview I'm on is now going all the way to edge of screen.

2. Product filter follows multiselect style behaviour .... which I'm not sure I anticipated, I just thought it would show you that product .... I can see arguments for both approaches ... would be good hear other views on this

3. I dont love the behaviour on the Status and Asset dropdown because we elected for checkbox style multi-select here, which I think is right, but I think if we do that we shouldnt close the dropdown as soon as you select one.  We should either apply live and let you close the dropdown by clicking outside of it, or we should have an apply button in the dropdown.  At the moment if I'm on "Open" and I want to view "Closed" I have to click once on Closed to end up with both selected, then open it again and deselect the open one.

apps/trading/client-pages/markets/markets-page.tsx
```diff
@@ -125,78 +166,355 @@ export const MarketsPage = () => {
...
-            <Proposed />
-          </ErrorBoundary>
-        )}
-        {activeTab === 'closed-markets' && (
-          <ErrorBoundary feature="markets-closed">
-            <Closed />
-          </ErrorBoundary>
-        )}
+        <ErrorBoundary feature="all-markets">
+          <OpenMarkets
```
mattrussell36, line 512: We should rename this to or remove one level of components here as this is no longer a component for just open markets

apps/trading/client-pages/markets/markets-page.tsx
```diff
@@ -125,78 +166,355 @@ export const MarketsPage = () => {
...
+          values={{ count: Math.abs(diff) }}
+          components={[
+            <button
+              key="reset-filters"
+              className="underline"
+              onClick={() => {
+                reset();
+              }}
+            >
+              Remove filters
```
mattrussell36, line 373: Translation

apps/trading/client-pages/markets/markets-page.tsx
```diff
@@ -125,78 +166,355 @@ export const MarketsPage = () => {
...
+                    <span className="text-xs">({chainName})</span>
+                  </MultiSelectOption>
+                );
+              })}
+            </MultiSelect>
+          </div>
+          {/** MARKET NAME FILTER */}
+          <div className="w-2/3">
+            <TradingInput
+              prependIconName={VegaIconNames.SEARCH}
```
mattrussell36, line 495: This icon looks different (chunkier) to the one in the market selector dropdown, which is the correct one

apps/trading/client-pages/markets/use-column-defs.tsx
```diff
@@ -126,7 +126,14 @@ export const priceValueFormatter = (
           data.decimalPlaces
         )} ${quoteAsset?.symbol}`;
   }
-  const quoteName = data && getQuoteName(data);
+  let quoteName: string | undefined = undefined;
+  if (data) {
+    try {
```
mattrussell36, line 131: Requiring a try catch here to me indicates something isn't quite right with the data object or if `data` can be undefined it would be better to just return early with an empty string

libs/ui-toolkit/src/components/trading-select/select.tsx
```diff
@@ -175,3 +177,97 @@ export const MiniSelectOption = forwardRef<
     <SelectPrimitive.ItemText>{children}</SelectPrimitive.ItemText>
   </SelectPrimitive.Item>
 ));
+
+type MultiSelectProps = React.ComponentProps<typeof DropdownPrimitive.Root> & {
```
mattrussell36, line 181: This new component uses a different radix primitive so I think it would be better in its own directory or with the other TradingDropdowns

apps/trading/client-pages/markets/markets-page.tsx
```diff
@@ -125,78 +166,355 @@ export const MarketsPage = () => {
...
+const useMarketPageFilters = create<MarketPageFiltersStore>()((set) => ({
+  ...DEFAULT_FILTERS,
+  setMarketTypes: (marketTypes) => set({ marketTypes }),
+  setMarketStates: (marketStates) => set({ marketStates }),
+  setAssets: (assets) => set({ assets }),
+  setSearchTerm: (searchTerm) => set({ searchTerm }),
+  reset: () => set(DEFAULT_FILTERS),
+}));
+
+const filter = (
```
mattrussell36, line 215: non blocking: I was expecting to share the filter logic between market selector dropdown and the market list. Can we not have a single hook that manages this

mattrussell36 (changes requested):

mattrussell36:
AgGrid does provide an api for external filters, it may be worth passing the filter function in as an option. I think the grid will be smoother as a result as you aren't having to rebuild the grid entirely every time a filter is applied 

https://www.ag-grid.com/javascript-data-grid/filter-external/

apps/trading/client-pages/markets/markets-page.tsx
asiaznik: Not needed here, this text will be replaces by `<0>Remove filters</0>` from translations.

mattrussell36 (approved):

PR #6696, opened 2024-07-15
fix(trading): total locked volume calc

mattrussell36:
# Related issues 🔗

N/A

# Description ℹ️

Queries asset pool to get total locked volume rather than using the workaround parties/assets queries


libs/environment/src/chains.ts
```diff
@@ -11,3 +11,33 @@ export const CHAIN_IDS: {
   [Networks.CUSTOM]: 'vega-stagnet1-202307191148',
   [Networks.MAINNET]: 'vega-mainnet-0011',
 };
+
+export const ASSET_POOL_ADDRESSES: {
+  [N in Networks]: { [chainId: number]: string };
+} = {
+  [Networks.DEVNET]: {
+    11155111: '0x6EA5BdbF96DcCbADb094A724a5f971C56B328c85',
```
dexturr, line 19: might be worth adding these as a separate enum and using the values for these keys rather than magic numbers

dexturr (approved):

PR #6706, opened 2024-07-17
feat(trading): portfolio assets tab

### LINKED ISSUE #6683: Focus portfolio page sidebar on asset breakdown
## The Chore

Currently the portoflio page sidebar has become a dumping ground for all asset actions and as a result there are now a lot of tabs.

image

This is no longer necessary because the asset breakdown offers contextual ways to trigger all these flows. 

## Acceptance criteria

- [x] Focus the sidebar on the asset breakdown view
- [x] Sort the asset list by balance in descencing order
- [x] Add a search bar which filters live
- [x] Consider if the top asset should automatically as below otherwise it might not be obvious the user can click it
- [x] The action buttons should launch those actions IN the sidebar not send you to another page
- [x] For the deposit flow an additional interim screen will be needed that gives the user the ability to select between same and chain cross chain deposit options as to big buttons.  The text on the same / cross chain deposit screens could be used as explanation here.
- [x] Remove all the other tabs besides Assets

### Sketch 

image

asiaznik:
# Related issues 🔗

Closes #6683

# Description ℹ️

* removes other tabs on Portfolio sidebar
* adds Search field to the assets breakdown
* adds combined deposit button

### DIFF
--- a/apps/trading/e2e/.env
+++ b/apps/trading/e2e/.env
@@ -1,4 +1,4 @@
 CONSOLE_IMAGE_NAME=vegaprotocol/trading:develop
 VEGA_VERSION=v0.76.8
-LOCAL_SERVER=true
+LOCAL_SERVER=false
 VEGA_ENV=STAGNET1
--- a/apps/trading/e2e/.env
+++ b/apps/trading/e2e/.env
@@ -1,4 +1,4 @@
 CONSOLE_IMAGE_NAME=vegaprotocol/trading:develop
 VEGA_VERSION=v0.76.8
-LOCAL_SERVER=false
+LOCAL_SERVER=true
 VEGA_ENV=STAGNET1
--- a/libs/accounts/src/lib/account-card.tsx
+++ b/libs/accounts/src/lib/account-card.tsx
@@ -26,7 +26,7 @@ import {
   useState,
   type ComponentProps,
   forwardRef,
-  ButtonHTMLAttributes,
+  type ButtonHTMLAttributes,
 } from 'react';
 import classNames from 'classnames';
 import { useChainId } from '@vegaprotocol/wallet-react';
--- a/libs/accounts/src/lib/account-card.tsx
+++ b/libs/accounts/src/lib/account-card.tsx
@@ -6,6 +6,10 @@ import {
   VegaIcon,
   VegaIconNames,
   ProgressBar,
+  TradingDropdown,
+  TradingDropdownItem,
+  TradingDropdownContent,
+  TradingDropdownTrigger,
 } from '@vegaprotocol/ui-toolkit';
 
 import {
@@ -18,7 +22,7 @@ import { Emblem } from '@vegaprotocol/emblem';
 import { AccountsActionsDropdown } from './accounts-actions-dropdown';
 import { type AssetFieldsFragment } from '@vegaprotocol/assets';
 import { useDataProvider } from '@vegaprotocol/data-provider';
-import { useState } from 'react';
+import { type ReactNode, useState } from 'react';
 import classNames from 'classnames';
 import { useChainId } from '@vegaprotocol/wallet-react';
 import { getExternalChainShortLabel } from '@vegaprotocol/environment';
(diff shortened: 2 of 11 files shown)


JonRay15:
Reviewed connected to mainner to excuse the dead icons. 

Tabs are gone ... Sort looks good

image

Search works

image

I do wonder if it is not obvious enough that you need to click to expand it, and this is compounded by that little menu in the top right which a user might think is where they should click rather than _anywhere else_ and when they do ultimately they get a less nice experience. 

image

On the deposit, I think this little dropdown works reasonably well. Probably better than a whole dedicated screen just to make the choice 

image

BUT ... I do think there are two issues:

1. If you just click and dont notice the dropdown you end up on same chain deposit and might not realise the cross chain is there
2. If you select cross-chain it doesnt remember that for next time and the button still will take you to same chain 

I think either:

1. A simple click should just open that little dropdown menu so you _always_ have to make the selection, then you dont need seperate button for the dropdown
2. The button should remember your last selection and take you there again next time.  If we went with this we should consider adding a button maybe in the top right that says "Switch to same-chain deposit" / "Switch to cross-chain deposit" so its easy to swap if you'ce been stickily led to somethig you didnt want to be

JonRay15:
This all looks good to me now

Lets merge it and if we decide we need to rename the deposit types we can easily do that later

JonRay15 (approved):

libs/accounts/src/lib/account-card.tsx
```diff
@@ -35,19 +40,79 @@ const Button = ({
...
+} & ComponentProps<typeof TradingButton>) => (
+  <TradingButton size="custom" className="p-[6px]" onClick={onClick} {...props}>
     <div className="flex flex-col items-center p">
       <VegaIcon name={icon} size={16} className="mb-1" />
       <span className="text-xs">{label}</span>
     </div>
   </TradingButton>
 );
 
+const CombinedButton = ({
```
mattrussell36, line 57: Not sure why we need all this logic. Why not just render the dropdown and the button where needed? You may need to ensure the ref is passed down to <Button>

```
      {expandable && expanded ? (
        <div className="grid gap-1 grid-cols-4 p-3 pt-0">
          <TradingDropdown
            trigger={
              <TradingDropdownTrigger asChild>
                <Button label={t('Deposit')} icon={VegaIconNames.DEPOSIT} />
              </TradingDropdownTrigger>
            }
          >
            <TradingDropdownContent side="bottom" align="start">
              <TradingDropdownItem
                onClick={() => {
                  actions.onClickDeposit?.(asset.id);
                }}
              >
                <VegaIcon name={VegaIconNames.DEPOSIT} /> {t('Deposit')}
              </TradingDropdownItem>
              <TradingDropdownItem
                onClick={() => {
                  actions.onClickCrossChainDeposit?.(asset.id);
                }}
              >
                <VegaIcon name={VegaIconNames.DEPOSIT} />{' '}
                {t('Cross-chain deposit')}
              </TradingDropdownItem>
            </TradingDropdownContent>
          </TradingDropdown>
          <Button
            data-testid="account-action-swap"
            onClick={() => actions.onClickSwap?.(asset.id)}
            label={t('Swap')}
            icon={VegaIconNames.SWAP}
          />
          <Button
            data-testid="account-action-transfer"
            onClick={() => actions.onClickTransfer?.(asset.id)}
            label={t('Transfer')}
            icon={VegaIconNames.TRANSFER}
          />
          <Button
            data-testid="account-action-withdraw"
            onClick={() => actions.onClickWithdraw?.(asset.id)}
            label={t('Withdraw')}
            icon={VegaIconNames.WITHDRAW}
          />
        </div>
      ) : null}

```

mattrussell36 (changes requested):

apps/trading/e2e/.env
```diff
@@ -1,4 +1,4 @@
 CONSOLE_IMAGE_NAME=vegaprotocol/trading:develop
 VEGA_VERSION=v0.76.8
-LOCAL_SERVER=false
+LOCAL_SERVER=true
```
mattrussell36, line 3: Please revert

libs/accounts/src/lib/account-card.tsx
asiaznik: Just a leftover from the actual combined button, will clean this up shortly

mattrussell36 (approved):

PR #6722, opened 2024-07-25
feat(trading): reward card design and reward detail page

### LINKED ISSUE #6206: Reward card summary visual refresh
## The Chore

Need to rework these with Joelle's designs

### Reward cards

- [x] All reward cards should show the summary type data with teh button for teh detail ... even if there is only 1 reward behind it (so we have consistent experience)
- [x] The button should count how many rewards are behind it and include in the text as shown 
- [x] The lock icon removed since this is now in the reward detail
- [x] Instead the dispatch strategy is now in the top right and has icons for pro rata vs rank

image

### Reward details

- [x] Create new page with the sketch shown 
- [x] The table shows the actual detail from the `transfers` API and shows
  - [x] The settlement asset 
  - [x] The market name (looked up from the code)
  - [x] The start and end epoch
  - [x] The "Rewards paid" 
    - [x] This is taken from `transferInterval`
    - [x] If = 1 then display "Daily"
    - [x] If > 1 then display "X days" 
  - [x] The lock period
  - [x] The reward amount 
  - [x] The reward card (from `capRewardFeeMultiple`)

image

### Sketch

https://www.figma.com/design/ywcCZvXdvG28bg0AmCY9mF/Console-v3?t=bOVxi9w1DHzKnGOE-0

mattrussell36:
# Related issues 🔗

Closes #6206

# Description ℹ️

- Changes to support grouped reward cards always showing summary data even if there is only one reward behind it.
- Adds a reward detail page for a group of related rewards. This page is created using a querystring from which we can deduce the grouping. 
- Adjusts card component to enable transparent and different variants
- Adds useRewardsGrouped hook to handle grouping logic

# Demo 📺

Screenshot 2024-07-25 at 17 55 54


JonRay15:
Summary cards

We might have to change this count where it says "view rewards on 3 markets" ... because when you click through its actually 2 rewards across 10 total markets. 

image

image

Maybe just "View reward details"?

JonRay15:
Staking rewards

When you click on staking rewards the reward detail page doesnt load.

I suspect this is because iirc staking rewards dont have a dispatch strategy ... they're a bit of a freak, the shape on the transfer API is different, so you probably need specific handling for this detail page.
 
image

JonRay15:
The validator ranking reward isnt showing the reward paid column, ie. "Daily" ... it is just blank.

The transfer interval on that reward is null ... so I suspect we just need handling for that.  The reward on mainnet is definitely going out every day, so I suspect the answer is that if transferInterval is NULL then it is treated as 1 and we should show DAILY.

```
"kind": {
                "startEpoch": 972,
                "endEpoch": null,
                "dispatchStrategy": {
                  "dispatchMetric": "DISPATCH_METRIC_VALIDATOR_RANKING",
                  "dispatchMetricAssetId": "",
                  "marketIdsInScope": null,
                  "entityScope": "ENTITY_SCOPE_INDIVIDUALS",
                  "individualScope": "INDIVIDUAL_SCOPE_ALL",
                  "teamScope": null,
                  "nTopPerformers": "",
                  "stakingRequirement": "",
                  "notionalTimeWeightedAveragePositionRequirement": "",
                  "windowLength": 1,
                  "transferInterval": null,
                  "lockPeriod": 0,
                  "distributionStrategy": "DISTRIBUTION_STRATEGY_PRO_RATA",
                  "capRewardFeeMultiple": null,
                  "rankTable": null
                }
              },
```

asiaznik (approved):

PR #6728, opened 2024-07-29
feat(trading): ticket fixes and changes

### LINKED ISSUE #6720: Deal ticket snags
## The Chore

Various minor snags on review of the new deal ticket changes.

## Tasks

- [x] Add a tooltip to "TIF" saying "Set the time in force (TIF) of the order"
- [x] Probably that tooltip needs to either explain the TIFs available for that order type (see Hyperliquid example below) or should direct to the docs.
- [x] Do somethiing about tooltip behaviour as they can be hard to get rid of. 
  - [x] Position them off to the side so they don't cover the ticket options
- [x] The clickable area for the post only etc checkboxes is pretty wild ... it extends across the entire row I think
- [x] I dont think the size input is setting the slider ... is that intentional? In my example it is set to 500 which should be about 10%
- [x] The price and size fields on the seconf part of the limit OCO are in the wrong order, they shoud match the top one
- [x] There is no size slider on the bottom part of the OCO on either stop limit or stop market
- [x] The fees & notional are not appearing on stop limit or stop market either in normal or OCO mode.

### Weird TIF text

image

### Hyperliquid TIF

image

### Slider behaviour

image

### OCO fields

image

### Stop fees etc

image

mattrussell36:
# Related issues 🔗

Closes #6720 

# Description ℹ️

Various fixes and improvements for the deal ticket
- Allow expiresAt field for GFN
- Rename ticket 'default' to 'derivative'
- Offset tooltips to the left of checkboxes so they don't obscure the ticket
- Add a tooltip for time in force
- Update the slider position when inputting the size
- Fix order of fields for stop orders
- Add size sliders for all stop order size fields (including oco)
- Allows switching to notional mode for stop orders
- Add fees and notional feedback values for stop orders, including a summary of the order to clarify which values relate to which order
- Fix an issue where validation for size override (percentage based size) was using step size


edd (approved):
**nitpick (non-blocking)**: It's a chunky PR so you can't totally trust my review, but it's pretty neat and well organised. Linting errors aside I think this looks fine.

JonRay15:
Dont know if this is related to the work for this PR or not ... but have we always had this weird message?  And has it always looked like this?

image

In mainnet we have this button on market orders prompting a swicth ... which incidentally does not work! 

image

And on limits we have this 

image

So it does look like the new refactor has broken that blue info box.

JonRay15:
TIF tooltip ... I think this is a little confusing when not all modes are available on all order types. 

image

I think either:

Ideal - We should show only the TIF options relevant for that order type (limit / market etc)
Acceptance - We should at least include a second sentence at the top so it says "Set the time in force (TIF) of the order. Available TIF options vary by order type"

JonRay15:
The TPSL and iceberg fields are still indented which we agreed we didnt want to do 

image

JonRay15:
This preview on the button needs to be removed per discussion 

image

JonRay15:
On the slider, fix the slider so that if you swap between amount vs percentage it always maintains the value instead of zeroing it 

For stop limit; this conversion should be done based on the set limit price
For stop market; the conversion will need to be done using the trigger price as the approximation of execution price (it wont be perfect but its the best we have)

image

PR #6736, opened 2024-07-31
fix(trading): fee table label changes

### LINKED ISSUE #6707: Improve fee table labels and visual hierarchy, add maker rebate column
The columns in this fees table are not labelled in line with industry norms, and it's not clear that three of the columns are a breakdown of another column rather than separate fees:

Screenshot 2024-07-18 at 11 11 46

Labels should be updated as follows:

`Total fee after discount` → `Taker fee after discount`
`Total fee before discount` → `Taker fee before discount`
 
We need to make the fee components much more clearly a breakdown of the fee before discount, perhaps we can use larger fonts for the two total fees and a more muted colour for the fee components. If there was a way to make this hierarchy/breakdown even more clear, or even to hide the breakdown initially, that'd be good.

It would also be worth adding a maker rebate column which would need to show a range from the lowest (with maxed out discounts) to the highest (the maker fee component amount) rebate that a maker woudl receive.

This table needs to be redesigned when the new fee components in https://github.com/vegaprotocol/specs/issues/2295 and the dynamic maker fee landhttps://github.com/vegaprotocol/specs/issues/2296 land, but the above should be good for now.

## Acceptance Criteria

- [x] Change `Total fee after discount` → `Taker fee after discount` and pin it to the far left
- [ ] Change `Total fee before discount` → `Taker fee before discount` and move it to be the second column
- [ ] Other columns are unchanged to the right
- [ ] Do something to emphasise those first two columns or de-emphasize the last three (brightness / size)

dexturr:
# Related issues 🔗

Closes #6707 

# Demo 📺

image

# Technical 👨‍🔧

Used muted text class in order to de-emphasise the final 3 columns.


libs/i18n/src/locales/en/trading.json
```diff
@@ -660,8 +660,8 @@
   "Total amount of rewards paid out to this team in qUSD": "Total amount of rewards paid out to this team in qUSD",
   "Total discount": "Total discount",
   "Total distributed": "Total distributed",
-  "Total fee after discount": "Total fee after discount",
-  "Total fee before discount": "Total fee before discount",
+  "Taker fee after discount": "Taker fee after discount",
```
mattrussell36, line 663: Can you move these so the file is kept alphabetical

apps/trading/components/fees-container/market-fees.tsx
```diff
@@ -63,15 +63,24 @@ const useFeesTableColumnDefs = (): ColDef[] => {
...
+          cellRenderer: ({ value }: { value: string }) => (
+            <span className="text-muted">{value}</span>
+          ),
           valueFormatter: ({ value }: { value: number }) => value + '%',
         },
         {
           field: 'totalFee',
-          headerName: t('Total fee before discount'),
+          headerName: t('Taker fee before discount'),
+          cellRenderer: ({ value }: { value: string }) => (
```
mattrussell36, line 81: I'd just use the cellClass property rather than a cellRenderer

mattrussell36 (changes requested):

JonRay15:
Missing a few things

- Taker fee BEFORE discount wants to go to the left to become the second column
- Also I notice that we dont show the % symbole on the infra / maker / taker fee before discount
- And we wanted some way to emphasise the taker fee after discount .... maybe make it bold?  Or highlight the cell? or provide some extra border on the column?  Something to emphasise that it is really the key thing to worry about

image

mattrussell36 (approved):

edd (approved):

PR #6737, opened 2024-07-31
chore(trading): usdt arb only via squid, non usdt arb deposit notific…

### LINKED ISSUE #6688: Add warning in deposit flow when depositing a useless asset
When a user is depositing, we should try to make sure they deposit something that will be useful.

- [x] On squid router block everything other than USDT Arb
- [x] On the normal deposit flow add a warning if the selected asset is not USDT Arb; "The majority of markets on the network settle in USDT Arb, are you sure you wish to deposit?"

asiaznik:
…ation

# Related issues 🔗

Closes #6688 

# Description ℹ️

* allows only USDT Arb to be deposited via Squid
* displays notification when trying to deposit non USDT Arb asset

# Demo 📺

Screenshot 2024-07-31 at 17 40 26
Screenshot 2024-08-01 at 13 19 49


apps/trading/lib/utils/is-asset-usdt-arb.ts
```diff
@@ -0,0 +1,14 @@
+import type { AssetFieldsFragment } from '@vegaprotocol/assets';
+import {
+  ARBITRUM_CHAIN_ID,
+  ARBITRUM_SEPOLIA_CHAIN_ID,
+} from '@vegaprotocol/web3';
+
+export const isAssetUSDTArb = (asset: AssetFieldsFragment) => {
+  const source = asset.source.__typename === 'ERC20' ? asset.source : undefined;
+  return (
+    asset.symbol.toLowerCase() === 'usdt' &&
```
mattrussell36, line 10: We've run into clashes of symbol before. Though its very unlikely on mainnet, I'd prefer if we just had the arb usd asset ID hardcoded in a config file and we used that, we'd then only need to check the id

apps/trading/components/deposit-container/deposit-container.tsx
```diff
@@ -220,32 +222,43 @@ const DepositForm = ({
...
+                        <AssetOption asset={a} />
+                      </TradingRichSelectOption>
+                    );
+                  })}
+                </TradingRichSelect>
+              );
+            }}
+          />
+          {asset && !isAssetUSDTArb(asset) && (
+            <Notification
```
mattrussell36, line 248: We are moving away from this component in forms. Can we instead render text similar to TradingInputError but in a warning (orange) color? Perhaps a new TradingInputWarning component, or perhaps TradingInputError could be changed to InputFeedback with a variant prop

mattrussell36 (changes requested):

mattrussell36 (approved):

JonRay15:
Warning is displayed ... tick

image

On squid we have a locked "receive" as expected. 

image

I wonder if we should have some extra explanation here?

Currently we say "To ensure users enjoy the cheapest fees possible this mode only supports deposits into Arbitrum assets"

We should change this to "This form only supports deposits in Arbitrum USDT since the majority of markets on Vega settle in this asset.  If you wish to deposit a different asset please use the normal deposit flow."

PR #6773, opened 2024-08-09
feat(rest): new rest api library

asiaznik:
# Description ℹ️

* adds new lib
* adds a collection of queries and hooks that read data from rest API


libs/rest/src/hooks/use-tx.ts
mattrussell36, line 1: I think we should probably not include this hook and use one of the options already in the monorepo. If thats not possible lets isolate this hook to the amm pages

libs/rest/src/queries/transactions.ts
mattrussell36, line 1: Delete

libs/rest/src/query-client.ts
mattrussell36, line 1: Delete as query client already available

mattrussell36 (changes requested):
I see no reason why we need to introduce a new env var for this. Same goes for the extra ui to handle setting two urls in the nodes switcher which also won't be needed.

Two urls in the node switcher wont make a lot of sense to users who don't care if its rest or gql, they just want to pick a node. All the data nodes follow the same url patterns for different endpoints, so why not just alter VEGA_URL to not include the graphql path? This will simplify our code and be better for users.

- Change VEGA_URL to _not_ include the `/graphql` path
- Append the `/graphql` path in the `createClient` function, or wherever the apollo client is created in each app.
- Just use VEGA_URL for rest, or append the versioning path if required in the `getBaseUrl` function
- Remove the `/graphql` path from VEGA_URL in env vars

libs/rest/src/paths.ts
```diff
@@ -0,0 +1,67 @@
+import trimEnd from 'lodash/trimEnd';
+import path from 'path';
+import { getBaseUrl } from './rest-config';
+import tradingDataApi from '@vegaprotocol/rest-clients/openapi/trading_data_v2.openapi.json';
```
mattrussell36, line 4: Good idea to use this

libs/environment/src/components/node-switcher/node-switcher.tsx
```diff
@@ -139,31 +152,35 @@ export const NodeSwitcher = ({ closeDialog }: { closeDialog: () => void }) => {
 };
 
 interface CustomRowWrapperProps {
-  inputText: string;
-  setInputText: (text: string) => void;
-  nodes: string[];
+  graphQLApiUrl: string;
```
mattrussell36, line 155: See my main comment, but I don't think we need to change anything in the node switcher, if we continue to use VEGA_URL without the gql path

asiaznik:
> I see no reason why we need to introduce a new env var for this. Same goes for the extra ui to handle setting two urls in the nodes switcher which also won't be needed.
> 
> Two urls in the node switcher wont make a lot of sense to users who don't care if its rest or gql, they just want to pick a node. All the data nodes follow the same url patterns for different endpoints, so why not just alter VEGA_URL to not include the graphql path? This will simplify our code and be better for users.
> 
> * Change VEGA_URL to _not_ include the `/graphql` path
> * Append the `/graphql` path in the `createClient` function, or wherever the apollo client is created in each app.
> * Just use VEGA_URL for rest, or append the versioning path if required in the `getBaseUrl` function
> * Remove the `/graphql` path from VEGA_URL in env vars

As always there's an exception to the rule :D

GQL: https://vega.xprv.io/datanode,
REST: https://vega.xprv.io/datanode/query

I remember relying on a url convention has bitten us in the past for that ☝🏻 particular reason, hence I wanted to foolproof it with having explicit values for gql and rest.

libs/rest/src/paths.ts
asiaznik: Yeah, I can generate additional file there that would consist of only the paths.

mattrussell36 (approved):

PR #6775, opened 2024-08-12
feat(trading): asset activity withdrawal completion flow

### LINKED ISSUE #6101: Transaction list in portfolio

mattrussell36:
# Related issues 🔗

Closes #6101 

# Description ℹ️

- Makes useTransfers just use generated query so that polling does not reset the data causing extra mounts/unmounts
- Remove the connectkit button and instead prompt connection and send tx using the same click handler.
- Replace 'view' button link with info icon
- Change complete button link to secondary button
- Highlight pinned withdraw rows with a light blue background

# Screenshot
Screenshot 2024-08-12 at 12 46 41

### DIFF
--- a/apps/trading/components/asset-activity/asset-activity.tsx
+++ b/apps/trading/components/asset-activity/asset-activity.tsx
@@ -247,6 +247,7 @@ export const AssetActivityDatagrid = ({
         rowData={rowData}
         pinnedTopRowData={pinnedTopRowData}
         overlayNoRowsTemplate={t('No data')}
+        rowClass="ag-highlight-pinned"
       />
       <Pagination
         count={rowData.length}
--- a/apps/trading/components/asset-activity/withdrawal-status-cell.tsx
+++ b/apps/trading/components/asset-activity/withdrawal-status-cell.tsx
@@ -13,6 +13,7 @@ import { useAccount, useReadContract } from 'wagmi';
 import { BRIDGE_ABI } from '@vegaprotocol/smart-contracts';
 import { useEffect, useRef, useState } from 'react';
 import { useModal } from 'connectkit';
+import { TradingButton as Button, Intent } from '@vegaprotocol/ui-toolkit';
 
 type Props = {
   data: RowWithdrawal;
@@ -152,7 +153,9 @@ const WithdrawalStatusOpen = ({ data, openDialog }: Props) => {
     return (
       <span className="flex gap-1 items-center">
         {t('Pending')}:{' '}
-        <button
+        <Button
+          intent={Intent.Secondary}
+          size="extra-small"
           onClick={() => {
             if (ethWalletStatus === 'disconnected') {
               modal.setOpen(true);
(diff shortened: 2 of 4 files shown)


JonRay15:
Environment pain stopping me from testing this properly but lets assume it works and we can always re-review it later.

JonRay15 (approved):

PR #6791, opened 2024-08-14
feat: amm ui

asiaznik:
# Related issues 🔗

Closes #[Issue number here]

# Description ℹ️

# Demo 📺

# Technical 👨‍🔧


libs/utils/src/lib/cn.ts
```diff
@@ -0,0 +1,6 @@
+import { type ClassValue, clsx } from 'clsx';
```
mattrussell36, line 1: Just a note that I've added this func to the next branch already in ui-toolkit

PR #6797, opened 2024-08-19
feat(browser-wallet,trading): create derived key from meta mask

### LINKED ISSUE #6754: Connect embedded wallet to ethereum
## The Chore

The very first MVP just embed the existing wallet in console, but wallet creation / seed phrase etc is all manual inside the wallet the same as it is now.  This ticket is to use an ethereum wallet to sign a TX to create the seed phrase. 

## Tasks

- [ ] Connect to ethereum wallet 
- [ ] Sign TX to create seed
- [ ] It works

dexturr:
# Related issues 🔗

Closes #6754

# Description ℹ️

Adds ability to create a mnemonic from a signed message derived from an Ethereum wallet

https://github.com/user-attachments/assets/e4955420-5bc9-4ca4-8e4c-bb7662aa6b02


libs/browser-wallet/src/frontend/hooks/suggest-mnemonic/index.ts
```diff
@@ -2,6 +2,57 @@ import { useCallback, useEffect, useState } from 'react';
...
 import { RpcMethods } from '@/lib/client-rpc-methods';
+import { useSignTypedData } from 'wagmi';
+
+/**
+ * Derives a mnemonic from the user's connected Ethereum wallet
+ */
+export const useDeriveMnemonic = (chainId: number) => {
+  const { signTypedDataAsync } = useSignTypedData();
+  const { request } = useJsonRpcClient();
+  const [error, setError] = useState<string | null>(null);
```
mattrussell36, line 13: I don't think you need these error and loading states. wagmi hooks leverage react-query, so if you look at the returned data from the hook you will see lots of helpful state

mattrussell36 (changes requested):

libs/browser-wallet/src/frontend/routes/onboarding/components/mnemonic-page/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { WalletCreated } from './wallet-created';
+import { useJsonRpcClient } from '@/contexts/json-rpc/json-rpc-context';
+import { createWallet } from '@/lib/create-wallet';
+
+export const locators = {
+  saveMnemonicDescription: 'save-mnemonic-description',
+};
+
+export const MnemonicPage = ({ mnemonic }: { mnemonic: string | null }) => {
+  const [showSuccess, setShowSuccess] = useState(false);
```
mattrussell36, line 18: non blocking: I'd like to try and avoid writing success/error/loading state  handling over and over again. You could look at using react query here (useMutation)

libs/browser-wallet/src/frontend/hooks/suggest-mnemonic/index.ts
dexturr: Discussed on slack, we want to move everything to using react query where possible but will do as a follow up task

mattrussell36 (approved):

Issue #6908, opened 2024-09-16
Control network when signing message on "connect with ethereum" mode


JonRay15:
## The Chore

There are two issues with this mode right now:

1. When we are upgrading the build it is changing the message we're signing (in the latest case "Vega Onboarding" changed to "Onboarding") and this is meaning we generate different keys .... losing any funds the user had deposited previously.
2. The signature is being signed on whichever chain id you happen to be connected to on MM at that time which means that you get a different key if you happen to be on another network at the time

## Tasks

- [x] Ensure the signature is completely deterministic at all times by triggering WAGMI to set it to ethereum mainnet before signing the transaction
- [x] When generating the message we want the format "`chainName` Onboarding" ... this might mean adding a chain name to whatever config file you've been using so we can store a user friendly chain name against the chain ID (or maybe this already exists)
- [x] Once we've made that change NEVER CHANGED THIS AGAIN unless there's very good reason
- [x] In addition to this I suggest we actually add back in the export key option in settings because this would protect early users from something happening if they had that key saved

### Example of AVAX sign

![image](https://github.com/user-attachments/assets/017bd472-54b7-4ac2-9a51-3647249d62da)

JonRay15:
Network switcher kicked in properly 

![image](https://github.com/user-attachments/assets/f3ddba6d-d661-4f65-b624-8e26b0dacc4c)

Onboarding message was updated.

Connection successful.

JonRay15:
Export recovery phrase doesnt work.

![image](https://github.com/user-attachments/assets/60e1bd23-5ede-4e64-8f7b-faae25b95779)

Bumping back to IN PROG for Dex tomorrow morning.

JonRay15:
There is still an open bug on export recovery but its on a diff ticket now ... this works so closing.

PR #6910, opened 2024-09-17
feat(browser-wallet,trading): switch chain

dexturr:
# Related issues 🔗

#6908

# Description ℹ️

1. Ensure that the chain is always Arbitrum mainnet whenever we create a wallet
2. Pass app name into the message to be signed in order
3. Allow for exporting of recovery phrases


libs/browser-wallet/src/frontend/routes/auth/settings/home/sections/export-recovery-phrase-section/export-recovery-phrase-section.tsx
```diff
@@ -15,11 +16,24 @@ export const locators = {
 export const ExportRecoveryPhraseSection = () => {
   const [open, setOpen] = useState(false);
   const [recoveryPhrase, setRecoveryPhrase] = useState<string | null>(null);
+  const { request } = useJsonRpcClient();
+  useEffect(() => {
```
mattrussell36, line 20: non blocking: I would prefer this was run on an action, such as the click to open the dialog

mattrussell36 (approved):

PR #6915, opened 2024-09-17
fix(6896): notional asset symbol

### LINKED ISSUE #6896: For future & perp metadata being used incorrectly
## The Chore

## Tasks

- [x] Price is currently using SETTLEMENT ASSET and it should be using QUOTE from the metadata
- [x] Size has two potential modes:
  - [ ] Size is fine ... it is using BASE
  - [x] ~Notional it is not right ... it should be using QUOTE again but its using SETTLEMENT~

## Final ACs

On Deal ticket
- [ ] Size unit = "Contracts"
- [ ] Notional unit = Settlement asset

On trade table
- [ ] Price = quote in metadata
- [ ] Notional = Settlement asset

asiaznik:
# Related issues 🔗

Closes #6896

### DIFF
--- a/apps/trading/components/ticket/fields/notional.tsx
+++ b/apps/trading/components/ticket/fields/notional.tsx
@@ -91,7 +91,16 @@ const NotionalLabel = () => {
   const t = useT();
   const ticket = useTicketContext();
   const label = t('Notional');
-  const symbol = ticket.quoteSymbol;
+
+  let symbol = undefined;
+  switch (ticket.type) {
+    case 'spot':
+      symbol = ticket.quoteAsset.symbol;
+      break;
+    case 'default':
+      symbol = ticket.settlementAsset.symbol;
+      break;
+  }
 
   return <InputLabel label={label} symbol={symbol} />;
 };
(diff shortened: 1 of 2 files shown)


apps/trading/components/ticket/fields/notional.tsx
```diff
@@ -91,7 +91,16 @@ const NotionalLabel = () => {
   const t = useT();
   const ticket = useTicketContext();
   const label = t('Notional');
-  const symbol = ticket.quoteSymbol;
+
+  let symbol = undefined;
```
mattrussell36, line 95: Why not a ternary here. There are only two options so is concise and readable

apps/trading/components/ticket/info/notional.tsx
```diff
@@ -14,7 +14,17 @@ export const Notional = ({
   const form = useForm();
   const ticket = useTicketContext();
   const notional = form.watch(name);
-  const symbol = ticket.quoteSymbol;
+
+  let symbol = undefined;
```
mattrussell36, line 18: Same as in notional.tsx

mattrussell36 (changes requested):

mattrussell36 (approved):

Issue #6906, opened 2024-09-16
Market order & Toast issues with embedded wallet


JonRay15:
## The Chore

Have observed two issues:

## Tasks

- [ ] When making a market order it does not seem to send the order at all and ask me to approve
- [ ] When making a limit order I can get it to send but the toast stays stuck on orange approval step and never moves to green even though I can see it hhas gone through >> NOTE I've seen some of these working now in FG, so maybe it is not a problem.

JonRay15:
Both of these are fixed now

PR #6916, opened 2024-09-17
fix(trading,browser-wallet): market order

dexturr:
# Related issues 🔗

#6906

### DIFF
--- a/libs/browser-wallet-backend/src/src/client-ns.js
+++ b/libs/browser-wallet-backend/src/src/client-ns.js
@@ -132,7 +132,7 @@ export default function init({
       },
       async 'client.send_transaction'(params, context) {
         const receivedAt = new Date().toISOString();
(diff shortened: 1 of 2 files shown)


libs/browser-wallet-backend/src/src/client-ns.js
```diff
@@ -132,7 +132,7 @@ export default function init({
       },
       async 'client.send_transaction'(params, context) {
         const receivedAt = new Date().toISOString();
-        doValidate(clientValidation.sendTransaction, params);
+        // doValidate(clientValidation.sendTransaction, params);
```
dexturr, line 135:
```suggestion
        // TODO the validation needs to be moved into this repo so this is consistent with current protos
        // doValidate(clientValidation.sendTransaction, params);
```

mattrussell36 (approved):

PR #6951, opened 2024-09-18
fix(trading): do not show existing referees text for solo players

### LINKED ISSUE #6790: Tweaks to Solo play / lone wolf
## The Chore

Want to remove the team related terminology from Solo play

## Tasks

  - [x] Change the options for signup to the three in the sketch in that order
  - [x] When you hit play as individual
    - [x] On the create team screen change "create solo team" to "Create my profile"
    - [x] Instead of Team Name display "Trading alias"
    - [x] Replace "Provide a link so users can learn more about your team" with "Provide a link so users can learn more about you"
    - [x] When you submit the TX ensure it is setting a whitelist with only your key on it

Image

edd:
# Related issues 🔗

Closes #6790

# Description ℹ️

Hide some additional text that does not make sense when the create team upgrade is for a
'solo team'.

### DIFF
--- a/apps/trading/client-pages/competitions/competitions-create-team.tsx
+++ b/apps/trading/client-pages/competitions/competitions-create-team.tsx
@@ -88,7 +88,7 @@ export const CompetitionsCreateTeam = () => {
               </Button>
             </>
           )}
-          {isUpgrade && (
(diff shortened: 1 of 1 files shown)


JonRay15:
This is fixed

PR #6970, opened 2024-09-19
fix(trading): update team from private back to public

### LINKED ISSUE #6965: Update team cannot re-open a team
## The Chore

When on the "Update Team" page if you try to set a team from private to not private it doesnt work.

ie. I go to a team that has the private checkbox ticked, I untick it, and I save changes. 

The transaction that is sent has 

```
{
  "updateReferralSet": {
    "id": "7e42dc8168827a39c69318714a21ec54daed38a3d1d66cc473c9dab9f32249c6",
    "isTeam": true,
    "team": {
      "name": "Testet 6 Team",
      "teamUrl": "",
      "avatarUrl": "",
      "closed": false,
      "allowList": []
    }
  }
}
```

But false is incorrect as this is NULL so will be interpreted as no change.

## Tasks

- [x] When user is trying to change a private team to a public one detect that is what they've changed and send closed = `ptr-to-false`
- [x] Somehow I feel like a toggle for private / public might be better than the current checkbox in terms of UX ... if you agree maybe try changing it

asiaznik:
# Related issues 🔗

Closes #6965 

# Demo 📺

image

### DIFF
--- a/apps/trading/client-pages/competitions/team-form.tsx
+++ b/apps/trading/client-pages/competitions/team-form.tsx
@@ -68,7 +68,7 @@ const prepareTransaction = (
             name: fields.name.trim(),
             teamUrl: fields.url,
             avatarUrl: fields.avatarUrl,
-            closed: fields.private,
+            closed: fields.private || 'ptr-to-false',
             allowList: fields.private
               ? parseAllowListText(fields.allowList)
               : [],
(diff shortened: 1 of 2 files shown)


JonRay15 (approved):

JonRay15:
All works.

outbound TX verified

```
{
  "updateReferralSet": {
    "id": "fc4a48a74a52df97d672044faf71374a369d5488ee928b8494450f0645122ff2",
    "isTeam": true,
    "team": {
      "name": "Lets play",
      "teamUrl": "",
      "avatarUrl": "",
      "closed": "ptr-to-false",
      "allowList": []
    }
  }
}
```

PR #6984, opened 2024-09-23
feat(trading,browser-wallet): enable auto consent by default, fix auto consent always failing

dexturr:
Fixes the fact auto consent was always disabled due to a syntax error
Enables autoConsent by default


edd:
@dexturr Testing on the preview link, I connected the testnet wallet. I don't have any funds to test with, but in the settings I see this:

Screenshot 2024-09-23 at 10 08 55

That suggests this isn't fixed to me, how else can I verify it?

edd (approved):
Screenshot 2024-09-23 at 10 24 09

Working nicely now I'm using the correct wallet

